Issue #59, opened 2013-05-10
Use contact point positions computed by FCL


jslee02:
1. DART creates meshes for primitive geometries and conduct collision detection using FCL. It sometimes makes unnatural looking result because the collision result is not analytical solution. For example, a sphere does not roll on the ground smoothly. Now FCL supports primitive geometries and analytical solution. Let's use it in DART.
2. DART calculates contact points with triangle-triangle contact information from FCL because older version of FCL did not provide the contact points. Let's use it in DART, too!

jslee02:
Changed title to confine the issue because primitive shape issue duplicates with #19.

Pending until [FCL's incorrect contact point computation](https://github.com/flexible-collision-library/fcl/issues/106) is resolved.

PR #156, opened 2014-03-17
Add optional collision detector: BulletCollisionDetector

jslee02:
DART has 3 collision detector types:
- [`FCLMeshCollisionDetector`](https://github.com/dartsim/dart/tree/master/dart/collision/fcl_mesh) (default)
- [`FCLCollisionDetector`](https://github.com/dartsim/dart/tree/master/dart/collision/fcl)
- [`DARTCollisionDetector`](https://github.com/dartsim/dart/tree/master/dart/collision/dart)

Both of `FCLMeshCollisionDetector` and `FCLCollisionDetector` are based on [`fcl`](https://github.com/flexible-collision-library/fcl). The difference between them is that `FCLMeshCollisionDetector` convert all the primitive shapes to meshs for collision detection while `fcl` uses primitive shapes of 'fcl'.

Some simulation requires analytic collision information using primitive shapes for example rolling wheel for vehicle simulation. In this sense, `FCLCollisionDetector` or `DARTCollisionDetector` is good solution instead of `FCLMeshCollisionDetector`. However, they have issues to use as default collision detector of DART.

`fcl` returns only one contact point for one collision pairs. This leads to unexpected behavior. See [#59] for the detail. Whereas `DARTCollisionDetector` does not cover all the shapes of DART. It supports only `BoxShape` and `Sphere`.

Under current circumstance, `BulletCollision` may be considerable solution. It supports primitive shapes and mesh together and returns reasonable number of contact points for primitive shapes. Of course [`Bullet`](http://bulletphysics.org/wordpress/) is another physics engine as it is. We uses `BulletCollision` component only for collision detection and separate [debian package](http://packages.ubuntu.com/search?keywords=bullet&searchon=names&suite=all&section=all) of it is delivered since Ubuntu 13.10. 

This may not our long term solution but I think it is worth in case of that analytic collision solution for primitive shapes and mesh support are required at the same time. Also this pull request does not increase mandatory dependency. When `BulletCollision` is installed on the machine, then cmake detect it and add `BulletCollsisionDetector` of DART.

### DIFF
--- a/dart/CMakeLists.txt
+++ b/dart/CMakeLists.txt
@@ -5,7 +5,7 @@ if(MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
 endif()
 
-set(CORE_SUBDIRS common math dynamics renderer collision lcpsolver constraint integration simulation)
+set(CORE_SUBDIRS common math dynamics renderer lcpsolver constraint integration simulation)
 set(NONCORE_SUBDIRS gui planning utils)
 # optimizer is also a NONCORE_SUBDIR, but it will be handled separately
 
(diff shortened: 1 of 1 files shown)


ljklonepiece:
Hi there, just wondering is there any chance DARTCollisionDetector will support MeshShape objects soon?

Issue #211, opened 2014-07-03
libdart-core4.1 should conflict with libdart-core4.0


scpeters:
I'm getting a failure when trying to install libdart-core4.1:
~~~
Unpacking libdart-core4.1 (4.1.0-ppa1~trusty) ...
dpkg: error processing archive /var/cache/apt/archives/libdart-core4.1_4.1.0-ppa1~trusty_amd64.deb (--unpack):
 trying to overwrite '/usr/lib/libdart-core4.so', which is also in package libdart-core4.0 4.0.0-ppa2~trusty
~~~
@j-rivero

We can probably make a similar change as the one in #202 / #203

jslee02:
I believe this is fixed by #286 since `libdart-core4.x` installs `libdart-core4.so.*` (not `libdart-core4.so`).

scpeters:
Fixed by #286

PR #286, opened 2014-12-11
Version numbering of the shared libraries in debian packages

### LINKED ISSUE #281: Debian install error for libdart-core4.2
I just tried upgrading to the new version of `libdart-core4-dev`, but I got an error:
~~~
The following packages have unmet dependencies:
 libdart-core4-dev : Depends: libdart-core4.2 (= 4.2.0-ppa0~trusty) but it is not going to be installed
E: Unmet dependencies. Try 'apt-get -f install' with no packages (or specify a solution).
~~~

So I tried installing `libdart-core4.2` and it has a conflict with `libdart-core4.1`:
~~~
Unpacking libdart-core4.2 (4.2.0-ppa0~trusty) ...
dpkg: error processing archive /var/cache/apt/archives/libdart-core4.2_4.2.0-ppa0~trusty_amd64.deb (--unpack):
 trying to overwrite '/usr/lib/libdart-core4.so', which is also in package libdart-core4.1 4.1.1-ppa0~trusty
~~~

Normally I would say to add a `Conflicts with` to the control file, but I thought libraries could be installed side-by-side. Is it a problem with having the major number `4` in the library name instead of being separated by `.`'s?

@j-rivero

jslee02:
Version numbering of the shared libraries in debian packages

This PR changes naming of the shared libraries. Since ABI breaking is done per minor version up, minor version number needs to be appear to the shared libraries.

For example:
* `libdart-core4.2` package installs `libdart-core.so.4.2`
* `libdart-core4-dev` package installs `libdart-core.so.4` symlink

This PR should fixes #281.


jslee02:
I think, this pr is ready to merge, if I correctly followed @j-rivero 's comment.
+1

scpeters:
Fixed #211

PR #288, opened 2014-12-16
DegreeOfFreedom class

mxgrey:
This pull request introduces the DegreeOfFreedom class which serves as a proxy class for getting/setting the data (e.g. position, velocity, acceleration, force, limits, etc) of individual generalized coordinates (aka degrees of freedom). This pull request also gives names to each degree of freedom and allows each degree of freedom to be accessed by name. The name of each degree of freedom in a given Skeleton is kept unique by the NameManager class.

By default, each degree of freedom is given a name based on the name of the joint it belongs to. For SingleDofJoints, their single dof is just given the same name as the joint itself. For MultiDofJoints, the dof is given a name with the pattern "\<joint_name\>_\<affix\>" where "\<joint_name\>" is simple the name of its joint, and "\<affix\>" indicates its role in the joint. For example, a TranslationalJoint named "joint1" will be given degrees of freedom with the following names: "joint1_x", "joint1_y", and "joint1_z".

The user can rename any DegreeOfFreedom at any time, however names that match already existing DOF names will be altered by the NameManager to prevent a lookup conflict.

When the user changes a joint name using Joint::setName(~), they are given the option of whether or not to have the joint's degrees of freedom automatically renamed as well (default is true).

Something to consider: Currently, if the Skeleton's Joint NameManager needs to change a joint's name to prevent a naming conflict, it will also rename that Joint's DegreesOfFreedom to match the new name. This could potentially wipe out custom names that the user might have given to that Joint's DegreesOfFreedom. Should we have a flag that the user can set to prevent this behavior?


dart/dynamics/Skeleton.h
```diff
@@ -591,8 +606,15 @@ class Skeleton
   //----------------------------------------------------------------------------
   friend class BodyNode;
   friend class Joint;
+  friend class DegreeOfFreedom;
 
 protected:
+  /// Register a joint with the Skeleton. Internal use only.
+  void registerJoint(dart::dynamics::Joint* _newJoint);
```
jslee02: Skeleton and Joint are in the same namespace. Why the full namespace is specified here?

dart/dynamics/Joint.h
```diff
@@ -119,6 +126,14 @@ class Joint
   /// Get number of generalized coordinates
   DEPRECATED(4.1) virtual size_t getDof() const = 0;
 
+  /// Get an object to access the _index-th degree of freedom (generalized
+  /// coordinate) of this Joint
+  virtual DegreeOfFreedom* getDof(size_t _index) = 0;
+
+  /// Get an object to access the _index-th degree of freedom (generalized
+  /// coordinate) of this Joint
+  virtual const DegreeOfFreedom* getDof(size_t _index) const = 0;
```
jslee02: Skeleton can get DegreeOfFreedom by the name. Shouldn't we provide the same functionality for Joint class?

dart/dynamics/UniversalJoint.h
```diff
@@ -54,6 +54,9 @@ class UniversalJoint : public MultiDofJoint<2>
   UniversalJoint(const Eigen::Vector3d& _axis0 = Eigen::Vector3d::UnitX(),
                  const Eigen::Vector3d& _axis1 = Eigen::Vector3d::UnitY(),
                  const std::string& _name = "Universal joint");
+  // TODO(MXG): We should be more consistent about indexing starting from 0 or 1
+  // In the constructor arguments, we start with axis0, but in the member
+  // functions we start with axis1.
```
jslee02: How about we always use zero-based numbers?

jslee02:
Added some commits for style fixes, more comments, and adapting to your (todo) comment in the code, and also made some line notes.

Regarding the flag to prevent possible wiping out custom names, I'm not sure what's the best. We may need more discussion on the DegreeOfFreedom name management. Let's leave it as an issue for now.

dart/dynamics/Joint.h
mxgrey: I was debating whether or not to do that. I decided not to because the Skeleton is responsible for DOF name management (because each DOF in the skeleton must have a unique name), so accessing DOFs by name should be done through the skeleton.

Having the joints also keep track of DOF names seems redundant. And if we gave the joint class a function that uses the skeleton to look up DOFs it could be strange/confusing if someone asks a joint for a DOF that belongs to another joint.

In my mind, the thing that makes the most sense is asking the Skeleton for access to a DOF if you want to access by name, and you should only go through the Joint to access a DOF if you want to access it by its index within that Joint.

Let me know if that doesn't make sense.

dart/dynamics/UniversalJoint.h
mxgrey: I'm fine with that, although it will require an API change, since the function names themselves are indexing starting from 1. If an API change like that is okay right now, then I can make that change.

jslee02: Wait, I thought you mentioned about the name affix for DegreeOfFreedom of UniversalJoint, but it was about setAxis1() and setAxis2() (and corresponding getters).

I looked up to find other names have numbers. For example, [Contact](https://github.com/dartsim/dart/blob/master/dart/collision/CollisionDetector.h#L60) and [ContactConstraint](https://github.com/dartsim/dart/blob/master/dart/constraint/ContactConstraint.h#L56) has members indexing starting from 1. So let's change the parameter names to _axis1 and _axis2. It will not change the API as well.

dart/dynamics/Joint.h
jslee02: It makes sense. Let's not to add the function to Joint.

I'm just wondering that the Joint might be responsible for DOF name management since DOF is belongs to Joint. It would be okay that the DOF name is unique in the Joint. The name uniqueness in Skeleton is guaranteed by **full name**. The full name would be like "[SkeletonName].[JointName].[DofName]". But this idea is not finalized yet.

mxgrey: I see what you're saying. It's true that the DOF belongs to its joint, but one of the purposes of the DOF class is to offer a more flexible way of accessing joint properties. In particular, I think it's valuable for the user to be able to customize DOF names however they'd like to, and then access those names through the Skeleton::getDof(string) function. For that to work, we need the Skeleton to manage DOF name uniqueness. We could enforce name uniqueness requiring a [JointName].[DofName] style as you're suggesting, but that would detract from customizability.

mxgrey:
The new commit contains two major changes:

1) I've added a *dof* element into the skeleton file format. This allows skeleton files to specify things like name and position/velocity/acceleration limits for individual degrees of freedom in each joint. The format is agnostic to the type of joint. The old approach still works (i.e. using axis, axis2, axis3, etc. elements), but any information provided in the new *dof* element will override any information provided by the old format.

2) The user now has the ability to lock-in their customized DOF names so that they will not automatically get overwritten. I found that having this feature was necessary in order to correctly parse DOF names in skeleton files, so now the user has access to this feature as well.

mxgrey:
In the latest changes, if a user calls the "DegreeOfFreedom::setName(string)" function and leaves the second argument as default, it will preserve the custom name that they gave to that DegreeOfFreedom. Related to this, if a custom name is specified in the skel file, it will have that name be preserved. Preserving a name prevents DART from automatically changing the name when its joint properties change.

Error checking for the dof element in skel parsing has been improved. If an attribute has the wrong type, it will be reported. Also, the local_index attribute is required for dof elements belonging to multi-dof joints.

jslee02:
Waiting to make a decision to pick the terminology between "force" or "effort".

mxgrey:
All uses of "Effort" have now been switched to "Force" to match the rest of DART's conventions.

mxgrey:
Manually merged the master branch into here so that the pull request can be auto-merged

jslee02:
+1

PR #291, opened 2014-12-31
Fix bug in collision checking when a body node has multiple collision shapes

tobiaskunz:
Fix bug: Only the first collision shape of a body node is considered when checking for collision without calculating contact points

The change is smaller than it seems here. When ignoring whitespace changes, only 3 lines are changed.

I have only tested this bug fix with DART 3, but it should be the same for DART 4.

### DIFF
--- a/dart/collision/fcl_mesh/FCLMeshCollisionNode.cpp
+++ b/dart/collision/fcl_mesh/FCLMeshCollisionNode.cpp
@@ -177,7 +177,12 @@ bool FCLMeshCollisionNode::detectCollision(FCLMeshCollisionNode* _otherNode,
         collision = true;
 
       if (!_contactPoints)
-        return collision;
+      {
+        if (collision)
+          return true;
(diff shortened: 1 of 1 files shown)


jslee02:
+1

I'm just curious. How did you revert the previous commit?

tobiaskunz:
I reverted my working tree back to the master branch, created a new commit and pushed it to the remote branch of the pull request (fix_multiple_collision_shapes). This would usually fail because the remote branch is ahead one commit and you are asked to do a merge. The force option (-f) let's you ignore that error and let's you force your local branch onto the remote branch without merging and with all previous changes to the remote branch being lost.

This is usually a bad idea to do if other people are working on the same remote branch (like for example the master branch). But I think in this case it is okay.

jslee02:
I see, thanks. It seems like a good trick to revert small changes.

Anyways, this PR looks good to merge.

PR #298, opened 2015-01-07
Hybrid dynamics

jslee02:
This pull request adds hybrid dynamics. In the hybrid dynamics problem, we are given either joint force or joint acceleration at each joint, and the task is to calculate the unknown accelerations and forces. In effect, hybrid dynamics is like performing forward dynamics at some of the joints and inverse dynamics at the rest. Hybrid dynamics is used to introduce prescribed motions into a rigid-body system. [1]

This pull request also introduces joint actuator types: force, passive, acceleration, and velocity (servo is not implemented yet). The hybrid dynamics performs forward dynamics for force and passive types, and inverse dynamics for acceleration and velocity types. Here is description on the command input and the output for each actuator types.

* `FORCE`: The command input is joint force, and the output is joint acceleration. If the command is always set to zero, then the force joint is identical to passive joint. 
* `PASSIVE`: The command input is joint force and always zero (ignore command from the user), and the output is joint acceleration. 
* `ACCELERATION`: The command input is joint acceleration, and the output is joint force.
* `VELOCITY`: The command input is joint velocity, and the output is joint force.
* `LOCKED`: Locked joint always set the velocity and acceleration to zero so that the joint doesn't move from the current position (locked), and the output is joint force.

The joint actuator type can be specified by calling `Joint::setActuator()`. The default type is FORCE for backward compatibility.

We define `FORCE`/`PASSIVE`/`SERVO` joints are dynamic joints while `ACCELERATION`/`VELOCITY`/`LOCKED` joints are kinematic joints. (see also: `Joint::isKinematic()` and `Joint::isDynamic()`)

Example app, `hybridDyanmics` is added, and basic unit tests are added as well. `SkelParser` can parse the actuator type as an optional attribute of `<joint>`. If it is not specified, force type is used as default. The description of the updated skel file format can be found at [wiki](https://github.com/dartsim/dart/wiki/skel-0.1). Here is an example of specifying joint actuator type:
```xml
<joint type="revolute" name="joint1" actuator="force">
    <parent>link 1</parent>
    <child>link 2</child>
</joint>
```

Minor changes:
* Enabled `JAVADOC_AUTOBRIEF` option of doxygen
* Deprecated `Joint::mWrench`
* Deprecated `castUIntToInt()`

[1] Roy Featherstone, *Rigid Body Dynamics Algorithms*, Springer, New York, 2008.


jslee02:
I found two things to be done:
* Joint position limit should be ignored when the actuator type is ACCELERATION or VELOCITY
* Check routine is needed to determine that a body is responsive to impulse depending on the actuator type of its parent joint.

jslee02:
The last commit contains following changes:
* It defines motion-prescribed joint, which means the joint takes prescribed motion as command input. ACCELERATION/VELOCITY joints are the prescribed joint. `Joint::isMotionPrescribed()` returns whether the joint is prescribed.
* If the joint is motion-prescribed joint then the constraint solver ignore joint position limit.
* If two bodies are colliding whose all the ancestor joints are motion-prescribed joints, then constraint solver ignore the contact constraint as well because the bodies are not reactive to contact impulse.

Waiting to pass build tests of Travis-CI and Appveyor.

jslee02:
* Added "locked" actuator type whose velocity and acceleration are always zero so that the joint doesn't move at all from the current position.
* Renamed `Joint::isMotionPrescribed()` to `Joint::isKinematic()`, and added 'Joint::isDynamic()`.

jslee02:
Ready to merge
+1

PR #300, opened 2015-01-15
BSpline class

jslee02:
This pull request adds `BSpline` class, which is a wrapper class of `Eigen::Spline`. `BSpline` provides essential interfaces such as modifiers of knots and control points that `Eigen:Spline` doesn't provide itself.


jslee02:
Failing build on Windows because the version of eigen in the dart-prerequisites is too low. DART 4.3. requires eigen-3.1.0 or greater.

jslee02:
This PR is quite outdated. I will recreate another PR soon. Closing for now.

PR #302, opened 2015-01-17
Fix build issues on Mac

### LINKED ISSUE #282: Build error on Mac
The dependency settings between namespaces are not completed in cmake configuration. This seems fine on Linux but not on Mac.

jslee02:
This pull request solves build issues on Mac.

* Cyclic dependencies between namespace libraries
By #278, namespace-wise cmake files were used to support various build options with DART's optional dependencies. The each cmake files generated libraries per namespace like libdart_common and libdart_math. Some libraries, however, are dependent on each other and it causes cyclic dependencies between the libraries. Somehow, Mac is the only platform complains this. This PR generates single libraries for each `dart` and `dart-core` while maintaining other useful features introduced by #278.

* Case-insensitive
One of useful feature introduced by #278 is that DART generates all-inclusive headers for each namespaces like `constraint.h` and `dynamics.h` that includes all the headers in the corresponding namespaces. These autogenerated headers are all lower cases. The problem is that `constraint` namespace has `Constraint.h` and it conflicts with `constraint.h` on Mac. So, this PR removes the extensions of the autogenerated headers like `constraint` and `dynamics`.

* Source file without any symbol
Mac complained to the source file that doesn't have any symbols in them. So `ClosedLoopConstraint.h/cpp` are removed.

Fix #282.


scpeters:
This does fix the build, but I don't feel great about dropping the header file extension entirely. We faced the case sensitivity issue previously in gazebo (this must be a windows issue as well right?). For example, we had `Physics.hh` and `physics.hh`. We ended up renaming to `PhysicsIface.hh` and `physics.hh` (see [pull request 638](https://bitbucket.org/osrf/gazebo/pull-request/638/fix-for-775-deprecate-headers-that-require/diff#chg-tools/issue_775_generator.bash)).

I would figure out how to rename the headers, or use different extensions `.h`, `.hh`, `.hpp`, etc.

@j-rivero

jslee02:
It sounds more reasonable renaming the headers rather than dropping all the header file extensions. I just renamed `Constraint` to `ConstraintBase`. In most case, the classes those names are same with the namespaces are interface classes. I'd like to use `-Base` for those classes.

You're right, Windows is also case-insensitive system.

j-rivero:
+1 to Steve proposal and final solution. Headers without an extension seems out to any convention to me.

jslee02:
Thanks for the comments!

PR #304, opened 2015-01-20
Fix install directories

jslee02:
This pull request fixes the install directories introduced by #302. By #302, all the headers were installed in single directories, e.g., in `/usr/local/include/dart/`, rather than take there own directories according to the namespaces. This is the reason that `dart 4.3.0` package was not able to be compiled in [the PPA](https://launchpad.net/~dartsim/+archive/ubuntu/ppa/+build/6732642).


jslee02:
+1

PR #315, opened 2015-01-24
Add in gtest source to the repo

### LINKED ISSUE #312: Add in gtest source to the build
It is [not recommended by google](https://code.google.com/p/googletest/wiki/FAQ#Why_is_it_not_recommended_to_install_a_pre-compiled_copy_of_Goog) to install a pre-compiled copy of gtest, since it needs to use the same compiler flags as DART code.

It seems fine on Ubuntu since Ubuntu provides header only gtest package, but homebrew just doesn't even recommend to install gtest header from them (gtest is on their [blacklist](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/blacklist.rb#L44-L47)). We should add in gtest source like [Gazebo did](https://bitbucket.org/osrf/gazebo/src/d504d2050c0f0058a2606e836871bd80c44b6f45/test/gtest/?at=default).

Thanks @scpeters !

jslee02:
This pull request adds gtest source to this repo to resolve #312.


jslee02:
+1

PR #319, opened 2015-01-27
Frame class and auto-updating for forward kinematics

mxgrey:
This pull request includes two significant new features: a Frame subclass (as well as an Entity subclass) and smart automatic updating for forward kinematics.

## Frame
The Frame class allows a greater amount of expression when defining and requesting data. For example, you can now request positions, velocities, and accelerations with request to an arbitrary reference frame by specifying a Frame class as an argument for various functions [(as an example, see the getLinearVelocity() function)](https://github.com/dartsim/dart/blob/6066b7de6cd2e3a0163671f07d1a427db9f5ba0c/dart/dynamics/Frame.cpp#L165).

#### Entity
The Entity class is simply a base class for anything that exists within a Frame. At the moment, that only includes Frames (because Frames exist within Frames), but in the future it can include things like a Point class (useful for expressing center of mass or collision points), an Axis class, a Transform class, and perhaps many other things.

## Auto-Updating
The other -- and possibly more significant -- feature of this pull request is auto-updating for forward kinematics. From the user point of view, the impact of this feature is simply that there is no longer ever a need to call Skeleton::computeForwardKinematics(). More specifically, forward kinematics will always be computed when it's needed and will *only* perform the computations that are necessary for what the user requests. This works by performing some bookkeeping internally with dirty flags whenever a position, velocity, or acceleration gets set, either internally or by the user.

On one hand, this results in some overhead due to the extra effort of bookkeeping, but on the other hand we have much greater code safety, and in some cases performance can be dramatically improved with the auto-updating (for example, the computeForwardKinematics() function was inefficient when only one portion of the BodyNodes needed to be updated rather than the entire Skeleton, which is common when performing inverse kinematics on a limb or on some subsection of a Skeleton).

## Speed Test
To determine whether the bookkeeping overhead is worth the auto-updating, I wrote the speedTest app (note, this is not a unit test because it can take a while to run). When comparing this branch with the [grey/speed_test branch](https://github.com/dartsim/dart/tree/grey/speed_test) (which is the same as the current master, but with the speed test added), the difference in performance between the two branches was not appreciable. I ran the speedTest on each branch ten times in "ideal" conditions (I freshly restarted my PC and ran nothing but the terminal) and neither branch was consistently faster than the other. Ultimately, the worst performance of this auto-updating branch was less than 1% slower than the worst performance of the speed_test branch.

By default, the speedTest app is testing the speed of dynamic simulation, but it also includes a mode for testing the speed of pure forward kinematics. Comparing the results in that mode, I consistently found that the speed_test branch tends to have *slightly* better performance for computing all three (position, velocity, and acceleration); it seems to be a tossup when computing position and velocity (without computing acceleration); and then this auto-updating branch tends to win by a significant margin when computing only position (likely because the amount of bookkeeping overhead is minimized when only positions are being computed).

## Caveats
It's important to note that the auto-updating comes with one important caveat for all DART developers now and in the future: there are certain data members that **must not** be modified or even accessed outside of the appropriate functions, even for internal use. Some of these include BodyNode::mWorldTransform, BodyNode::mVelocity, BodyNode::mAcceleration, and even others that might not be expected like Joint::mJacobian, Joint::mJacobianDeriv, BodyNode::mArtInertia. There are too many to list right here, but they should all be noted in the header comments. The reason these members should not be used directly is because they are carefully monitored by the auto-updating system, so anything that writes to them when it shouldn't will invalidate results, and anything that reads from them when it shouldn't might end up using outdated data.

This raises a question that I'd like to pose: Should we append something to the names of these "untouchable" data members similar to how the [Hungarian Notation](http://en.wikipedia.org/wiki/Hungarian_notation) encodes information in variable names? Normally I would not advocate such a naming convention, but it might help future (and even current) developers if there were a clear way of immediately knowing that they should not use a variable directly. I don't really have any good suggestions for what the code should be, though. Maybe append "_dnu" (for "do not use") to the ends of the variable names?

## TODO
Mark functions that are now deprecated (there are many…).

### Final Comment
The changes that were necessary for the auto-updating to work fast and reliably were pretty exhaustive. Don't hesitate to ask if you find any changes peculiar, or if you're wondering why some were necessary. I had good reasons for each and every change, and I don't mind explaining those reasons.


mxgrey:
For anyone wondering why Travis-CI was failing so many times, it turns out that (for reasons that are beyond my knowledge) Travis-CI seems to autogenerate a directory in the apps folder called "CMakeFiles". This directory was getting picked up by my new apps/CMakeLists.txt script as though it were a directory for an app (because apps/CMakeLists.txt now automatically parses the directories in the apps folder, so we no longer need to maintain a list of apps), but it was not valid because it did not contain a CMakeLists.txt, so Travis-CI was pooping out.

mxgrey:
Update: I've modified my earlier solution so that now *any* directory in the apps folder without a CMakeLists.txt will just get ignored. This should be a more robust and elegant solution.

dart/dynamics/Entity.h
```diff
@@ -0,0 +1,168 @@
...
+  bool dependsOn(const Frame* _someFrame) const;
+
+  /// Returns true if this Entity is set to be quiet.
+  ///
+  /// A quiet entity is unknown to its parent Frame. It will not be tracked by
+  /// its parent; it will not receive notifications from its parent, and it will
+  /// not be rendered. The advantage to a quiet Entity is that it has less
+  /// overhead when constructing and deconstructing, which makes it more
+  /// suitable for temporary objects.
+  bool isQuiet() const;
```
jslee02: Couldn't find any definition for this function in the source file nor use.

dart/dynamics/Entity.h
```diff
@@ -0,0 +1,168 @@
...
+                    const Eigen::Vector4d& _color = Eigen::Vector4d::Ones(),
+                    bool _useDefaultColor = true, int _depth = 0) const;
+
+  /// Get the parent (reference) frame of this Entity
+  Frame* getParentFrame();
+
+  /// Get the parent (reference) frame of this Entity
+  const Frame* getParentFrame() const;
+
+  /// True iff this Entity depends on (i.e. kinematically descends from)
```
jslee02: iff -> if

dart/dynamics/Frame.cpp
```diff
@@ -0,0 +1,478 @@
...
+}
+
+//==============================================================================
+Eigen::Isometry3d Frame::getTransform(const Frame* _withRespectTo) const
+{
+  if(_withRespectTo->isWorld())
+    return getWorldTransform();
+  else if(_withRespectTo == mParentFrame)
+    return getRelativeTransform();
+
```
jslee02: I'd like to add
```cpp
else if (_withRespectTo == this)
  return Eigen::Isometry3d::Identity();
```

dart/dynamics/Frame.cpp
```diff
@@ -0,0 +1,478 @@
...
+    return mAcceleration;
+
+  if(mNeedAccelerationUpdate)
+  {
+    mAcceleration = math::AdInvT(getRelativeTransform(),
+                                 getParentFrame()->getSpatialAcceleration())
+        + getPrimaryRelativeAcceleration()
+        + getPartialAcceleration();
+//        + getRelativeSpatialAcceleration()
+//        + math::ad(getSpatialVelocity(), getRelativeSpatialVelocity());
```
jslee02: Do we need to keep this commented code?

dart/dynamics/Frame.cpp
```diff
@@ -0,0 +1,478 @@
...
+}
+
+//==============================================================================
+template <typename T>
+static std::set<const T*> convertToConstSet(const std::set<T*>& _set)
+{
+  std::set<const T*> const_set;
+  typename std::set<T*>::const_iterator it=_set.begin(), end=_set.end();
+  for( ; it != end; ++it)
+    const_set.insert(*it);
```
jslee02: Since we support C++11, I advocate to use range-based for loop like:
```cpp
  for (const auto& element : _set)
    const_set.insert(element);
```
It makes more readable.

dart/dynamics/Entity.h
mxgrey: Hah, I guess there's no linking error because it isn't explicitly used yet.

dart/dynamics/Entity.h
mxgrey: I use "iff" to mean "if and only if". Is that too confusing? Would it be better to write out "if and only if"?

dart/dynamics/Frame.cpp
mxgrey: Good call.

dart/dynamics/Frame.cpp
mxgrey: Good call; I do this out of habit since I'm accustomed to C++11 yet.

scpeters:
Regarding the travis-ci failures, I noticed that you are using an [in-source build](https://github.com/dartsim/dart/blob/master/.travis.yml#L21)

~~~
cmake .
~~~

It is [recommended to use out of source builds](http://voices.canonical.com/jussi.pakkanen/2013/04/16/why-you-should-consider-using-separate-build-directories/), as is done by [simbody](https://github.com/simbody/simbody/blob/master/.travis.yml#L52-L57)

~~~
mkdir build
cd build
cmake ..
~~~

dart/dynamics/Entity.h
jslee02: Oh, I see. Never mind, I though it was typo.

jslee02:
Fixed, thanks @scpeters !

dart/dynamics/BodyNode.cpp
```diff
@@ -888,34 +960,16 @@ void BodyNode::init(Skeleton* _skeleton)
...
 //==============================================================================
-void BodyNode::draw(renderer::RenderInterface* _ri,
-                    const Eigen::Vector4d& _color,
-                    bool _useDefaultColor,
-                    int _depth) const
+void BodyNode::processNewEntity(Entity* _newChildEntity)
 {
-  if (_ri == NULL)
+  if(find(mChildBodyNodes.begin(), mChildBodyNodes.end(), _newChildEntity) !=
+     mChildBodyNodes.end())
```
jslee02: Do we need to check the duplicity here? It's already checked in `BodyNode::addChildBodyNode(~)` and also this function does not add the new entity to the list of child body nodes.

dart/dynamics/Frame.h
```diff
@@ -0,0 +1,297 @@
...
+#include "dart/math/MathTypes.h"
+
+namespace dart {
+namespace dynamics {
+
+/// The Frame class serves as the backbone of DART's kinematic tree structure.
+///
+/// Frame inherits Entity, so it exists within a reference Frame. This class
+/// keeps track of both its local (relative) and global (world) transforms,
+/// velocities, and accelerations. It also notifies every child Entity when
```
jslee02: I'm not sure whether you finished the sentence here.

dart/dynamics/Frame.h
```diff
@@ -0,0 +1,297 @@
...
+  /// Get the spatial velocity of this Frame relative to some other Frame. It
+  /// can be expressed in the coordinates of any Frame.
+  Eigen::Vector6d getSpatialVelocity(const Frame* _relativeTo,
+                                     const Frame* _inCoordinatesOf) const;
+
+  /// Get the linear portion of classical velocity of this Frame relative to
+  /// some other Frame. It can be expressed in the coordinates of any Frame.
+  Eigen::Vector3d getLinearVelocity(
+      const Frame* _relativeTo = Frame::World(),
+      const Frame* _inCoordinatesOf = Frame::World()) const;
```
jslee02: If I'm correct, we are willing to replace all the kinematics related member functions of `BodyNode` with `Frame`'s. This function should be able to return a linear velocity at an arbitrary point on the body node like `BodyNode::getBodyLinearVelocity(const Eigen::Vector3d& _offset)` does.

jslee02:
I figured out that this soft body dynamics doesn't work with this pull request. The main reason is that the inherited virtual functions of `SoftBodyNode` are not calling in the new dynamics update routine.

dart/dynamics/Frame.h
mxgrey: I was actually planning on offering this functionality with the Point class that I've mentioned in the past. It would look something like this:

Eigen::Vector3d linearVelocity = Point(Eigen::Vector3d(1,0,0), referenceFrame).getLinearVelocity();

I could also change the functions to work the way you've proposed, but there are two reasons I didn't: (1) The number of arguments starts to look kind of crazy, and (2) If we're always accounting for an offset, that means there are some extra computations that we are always doing.

I suppose both of those problems can be solved by offering two different versions of the function: one with an offset argument (it would be the first argument in the function) and one without an offset argument. Does that seem like a decent compromise?

dart/dynamics/BodyNode.cpp
mxgrey: The purpose of this function is probably not obvious, so I'll add an explanation in the comments of the code. The point of this function is to sort out the arbitrary Entities that get added from the BodyNode children that are getting added.

This might sound silly, but there are some performance optimizations that can be achieved in (for example) the update notifications by sorting out the BodyNode child Entities from the other arbitrary types of child Entities. It has to do with compile-time resolution of virtual functions. If I call notifyTransformUpdate() on an arbitrary Entity pointer, a runtime vtable lookup will be necessary, which means the compiler can't do any inline optimizations. On the other hand, if I call notifyTransformUpdate() on a BodyNode pointer, it can resolve the vtable lookup at compile time, which allows the compiler to inline.

dart/dynamics/Frame.h
mxgrey: Woops, I must have gotten distracted mid-sentence.

mxgrey:
Ah, I hadn't thought much about the SoftBodyNodes. I'm sure I can get those functional again.

dart/dynamics/Frame.h
jslee02: Offering two different versions of the function sounds good!

dart/dynamics/BodyNode.cpp
jslee02: I can see storing Entities and BodyNodes has benefits in the performance. 

What I meant was that the check looks redundant. `BodyNode::processNewEntity(~)` seems to be called by only `addChildBodyNode(~)` and the duplicity check is already done in the function. Is there any possibility that `BodyNode::processNewEntity(~)` can be called in other functions?

mxgrey: Actually, BodyNode::processNewEntity(~) gets called by Entity::changeParentFrame(~). This means that it will get called for *any* Entity that gets added as a child of the BodyNode.

For example, suppose I make SimpleFrame instance, and I want its parent Frame to be a BodyNode. We'll call Entity::changeParentFrame(~) on the SimpleFrame instance. This will add the SimpleFrame to the mChildEntities vector of the BodyNode. However, the mChildEntities vector contains child BodyNodes as well as all other types of Entities that are children of the BodyNode. Entity::changeParentFrame(~) does not know that the BodyNodes want to sort the BodyNodes out from the rest of the Entities, so we have it call processNewEntity(~). If we were adding a child BodyNode, then it would already exist in mChildBodyNodes, because it would have been added to that vector when we called BodyNode::addChildBodyNode(~). So we check if it's already in there. If it *is* in there, then we quit, because it's already sorted out. If it's *not* in there, then we check if it's already been added to the mNonBodyNodeEntities vector, which is where we want it to be. If it's not already added, then we add it to that vector.

This would probably be an easier thing to explain if I could draw a diagram.

However, I'm glad that you brought my attention to these vectors, because I just realized that we do not yet correctly process child Entities that are being removed from BodyNodes, so I'm going to go take care of that now.

karenliu:
+1

mxgrey:
I'm still working on incorporating the auto-updating for SoftBodyNodes. I hope to have that finished tonight.

mxgrey:
With this latest commit, the SoftBodyNodes appear to be working within the auto-updating framework. At the very least, running the softBodies app in this branch seems to be identical to running it in master.

mxgrey:
This might be a good time for me to pose this question again:

What do people think of altering the names of any internal variables that **must not** be used outside of their set/get functions? The auto-updating features result in really good robustness to user error, but at the cost of greater sensitivity to developer error, because if a developer haphazardly uses an auto-updating variable outside of its set/get/update functions, then it will almost certainly produce wrong results.

To prevent that, I would propose adopting some naming convention that makes it clear which variables are safe to use internally and which aren't. I think the best thing I've come up with for the moment would be a prefix of "_au" where "au" stands for "auto-update" and the underscore makes sure that the variable names only show up at the very end of an autocomplete list (assuming your autocomplete sorts variable names alphabetically). For example, since mWorldTransform is an auto-updated variable, we would change its name to _auWorldTransform. This way, it can be immediately clear to all current and future developers whether a variable is safe to use internally (because it would start with an "m") or whether it can only be interacted with through set/get functions (because it would start with an "_au").

What does everyone think?

mxgrey:
I'm working on velocity and acceleration functions for offsets (fixed points) in Frames, and I was wondering if there would be any value to a getSpatialVelocity(offset) or getSpatialAcceleration(offset) function. Would anyone have a use for that, or is it sufficient to just make getLinearVelocity(offset) and getLinearAcceleration(offset) functions?

jslee02:
For now, I don't think offset parameter is required for spatial velocity and acceleration functions.

Regarding the special naming for auto-updating variables, I have have questions. All the auto-updating variables are modified by only the declared classes not by inheriting classes? If so, how about making them as private member variables?

mxgrey:
> All the auto-updating variables are modified by only the declared classes not by inheriting classes?

Not always. For example, Joint::mLocalTransform is an auto-updated member, but it needs to be set by the updateTransform() function which is virtual and gets overridden by each joint type. So Joint::mLocalTransform and almost all of the other Joint members cannot be made private.

Also, privatizing them doesn't make them safe from misuse within the class that they belong to, which is my biggest concern. I think it's natural for a developer to assume that they have the freedom to directly use any member variables of the class that they're working on. Also, there are many member variables that *can* be used directly with no negative consequences. So I think a naming convention would help to differentiate which variables can and cannot be used directly. That should help prevent a lot of confusion and second-guessing while developers are working on a given class.

jslee02:
If auto-updating variables should be interacted through get/set functions, then it is necessary to make sure which functions are allowed to modify the variables. If not, it is still prone to be misused. It seem like get function is the allowed function, right? For example, auWorldTransform is must be modified by only getWorldTransform(). 

Also, underscore is preserved for parameter name so I would like to use "au" for the auto-updating variables.

mxgrey:
> It seem like get function is the allowed function, right?

In some cases, but not always. For example, mArticulatedInertia is now an auto-updated variable, but it can't be updated in a single "get" function, because it's updated in by a Skeleton class function that iterates through all the BodyNodes. Within that Skeleton function, the BodyNode::mArticulatedInertia members need to be directly accessed and changed. So we do need to be aware of which specific functions can alter the values of which specific data members. I already have that worked out at the moment, but maybe I should put it into documentation somewhere for posterity?

> Also, underscore is preserved for parameter name so I would like to use "au" for the auto-updating variables.

Yeah, this is something that I kind of wanted to talk about (although I was going to save it for later). I've never quite understood why parameter names are given underscores. I would understand it if member variables didn't have any designator, because an underscore in parameter names would ensure that there is no conflict between parameter names and member variable names, but all member variables start with an "m", so that should already prevent any conflicts. In most conventions that I've seen, a leading underscore is used to indicate a protected/private member variable, so I find it awkward that they're being used for parameter names. I also feel like it hurts auto-complete readability a little bit (when seeing a preview of the function declaration in an IDE), but that might just be a personal issue.

In any case, the main reason that I'd rather not start them with just "au" is that I think most auto-completers will list suggestions in alphabetical order, so things that start with "au" would be at the very top of the list, but we're talking about variables that we want to make sure people *don't* use in general, so it would seem better to me if they were hidden away at the bottom. Underscore is usually sorted to the bottom of an alphabetical list, so leading with an underscore would ensure that those variables are relatively hidden.

mxgrey:
In order to create the Frame::getLinearAcceleration(offset) function, I actually found it easiest and most efficient to also create a Frame::getSpatialVelocity(offset) and Frame::getSpatialAcceleration(offset) function. So those exist now, regardless of whether or not an end-user has need for them.

mxgrey:
There appear to be either FCL or boost-related build issues on Travis-CI. In two of the build attempts, it failed to find a header named "boost/numeric/conversion/detail/preprocessed/numeric_cast_traits_common.hpp". Unfortunately, I don't know enough about how Travis-CI works to diagnose this problem, but I can try looking into it.

jslee02:
> There appear to be either FCL or boost-related build issues on Travis-CI. In two of the build attempts, it failed to find a header named "boost/numeric/conversion/detail/preprocessed/numeric_cast_traits_common.hpp". Unfortunately, I don't know enough about how Travis-CI works to diagnose this problem, but I can try looking into it.

I'm not sure why but it will be gone if retry to build.

mxgrey:
The latest commit adds Jacobian functions that match the new velocity and acceleration functions. They allow the user to specify a frame of reference as well as offsets, and they clearly distinguish between spatial and classical vector terms.

The only issue is that Appveyor has been crashing when it attempts to build the latest changes for Win32. There have been no problems for Win64 or for any of the build tests on Travis-CI. Moreover, I have successfully built this branch in Win32 on two different Windows machines with no issues. This leads me to believe that Win32 compilation on Appveyor is simply broken (maybe it needs a system update?) and should therefore be ignored for the time being.

If anyone else has a Windows development environment set up, it would be very helpful if you could attempt to build this branch and see if it gives you any issues. The issue on Appveyor is that it attempts to compile BodyNode.cpp but just hangs there for about five minutes and then quits with an exit code of 1. It does not give any description of why it failed.

jslee02:
> The only issue is that Appveyor has been crashing when it attempts to build the latest changes for Win32. There have been no problems for Win64 or for any of the build tests on Travis-CI. Moreover, I have successfully built this branch in Win32 on two different Windows machines with no issues. This leads me to believe that Win32 compilation on Appveyor is simply broken (maybe it needs a system update?) and should therefore be ignored for the time being.

I did some test on this issue and [it seems to be resolved](https://ci.appveyor.com/project/jslee02/dart/build/639). What I did is [replacing all the `int` in `BodyNode` to `size_t`](https://github.com/dartsim/dart/commit/abcd0f978175d4cc8b8f66281987cab60f24e4c8).

mxgrey:
Wow! Out of curiosity, how did you know to do that? Is this an issue that's been encountered before, or did you have some way of debugging it that I'm unaware of?

jslee02:
I just got a hint from [here](https://ci.appveyor.com/project/jslee02/dart/build/640/job/vxw5q3mj1fubh5i3/messages). It just a warning message about `int` and `size_t`, but it's from out side of our code and causes the error. So I thought it must have something to with BodyNode's `int` and `size_t`.

Actually, [I first wonder that there is a way to get more detailed build error from AppVeyor](http://help.appveyor.com/discussions/problems/1600-build-failure-on-win32-with) but seems not possible (or my question is not clear).

mxgrey:
Thanks for figuring this out! I had been ignoring that message because I couldn't imagine how a signed/unsigned conflict would cause Win32 to crash. I still don't understand how that's possible, but I'm certainly glad the issue is resolved.

jslee02:
I also don't understand how but yeah it's resolved. :)

At some point, we need to be more restrict to the data types DART uses.

mxgrey:
Agreed. I generally prefer using size_t since that's what gets used by STL, and it makes sense for ordered lists that start from 0 and go up. But we have the unfortunate conflict with Eigen which uses int, presumably so that it can have a -1 represent dynamically sized matrices.

PR #322, opened 2015-01-28
Added convenience functions to help with setting FreeJoint, EulerJoint, and BallJoint positions

mxgrey:
I found that setting FreeJoint and other joint types that have rotations can be confusing or can require more lines of code than it ought to, so I made some functions to help users convert transform and rotation types into joint position vectors that they can pass into Joint::setPositions(VectorXd).

### DIFF
--- a/dart/dynamics/BallJoint.h
+++ b/dart/dynamics/BallJoint.h
@@ -66,6 +66,12 @@ public:
     return math::logMap(_rotation);
   }
 
+  /// Convert a BallJoint-style position vector into a transform
+  static Eigen::Isometry3d convertToTransform(const Eigen::Vector3d& _positions);
+
+  /// Convert a BallJoint-style position vector into a rotation matrix
+  static Eigen::Matrix3d convertToRotation(const Eigen::Vector3d& _positions);
+
 protected:
   // Documentation inherited
   virtual void integratePositions(double _dt);
--- a/dart/dynamics/BallJoint.h
+++ b/dart/dynamics/BallJoint.h
@@ -54,6 +54,18 @@ public:
   /// Destructor
   virtual ~BallJoint();
 
+  /// Convert a rotation into a 3D vector that can be used to set the positions
+  /// of a BallJoint. The positions returned by this function will result in a
+  /// relative transform of
+  /// getTransformFromParentBodyNode() * _rotation * getTransformFromChildBodyNode().inverse()
+  /// between the parent BodyNode and the child BodyNode frames when applied to
+  /// a BallJoint.
+  template <typename RotationType>
+  static Eigen::Vector3d convertToPositions(const RotationType& _rotation)
+  {
+    return math::logMap(_rotation);
+  }
+
 protected:
   // Documentation inherited
   virtual void integratePositions(double _dt);
--- a/dart/dynamics/EulerJoint.h
+++ b/dart/dynamics/EulerJoint.h
@@ -68,6 +68,38 @@ public:
   /// renmaed according to the axis order.
   void setAxisOrder(AxisOrder _order, bool _renameDofs = true);
 
+  /// Convert a rotation into a 3D vector that can be used to set the positions
+  /// of an EulerJoint with the specified AxisOrder. The positions returned by
+  /// this function will result in a relative transform of
+  /// getTransformFromParentBodyNode() * _rotation * getTransformFromChildBodyNode().inverse()
+  /// between the parent BodyNode and the child BodyNode frames when applied to
+  /// an EulerJoint with the correct axis ordering.
+  template <typename RotationType>
+  static Eigen::Vector3d convertToPositions(
+      const RotationType& _rotation, AxisOrder _ordering)
+  {
+    switch(_ordering)
+    {
+      case AO_XYZ:
+        return math::matrixToEulerXYZ(_rotation);
+      case AO_ZYX:
+        return math::matrixToEulerZYX(_rotation);
+      default:
+        dtwarn << "[EulerJoint::convertToPositions] Unsupported AxisOrder ("
+               << _ordering << "), returning a zero vector\n";
+        return Eigen::Vector3d::Zero();
+    }
+  }
+
+  /// This is a version of EulerJoint::convertToPositions(const RotationType&,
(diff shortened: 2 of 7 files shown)


dart/dynamics/BallJoint.h
```diff
@@ -54,6 +54,18 @@ class BallJoint : public MultiDofJoint<3>
...
   virtual ~BallJoint();
 
+  /// Convert a rotation into a 3D vector that can be used to set the positions
+  /// of a BallJoint. The positions returned by this function will result in a
+  /// relative transform of
+  /// getTransformFromParentBodyNode() * _rotation * getTransformFromChildBodyNode().inverse()
+  /// between the parent BodyNode and the child BodyNode frames when applied to
+  /// a BallJoint.
+  template <typename RotationType>
+  static Eigen::Vector3d convertToPositions(const RotationType& _rotation)
```
jslee02: Why this function is made as a template function?

dart/dynamics/EulerJoint.h
```diff
@@ -71,6 +71,38 @@ class EulerJoint : public MultiDofJoint<3>
...
+               << _ordering << "), returning a zero vector\n";
+        return Eigen::Vector3d::Zero();
+    }
+  }
+
+  /// This is a version of EulerJoint::convertToPositions(const RotationType&,
+  /// AxisOrder) which will use the AxisOrder belonging to the joint instance
+  /// that it gets called on.
+  template <typename RotationType>
+  Eigen::Vector3d convertToPositions(const RotationType& _rotation)
```
jslee02: Why not static member function?

jslee02:
How about to create inverse conversion functions like `FreeJoint::convToTransfom(~)` and `BallJoint::convToRotation(~)`, and use all the conversion functions inside the joint classes for consistency.

For example,
```cpp
void FreeJoint::integratePositions(double _dt)
{
  mQ.linear()      = mQ.linear() * math::expMapRot(mJacobian.topRows<3>()
                                                   * mVelocities * _dt);
  mQ.translation() = mQ.translation() + mVelocities.tail<3>() * _dt;

  //mPositions.head<3>() = math::logMap(mQ.linear());
  //mPositions.tail<3>() = mQ.translation();
  mPosition = convertToTransfom(mQ);
}

```

dart/dynamics/BallJoint.h
mxgrey: It's templated because Eigen uses a variety of ways to express rotations, and using a templated argument ensures that all of them can be covered by one function. I modelled it after the Isometry3d::rotate(~) function.

It would probably still work if it used a Matrix3d argument, since Eigen is pretty smart about converting its types, but the template usage here doesn't really hurt anything, because you don't need to use the template explicitly. If you look at [an example of usage](https://github.com/dartsim/dart/blob/a3cca9d907ff2fe8e482e2864f011f9b1ee3837e/unittests/testJoints.cpp#L576) you'll see that C++ can infer the template argument so you don't even need to treat it like a templated function.

dart/dynamics/EulerJoint.h
mxgrey: This version of the function can't be static because it uses mAxisOrder. This way you can call convertToPositions(~) directly on an instance of EulerJoint instead of needing to worry about passing in the AxisOrder of your joint.

mxgrey:
Making inverse versions and using them all inside the update functions makes perfect sense.

mxgrey:
I've added a convertToTransform(~) function for the three joint types, as well as a convertToRotation(~) function for BallJoint and EulerJoint. They're also being used in the update functions for those joints.

dart/dynamics/FreeJoint.cpp
```diff
@@ -58,14 +58,32 @@ FreeJoint::~FreeJoint()
 }
 
 //==============================================================================
+Eigen::Vector6d FreeJoint::convertToPositions(const Eigen::Isometry3d& _tf)
+{
+  Eigen::Vector6d x;
+  x.head<3>() = math::logMap(_tf.rotation());
```
jslee02: I've been using `Eigen::Isometry3d::linear()` to get the rotation part of a transformation. Do you know what's the difference between `linear()` and `rotation()`?

dart/dynamics/BallJoint.h
jslee02: > Eigen uses a variety of ways to express rotations

Do you mean like `Quaternion3d`? Is there another types that express rotations in Eigen?

dart/dynamics/EulerJoint.h
jslee02: Oh, I missed that it uses non-static member variable.

dart/dynamics/BallJoint.h
mxgrey: Right, Quaternion3d is an example. AngleAxisd is another one. I'm not sure if there are more than those three.

jslee02: I see. It seems all the ration expressions can be converted to `Eigen::Matrix3d` so that it works with `math::logMap(Eigen::Matrix3d)`.

dart/dynamics/FreeJoint.cpp
mxgrey: Good question; I haven't given it much thought, I just use rotation() out of habit.

Glancing through the Eigen headers, it looks like the difference is that linear() just returns a reference to the rotational part of the Isometry3d, whereas rotation() enforces orthonormality before returning a copy of the linear part (so it's not a reference).

It sounds like it's a trade off between efficiency -- where linear() is the winner -- versus numerical accuracy -- where rotation() is the winner. Also, if you want to set the linear part directly, you have to use linear().

It might be better to use linear() here since it would be more efficient, and I suspect it's reasonable to expect that the incoming Isometry3d has an okay rotational part.

jslee02: Let's use linear() consistently to see more clear if the inaccuracy causes problem or not.

jslee02:
Changed `rotation()` to `linear()`. As I mentioned, let's use linear() consistently, and let's see if the orthogonality does matter.

If the build tests are happy, then +1.

PR #323, opened 2015-01-28
[Fix #321] Remove unsupported axis orders of EulerJoint

### LINKED ISSUE #321: EulerJoint::AxisOrder enumeration is deceptive
There are currently 4 enumerations for EulerJoint::AxisOrder, but only two of those enumerations appear to be supported in the code. Specifically, EulerJoint::updateLocalTransform(), EulerJoint::updateLocalJacobian(), and EulerJoint::updateLocalJacobianDeriv() only handle AO_XYZ and AO_ZYX in their code.

To me it seems like any axis orderings that are not supported should not be enumerated, or else the user will be left thinking that they can be used. If we extend support to those orderings in the future, I don't think there's any issue with adding them back into the enumeration at that time.

jslee02:
Fixes #321

### DIFF
--- a/dart/dynamics/EulerJoint.cpp
+++ b/dart/dynamics/EulerJoint.cpp
@@ -84,20 +84,10 @@ void EulerJoint::updateDegreeOfFreedomNames()
       affixes.push_back("_y");
       affixes.push_back("_x");
       break;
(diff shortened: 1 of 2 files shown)


karenliu:
+1

PR #324, opened 2015-01-29
added the feature to plot 2D trajectories; see bipedStand for example.

karenliu:


jslee02:
Made changes:
* Change copyright year to 2015
* Delete created graph windows
* Fix segfault from creating graph window when the simulation hasn't started yet

If the build test are happy then +1.

PR #325, opened 2015-01-29
Fixed a small bug in loading WorldFile

karenliu:


jslee02:
I added an unit test for `FileInfoWorld`. It check whether the world's recording and the saved recording are identical. fullbody1 is used to the test.

Also, `FileInfoWorld` deletes `Recording` before creates a new one to prevent memory leak. If `FileInfoWorld` create the first Recording then it will try to delete null pointer, but it's fine because [`delete` doesn't do anything for null pointer](http://www.parashift.com/c++-faq/delete-handles-null.html).

If all the build tests are happy then +1.

PR #331, opened 2015-02-11
Fixed naming issues for Skeletons in World and created regression tests

mxgrey:
This fix addresses two issues:

1) Skeleton names were not getting removed from the name manager in the World when those Skeletons were removed.

2) Skeletons were not being reassigned names when a Skeleton with a repeated name was added to the World

This fixes have been made and a regression test was created to ensure that these issues are fixed and remain fixed.


jslee02:
+1

Issue #333, opened 2015-02-13
Bug in FreeJoint helper


ana-GT:
* File: dart/dynamics/FreeJoint.cpp
* Functions compromised: convertToPositions (line 61) and converToTransform (line 70)
* Problem description: I am generating a set of transformations with rotations around Z with angles in the range  [0,2*PI], increasing at each step. I use these Transformations to set the joint positions of a freeJoint (using the same formula as used in convertToPositions). 

The joint positions I get are wrong for some angles. The joint positions correspond to the intended Transformation rotated 90 around Z. This happens usually after half of the range (either if I use [0,2PI] or [-PI,PI]. An example in which the bug happens is copied at the end of this issue. In this example I use:

Input transform:
```
-0.815214  -0.57916         0  0.437796
 -0.57916  0.815214         0  0.488175
        0         0        -1      1.03
        0         0         0         1
```
Positions obtained using convertToPositions():
```
-0.954925  -2.99295        -0  0.437796  0.488175      1.03
```
Output transform obtained using convertToTransform() (should be the same as input transform):
```
   -0.815214      0.57916 -5.39747e-16     0.437796
     0.57916     0.815214  1.72211e-16     0.488175
 5.39747e-16 -1.72211e-16           -1         1.03
6.95317e-310 1.38338e-322 6.92085e-310 2.12999e-317
```
Notice the flipped signs in x an y columns.

The problem described above is related to convertToPositions. I also noticed that in convertToTransform, there is a line missing: Tf.setIdentity(). Otherwise, the last row of Tf will be filled with garbage (as in the example above).

* Final note: Please, if somebody can fix this quickly I would appreciate it. I spent most of today searching for the bug in  my own code until I realized that the bug was (very possibly and unless I am mistaken) in dart so I would like to continue with my work once the bug is fixed. Thanks!

PS.- Bug test file:
```cpp
#include <dart/math/Geometry.h>
#include <iostream>

Eigen::Vector6d convertToPositions(const Eigen::Isometry3d& _tf);
Eigen::Isometry3d convertToTransform(const Eigen::Vector6d& _positions);

int main( int argc, char* argv[] ) {

  Eigen::Isometry3d Tf;
  Eigen::VectorXd q;
  Eigen::Isometry3d Tf_back;

  Tf.setIdentity();
  Tf.translation() << 0.1, 0.2, 0.3;

  double dang = 10*M_PI/180.0;
  int n = (int) (2*M_PI /dang);

[...]

Issue #88, opened 2013-09-10
Performance test: dart::math::expAngular() vs Eigen::AngleAxis()


jslee02:
DART has own exponential mapping function, dart::math::expAngular(), and Eigen also has Eigen::AngleAxis() which has same functionality. We may need to know which one performs better.

mxgrey:
The speed test I wrote could come in handy here. We could use the same speed test on two branches, one branch using dart::math:expAngular() and the other branch using Eigen::AngleAxisd(). However, the speed test would not check for numerical accuracy; I don't know if that's a concern or not.

mxgrey:
It turns out that Eigen::AngleAxisd is in fact faster and more robust than DART's existing implementation. Pull request #334 replaces the existing implementation with Eigen::AngleAxisd.

jslee02:
I believe this is resolved.

PR #334, opened 2015-02-14
Use Eigen::AngleAxisd for exponential maps

mxgrey:
A simple fix to FreeJoint::convertToTransform(~) plus a regression test.


mxgrey:
The latest change addressed the second part of #333. It seems that DART's built-in logMap function is not robust to imprecision in the rotation matrices that it's given. In the example given in the bug report, the built-in logMap function decided to reflect the rotation that was fed to it.

So instead, I've changed the logMap function to use Eigen::AngleAxisd which is very robust (and which solved the issue brought up in #333). In response to #88 I ran a speedtest to compare DART's built-in implementation to the speed of Eigen::AngleAxisd. Interestingly, the Eigen version was consistently faster. The dynamics speed test had, on average, a 15% reduction in time and the kinematics test had, on average, a 4% reduction in time using Eigen::AngleAxisd.

It seems that Eigen::AngleAxisd is faster and more robust than the existing DART implementation, so it would be a good idea to adopt it with this pull request.

jslee02:
Can you split this into two separate pull requests: Fix FreeJoint::convertToTransform() and Eigen::AngleAxis?

Fix FreeJoint::converToTransform() looks fine, but I'd like to do more tests on Eigen::AngleAxis such as the angle range, the robustness near zero angle.

mxgrey:
The first commit from this pull request is now available in #338 as its own pull request.

jslee02:
I think I messed up this PR and it's too old where the changes are small. Could you recreate clean PR?

PR #343, opened 2015-03-02
Subscriptions for destructions and notifications

mxgrey:
This pull request introduces a subscription concept, aimed primarily at Entities (but extendable to anything that inherits the Subscription class).

The motivation behind this change is that, in a complex framework, it can be difficult (or impossible) to know whether a pointer that you are holding is still valid. If an object is deleted in one part of the code, but another part of the code is holding on to a pointer to that object, there might not be an easy way to communicate that the pointer is now invalid. Additionally, when changes are made to an object in one part of the code, it might be useful to receive a notification about it in another part of the code (similar to the already existing auto-updating features, but more extensible).

To get these features, three new classes are introduced in this pull request: Subscription, Subscriber, and sub_ptr.

Subscription is a base class that is able to send out notifications. Any class that inherits Subscription can be subscribed to. A class that inherits Subscription will *always* send out a destruction notification upon destruction, so all of its Subscribers will be informed when it gets deleted. Inheriting classes also have the ability to send out arbitrary notifications which are simply integers. The inheriting class is expected to enumerate the meanings of its integer notifications.

Subscriber is a base class that facilitates receiving and managing subscriptions. It offers virtual functions that can be overridden in order to customize the handling of incoming notifications.

sub_ptr is a templated "subscribed" smart pointer that works on any class that inherits Subscription. It allows the user to create a pointer to a Subscription-inheriting class, and that pointer will turn to a nullptr when the class it points to is deleted (no matter where or how it happens). sub_ptr will also keep track of the very last notification sent out by the Subscription.

There is no overhead added by the new features in this pull request, except for what the end-users themselves decide to create. If the Subscription model goes unused, it will perform no extra work.

#### Why not reference-counting / shared pointers?

This Subscription model is **not** a memory management tool. It will not help to prevent memory leaks; it only helps to prevent segfaults that could be caused by dereferencing an invalidated pointer. While I do think that DART could benefit greatly from more rigorous memory management, I do not believe that reference-counting or shared pointers are the right solution, because the resources in DART have a very rigid hierarchy of ownership (e.g. Joints are owned by their child BodyNode; BodyNodes are owned by their parent Skeletons; and Skeletons are owned by their World), and reference-counting or shared pointers would undermine that hierarchy. I do have a proposal to help with memory management in DART, but it would be unrelated to this pull request.


mxgrey:
As a point of comparison, sub_ptr is conceptually similar to the std::weak_ptr concept, except sub_ptr is even *weaker* (it does not have ownership of the object at any time, even temporarily), and sub_ptr does not rely at all on a shared_ptr framework.

dart/common/Subscriber.cpp
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright (c) 2014-2015, Georgia Tech Research Corporation
```
jslee02: Change the copyright year to 2015. The beginning year should be the year of the code was initially created, and the ending year should be the year of the release of DART.

dart/common/Subscriber.cpp
```diff
@@ -0,0 +1,110 @@
...
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dart/common/Subscription.h"
+#include "dart/common/Subscriber.h"
+
+namespace dart {
+namespace common {
+
```
jslee02: Add `//==============================================================================`

dart/common/Subscriber.h
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright (c) 2014-2015, Georgia Tech Research Corporation
```
jslee02: Change the copyright year to 2015.

dart/common/Subscription.cpp
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (c) 2014-2015, Georgia Tech Research Corporation
```
jslee02: Change the copyright year to 2015.

dart/common/Subscription.h
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (c) 2014-2015, Georgia Tech Research Corporation
```
jslee02: Change the copyright year to 2015.

dart/common/sub_ptr.h
```diff
@@ -0,0 +1,140 @@
+/*
+ * Copyright (c) 2014-2015, Georgia Tech Research Corporation
```
jslee02: Change the copyright year to 2015.

unittests/testSubscriptions.cpp
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (c) 2014, Georgia Tech Research Corporation
```
jslee02: Change the copyright year to 2015.

dart/common/Subscription.h
```diff
@@ -0,0 +1,80 @@
...
+#define DART_COMMON_SUBSCRIPTION_H_
+
+#include <set>
+
+namespace dart {
+namespace common {
+
+class Subscriber;
+
+class Subscription
```
jslee02: I think `Publisher` matches with `Subscriber` more than `Subscription`. (or `Subject`-`Observer`).

dart/common/Subscription.cpp
```diff
@@ -0,0 +1,93 @@
...
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dart/common/Subscription.h"
+#include "dart/common/Subscriber.h"
+
+namespace dart {
+namespace common {
+
```
jslee02: Add `//==============================================================================`

dart/common/Subscriber.h
```diff
@@ -0,0 +1,89 @@
...
+  virtual void handleDestructionNotification(const Subscription* _subscription);
+
+  /// Add a Subscription for this Subscriber
+  void addSubscription(const Subscription* _subscription);
+
+  /// Remove a Subscription from this Subscriber
+  void removeSubscription(const Subscription* _subscription);
+
+  /// Remove all Subscriptions from this Subscriber
+  void clearSubscriptions();
```
jslee02: How about `removeAllSubscriptions()`? We've been using `removeAll[~]` for this kind of function. For example, `removeAllSkeletons()` and `removeAllConstraints()`.

dart/common/Subscription.h
```diff
@@ -0,0 +1,80 @@
...
+
+  friend class Subscriber;
+
+  /// Destructor will notify all Subscribers that it is destructing
+  virtual ~Subscription();
+
+protected:
+
+  /// Send a notification to all Subscribers
+  void sendNotification(int _notice) const;
```
jslee02: Is there any case that we need to pass additional parameter to `sendNotification()`? For example,
```cpp
void SomeNewSubscription::processExternalData(SomeData* _data)
{
  // Do something without saving the data in this subscription

  // but the subscriber needs the data so we hand it over
  sendNotification(GOT_EXTERNAL_DATA, _data);
}
```
I couldn't find any use case but I'd like to double check it with you.

dart/dynamics/Entity.h
```diff
@@ -63,11 +63,23 @@ class Shape;
 /// may have different policies about how/if their reference frame or name
 /// can be changed. Use the Detachable class to create an Entity whose reference
 /// Frame can be changed arbitrarily.
-class Entity
+class Entity : public common::Subscription
 {
 public:
   friend class Frame;
 
+  enum Notification {
```
jslee02: How about to use `enum class` as
```cpp
enum class Notification : int {
  // items
}
```
for [type safety and scoped enumeration](http://www.cprogramming.com/c++11/c++11-nullptr-strongly-typed-enum-class.html)?

dart/common/Subscription.h
mxgrey: You're right, "Publisher" would be better for distinguishing between the Subscriber and the thing being Subscribed to.

dart/common/Subscription.h
mxgrey: I thought about it, but there are a couple reasons I decided to only offer enumerated notifications:

1. The Subscription concept isn't meant for heavy duty message passing, nor is it meant for robust or general purpose data transfer. There are already many tools for that, like Qt's signals and slots. The notifications are just meant to be a light-weight tool for informing interested parties when they should update.

2. I think an enumerated flag and a pointer to the object that sent it should be sufficient for subscribers to obtain the data that they need from that object. A use case that demands more than that would be better off using a more robust and comprehensive pipeline.

dart/dynamics/Entity.h
```diff
@@ -63,11 +63,23 @@ class Shape;
...
+
+    FRAME_CHANGE_NOTICE = 1,
```
jslee02: I prefer to use tense to represent a name of event (notification), for example, `FRAME_CHANGING` and `FRAME_CHANGED`. `FRAME_CHANGING` tells the subscribers that the frame will be changed and `FRAME_CHANGED` it's already changed.

```cpp
sendNotification(FRAME_CHANGING);

// Change the frame

sendNotification(FRAME_CHANGED);
```

For now, `FRAME_CHANGING` is not necessary but I'd like to use the name `FRAME_CHANGED` instead of `FRAME_CHANGE_NOTICE`.

dart/dynamics/Entity.h
mxgrey: I like the idea of strongly typed enums. How about we do it like this:

```
enum class Notice : int {
  TRANSFORM,
  VELOCITY,
  // ... etc ...
}

sendNotification(Entity::Notice::TRANSFORM);
```

jslee02: You're one is correct. `sendNotification(Entity::TRANSFORM)` will not work since the enum name is missing.

dart/dynamics/Entity.h
mxgrey: FRAME_CHANGED sounds good. The idea of a FRAME_CHANGING notice is interesting... I suppose some agent might want to take action if object that it's tracking is about to change its frame but hasn't done so yet. I'll think about how to incorporate that, and whether there could be any negative consequences.

dart/dynamics/Entity.h
mxgrey: There seems to be a drawback with strongly typed enums: The compiler refuses to implicitly cast them to int, even if it is declared with the size of "int". This means that a type cast is needed to pass an enumerated value into the Publisher::sendNotification(int) function, and a typecast will again be needed to compare that int with an enumerated value.

I try to avoid type casting when possible, because it tends to feel dirty to me. I guess the question now is whether the advantages of strongly typed enumerations outweigh the grossness of casting them over and over again.

dart/common/Subscription.h
jslee02: I got it. This light-weight tool seems sufficient for the current usage.

Do we assume that the number of subscribers will stay in small number? If not, we may need to worry about many of unnecessary function calls. The current design calls `receiveNotification` of all the subscribers regardless that the subscribers are interested in the notification. For this case we might need different design.

Again, this design seems sufficient for now. I'm leaving a node just in case.

dart/dynamics/Entity.h
jslee02: Or we can change the parameter type to the enumeration like:
```cpp
void Subscription::sendNotification(Notification _notice) {}
void Subscriber::receiveNotification(Subscription* _subscription, Notification _notice) {}
```
But this is not extensible in terms of Notification type.

If we need to choose one of the advantages of strongly typed enumerations and avoiding grossness of casting, we should go for the avoiding grossness of casting.

By the way, what do you think about class `Event` I mentioned in other reply. It is composition design while Subscription is inheritance. Generally [composition is more flexible than inheritance], and `Event` can prevent unnecessary notifications of a notice for the subscriber that is not interested in the notice.

mxgrey: Your event concept does look like a good replacement for the general-purpose notification sending. The only thing I'd be concerned about is destruction notifications, which was the primary motivation behind the subscription concept: Inheritance can guarantee that the destruction notification gets sent out when the object is destructed, and it can guarantee that this notification will occur for any class that inherits the Publisher class.

The destruction notification *could* be done through an Event object, but it would require (1) Creating an Event object for it, and (2) Making sure the Event is triggered during destruction (and *only* during destruction). Also, for the templated sub_ptr class to work, there would be a third requirement (3) The destruction Event object is accessed in the same exact way for every class that wants to work with sub_ptr. On the other hand, using inheritance for the specific purpose of destruction notifications would remove all these requirements, and simply involve inheriting the Publisher class.

So I think I would propose a compromise: continue to use inheritance for destruction notifications, but get rid of the current general-purpose notification system in favor of using the Event class. I think this compromise would simultaneously maximize safety, efficiency, and generalizability. If we do this, we might want to change the names of the Subscriber and Publisher classes to be more specific to the fact that they're only tracking destruction.

jslee02: The compromise sounds good. I implemented Signal class in [grey/subscription_signal](https://github.com/dartsim/dart/commit/4bc93dcaf7e19349917ef55f8cb929516effcf72) branch. Event was renamed to Signal because its design is quiet close to Signal/Slot of Qt and Boost.Signals2.

While I was implementing Signal, I also considered Boost.Signals2 as an alternative since don't want to *reinvent the wheel*. Boost.Signals2 is mature library and thread safe. However, I found some comments that Boost.Signals2 shows low performance [[1]](https://testbit.eu/cpp11-signal-system-performance/) [[2]](http://www.cpgf.org/document/callback-benchmark.html). 

Here are some performance test with Publisher, Signal, and Boost.Signals2:
```
[ Number of listeners: 10 (slots), 50k calls]
Publisher                 : 0.00658989
Signal                    : 0.00805116
Boost.Signal              : 0.035176
Boost.Signal (dummy mutex): 0.0199561

[ Number of listeners: 100 (slots), 50k calls]
Publisher                 : 0.0706949
Signal                    : 0.0816331
Boost.Signal              : 0.274847
Boost.Signal (dummy mutex): 0.146624

[ Number of listeners: 1000 (slots), 50k calls]
Publisher                 : 0.822108
Signal                    : 0.843535
Boost.Signal              : 2.96526
Boost.Signal (dummy mutex): 1.62066
```
The unit is second, and the test code can be found [here](https://github.com/dartsim/dart/blob/js/signal_testing/unittests/testSubscriptions.cpp#L237).

Each test invoked 50k times for different number of slots. `Boost.Signal (dummy mutex)` were tested [without thread safety](http://theboostcpplibraries.com/boost.signals2-multithreading).

As for now, it seems Signal looks reasonable solution in efficiency and generalizability. On the other hand, I'm still open to other solutions if it's more proper for DART.

mxgrey: The Signal class looks really good to me. I do plan on doing some multithreading down the road, but I would be using QThreads for that, so I would probably connect Qt signals/slots to the DART signals to facilitate that.

The only question left would be what to name the Subscriber/Publisher classes because I think those names sound too general for something with such a specific purpose.

jslee02: I will create a separate PR for Signal class.

> The only question left would be what to name the Subscriber/Publisher classes because I think those names sound too general for something with such a specific purpose.

Agree. The key function of Publisher is `sendDestructionNotification()` so it could be like DestructionNoticable/DestructionObserver.

jslee02:
Regarding destruction notifications, I'd like to propose smart pointer approach instead of using class inheritance.

I pasted the original motivation to remind:
> The motivation behind this change is that, in a complex framework, it can be difficult (or impossible) to know whether a pointer that you are holding is still valid. If an object is deleted in one part of the code, but another part of the code is holding on to a pointer to that object, there might not be an easy way to communicate that the pointer is now invalid. 

This means we need a way to observe whether the holding object is valid or not. 

Here is more specific example with code:
```cpp
class Skeleton
{
public:
  Skeleton() : mBodyNode(new BodyNode()) {}
  ~Skeleton() { delete mBodyNode; }  // mBodyNode is destructed here
  BodyNode* getBodyNode() { return mBodyNode; }
protected:
  BodyNode* mBodyNode;
};

void bar(BodyNode* _bodyNode)
{
  std::cout << "My name is " << _bodyNode->mName << std::endl;
}

void foo()
{
  Skeleton* skel = new Skeleton();
  BodyNode* body = skel->getBodyNode();
  bar(body);  // okay

  delete skel;  // the object 'body' is pointing is destroyed here, but 'body' doesn't know it
  bar(body);  // segfault!
}
```

There already exist smart pointers in STL (`std::shared_ptr`, `std::weak_ptr`, and `std::unique_ptr`). The proposing new smart pointers, however, fulfill all the requirements. Here I summarized the requirements to solve our problem.

#### Requirements
(a) A Skeleton has BodyNodes pointer and is responsible to the creation and destruction. The ownership shouldn't be shared.
(b) A Skeleton should be able to return BodyNode pointer to others but shouldn't share the ownership.
(c) The returned BodyNode pointer is not allowed to destruct the actually object.
(d) The returned BodyNode pointer should be able to check if the pointer is valid or not.

#### Possible solution
A. `shared_ptr` and `weak_ptr`
Let's assume that Skeleton stores BodyNode pointer as `shared_ptr` and returns it as `weak_ptr`. Then (b), (c), and (d) will be solved. It seems (a) is also solvable if Skeletons never return BodyNode pointer as `share_ptr` but it's not. Here is an example:
```cpp
std::weak_ptr<BodyNode> weakBody = skel->getBodyNode();  // still skel has the ownership with reference count 1

// to do actual work with weakBody it needs to converted share_ptr
std::shared_ptr<BodyNode> shareBody = weakBody.lock();  // now the ownership is shared !!
```

B. `unique_ptr`
If Skeleton store BodyNode pointer with `unique_ptr`, there is no way to return BodyNode pointer holding the ownership. We ought to transfer the ownership or return raw pointer.

C. `sub_ptr` and `ob_ptr` (proposing)
I'd like to introduce [two classes (subject_pointer (a.k.a. sub_ptr) and observer_pointer (a.k.a. ob_ptr) ) ](https://github.com/dartsim/dart/commit/de62fea17f841f2cc0c9671972bef6e3a9ec1531) that take the middle ground between A and B. `sub_ptr` is similar to `unique_ptr` in terms of single ownership, and `ob_ptr` is similar to `weak_ptr` in terms of shared observation.

                                             | shared ownership  | shared observation
------------------------------------- | ------------------------ | --------------------------
`unique_ptr`                          |              X              |                X
`sub_ptr` and `ob_ptr`          |              X              |                O
`shared_ptr` and `weak_ptr` |               O              |                O

Here is a example code that utilizes the new classes:
```cpp
struct BodyNode { std::string mName; }
class Skeleton
{
public:
  Skeleton() : mBodyNode(new BodyNode()) {}
  ~Skeleton() {}  // mBodyNode will be automatically destroyed if assigned
  ob_ptr<BodyNode> getBodyNode() { return mBodyNode; }
protected:
  sub_ptr<BodyNode> mBodyNode;
};

void bar(ob_ptr<BodyNode> _bodyNode)
{
  if (_bodyNode.valid())
    std::cout << "My name is " << _bodyNode->mName << std::endl;
}

void foo()
{
  Skeleton* skel = new Skeleton();
  ob_ptr<BodyNode> body = skel->getBodyNode();
  bar(body);  // okay

  delete skel;  // the object 'body' is pointing is destroyed here, but still 'body' can handle it
  bar(body);  // okay
}
```
Now (a), (b), (c), (d) are all satisfied.

This smart pointer approach can be applied to any class without class inheritance. So it's more general solution while still solving the target problem (to prevent segfaults that could be caused by dereferencing an invalidated pointer).

Please let me know what do you think.

mxgrey:
To me this actually sounds like a more intensive structural change than the inheritance approach.

If we take this approach, we'll need to design a very specialized reference-counting pointer (which could be a challenge when it comes to threading) and then modify every pointer-returning function to return an ob_ptr value (and there are *many* pointer-returning functions). Also, this would not allow the custom handling of object destructions. For example, in my osgDart framework, I have mappings from DART objects to OSG objects, and I want the OSG objects and their mappings to be destroyed as soon as the DART object that they correspond to is destroyed. This is easily done by overriding a virtual function using the Subscriber class.

If I understand correctly, the motivation for avoiding inheritance is to offer greater usability by allowing the method to work on any arbitrary class. But I don't think this motivation is truly met, because this method still requires that a very particular set of rules needs to be followed by any object that wants to benefit from it. So it foregoes the inheritance requirement by imposing a greater number of code management requirements.

We'll mostly want to get destruction notifications for classes that we have complete control over, in which case adding ``: public Publisher`` to the class definition is trivial. And even if we want a third-party class to benefit from it, it's not hard to extend it. Suppose ``Foo`` is a class from a third-party library, then all we have to do is create ``class PublishingFoo : public Foo, public Publisher`` and have it inherit the constructor of Foo. Then the ``PublishingFoo`` class offers everything the Foo class does, plus the destruction notifications. This seems much easier to me than intertwining specialized pointer classes throughout the API.

(I'm only using the terms "Publisher" and "Subscriber" because I still haven't thought of better terms for them, but I definitely want to rename them.)

jslee02:
> I want the OSG objects and their mappings to be destroyed as soon as the DART object that they correspond to is destroyed.

Indeed, if the destruction notification should be sent as soon as the object is destructed, then `ob_ptr` doesn't fit.

I have one concern then. Is it possible that a class can be a publisher and a subscriber at the same time without any difficulties?

mxgrey:
Yep, any class can definitely be both with no problems. In fact, I've already been doing that.

I just thought of one potential issue with the inheritance approach: the diamond problem. But that's easily solved; we just need make sure we always virtually inherit the Publisher or Subscriber class. Since both Publisher and Subscriber only have a default constructor, we don't need to worry about any virtual inheritance complications.

Also, I'm thinking of changing the class names to Subject (for Publisher) and Observer (for Subscriber), because I do like those terms.

jslee02:
Subject will support general notifications also, or just destruction notifications?

mxgrey:
It would just support destruction notifications.

jslee02:
If Subscriber is changed to Observer then sub_ptr would be better with new name pertain to observer. Something like ob_ptr or obs_ptr.

mxgrey:
I actually think sub_ptr still works, because it would be a pointer to a Subject, therefore sub_ptr.

jslee02:
I was confused because it inherits Observer. You're right, it points Subjects so sub_ptr.

PR #345, opened 2015-03-06
Fixing segfault in SoftSdfParser

mxgrey:
The atlasSimbicon app is failing in master because of a segfault in the SoftSdfParser. It tries to call "getTransform()" on a BodyNode that does not yet have a parent joint, and this is guaranteed to result in a segfault, because the parent joint is strictly responsible for the relative transform of its child BodyNode. The parser has been corrected to set the transform-from-parent of the FreeJoint according to the initial transform as read by the SdfParser. However, we might want to consider setting that transform to Identity and using the initial transform to set the initial joint values of the FreeJoint. Either way works, but one might be semantically preferable over the other.

I also modified the warning message for the soft boxes to be a bit more clear.


jslee02:
:+1:

Waiting to pass build tests.

Issue #346, opened 2015-03-12
LineShape for easily creating line visualizations


mxgrey:
I'd like to create a LineShape that allows users to put lines into the Frame/Entity kinematic tree and be visualized. I don't anticipate making lines compatible with dynamic collision checking or handling, because they are 1D shapes (so they are non-physical), but I think it would be fair to use them as visualization shapes.

jslee02:
LineShape as visualization shape make sense but not for collision shape. First, it is not supported by collision checker. Although we can do collision check with LineShape, the dynamics can't be handled because of the zero momenta of inertia along the line's axis. For example, if we apply angular force along the axis of line then the acceleration would be infinite with the zero inertia. To support line dynamics model, new body node type would be required whose angular motion along the axis is ignored.

Anyways, LineShape as visualization shape sounds good. Please add error handling code for the case of attempting to add LineShape as collision shape.

mxgrey:
> Please add error handling code for the case of attempting to add LineShape as collision shape.

Should I just have it spit out a warning, or should I also refuse to allow LineShapes to be added as collision shapes?

jslee02:
Ignoring the attempting to add with a warning would be good.

jslee02:
I have a question. If it is able to collision check with LineShape then do you think it would be helpful for kinematics (or planning)? I have a plan to implement sensors for DART at some point. Then collision checking with ray should be implemented, and line can be supported easily.

mxgrey:
That's a really good point. Line segment collision checking could be used to simulate proximity sensors, or to take point samples of objects to simulate LIDAR. I could easily imagine that being very useful.

Then if we're supporting collision detection with line segments, we might want to once again consider whether it makes any sense to allow lines to be added to BodyNodes as collision shapes. I'm still strongly inclined to say no, because 1D lines are still fundamentally non-physical, and there might end up being edge cases that are difficult to handle.

mxgrey:
By the way, my plan was to implement line segments (I guess I should have been more clear about that). Do we also want to support infinite lines and rays as shapes?

jslee02:
If FCL can support line segment, it would be best to make line segment to work with other collision shapes. I'll take a look at it if I can do.

jslee02:
I think adding new shapes just for visualization is fine as long as we can draw it. So adding them sounds good to me.

mxgrey:
[...]

PR #349, opened 2015-03-13
Implemented LineSegmentShape for visualizing line segments

mxgrey:
Line segments can now be added as visualization shapes for Entities (but not as collision shapes for BodyNodes). Collision checking with LineSegmentShape has **not** been implemented yet.

### DIFF
--- a/dart/dynamics/LineSegmentShape.cpp
+++ b/dart/dynamics/LineSegmentShape.cpp
@@ -138,6 +138,18 @@ void LineSegmentShape::removeVertex(size_t _idx)
 //==============================================================================
 void LineSegmentShape::setVertex(size_t _idx, const Eigen::Vector3d& v)
 {
+  if(_idx >= mVertices.size())
+  {
+    dtwarn << "[LineSegmentShape::setVertex] Attempting to set vertex #" << _idx
+           << ", but ";
+    if(mVertices.size() == 0)
+      dtwarn << "no vertices exist in this LineSegmentShape yet.\n";
+    else
+      dtwarn << "the vertices of this LineSegmentShape only go up to #"
+             << mVertices.size()-1 << ".\n";
+
+    return;
+  }
   mVertices[_idx] = v;
 }
 
--- a/dart/dynamics/LineSegmentShape.cpp
+++ b/dart/dynamics/LineSegmentShape.cpp
@@ -1,0 +1,324 @@
+/*
+ * Copyright (c) 2015, Georgia Tech Research Corporation
+ * All rights reserved.
+ *
+ * Author(s): Michael X. Grey <mxgrey@gatech.edu>
+ *
+ * Georgia Tech Graphics Lab and Humanoid Robotics Lab
+ *
+ * Directed by Prof. C. Karen Liu and Prof. Mike Stilman
+ * <karenliu@cc.gatech.edu> <mstilman@cc.gatech.edu>
+ *
+ * This file is provided under the following "BSD-style" License:
+ *   Redistribution and use in source and binary forms, with or
+ *   without modification, are permitted provided that the following
+ *   conditions are met:
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND
+ *   CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ *   INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ *   MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dart/dynamics/LineSegmentShape.h"
+#include "dart/common/Console.h"
+#include "dart/renderer/RenderInterface.h"
+#include "dart/math/Helpers.h"
+
+namespace dart {
+namespace dynamics {
+
+//==============================================================================
+LineSegmentShape::LineSegmentShape(float _thickness)
+  : Shape(LINE_SEGMENT),
+    mThickness(_thickness),
+    mDummyVertex(Eigen::Vector3d::Zero())
+{
+  computeVolume();
+}
+
+//==============================================================================
+LineSegmentShape::LineSegmentShape(const Eigen::Vector3d& v1,
+                                   const Eigen::Vector3d& v2,
+                                   float _thickness)
+  : Shape(LINE_SEGMENT),
+    mThickness(_thickness)
+{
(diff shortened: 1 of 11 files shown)


mxgrey:
This pull requests addresses issue #346

jslee02:
Why LineSegmentShape has vertices? It wasn't supposed to contain single line segment?

mxgrey:
I designed it to allow any number of vertices with any number of connections. This makes for more efficient data structures when something more than just a simple line segment is desired.

At the same time, the constructor that accepts two "const Eigen::Vector3d&" arguments will produce a simple line segment between those two points with no need for the user to deal with adding vertices or making connections.

mxgrey:
I'm actually tempted to make a constructor or member function with a variadic template that allows the user to pass in arbitrarily many vertices as arguments, and then strings those vertices into a line strip. It would be pretty easy, but variadic templates are so new that I'd be afraid of confusing users.

jslee02:
Definitely this implementation is efficient for multiple line segments especially for rendering. However, I'm not clear how to do collision check with multiple line segments when their are related rigidly. Are the line segments connected like a rough curve, or it's just arbitrary?

Then, how about creating LineSegmentShape for a single line segment, and rename the current one to LineSegmentsShape (or LineSegmentSetShape)?

mxgrey:
I think the way to handle collision checking for the multiple line segments would just be to perform collision checking using each vertex pair as a separate line segment.

I'd rather not have separate shapes classes for one line segment versus multiple line segments, since the one line segment is just a special case of the multiple line segments. I feel that we'd have an unnecessary extra class, and I don't think the API would benefit significantly.

jslee02:
Yeah, the collision detection with multiple line segments would be similar to mesh except for the primitive element is line while mesh is triangle.

I'd like to make FCL to support all the shapes of DART including LineSegmentShape. I wonder that multiple-line-segments shape makes sense as a general purpose of collision shape so that FCL accepts the shape. So far, I've never seen collision detector that support multiple line segment shape.

mxgrey:
It might take a little massaging. By that I mean, instead of a 1 to 1 mapping of Shapes to FCL collision objects, we would create an FCL collision object for each line segment in the shape. I'll have to look into how DART is interfacing with FCL before I can offer a more descriptive suggestion for implementing this.

jslee02:
Thanks, and I can see your point. By the way, making FCL to support LineSegmentShape is not urgent. So you don't need to spend too much time to look into the collision interface for FCL unless you really want to. :)

mxgrey:
It looks like Appveyor had an internet connection issue while attempting to run the test for the pull request.

jslee02:
> It looks like Appveyor had an internet connection issue while attempting to run the test for the pull request.

I triggered [re-build](https://ci.appveyor.com/project/jslee02/dart/build/733).

jslee02:
Regarding the class name again, LineSegmentsShape seems more suitable to me since it can contain multiple line segments. How do you think?

dart/dynamics/LineSegmentShape.cpp
```diff
@@ -0,0 +1,336 @@
...
+{
+  addVertex(v1);
+  addVertex(v2);
+  computeVolume();
+}
+
+//==============================================================================
+void LineSegmentShape::setThickness(float _thickness)
+{
+  mThickness = _thickness;
```
jslee02: assert or warning would help preventing to set non-positive thickness

dart/dynamics/LineSegmentShape.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+//==============================================================================
+float LineSegmentShape::getThickness() const
+{
+  return mThickness;
+}
+
+//==============================================================================
+size_t LineSegmentShape::addVertex(const Eigen::Vector3d& v)
```
jslee02: v --> _v

dart/dynamics/LineSegmentShape.cpp
```diff
@@ -0,0 +1,336 @@
...
+  size_t parent = mVertices.size();
+  if(parent > 0)
+    return addVertex(v, parent-1);
+
+  mVertices.push_back(v);
+  return 0;
+}
+
+//==============================================================================
+size_t LineSegmentShape::addVertex(const Eigen::Vector3d& v, size_t _parent)
```
jslee02: v --> _v

dart/dynamics/LineSegmentShape.cpp
```diff
@@ -0,0 +1,336 @@
...
+    dtwarn << "No vertex will be removed.\n";
+
+    return;
+  }
+
+  mVertices.erase(mVertices.begin()+_idx);
+}
+
+//==============================================================================
+void LineSegmentShape::setVertex(size_t _idx, const Eigen::Vector3d& v)
```
jslee02: v --> _v

dart/dynamics/LineSegmentShape.h
```diff
@@ -0,0 +1,133 @@
...
+/// Note: LineSegmentShape may NOT be used as a collision shape for BodyNodes,
+/// but it may be used for visualization purposes.
+class LineSegmentShape : public Shape
+{
+public:
+  /// Default constructor
+  LineSegmentShape(float _thickness = 1.0);
+
+  /// Constructor for creating a simple line segment that connects two vertices
+  LineSegmentShape(const Eigen::Vector3d& v1, const Eigen::Vector3d& v2,
```
jslee02: v1, v2 --> _v1, _v2

dart/dynamics/LineSegmentShape.h
```diff
@@ -0,0 +1,133 @@
...
+                   float _thickness = 1.0);
+
+  /// Set the line thickness/width for rendering
+  void setThickness(float _thickness);
+
+  /// Get the line thickness/width used for rendering
+  float getThickness() const;
+
+  /// Add a vertex as a child to the last vertex that was added
+  size_t addVertex(const Eigen::Vector3d& v);
```
jslee02: v --> _v

dart/dynamics/LineSegmentShape.h
```diff
@@ -0,0 +1,133 @@
...
+
+  /// Add a vertex as a child to the specified vertex
+  size_t addVertex(const Eigen::Vector3d& v, size_t _parent);
```
jslee02: v --> _v

dart/dynamics/LineSegmentShape.h
```diff
@@ -0,0 +1,133 @@
...
+
+  /// Remove a vertex from the list of vertices. IMPORTANT: Note that this
+  /// alters the indices of all vertices that follow it in the list, which also
+  /// clobbers the validity of the list of connections for all those vertices.
+  /// A safer and more efficient method might be to recycle vertices by moving
+  /// them around with setVertex() and/or altering their connections.
+  void removeVertex(size_t _idx);
+
+  /// Change the location of the specified vertex
+  void setVertex(size_t _idx, const Eigen::Vector3d& v);
```
jslee02: v --> _v

dart/dynamics/LineSegmentShape.h
```diff
@@ -0,0 +1,133 @@
...
+
+/// The LineSegmentShape facilitates creating graphs of line segments. The graph
+/// can consist of a single line segment or many interconnected line segments.
+/// Note: LineSegmentShape may NOT be used as a collision shape for BodyNodes,
+/// but it may be used for visualization purposes.
+class LineSegmentShape : public Shape
+{
+public:
+  /// Default constructor
+  LineSegmentShape(float _thickness = 1.0);
```
jslee02: 1.0 --> 1.0f

dart/dynamics/LineSegmentShape.h
```diff
@@ -0,0 +1,133 @@
...
+
+  /// Constructor for creating a simple line segment that connects two vertices
+  LineSegmentShape(const Eigen::Vector3d& v1, const Eigen::Vector3d& v2,
+                   float _thickness = 1.0);
```
jslee02: 1.0 --> 1.0f

dart/dynamics/LineSegmentShape.h
```diff
@@ -0,0 +1,133 @@
...
+  /// Line thickness for rendering
+  float mThickness;
+
+  /// Vector of vertices
+  std::vector<Eigen::Vector3d> mVertices;
+
+  /// Vector of connections
+  std::vector<Eigen::Vector2i> mConnections;
+
+  /// A dummy vertex that can be returned when an out-of-bounds vertex is requested
```
jslee02: It's over 80 columns.

jslee02:
I added notes. They all are about styles.

mxgrey:
I had been debating between the name "LineSegmentShape" and "LineSegmentsShape". I went with "LineSegmentShape" for a couple reasons:

(1) Semantically, you can think of it as a "line segment shape", because it is a shape for line segments, analogous to how a "pedestrian walkway" is a walkway for pedestrians, or how a "visitor parking deck" is a parking deck for visitors. The words "Line Segment" are being used [like an adjective](http://en.wikipedia.org/wiki/Noun_adjunct) to describe "Shape", and noun adjuncts are traditionally singular (as mentioned in the link).

(2) Having the "sS" right next to each other is very unappealing to me aesthetically.

jslee02:
It sounds reasonable. Thanks for the explanation.

jslee02:
+1

PR #350, opened 2015-03-13
Signal class

jslee02:
This pull request introduces the Signal class that provides signal/slot mechanism.

Basic usage:
```cpp
void foo1() {}
void foo2(int) {}
void foo3(int, float) {}
double foo4() {}
double bar() {}

int main()
{
  Signal<void()> s1;
  Signal<void(int)> s2;
  Signal<void(int, float)> s3;
  Signal<double()> s4;

  Connection c1 = signal1.connect(&foo1);
  Connection c2 = signal2.connect(&foo2);
  Connection c3 = signal3.connect(&foo3);
  Connection c4 = signal4.connect(&foo4);

  s1.raise();  // foo1() called
  s2(10);  // foo2(10) called
  s3(5, 1.5)  // foo3(5, 1.5) called
  double res = s4();  // foo4() called

  c1.disconnect();  

  s1.raise();  // nothing called

  signal4.connect(&bar);
  
  s4();  // foo4() and bar() called
}
```

This Signal class is quiet similar to [Boost.Signals2](http://www.boost.org/doc/libs/1_57_0/doc/html/signals2.html) but is light-weight version and faster than Boost.Signals2.

This Signal is not thread safe. It will be considered when we start to make DART thread safe.


jslee02:
+1

PR #355, opened 2015-03-17
Move glut dependency from dart-core to dart

### LINKED ISSUE #329: OSX build problem related to GLUT
A user had trouble building gazebo ( https://github.com/osrf/homebrew-simulation/issues/36 ) due to a linking problem with GLUT when not building with `BUILD_CORE_ONLY=ON`. Are the GLUT libraries being exported as dependencies in the cmake config or pkg-config files? If they are not needed by downsteam users, we should just link the GLUT libraries with the apps and not the dart library.

As an aside, GLUT is only used by the `dart/gui` and `apps` code, which should only be built when `BUILD_CORE_ONLY=ON`. However, I see that [GLUT is always searched for](https://github.com/dartsim/dart/blob/master/CMakeLists.txt#L128-L140) (even when not `CORE_ONLY`), and it's library and header files are also always used.

jslee02:
Fixes #329.

### DIFF
--- a/dart/gui/LoadGlut.h
+++ b/dart/gui/LoadGlut.h
@@ -40,7 +40,7 @@
 #if WIN32
   #include <cstdlib> // To disable glut::exit() function
   #include <GL/glut.h>
-#elif defined(__linux)
+#elif defined(__linux__)
   #include <GL/glut.h>
 #elif defined(__APPLE__)
   #include <Glut/glut.h>
--- a/dart/gui/LoadGlut.h
+++ b/dart/gui/LoadGlut.h
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2011-2015, Georgia Tech Research Corporation
+ * Copyright (c) 2015, Georgia Tech Research Corporation
  * All rights reserved.
  *
- * Author(s): Sehoon Ha <sehoon.ha@gmail.com>
+ * Author(s): Jeongsoek Lee <jslee02@gmail.com>
  *
  * Georgia Tech Graphics Lab and Humanoid Robotics Lab
  *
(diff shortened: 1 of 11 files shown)


dart/gui/LoadGlut.h
```diff
@@ -0,0 +1,52 @@
...
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef DART_GUI_LOADGLUT_H_
+#define DART_GUI_LOADGLUT_H_
+
+#if WIN32
+  #include <cstdlib> // To disable glut::exit() function
+  #include <GL/glut.h>
+#elif defined(__linux)
```
scpeters: `__linux__` is the [POSIX compliant macro](http://sourceforge.net/p/predef/wiki/OperatingSystems/) as @j-rivero pointed out in https://github.com/simbody/simbody/pull/338

dart/renderer/LoadOpengl.h
```diff
@@ -34,19 +34,21 @@
...
+#ifndef DART_RENDERER_LOADOPENGL_H_
+#define DART_RENDERER_LOADOPENGL_H_
 
 #if WIN32
-#include <cstdlib> // To disable glut::exit() function
-#include <GL/glut.h>
+  #include <windows.h>
+  #include <GL/gl.h>
+  #include <GL/glu.h>
 #elif defined(__linux)
```
scpeters: `__linux__` here too

dart/gui/LoadGlut.h
jslee02: It is replaced by __linux__, thanks!

scpeters:
+1 pending CI result

scpeters:
thanks!

PR #359, opened 2015-03-19
Added alpha channel and Color functions

mxgrey:
I've added an alpha channel for Shape colors. It turns out the whole OpenGL renderer pipeline was completely ready for an alpha channel, it just wasn't being used.

I also added some Color generating functions. Rather than create a Color class, I simply made a namespace called Color in dart/math/Helpers.h and put some functions in that namespace. I decided to inline all the functions so that end users can just check the header to see what actual color codes are being used for each of the color functions.

### DIFF
--- a/apps/simpleFrames/Main.cpp
+++ b/apps/simpleFrames/Main.cpp
@@ -78,10 +78,11 @@ int main(int argc, char* argv[])
   myWorld.addEntity(&A);
 
   SimpleFrame arrow(Frame::World(), "arrow");
-  arrow.addVisualizationShape(new ArrowShape(Eigen::Vector3d(0.1,-0.1, 0.0),
-                                             Eigen::Vector3d(0.1, 0.0, 0.0),
-                                             ArrowShape::Properties(0.002, 1.8),
-                                             Eigen::Vector3d(1.0, 0.5, 0.5)));
+  arrow.addVisualizationShape(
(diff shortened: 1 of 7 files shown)


mxgrey:
Also, I put a usage example in the softBodies app.

jslee02:
I'd like to create actual Color class so we don't need to worry about the validness with color arithmetic operation. Also, Color class can be in common namespace.

mxgrey:
I think being able to do normal vector arithmetic with colors is valuable. I'm not seeing the purpose behind a unique class for them.

jslee02:
Color class will be able to normal vector arithmetic operations but it will clamp when the result is not valid. For example, result of (0, 0, 0) - (1, 0, 0) = (-1, 0, 0) is not valid so it will be clamped as (0, 0, 0).

mxgrey:
I'm kind of worried that could be overbearing. At the end of the day, it seems like it's the user's responsibility to provide sane color values. I could imagine someone wanting to determine their colors programmatically, and having a color value that's temporarily negative (between arithmetic operations). But then if the color class suppresses that, it could mess up the computation.

jslee02:
It's possible. Okay, let's merge this.

jslee02:
There are conflicts with master. Could you merge master branch into this PR first?

mxgrey:
Merged.

PR #369, opened 2015-03-26
Eradicating memory leaks & making classes copy-safe and clonable

mxgrey:
~~This is a premature pull request for review and discussion purposes, not ready to be merged yet~~

### This pull request is ready for review

This branch and pull request are dedicated to removing memory leaks from DART as well introducing the ability to copy properties and create clones of objects. Spawning clones will be useful for making efficient multithreaded applications.

The changes being made here are pervasive, and some will require breaking the API, specifically the adoption of std::shared_ptr for the Shape and Skeleton classes.

Here is a brief list of the conceptual changes:
* BodyNode and Joint classes will all have a Properties struct that can be safely exchanged between different instances, and this struct will be used to standardize the constructors for the different types.
* BodyNode and Joint instances will be fully managed by their Skeleton. BodyNode and Joint instances will only be creatable by the Skeleton class (the current procedure for assembling Skeletons will be deprecated) and they will be destroyed when the Skeleton class destructs
* Shape instances and Skeleton instances will be treated as shared resources, and they will be managed by std::shared_ptr. BodyNodes and Joints are not shared resources because they are worthless/meaningless without being part of a single specific Skeleton.
* It will be possible to clone Worlds and Skeletons.
* Cloning a Skeleton entails creating a new Skeleton instance that is filled with new Joint instances and new BodyNodes instances which copy the properties of the original Skeleton's Joints and BodyNodes. Shape resources will be shared between Skeleton clones to minimize memory usage and copying overhead.
* Cloning a World entails creating a new World instance that is filled with new Skeletons which are clones of the Skeletons of the original World.

#### To be determined:
It would also be good to have a concept of a "State" to complement the "Properties" concept. I am still hashing out the details for how a State class would work, so suggestions are welcome. At the moment, I think the State should belong to the Skeleton, but it's not clear to me exactly how much information it should contain. It could either be just Generalized Position, Velocity, Acceleration, and Force vectors, or it could also include information like external forces.


mxgrey:
The SDF Parsers have undergone a major overhaul. The public API is unchanged (except for providing shared_ptrs instead of raw pointers for Skeletons; we'll be doing the same for Worlds soon).

Considering the extent of changes that have been made, if anyone has an SDF file that they like, I'd appreciate testing it with the new parser. The atlasSimbicon app appears to be working fine, but I'd like to know if there are any edge cases that clobber the changes I've made.

mxgrey:
SkelParser has been overhauled to work with the new Skeleton construction method. The only catch is that joint numbering will not necessarily be consistent between the new parsing and the old parsing, so any programs that use hard-coded joint or dof numbers will need to be mended.

mxgrey:
I've modified the way the renovated SkelParser assembles Skeletons. Now it will make a best-effort attempt to order all Joints in the Skeleton based on how they are ordered in the file that's being parsed. However, if the file places any child Joints before their parent Joints, then the file's order will not be retained. I **think** this will produce an ordering that is exactly the same as the original SkelParser, which means all old code with hard-coded joint indices should still be valid with the new parser, but I cannot guarantee that as of yet.

mxgrey:
At this point, I believe I've implemented everything that I intend to (including cloning Skeletons and Worlds), so it's probably ready for a code review.

The only things that are left to do would be creating tests to make sure that cloning is working as expected, and removing all uses of the deprecated constructors (of which there are a lot).

Also, we still need to come up with a concept for State, which I think should include something like Skeleton::State and World::State. Those would allow the cloning to be more effective, and allow us to transfer information between clones more easily. However, it's not directly related to memory management, so I don't intend to tackle it within this particular pull request.

mxgrey:
Tests have now been written for world cloning (which I think also suffices as a test for Skeleton, BodyNode, and Joint cloning, because the simulation can't possibly work without *all* of those clone functions working).

The test is strongly suggesting that simulating a cloned world will have a 100% numerically deterministic result (the test is comparing simulation results with 0 tolerance and succeeding), which I think is pretty good news.

mxgrey:
Testing a few applications with valgrind, I believe I've now removed all notable memory leaks from DART. However, I have *not* tested memory leaks in any of the OpenGL code, because (1) I hate raw OpenGL and (2) running it through valgrind is so slow that I have to consider it a waste of time.

mxgrey:
I've removed all uses of the deprecated public constructors for BodyNode and Joint types.

The only thing remaining is to decide how to design the State concept, although that might not be completely necessary for this particular pull request.

mxgrey:
The latest commit introduces the BodyNodePtr which can be used to ensure that a specific BodyNode (and by extension, its Skeleton and everything within its Skeleton) does not get deleted. It does this by holding a shared_ptr to the Skeleton that the BodyNode belongs to.

BodyNodePtr can be implicitly cast to and from a raw BodyNode pointer. When the BodyNode that it references is moved between Skeletons, the BodyNodePtr will change its shared_ptr to the new Skeleton.

When a BodyNode is removed from its Skeleton (using ``BodyNode::remove()`` or using ``BodyNode::split()``), if a BodyNodePtr is referencing it, then the new Skeleton will be kept alive even if the shared_ptr returned by ``BodyNode::remove()`` or ``BodyNode::split()`` is not captured.

Put simply, the BodyNodePtr allows users to handle memory management by holding onto BodyNodePtrs instead of holding onto SkeletonPtrs. I considered offering the same thing for Joint and DegreeOfFreedom (where you could keep a Skeleton alive by holding onto a JointPtr), but this is problematic when it comes to changing the parent Joint type of a BodyNode, because that requires deleting the parent Joint and creating a new one.

For a smart BodyNode pointer that is conceptually similar to weak_ptr, you can use ``dart::sub_ptr<BodyNode>``.

Examples of usage can be found in testSkeleton.cpp

mxgrey:
The latest commit has redone the BodyNodePtr so that it is thread safe and has minimal overhead (no longer depends on callbacks).

The BodyNodePtr increments and decrements a reference count in BodyNode that it refers to. Whenever this reference count is above 0, the BodyNode itself will hold a ``shared_ptr<Skeleton>`` to the Skeleton that owns it. If the reference count drops to 0, this ``shared_ptr<Skeleton>`` is released. This reference counting is performed atomically, so it should be thread safe.

The WeakBodyNodePtr was implemented by holding a shared_ptr< weak_ptr<Skeleton> > to the Skeleton that owns the BodyNode. Whenever the BodyNode is moved to a new Skeleton, it will update the weak_ptr<Skeleton> that is held by the shared_ptr< weak_ptr<Skeleton> > of all the WeakBodyNodePtrs that refer to it. This way, the WeakBodyNodePtr can lock() the weak_ptr<Skeleton>, and if the return value of the lock() is valid, then the BodyNode it's referring to must still be alive and active. The *only* time a BodyNode gets deleted is in the destructor of the Skeleton that owns it, so it is *impossible* for the BodyNode to be dead if the Skeleton that most recently owned it is still alive. There is also a mutex to go along with each weak_ptr<Skeleton>, so the WeakBodyNodePtr should also be thread safe.

So far the scheme has been tested and verified for a single-threaded program, but I think it's worth testing it for a multi-threaded program as well. That said, I've never attempted to write a unit test with concurrency (and in fact I've read from some sources that it's a bad thing to do), so if anyone has suggestions regarding that matter, I'd appreciate any input.

mxgrey:
The latest commit introduces the MetaSkeleton API (but not the ReferentialSkeleton yet), and it splits up the Skeleton caching of dynamic properties so that dirty flags and data caching are handled *per independent tree* instead of per entire Skeleton. This will allow there to be absolutely no penalty for having arbitrarily many independent BodyNode trees within a single Skeleton. Previously, getting a dynamics-related value (like mass matrix) for a BodyNode anywhere in a Skeleton would require computing the dynamics of the entire Skeleton, but now the Skeleton will only compute the dynamics for the specific tree of the BodyNode that you're inquiring about. The task of intelligently computing dynamics-related values can't be made any more efficient than this without a pointlessly cumbersome increase in bookkeeping overhead.

The latest commit also makes it easier to get the individual root BodyNodes and see the independent kinematic trees within a Skeleton.

I think the last thing that needs to be done is the ReferentialSkeleton implementations, and then we'll be set for the goals of the 5.0 release.

This means that the 5.0 release will not include the EndEffector class that I'm working on, but I don't think that's an issue since that will be an addition to the API, and won't result in any API collisions.

mxgrey:
The latest push adds a final boolean argument (named ``_recursive``) to the BodyNode::copyTo and BodyNode::copyAs functions. The default value for the argument is true, but if you set it to false, then only a single BodyNode will be copied. None of its descendents will be copied with it.

mxgrey:
With the latest commit, I've finished everything that I had planned for the 5.0 release.

This pull request is ready for review. I wish the best of luck to anyone who attempts to review it; there are so many changes that diffs are going to look like a nightmare.

mxgrey:
This branch seems to be failing **testDynamics** with the configuration of Mac / Clang / Release. No other configuration is failing the test (not even Mac / Clang / Debug).

Would any Mac users be able to test this configuration on their machine and see what testDynamics prints out as failing?

mxgrey:
My guess was right. There was one randomized case that resulted in extremely large numbers for the impulse dynamics. It only showed up on Mac because of platform-dependent differences in rand(). The error in the comparison was just a bit beyond the tolerance, but loosening the tolerance has allowed the tests to pass on Travis again.

dart/common/NameManager.h
```diff
@@ -189,6 +227,12 @@ class NameManager
...
 
+  /// Return true if the object is contained
+  bool hasObject(T _obj) const
+  {
+    return (mReverseMap.find(_obj) != mReverseMap.end());
+  }
+
   /// Get the number of the objects currently stored by the NameManager
   size_t getCount() const
   {
```
jslee02: Any possibility the size of mMap and mReverseMap are different? Does it make sense to check it with `assert()` to prevent breaking the assumption by accident?

dart/common/NameManager.h
```diff
@@ -145,42 +148,77 @@ class NameManager
   }
 
   /// Add an object to the map
-  bool addName(const std::string& _name, T* _obj)
+  bool addName(const std::string& _name, T _obj)
```
jslee02: How about passing const reference for T?

dart/constraint/ConstrainedGroup.h
```diff
@@ -107,7 +108,7 @@ class ConstrainedGroup
   std::vector<ConstraintBase*> mConstraints;
 
   ///
-  dynamics::Skeleton* mRootSkeleton;
+  std::shared_ptr<dynamics::Skeleton> mRootSkeleton;
```
jslee02: Use `dynamics::SkeletonPtr`?

dart/dynamics/Joint.h
```diff
@@ -557,8 +620,19 @@ class Joint : public virtual common::Subject
...
+  /// Create a clone of this Joint. This may only be called by the Skeleton
+  /// class.
+  virtual Joint* clone() const = 0;
+
+  /// Called by the Skeleton class
+  virtual void registerDofs() = 0;
+
   /// Initialize this joint. This function is called by BodyNode::init()
-  virtual void init(Skeleton* _skel);
+  virtual void init(std::shared_ptr<Skeleton> _skel);
```
jslee02: It seems unnecessary. This function was used to store the parent Skeleton but now Joint access the parent Skeleton through the child BodyNode. So we can deprecate this function.

mxgrey: I kept it in just in case someone ever makes a Joint type that requires initialization for some reason. But maybe that's an extremely unlikely use case.

Since Joint::init is completely internal (protected), we can just remove it instead of deprecating it.

dart/constraint/ConstrainedGroup.h
mxgrey: dynamics::SkeletonPtr is defined in the Skeleton.h header. In places where the Skeleton.h header wouldn't be needed for anything else, I opted to use std::shared_ptr<Skeleton> instead of SkeletonPtr to try to cut down on unnecessary header inclusions.

In theory, it should reduce compilation times that result from changes to the Skeleton.h header. I actually think we might be able to do a lot better in this respect:

We might want to create a SkeletonPtr.h file that just contains the following:

```
#include <memory>

namespace dart {
namespace dynamics {
class Skeleton;
typedef std::shared_ptr<Skeleton> SkeletonPtr;
} // namespace dynamics
} // namespace dart
```

Then we can just include SkeletonPtr.h to access SkeletonPtr, without having a header dependency on Skeleton.h

dart/dynamics/Joint.h
jslee02: Do we just care only public API? I wonder what if someone inherit a class and attempts to use removed protected API.

mxgrey: That's a fair point; I thought we were assuming that no one uses it.

I guess if we deprecate it for a while, we can wait and see if anyone complains that we've messed up their custom Joint type. And if no one ever complains, *then* we can remove it.

dart/constraint/ConstrainedGroup.h
jslee02: I agree that including unnecessary header increases the compilation times.

Then we might create many of `[class]Ptr.h` files. How about creating `DynamicsTypes.h` and put all the smart pointer types into the file like:
```cpp
// DynamicsTypes.h

#include <memory>

namespace dart {
namespace dynamics {

class Skeleton;
typedef std::shared_ptr<Skeleton> SkeletonPtr;

class Shape;
typedef std::shared_ptr<Shape> ShapePtr;

// and so on

} // namespace dynamics
} // namespace dart
```

dart/common/NameManager.h
mxgrey: I was extremely careful to ensure that mMap and mReverseMap are always in sync with each other (they always insert and erase corresponding entries at the same time). The only way it could be violated is if the class is inherited, and the derived class does something insane with them. We could prevent that by making those variables private instead of protected, if that's a concern.

That said, I wouldn't be opposed to having a debug mode check somewhere ensuring that all the entries are in sync, but I would question whether the increased runtime would be worth it.

dart/constraint/ConstrainedGroup.h
mxgrey: That sounds good to me, although I'm pretty sure SkeletonPtr and ShapePtr are the only types that would be able to go in there, because BodyNodePtr operates in such a way that it needs to have the BodyNode class fully defined before it can be instantiated, and JointPtr depends directly on BodyNodePtr. So I'd be a little concerned that an end user would see "DynamicsTypes.h" and assume that they can get BodyNodePtr and JointPtr from it as well, when they actually can't.

dart/common/NameManager.h
jslee02: If it causes noticeable overhead then we might need to find other way, but `assert(mMap.size() == mReverseMap.size())` wouldn't be the case I think. But still it might not be able to capture the violation if `getCount()` is overloaded by the inherited class.

It seems fine without a debug mode check, I thought it would save the time when it causes problem.

dart/constraint/ConstrainedGroup.h
mxgrey: If we want to get really fancy, one option might be to abstract out the reference counting functionality of BodyNode that is needed for BodyNodePtr. If the referencing counting is put into a base class that gets inherited by BodyNodePtr, then maybe BodyNodePtr can just depend on the very simple base class instead of depending on a fully defined BodyNode class. Then we can safely put BodyNodePtr and JointPtr into DynamicsTypes.h.

I don't know for sure if this would work, but I think it's doable with some clever templating tricks.

dart/common/NameManager.h
mxgrey: getCount() can't actually be overloaded since it's not virtual, so nothing to worry about there.

Just checking the sizes shouldn't be expensive at all. I was imagining running through the maps and checking for a one-to-one correspondence on their entries. I'd have no objections to just a quick size comparison.

dart/dynamics/Linkage.h
```diff
@@ -0,0 +1,198 @@
...
+#define DART_DYNAMICS_LINKAGE_H_
+
+#include "ReferentialSkeleton.h"
+#include <unordered_set>
+
+namespace dart {
+namespace dynamics {
+
+/// A Linkage is a ReferentialSkeleton with the special property that all the
+/// BodyNodes included in it form a contiguous graph. This propery is only
```
jslee02: propery --> property?

dart/common/NameManager.h
jslee02: Right, but a non-virtual function still can be hidden by the derived class.

getCount() is for returning the size so comparing the sized would be enough.

jslee02:
Does it make sense to add child joint accessors to BodyNode? It would be like:

```cpp
size_t BodyNode::getNumChildJoints() const
{
  return getNumChildBodyNodes();
}

Joint* BodyNode::getChildJoint(std::size_t _index)
{
  BodyNode* childBodyNode = getChildBodyNode(_index);

  if (childBodyNode)
    return childBodyNode->getParentJoint();
  else
    return nullptr;
}

const Joint* BodyNode::getChildJoint(std::size_t _index) const
{
  return const_cast<BodyNode*>(this)->getChildJoint(_index);
}
```

mxgrey:
I think child joint accessors in BodyNode would be an extremely good idea. :+1:

mxgrey:
I've made some changes based on the feedback. Next I'll see about whether BodyNodePtr can be defined without a full definition of BodyNode.

dart/constraint/ConstrainedGroup.h
mxgrey: What would you think of a file named "Ptr.h", so that it's called "dart/dynamics/Ptr.h" when being included?

jslee02: I'm still inclined to use `DyanmicsTypes.h`, and put the smart pointer types and other type definitions such as `INVALID_INDEX`.

Also, I think we could use const variable instead of unnamed enum for type safety as:
```cpp
const size_t invalidIndex = static_cast<size_t>(-1);
```
enum's default type is int unless we use enum class with name scoping like:
```cpp
enum class Index : size_t
{
  INVALID_INDEX = static_cast<size_t>(-1)
};
```

jslee02: `DynamicsTypes.h`  is just my preference rather than create different type defining files like `InvalidIndex.h` if it doesn't cause too much coupling. But I'm fine with `Ptr.h` and `InvalidIndex.h`.

Beside on that, how do you think about of using const variable instead enum?

mxgrey: > Beside on that, how do you think of using const variable instead enum for INVALID_INDEX?

const variable is definitely better. I had my head in C land (rather than C++ land) when I made InvalidIndex.

> DynamicsTypes.h is just my preference rather than create different type defining files like InvalidIndex.h if it doesn't cause too much coupling. But I'm fine with Ptr.h and InvalidIndex.h.

I would normally agree, but I feel like the existing ``dynamics.h`` serves the role of a ``DynamicsTypes.h``. I made a separate header for InvalidIndex just because I couldn't think of anything else to put in there. If we find ourselves with more simple types belonging to the dynamics namespace, I'd be okay with consolidating them into a single header.

mkoval: I suggest using constexpr here. That way we get all of the benefits of an
enum (e.g. being able to use it as a template parameter), without
introducing a superfluous enum.

(Although, in this particular case, it's pretty unlikely that we will ever
do that.)

jslee02: > If we find ourselves with more simple types belonging to the dynamics namespace, I'd be okay with consolidating them into a single header.

I totally agree with this. Let's see we need more types.

mxgrey: @mkoval What would be the advantage of a constexpr rather than a const variable here?

mxgrey:
It turns out that VS2013 doesn't support constexpr.

I suppose we should either change INVALID_INDEX to just a regular const variable until VS2015, or use preprocessor directives to use const instead of constexpr just for VS2013.

Does anyone know a preprocessor definition that can reliably identify if the compiler is Visual Studio?

jslee02:
You can find an example with the preprocessor definitions [here](https://github.com/dartsim/dart/blob/master/dart/common/Deprecated.h#L49) as:
```cpp
#if defined(_MSC_VER)
```

mxgrey:
I recently removed the call to Skeleton::init(~) from World::addSkeleton(~), because in principle it should no longer be necessary to initialize a Skeleton before using it. However, running some of the apps, I've noticed that this has clobbered the controllers that depended on hard-coded indexing, because the BodyNodes tend to get rearranged after Skeleton::init(~) is called.

I think it's very undesirable that adding a Skeleton to a world runs the risk of changing the indexing of the BodyNodes and generalized coordinates in the Skeleton, so I want to keep Skeleton::init(~) out of World::addSkeleton(~). However, this could kill backwards compatibility for any code that hardcoded their indexing. It's not a big deal to fix this for the apps that are provided natively in the DART repository, but the code of end users would be at risk. But even at that, old code is unlikely to be backwards compatible with the latest major release of DART anyway, so it seems like now would be the best time to break that compatibility.

What does everyone else think?

mxgrey:
In my latest commits, I've created the dart::dynamics::Group class (which I had originally planned on naming ``Disjointment``, but decided that ``Group`` would be less silly) to allow arbitrarily setting the indexing of a ReferentialSkeleton.

I've also fixed the indexing issues that arose with the atlasSimbicon app and the bipedStand app.

mxgrey:
I've noticed something that might raise some red flags: The commit af95121f302439d61af4cc41e3f36ade6c216ca4 has altered the behavior of the softBodies app. In the past, the cylinder ends standing on its bottom, but ever since the change to the generalized coordinate differencing, it falls on its side.

The cylinder is in an unstable state that crucial point in the simulation, so either end result is entirely plausible, and I don't know that there's an objective way of determining which result is "correct". I suspect that once you account for things like integration error and the simplifications inherent in our friction and collision handling dynamics, either result can be considered "correct". But it might be worth considering what about that particular commit could have resulted in this quantitative change, and whether the change was something good, bad, or neutral.

dart/simulation/World.cpp
```diff
@@ -74,11 +76,49 @@ World::~World()
...
   {
-    delete (*it);
+    dynamics::Frame* current_parent =
+        worldClone->getFrame(i)->getParentFrame();
+
+    dynamics::SimpleFramePtr parent_candidate =
+        worldClone->getFrame(current_parent->getName());
+
+    if(parent_candidate)
+      worldClone->getFrame(i)->setParentFrame(parent_candidate.get());
```
jslee02: Any possibility if a SimpleFrame fails to find its parent in the cloned world? I think it shouldn't be.

mxgrey: If the parent of the SimpleFrame is another SimpleFrame which has **not** been added to the original World, then it will not be found in the new World.

The motivation behind this is you can decide which SimpleFrames are tied to their World versus which that aren't. You can have some SimpleFrames which will be copied when their World is cloned while other SimpleFrames are used across all the World copies.

jslee02: It make sense. So the parent frame of a SimpleFrame is able to be in a different World from the World of the SimpleFrame.

Beside on this, I came up with some idea that might be useful or not. That is ***master world***. Multiple worlds can have a master world like master slide of presentation tools. The difference from cloning is all the worlds share the skeletons and simple frames in the master world rather than actually creating the instances. This might be useful if we want to simulate with different parameters while having invariant things in the master world.

dart/simulation/World.h
```diff
@@ -107,57 +126,47 @@ class World : public virtual common::Subject
...
   /// Get the number of Entities
-  size_t getNumEntities() const;
+  size_t getNumFrames() const;
 
   /// Add an Entity to this world
-  std::string addEntity(dynamics::Entity* _entity);
-
-  /// Remove an Entity from this world without deleting it
-  void withdrawEntity(dynamics::Entity* _entity);
+  std::string addFrame(dynamics::SimpleFramePtr _frame);
```
jslee02: How about `addSimpleFrame()`? `addFrame()` sounds like it takes `Frame` or any classes inherited `Frame`.

dart/simulation/World.cpp
mxgrey: > So the parent frame of a SimpleFrame is able to be in a different World from the World of the SimpleFrame.

It's more like, a SimpleFrame isn't necessarily tied to a World at all. If you decide to tie a SimpleFrame to a World, then a copy of it will be made when the World is cloned. But if you don't tie a SimpleFrame to a World, then it won't get cloned automatically.

I'm not sure I'm following how the master world concept would work, although it might be related to something that I'm thinking about for the future: **lazy copying**. If we were to lazily copy BodyNodes and Skeletons, then we could make it so that the properties of a copied BodyNode are just references to the properties of the original BodyNode, *until* the user decides to change one of the properties. At that point, the copied BodyNode will instantiate whatever it needs to, instead of just referencing the original. The implementation would be somewhat challenging, but certainly doable. The big question would be whether changing a property in the original should also carry over to the existing clones of the original.

dart/simulation/World.h
mxgrey: Yeah, we should probably change it to that. Originally, you *were* able to add any kind of Frame, but then I wanted to be able to clone the Frames whenever the World is cloned, which would be a problem if someone added BodyNode frames.

dart/simulation/World.cpp
jslee02: > It's more like, a SimpleFrame isn't necessarily tied to a World at all.

I understand it more clearly now. I exclusion the case a SimpleFrame isn't tied to a World.

lazy copying sounds interesting, and I agree with that the implementation would be somewhat challenging.

> The big question would be whether changing a property in the original should also carry over to the existing clones of the original.

In my view, changing a property in the original shouldn't carry over to the clones, and they should be able to have different property values. It might be able to give a flag whether the property carry over to the clones, but keeping same property values in different world might be physically infeasible under certain circumstances. However, all the entities in the master world will always carry the same properties to all the dependent worlds. One use case I can imagine now is that we put entities for static environment to the master world and run different simulations in multiple worlds sharing the mast world. So whether carrying a property or not will be the differences between lazy copying and master world.

mxgrey:
I decided to go with a directory named ``detail``, because I think that name had the most support overall.

jslee02:
Could we change file name of `Ptr.h` to more informative like `Pointer[s].h` or `SmartPointer[s].h`?

Also, the author is not correct, and I would like to add comment that shows the final smart pointer class names at least one per each macro. Here is possible revised version what I imagine:

```cpp
/*
 * Copyright (c) 2015, Georgia Tech Research Corporation
 * All rights reserved.
 *
 * Author(s): Michael X. Grey <mxgrey@gatech.edu>
 *
 * Georgia Tech Graphics Lab and Humanoid Robotics Lab
 *
 * Directed by Prof. C. Karen Liu and Prof. Mike Stilman
 * <karenliu@cc.gatech.edu> <mstilman@cc.gatech.edu>
 *
 * This file is provided under the following "BSD-style" License:
 *   Redistribution and use in source and binary forms, with or
 *   without modification, are permitted provided that the following
 *   conditions are met:
 *   * Redistributions of source code must retain the above copyright
 *     notice, this list of conditions and the following disclaimer.
 *   * Redistributions in binary form must reproduce the above
 *     copyright notice, this list of conditions and the following
 *     disclaimer in the documentation and/or other materials provided
 *     with the distribution.
 *   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND
 *   CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
 *   INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
 *   MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
 *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
 *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
 *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
 *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
 *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
 *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
 *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 *   POSSIBILITY OF SUCH DAMAGE.
 */

#ifndef DART_DYNAMICS_PTR_H_
#define DART_DYNAMICS_PTR_H_

#include "dart/dynamics/detail/BodyNodePtr.h"
#include "dart/dynamics/detail/JointPtr.h"
#include "dart/dynamics/detail/DegreeOfFreedomPtr.h"

// This file is a lightweight means of providing the smart pointers which are
// commonly used within the dart::dynamics namespace. It is 'lightweight' in the
// sense that it does not depend on any types being fully defined, making this
// header suitable for inclusion in other headers which might only want access
// to the smart pointers without needing fully defined classes.

namespace dart {
namespace dynamics {

// -- Standard shared/weak pointers --
// Define a typedef for const and non-const version of shared_ptr and weak_ptr
// for the class X
#define DART_DYNAMICS_MAKE_SHARED_WEAK( X )                 \
  typedef std::shared_ptr< X >      X ## Ptr;               \
  typedef std::shared_ptr< const X > Const ## X ## Ptr;     \
  typedef std::weak_ptr< X >        Weak ## X ## Ptr;       \
  typedef std::weak_ptr< const X >   WeakConst ## X ## Ptr;

// Skeleton smart pointers
class Skeleton;
DART_DYNAMICS_MAKE_SHARED_WEAK(Skeleton)
// These pointers will take the form of:
// std::shared_ptr<Skeleton>        --> SkeletonPtr
// std::shared_ptr<const Skeleton>  --> ConstSkeletonPtr
// std::weak_ptr<Skeleton>          --> WeakSkeletonPtr
// std::weak_ptr<const Skeleton>    --> WeakConstSkeletonPtr

// MetaSkeleton smart pointers
class MetaSkeleton;
DART_DYNAMICS_MAKE_SHARED_WEAK(MetaSkeleton)

// ReferentialSkeleton smart pointers
class ReferentialSkeleton;
DART_DYNAMICS_MAKE_SHARED_WEAK(ReferentialSkeleton)

class Group;
DART_DYNAMICS_MAKE_SHARED_WEAK(Group)

class Linkage;
DART_DYNAMICS_MAKE_SHARED_WEAK(Linkage)

class Branch;
DART_DYNAMICS_MAKE_SHARED_WEAK(Branch)

class Chain;
DART_DYNAMICS_MAKE_SHARED_WEAK(Chain)

// Shape smart pointers
class Shape;
DART_DYNAMICS_MAKE_SHARED_WEAK(Shape)


// -- Custom BodyNode smart pointers --
#define DART_DYNAMICS_MAKE_BODYNODEPTR( X )                         \
  typedef TemplateBodyNodePtr< X >          X ## Ptr;               \
  typedef TemplateBodyNodePtr< const X >     Const ## X ## Ptr;     \
  typedef TemplateWeakBodyNodePtr< X >      Weak ## X ## Ptr;       \
  typedef TemplateWeakBodyNodePtr< const X > WeakConst ## X ## Ptr;

// BodyNode smart pointers
class BodyNode;
DART_DYNAMICS_MAKE_BODYNODEPTR(BodyNode)
// These pointers will take the form of:
// TemplateBodyNodePtr<BodyNode>            --> BodyNodePtr
// TemplateBodyNodePtr<const BodyNode>      --> ConstBodyNodePtr
// TemplateWeakBodyNodePtr<BodyNode>        --> WeakBodyNodePtr
// TemplateWeakBodyNodePtr<const BodyNode>  --> WeakConstBodyNodePtr

// SoftBodyNode smart pointers
class SoftBodyNode;
DART_DYNAMICS_MAKE_BODYNODEPTR(SoftBodyNode)


// -- BodyNode dependent smart pointers --
#define DART_DYNAMICS_MAKE_BN_DEPENDENT_PTR( X )                                          \
  typedef Template ## X ## Ptr < X , BodyNode >                   X ## Ptr;               \
  typedef Template ## X ## Ptr < const X , const BodyNode >       Const ## X ## Ptr;      \
  typedef TemplateWeak ## X ## Ptr < X , BodyNode >               Weak ## X ## Ptr;       \
  typedef TemplateWeak ## X ## Ptr < const X , const BodyNode >   WeakConst ## X ## Ptr;

// Joint smart pointers
class Joint;
DART_DYNAMICS_MAKE_BN_DEPENDENT_PTR(Joint)
// These pointers will take the form of:
// TemplateJointPtr<Joint>            --> JointPtr
// TemplateJointPtr<const Joint>      --> ConstJointPtr
// TemplateWeakJointPtr<Joint>        --> WeakJointPtr
// TemplateWeakJointPtr<const Joint>  --> WeakConstJointPtr

// DegreeOfFreedom smart pointers
class DegreeOfFreedom;
DART_DYNAMICS_MAKE_BN_DEPENDENT_PTR(DegreeOfFreedom)

} // namespace dynamics
} // namespace dart

#endif // DART_DYNAMICS_PTR_H_

```

jslee02:
Beside on `Ptr.h`, :+1: . 

There are still few issues but let's make them as separate issues or pull requests. I think the main purpose of this PR is already fulfilled.

apps/rigidCubes/Main.cpp
```diff
@@ -54,6 +54,15 @@ int main(int argc, char* argv[]) {
...
 
+  for(size_t i=0; i<myWorld->getNumSkeletons(); ++i)
+  {
+    dart::dynamics::SkeletonPtr skel = myWorld->getSkeleton(i);
+    for(size_t j=0; j<skel->getNumBodyNodes(); ++j)
+    {
+      std::cout << skel->getBodyNode(j)->getSpatialInertia() << "\n" << std::endl;
+    }
+  }
+
```
jslee02: Is this for debug?

mxgrey: Hahah, I would guess so. Probably from when I created the Spatial Inertia class. I'll remove it.

jslee02: Thanks. I do this a lot. :)

PR #390, opened 2015-05-23
Fixed the range check in getRootBodyNode.

mkoval:
This pull request fixes a bug in the range check in `Skeleton::getRootBodyNode`. The condition was inverted, which made it impossible to call this function.


mxgrey:
Whoops, I should create a regression test that makes sure Skeleton::getRootBodyNode is sane.

PR #391, opened 2015-05-23
Added virtual destructors to joint properties.

mkoval:
This pull request adds default virtual constructors to all joint `Properties` classes. These classes are polymorphic, so this is generally a good idea from a safety perspective. This is also necessary for RTTI to work on these classes (which we need to create Python bindings).


mxgrey:
I've always wished C++ destructors were virtual by default.

PR #392, opened 2015-05-24
Added getType and getStaticType functions to Joint

mkoval:
This pull request adds:

1. a virtual `getType` method to the `Joint` base class
2. a static `getStaticType` method to each concrete `Joint` class.

Both of these methods return a string representation of the class name. This is useful for debugging, for logic that needs to change depending upon the joint type, and for language bindings (where wrapping templates is difficult). This currently requires an exhaustive search over all `Joint` subclasses by checking the return value by `dynamic_cast`.

We were careful to make both of these functions return a const reference to a statically allocated string.

### DIFF
--- a/dart/dynamics/BallJoint.cpp
+++ b/dart/dynamics/BallJoint.cpp
@@ -66,6 +66,19 @@ BallJoint::~BallJoint()
 }
 
 //==============================================================================
+const std::string& BallJoint::getType() const
+{
+  return getStaticType();
+}
+
+//==============================================================================
+const std::string& BallJoint::getStaticType()
+{
+  static const std::string name = "BallJoint";
+  return name;
+}
+
+//==============================================================================
 BallJoint::Properties BallJoint::getBallJointProperties() const
 {
   return getMultiDofJointProperties();
--- a/dart/dynamics/BallJoint.h
+++ b/dart/dynamics/BallJoint.h
@@ -65,7 +65,7 @@ public:
   virtual ~BallJoint();
 
   // Documentation inherited
-  const std::string& getType() const override;
+  virtual const std::string& getType() const override;
 
   /// Get joint type for this class
   static const std::string& getStaticType();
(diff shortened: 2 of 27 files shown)


mxgrey:
I like the idea of the const reference to the static string inside the member function. Since it's not a reference to a member variable, there's no chance of an inheriting class or a friend class doing something silly and altering it in some way.

Although I do wonder if it would be better for the Joint::getType(), SingleDofJoint::getType(), ZeroDofJoint::getType(), and MultiDofJoint::getType(), versions to be pure abstract. That way we would enforce that all the fully derived classes must explicitly report their Joint type.

mkoval:
That makes sense to me. I removed `getType` from those joint types.

mkoval:
I just pushed some fixes (mostly missing `virtual` keywords) to my branch.

mxgrey:
There seems to be a bizarre GCC bug related to the use of ``default`` that's preventing it from compiling when BUILD_DEFAULT_ONLY is off. This is especially strange, because it claims to be erroring when compiling MultiDofJoint::Properties, but that should be getting compiled whether or not BUILD_DEFAULT_ONLY is off, so I can't imagine why it only seems to happen when it's off.

mxgrey:
The GCC bug I mentioned seems to be either platform dependent or it's been resolved in 4.9.2, because I can't manage to reproduce it on my own machine. I'll play around with the usage of the ``default`` keyword and see if we can convince Travis-CI to build it.

mkoval:
@mxgrey What is `BUILD_DEFAULT_ONLY`? Do you have any idea what caused this?

mxgrey:
Whoops, I meant to say BUILD_CORE_ONLY. I had the word "default" in my head so I wound up replacing "CORE" with "DEFAULT" while I was typing. I've edited my post above.

It seems to be an internal GCC bug. I saw reports for it from early 2014. It doesn't happen on my machine, and I'm using GCC 4.9, whereas Travis seems to use older versions. I'm guessing the bug has been removed in the latest versions of GCC.

PR #393, opened 2015-05-24
Remove GL_ACCUM flag from SimWindow

mkoval:
This pull request removes the `GL_ACCUM` flag from `SimWindow`. This flag was depcreated in OpenGL 3.1 and is not supported in Parallels VM.

Does anything in `dart::rendering` uses accumulation buffers? If not, can we remove this flag?

### DIFF
--- a/dart/gui/GlutWindow.cpp
+++ b/dart/gui/GlutWindow.cpp
@@ -80,8 +80,7 @@ void GlutWindow::initWindow(int _w, int _h, const char* _name) {
   mWinWidth = _w;
   mWinHeight = _h;
 
-  glutInitDisplayMode(GLUT_DEPTH | GLUT_DOUBLE |GLUT_RGBA  | GLUT_STENCIL
-                      | GLUT_ACCUM);
+  glutInitDisplayMode(GLUT_DEPTH | GLUT_DOUBLE | GLUT_RGBA | GLUT_STENCIL);
   glutInitWindowPosition(150, 100);
   glutInitWindowSize(_w, _h);
   mWinIDs.push_back(glutCreateWindow(_name));


jslee02:
It is used by [`Win3d::capturing()`](https://github.com/dartsim/dart/blob/6f4314404167f78d274098325e974048dafcfa5c/dart/gui/Win3D.cpp#L302) to take screenshots with [multi sampling anti-aliasing](https://en.wikipedia.org/wiki/Multisample_anti-aliasing). The screen capturing doesn't seem to work now, by the way.

I learned from googling that GLUT library is no longer in development and stopped at version 3.7 in 1998. Not only this accumulation buffer but lots of other API that DART uses are [deprecated in OpenGL 3.1](https://www.opengl.org/wiki/History_of_OpenGL#Deprecation_Model). FreeGLUT is an alternative to GLUT, but FreeGLUT is mostly concerned with providing a stable clone of GLUT still using deprecated OpenGL API (OpenGL 1, 2). That's why clang generates [deprecated-declarations warnings](https://travis-ci.org/dartsim/dart/jobs/64005224#L527) on Mac.

I think we need to move on alternative library such as [glfw](http://www.glfw.org/) for a long term solution. But we can disable accumulation buffer until screen capturing fixed as workaround. How Paralles VM doesn't support that flag? It causes segfault?

mxgrey:
I'll definitely have osgDart wrapped up before anyone would have time to implement a new rendering extension. If we want to get rid of the existing GLUT-based rendering but also want an alternative to OSG, we should make sure that the alternative offers something that isn't available (or isn't as good) with OSG.

For example, I've been hearing that OSG has difficulty installing or building on certain platforms, so it would make sense to have an alternative that reliably works on all the platforms we support even if it's not as feature rich. [Based on their FAQ](http://www.glfw.org/faq.html#what-is-glfw) it looks like glfw fits the bill for this: it's designed to be extremely portable without much emphasis on features beyond providing a window context.

jslee02:
It make sense. To choose glfw as the alternative to GLUT, we make sure that it's more portable at least more than osgDart.

By the way, I'm a fan of osgDart waiting for the release.

mxgrey:
My plan is to release osgDart simultaneously with my EndEffector class, because I'm using the two features to test each other. They should be out very early in June :)

mkoval:
I can confirm that OSG has issues in Parallels (and possibly also VirtualBox) on OS X.

I've never used `glfw`, so I have no idea if it has the same issues. It appears to be lower level, so I am optimistic that it will not.

Tagging @pyrym. I think he has used `glfw` in the past.

PyryM:
Antialiasing should really be turned on at a hardware+graphics API level rather than implemented yourself with accumulation buffers.

jslee02:
I believe this is resolved in #402. It replaces our own implementation with accumulation buffers with OpenGL's built-in method. Thanks @PyryM.

karenliu:
+1

PR #397, opened 2015-05-27
Store path to mesh in MeshShape

mkoval:
This pull request adds a `getMeshPath` method to `MeshShape`. This returns the path to mesh on disk if it is available. If not, it returns an empty string (the default behavior).

This information is necessary to render meshes with non-embedded textures. Assimp references external textures with a path relative to the directory containing the mesh.

### DIFF
--- a/dart/dynamics/MeshShape.cpp
+++ b/dart/dynamics/MeshShape.cpp
@@ -129,9 +129,9 @@ aiMaterial::~aiMaterial()
 namespace dart {
 namespace dynamics {
 
-MeshShape::MeshShape(const Eigen::Vector3d& _scale, const aiScene* _mesh)
+MeshShape::MeshShape(const Eigen::Vector3d& _scale, const aiScene* _mesh,
+                     const std::string &_path)
   : Shape(MESH),
-    mMesh(_mesh),
     mDisplayList(0),
     mScale(_scale)
 {
@@ -139,9 +139,11 @@ MeshShape::MeshShape(const Eigen::Vector3d& _scale, const aiScene* _mesh)
   assert(_scale[1] > 0.0);
   assert(_scale[2] > 0.0);
 
-  if(nullptr == mMesh)
-    return;
+  setMesh(_mesh, _path);
 
+  if (nullptr == mMesh) {
+    return;
+  }
   _updateBoundingBoxDim();
   computeVolume();
   initMeshes();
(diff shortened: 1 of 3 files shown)


mkoval:
I noticed that "testDynamics" is failing on Travis. I didn't make any changes to the dynamics code, so I suspect that this is caused by previous changes in the `memory_management` branch.

@mxgrey Is that possible?

mxgrey:
testDynamics was failing temporarily while I was merging the master branch (with the new finite differencing feature) into memory_management. But I believe I fixed the cause the of the failure before I even finalized the commit for the merge. Strangely, the test only failed in one configuration (Mac / clang / release) and passed in all the others. Unfortunately, there's no way (that I know of) to obtain more detailed information about what failed.

Considering how finicky clang has been on Travis lately, all I can think of doing is telling Travis to rerun it. If it fails again, then we'll need to find someone who can test that configuration on their own machine to get us more details about what's failing.

mxgrey:
I've now rerun it, and the same failure pops up. It's definitely a problem that originated in memory_management.

It only fails on Mac / Clang / release, so if anyone has a Mac that they can test it on, that would be very helpful. My best guess is that there are some differences in the implementation of ``rand()`` between platforms, and one of the randomized tests that gets produced by the Mac / Clang combination is just a little bit outside of the tolerance that we have set.

scpeters:
~~~
/Users/scpeters/ws/gazebo_dart/src/dart/unittests/testDynamics.cpp:1660: Failure
Value of: equals(deltaVel1, deltaVel2, 1e-5)
  Actual: false
Expected: true
~~~

scpeters:
Also, use `make test "ARGS=-VV"` to see all the console out from test files.

mxgrey:
Thanks, @scpeters ! Did it also print out the values of detlaVel1 and deltaVel2? Are they far off, or just barely shy of a tolerance of 1e-5?

Next time this happens, I'll push a commit with ``make test "ARGS=-VV"``.

mkoval:
@mxgrey Besides the `testDynamics` issues, do you have any comments on this pull request? :smile:

mxgrey:
Looks good. Does the path name only matter for URDF? I could add it to the Skel and SDF parsers as well.

mkoval:
We should add it anywhere that we load meshes from disk (which includes the `skel` and `sdf` parsers). I just forgot about them because I only use the URDF loader. :laughing:

mxgrey:
I can merge it in and add it to the other parsers.

scpeters:
@mxgrey sorry I didn't paste the whole output. I just looked for the failure and pasted that. I also deleted the build already. IMO, the `make test "ARGS=-VV"` is just a good idea anyway to maximize the usefulness of travis.

PR #406, opened 2015-06-15
Fixed bug with renaming Joints and created regression test

mxgrey:
A bug was found where renaming a Joint would always alter the DegreeOfFreedom names, even when the user told them not to be renamed. This pull request fixes that issue and introduces a regression test to prevent it from happening again.


jslee02:
:+1:

PR #415, opened 2015-06-17
Remove undefined behavior from SkelParser

mxgrey:
We were using an iterator without first checking whether it was valid, even though the container will drop down to size 0 by the time the parsing is finished. Now we quit parsing once the container is empty.


jslee02:
:+1:

PR #416, opened 2015-06-21
Tutorials

karenliu:


olzhas:
Can you please give an example with Timer ?

jslee02:
There are two use cases in DART: [[1](https://github.com/dartsim/dart/blob/d1ec56ce35e9b70d63e312560bfd5b1bbe071a30/unittests/testCommon.cpp#L44)], [[2](https://github.com/dartsim/dart/blob/d1ec56ce35e9b70d63e312560bfd5b1bbe071a30/unittests/testMath.cpp#L1242)]

As a side note, you can ask general question in [the issue tracker](https://github.com/dartsim/dart/issues/new).

PR #421, opened 2015-06-26
Improved CollisionNode's index validity check

jslee02:


mxgrey:
Looks reasonable :+1: 

With the upcoming ShapeNode changes, we should probably think about how to automatically manage these relationships. One possibility might be to have a ShapeNode be aware of which other ShapeNodes its collisions are disabled with. We could have ShapeNode contain a ``std::unordered_set <ShapeNode*>`` (or something similar) that contains all the other ShapeNodes with which collisions are disabled (assuming we want collisions to be enabled by default).

With that scheme, the disabled collision relationships could be carried over when a Skeleton or World gets cloned.

jslee02:
I agree that we should think about automatic management theses relationships. This is still workaround to prevent segfault due to invalid index. Also, we should consider how to make collision node is aware of property changes of associated ShapeNode such as size changing.

karenliu:
+1

PR #425, opened 2015-06-29
Standardized warning messages for MultiDofJoints

mxgrey:
There was a lot of inconsistency among the format of warning messages for MultiDofJoint functions, even though the warnings they were giving were basically identical: indexing issues. This pull request standardizes (most of) them with two macros.

### DIFF
--- a/dart/dynamics/detail/MultiDofJoint.h
+++ b/dart/dynamics/detail/MultiDofJoint.h
@@ -38,7 +38,7 @@
 #define DART_DYNAMICS_DETAIL_MULTIDOFJOINT_H_
 
 #define MULTIDOFJOINT_REPORT_DIM_MISMATCH( func, arg )              \
-  dterr << "[MultiDofJoint::" #func "] mismatch beteween size of "  \
+  dterr << "[MultiDofJoint::" #func "] Mismatch beteween size of "  \
         << #arg " [" << arg .size() << "] and the number of "       \
         << "DOFs [" << getNumDofs() << "] for Joint named ["        \
         << getName() << "]\n";                                      \
--- a/dart/dynamics/detail/MultiDofJoint.h
+++ b/dart/dynamics/detail/MultiDofJoint.h
@@ -37,6 +37,19 @@
 #ifndef DART_DYNAMICS_DETAIL_MULTIDOFJOINT_H_
 #define DART_DYNAMICS_DETAIL_MULTIDOFJOINT_H_
 
+#define MULTIDOFJOINT_REPORT_DIM_MISMATCH( func, arg )              \
+  dterr << "[MultiDofJoint::" #func "] mismatch beteween size of "  \
+        << #arg " [" << arg .size() << "] and the number of "       \
+        << "DOFs [" << getNumDofs() << "] for Joint named ["        \
+        << getName() << "]\n";                                      \
+  assert(false);
+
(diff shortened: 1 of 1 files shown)


dart/dynamics/detail/MultiDofJoint.h
```diff
@@ -37,6 +37,19 @@
 #ifndef DART_DYNAMICS_DETAIL_MULTIDOFJOINT_H_
 #define DART_DYNAMICS_DETAIL_MULTIDOFJOINT_H_
 
+#define MULTIDOFJOINT_REPORT_DIM_MISMATCH( func, arg )              \
+  dterr << "[MultiDofJoint::" #func "] mismatch beteween size of "  \
+        << #arg " [" << arg .size() << "] and the number of "       \
+        << "DOFs [" << getNumDofs() << "] for Joint named ["        \
+        << getName() << "]\n";                                      \
```
jslee02: Periods are omitted on purpose?

dart/dynamics/detail/MultiDofJoint.h
```diff
@@ -37,6 +37,19 @@
 #ifndef DART_DYNAMICS_DETAIL_MULTIDOFJOINT_H_
 #define DART_DYNAMICS_DETAIL_MULTIDOFJOINT_H_
 
+#define MULTIDOFJOINT_REPORT_DIM_MISMATCH( func, arg )              \
+  dterr << "[MultiDofJoint::" #func "] mismatch beteween size of "  \
```
jslee02: Not beginning with capital letter?

mxgrey: I'm indifferent. I have no problem with capitalizing the first letter.

dart/dynamics/detail/MultiDofJoint.h
mxgrey: I'm never really sure when to use periods in warning messages. I think usually if there is only one sentence for the whole message, I don't bother with a period, but I'll include periods if there are multiple sentences in the message.

That said, I have nothing against putting a period at the end.

dart/dynamics/detail/MultiDofJoint.h
jslee02: Okay, then let's capitalize the the first letter. It would be good when the message is not single sentence.

dart/dynamics/detail/MultiDofJoint.h
jslee02: I agree. Sometimes it is not a sentence but just a word or a phrase.

How about putting a period at the end consistently?

mxgrey: I guess if we enforce the use of complete sentences for warning messages (which would probably be good practice), then it would make sense to enforce ending those sentences with periods consistently. I'm on board.

jslee02: Sounds good. I will follow that rule. Please let me know if there are breaks in my future commits.

jslee02:
:+1:

mxgrey:
Looks like appveyor needs a "retry".

jslee02:
Appveyor has failed to build DART since 5 days ago. It seems to fail to install the dependencies of DART.

I would not say this PR has something to do with it. I'm merging this PR and will make an separate issue for Appveyor.

PR #429, opened 2015-07-02
Improved joint warnings for Single and MultiDofJoints

mxgrey:
Similar to the previous changes for MultiDofJoints, I've standardized and improved the warnings for SingleDofJoints.

This also fixes a bug that caused warning messages when a zero-valued command was set for passive actuators.

### DIFF
--- a/dart/dynamics/SingleDofJoint.cpp
+++ b/dart/dynamics/SingleDofJoint.cpp
@@ -42,6 +42,22 @@
 #include "dart/dynamics/Skeleton.h"
 #include "dart/dynamics/DegreeOfFreedom.h"
 
+#define SINGLEDOFJOINT_REPORT_DIM_MISMATCH( func, arg )                        \
+  dterr << "[SingleDofJoint::" #func "] Size of " << #arg << "[" << arg .size()\
+        << "] should be exactly 1 for Joint named [" << getName() << "].\n";   \
+  assert(false);
+
+#define SINGLEDOFJOINT_REPORT_OUT_OF_RANGE( func, index )         \
+  dterr << "[SingleDofJoint::" #func "] The index [" << index     \
+        << "] is out of range for Joint named [" << getName()     \
+        << "] which only has a zeroth index.\n";                  \
+  assert(false);
+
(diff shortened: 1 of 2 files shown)


jslee02:
We could consider to use `std::endl` instead of `\n` for new line. The difference is that `std::endl` causes flushing of the output buffer while `\n` doesn't.

In general, `\n` would be good for performance by avoiding to flush everytime. However, when we want to know where exactly the message is printed, `std::endl` would be better. For example, unittest prints warning messages after all the unittests are done so that we don't know which unittest caused which warning messages.

mxgrey:
> In general, \n would be good for performance by avoiding to flush everytime.

This is why I prefer ``\n``. I'd rather trust the computer to know when a good time to flush is, or trust the user to flush when it's good for them, instead of assuming that messages should be displayed immediately.

By default, ``dtwarn`` and ``dterr`` are streaming to ``std::cerr``, and the default behavior of ``std::cerr`` is to not do any buffering (i.e. flush immediately), so the only way ``dtwarn`` and ``dterr`` would not flush right away is if the user intentionally altered that behavior, in which case I don't think we should be forcing it to flush, because the user is probably doing their own flushing management.

mxgrey:
> For example, unittest prints warning messages after all the unittests are done so that we don't know which unittest caused which warning messages.

I'm guessing this is the fault of gtest. One thing we could consider is explicitly flushing ``std::cerr`` at the end of each test, if gtest is buffering ``cerr`` for some reason.

jslee02:
I see. Maybe it was happened for `dtmsg` which is streaming to `std::cout`. I'll double check.

mxgrey:
It's important to note that ``cerr`` and ``cout`` are completely asynchronous, so messages that are sent out at the same time to each could end up on the terminal in wildly different places. In theory, ``cerr`` messages should make it to the terminal before ``cout`` message, but in practice I don't think that's always been the case for me. I'm pretty sure some of the behavior depends on the terminal program itself.

jslee02:
The print order of gtest and `dtmsg`/`dtwarn`/`dterr` seems random but mostly in the right order. So let's leave flushing management to the user as you said.

PR #435, opened 2015-07-06
Added the SYSTEM flag to include_directories

mkoval:
This pull request flags `include_directories` on third-party packages with [the `SYSTEM` flag](http://www.cmake.org/cmake/help/v3.0/command/include_directories.html). This lets the compiler suppress warnings from those headers, typically by passing the directory using the `-isystem` flag instead of `-i`.

This is important when building DART with a modern version of Clang, which produce a large number of warnings every time Eigen is included.


mxgrey:
:+1: I like not being yelled at for other people's issues.

But just to be clear: If my code uses a deprecated function from one of those headers, would the warning be blocked? I'm having trouble finding specifics about the behavior.

mxgrey:
I can't find anything that says it explicitly, but I think the answer to my question is almost definitely a "No, those warnings will not be suppressed", because it seems that the ``-isystem`` flag does not recklessly block **all** warnings from a header.

mkoval:
This is what the [the Clang documentation](http://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-in-system-headers) says:

> Warnings are suppressed when they occur in system headers. By default, an included file is treated as a system header if it is found in an include path specified by `-isystem`, but this can be overridden in several ways.

GCC [says something similar](https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html):

> The header files declaring interfaces to the operating system and runtime libraries often cannot be written in strictly conforming C. Therefore, GCC gives code found in system headers special treatment. All warnings, other than those generated by ‘#warning’ (see Diagnostics), are suppressed while GCC is processing a system header. Macros defined in a system header are immune to a few warnings wherever they are expanded. This immunity is granted on an ad-hoc basis, when we find that a warning generates lots of false positives because of code in macros defined in system headers.

I can't find a concrete statement either way.

jslee02:
I tested an simple example of what @mxgrey said. The use of deprecated API warnings were not suppressed as expected.

:+1:

PR #438, opened 2015-07-07
Joint rest positions default to zero, even if that is out of their limits

mkoval:
The default `mRestPosition` defaults to zero. This causes a warning to be printed when a joint is loaded from URDF whose limits do not include the zero value. In that case, this commit defaults the rest position to the center of the joint range (if it is bounded) or one of its endpoints. The logic is a bit convoluted to handle infinite limits.

This pull request fixes the warning. However, even with these changes, the robot is loaded in the invalid zero configuration. My assumption was that `mRestPosition` would also be used to initialize the robot's initial configuration. Is that not the case?

Finally, I am wondering if we should always use this logic to initialize `mRestPosition`, even if zero is a valid value. Thoughts?

### DIFF
--- a/dart/utils/urdf/DartLoader.cpp
+++ b/dart/utils/urdf/DartLoader.cpp
@@ -383,6 +383,9 @@ dynamics::BodyNode* DartLoader::createDartJointAndNode(
     singleDof.mVelocityUpperLimit =  _jt->limits->velocity;
     singleDof.mForceLowerLimit = -_jt->limits->effort;
     singleDof.mForceUpperLimit =  _jt->limits->effort;
+
+    if(_jt->limits->lower > 0 || _jt->limits->upper < 0)
+      singleDof.mRestPosition = (_jt->limits->lower + _jt->limits->upper) / 2.;
   }
 
   if(_jt->dynamics)
--- a/dart/utils/urdf/DartLoader.cpp
+++ b/dart/utils/urdf/DartLoader.cpp
@@ -385,7 +385,19 @@ dynamics::BodyNode* DartLoader::createDartJointAndNode(
     singleDof.mForceUpperLimit =  _jt->limits->effort;
 
     if(_jt->limits->lower > 0 || _jt->limits->upper < 0)
-      singleDof.mRestPosition = (_jt->limits->lower + _jt->limits->upper) / 2.;
+    {
+      if(std::isfinite(_jt->limits->lower)
+         && std::isfinite(_jt->limits->upper))
+        singleDof.mRestPosition
+          = (_jt->limits->lower + _jt->limits->upper) / 2.;
+      else if(std::isfinite(_jt->limits->lower))
+        singleDof.mRestPosition = _jt->limits->lower;
+      else if(std::isfinite(_jt->limits->upper))
+        singleDof.mRestPosition = _jt->limits->upper;
+
+      // Any other case means that the limits are both +inf, both -inf, or
+      // either is NaN. This should generate a warning elsewhere.
+    }
   }
 
   if(_jt->dynamics)


mxgrey:
This is a matter I've considered in the past, but I've always put it off because every model that I've worked with has had all zeros as the default/initial configuration anyway.

The ``mRestPosition`` variable is really only meant to be used by the implicit joint spring mechanics (at least as far as I know). I suppose I would be okay with it also being used as the "default initial position" value. The only catch would be that someone might want their default initial value to be different from the rest position of their joint spring.

I'm thinking the best way to handle this would be to introduce a new property along the lines of ``mInitialPosition`` and have the joint values set to that when the ``resetPosition`` function gets called. Likewise we might also want a ``mInitialVelocity`` term to go along with the ``resetVelocity`` function. Having a ``mInitialAcceleration`` and ``mInitialForce`` might pose problems, though, especially initial force. We use the ``resetForce`` function to clear out the joint forces, so having an "initial force" would probably not jive well.

mkoval:
What are the implicit joint spring mechanics? I wasn't aware of this, but I now see a few places that reference this.

I think it makes sense to add `mInitialPosition` and agree that `mInitialForce` doesn't make sense. I don't personally need for `mInitialVelocity`, or `mInitialAcceleration`, so I'll leave that up to your judgement. I must be unlucky: HERB and Robonaut have joints that can't be in the zero configuration. :frowning:

mxgrey:
> What are the implicit joint spring mechanics?

When you're doing dynamic simulation in DART, you can specify a spring coefficient for each degree of freedom, and then a Hooke's law style spring force will be applied in your joints, on top of any joint force that you specify. The ``mRestPosition`` is the "equilibrium" position of those spring forces. Also, they're called "implicit" spring forces because they use an implicit Euler method for integration which makes them more numerically stable.

I think ``mInitialVelocity`` would be very good to have, as long as we're going to allow ``mInitialPosition``. I don't know for sure about ``mInitialAcceleration``, though... I'll wait to see if @jslee02 has an opinion about any of this.

mkoval:
@mxgrey What's the motivation for the springs being baked into joints at such a low level? I'd expect the user to specify that type of force themselves each simulation loop. Is there something that makes it inefficient to do so? Is this common in dynamic simulation engines? 

(Just trying to understand why it exists / if I should be using it.)

mxgrey:
I don't know if it's common among simulation engines, but DART has some pretty decent motivations for it.

If your mechanism has natural passive stiffness or damping/friction, then it's good to be able to incorporate that directly into the model. Besides the stiffness coefficient, DART also offers parameters for viscous damping and Coulomb friction. An end user could certainly implement these things in their own higher level controller, but DART is allowing the parameters to be tied directly to the virtual model, which allows for a more complete representation of the actual physical system.

mxgrey:
Since this would be a somewhat significant behavioral change, I'd like to see what @jslee02 thinks before moving on it. My proposal would be this:

We set ``mInitialPosition`` (or ``mInitialPositions`` for MultiDofJoints) based on the starting positions that are given by the file that gets loaded. Whenever a ``resetPositions()`` function is called, it will set all Joint positions to ``mInitialPosition`` instead of setting them to zero. We also do likewise for ``mInitialVelocity``. However, ``resetAcceleration`` and ``resetForce`` will remain the same: they will always zero out acceleration and the force.

mxgrey:
@jslee02 How would you feel about the concept of ``resetPositions()`` putting the joints at ``mInitialPositions`` instead of zero? If that sounds okay to you, I'll implement it and set up a pull request.

mxgrey:
The proposed changes of this pull request are available in #449

Issue #433, opened 2015-07-03
OMPL integration


olzhas:
Hello everyone,
I found that DART provides only one motion planning algorithm (RRT). There is OMPL (open motion planning library) that provides implementation of a state-of-art motion and path planning algorithms. I worked with it for some time. I can try to integrate with DART, if you guide me. I meant could you tell me what should it look like and I will code it, because I am not quite familiar with design patterns. 
Also I found that there was a discussion regarding OMPL and DART integration (http://wiki.ros.org/fuerte/Planning/Simulation%20and%20Control/JointMeeting), maybe you know what was the outcome?

mkoval:
@psigen and I wrote DART bindings for OMPL. It consists of two parts: (1) an OMPL `StateSpace` that wraps a set of `DegreeOfFreedom`s and (2) an OMPL `StateValidityChecker` that internally uses DART for collision checking. This is sufficient to call any of OMPL's geometric motion planners on a DART environment.

We don't currently have this code available in a usable state, but we are in the process of writing documentation and cleaning up the build system. I'll post a pointer to it once it's online.

mxgrey:
Seems like this can be marked as closed since the work is in progress, right? Of course we can reopen it if people want to continue discussing the matter.

jslee02:
How about leave this open and mark "in-progress" so that people know this is on going?

mxgrey:
Makes sense.

mkoval:
The current version of our interface is [available in this repository](https://github.com/personalrobotics/r3). We weren't planning on posting this publicly for a while, but I did my best to update the `README` with the information you need to build it.

There are a few caveats:

1. We haven't rigorously tested this code yet and, in particular, the `StateValidityChecker`.
2. Any `DegreeOfFreedom` with unbounded limits (i.e. +/- infinity) is assumed to be continuous revolution joint and map them to an `SO2StateSpace`. This means that the DOF's values will wrap after a full revolution. This will break if you have another joint with unbounded limits (e.g. a `FreeJoint`).
3. Only geometric motion planning is supported.

Please do file an issue on the repository if you run into trouble. We'll do our best to help.

mxgrey:
Just out of curiosity, is the limit wrapping issue because of the way we need to interface to OMPL? I'm assuming it's because RRT- and PRM-style geometric motion planners require finite joint limits, so OMPL is demanding finite joint limits.

[...]

PR #441, opened 2015-07-08
Added Joint::isCyclic to mark SO(2) topology

mkoval:
This implements the `isCyclic` function that we discussed in #433.

### DIFF
--- a/dart/dynamics/Joint.h
+++ b/dart/dynamics/Joint.h
@@ -347,6 +347,12 @@ public:
   /// Get upper limit for position
   virtual double getPositionUpperLimit(size_t _index) const = 0;
 
+  /// Gets whether a generalized coordinate is cyclic; i.e. has SO(2) topology.
+  virtual bool isCyclic(size_t _index) const = 0;
+
+  /// Gets whether the position of a generalized coordinate is limited.
+  virtual bool isPositionLimited(size_t _index) const = 0;
+
   /// \}
 
   //----------------------------------------------------------------------------
--- a/dart/dynamics/Joint.h
+++ b/dart/dynamics/Joint.h
@@ -347,10 +347,10 @@ public:
   /// Get upper limit for position
   virtual double getPositionUpperLimit(size_t _index) const = 0;
 
-  /// Gets whether a generalized coordinate is cyclic; i.e. has SO(2) topology.
+  /// Gets whether a generalized coordinate has SO(2) topology.
   virtual bool isCyclic(size_t _index) const = 0;
 
-  /// Gets whether the position of a generalized coordinate is limited.
+  /// Gets whether the position of a generalized coordinate has limits.
   virtual bool hasPositionLimit(size_t _index) const = 0;
 
   /// \}
--- a/dart/dynamics/Joint.h
+++ b/dart/dynamics/Joint.h
@@ -354,14 +354,14 @@ public:
   /// Get upper limit for position
   virtual double getPositionUpperLimit(size_t _index) const = 0;
 
-  /// Gets whether a generalized coordinate is cyclic. Returns true if and only
+  /// Get whether a generalized coordinate is cyclic. Return true if and only
   /// if this generalized coordinate has an infinite number of positions that
   /// produce the same local transform. Note that, for a multi-DOF joint,
   /// producing a cycle may require altering the position of this Joint's other
   /// generalized coordinates.
   virtual bool isCyclic(size_t _index) const = 0;
 
-  /// Gets whether the position of a generalized coordinate has limits.
+  /// Get whether the position of a generalized coordinate has limits.
   virtual bool hasPositionLimit(size_t _index) const = 0;
 
   /// \}
--- a/dart/dynamics/Joint.h
+++ b/dart/dynamics/Joint.h
@@ -42,6 +42,7 @@
 #include <vector>
 #include <memory>
 
+#include "dart/common/Deprecated.h"
 #include "dart/common/Subject.h"
 #include "dart/math/MathTypes.h"
 #include "dart/dynamics/SmartPointer.h"
@@ -244,7 +245,10 @@ public:
   /// PASSIVE/FORCE.
   ///
   /// \sa ActuatorType
-  void setPositionLimited(bool _isPositionLimited);
+  void setPositionLimitEnforced(bool _isPositionLimited);
+
+  /// Deprecated. Replaced by setPositionLimitEnforced.
+  DEPRECATED(5.0) void setPositionLimited(bool _isPositionLimited);
 
   /// Get whether enforcing joint position limit
   ///
@@ -252,7 +256,10 @@ public:
   /// PASSIVE/FORCE.
   ///
   /// \sa ActuatorType
-  bool isPositionLimited() const;
+  bool isPositionLimitEnforced() const;
+
+  /// Deprecated. Replaced by isPositionLimitEnforced.
+  DEPRECATED(5.0) bool isPositionLimited() const;
 
   /// Get a unique index in skeleton of a generalized coordinate in this Joint
   virtual size_t getIndexInSkeleton(size_t _index) const = 0;
--- a/dart/dynamics/Joint.h
+++ b/dart/dynamics/Joint.h
@@ -351,7 +351,7 @@ public:
   virtual bool isCyclic(size_t _index) const = 0;
 
   /// Gets whether the position of a generalized coordinate is limited.
-  virtual bool isPositionLimited(size_t _index) const = 0;
+  virtual bool hasPositionLimit(size_t _index) const = 0;
 
   /// \}
 
--- a/dart/dynamics/Joint.h
+++ b/dart/dynamics/Joint.h
@@ -354,7 +354,11 @@ public:
   /// Get upper limit for position
   virtual double getPositionUpperLimit(size_t _index) const = 0;
 
-  /// Gets whether a generalized coordinate has SO(2) topology.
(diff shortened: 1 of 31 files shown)


mxgrey:
I wonder about the implementation of ``isCyclic`` for ``FreeJoint`` and ``BallJoint``... The three generalized coordinates of a log map are deeply coupled, so I wonder if it would be better if all three generalized coordinates were all considered cyclic or non-cyclic together instead of individually.

mkoval:
As per our discussion in #433 I added an `isPositionLimited` method to `Joint`. Since `getPositionLowerLimit` and `getPositionUpperLimit` both return `0` for out of range indices, I made `isPositionLimited` return `true` for consistency. The same reasoning applies to `ZeroDofJoint`.

@mxgrey Can you explain coupling breaks the symmetry of `FreeJoint` and `BallJoint`? DART uses minimal coordinates, so `DegreeOfFreedom`s should *always* be uncoupled, even if they impose very unintuitive constraints on the pose of the child `BodyNode`.

Tagging @pkv @siddh in case they want to chime in on this.

mkoval:
I just noticed that there is already a no-arg function called `getPositionLimited`, which has nothing to do with the `isPositionLimited` function that I implemented:

```c++
/// Get whether enforcing joint position limit
///
/// The joint position limit is valid when the actutor type is one of
/// PASSIVE/FORCE.
///
/// \sa ActuatorType
bool isPositionLimited() const;
```

This has the potential to be very confusing. In an ideal world, I would like to rename `isPositionLimited` to `isPositionLimitEnforced`, but that would break backwards compatibility. Maybe we can rename the function I added to `hasPositionLimits`?

This is even a bigger problem on `DegreeOfFreedom`: both functions take no arguments. :laughing:

mxgrey:
Ouch. ``hasPositionLimit`` does sound good.

I think we can make an ``isPositionLimitEnforced`` function and then mark ``isPositionLimited`` as deprecated, because it will definitely be confusing with ``hasPositionLimit``.

mxgrey:
> DART uses minimal coordinates, so DegreeOfFreedoms should always be uncoupled, even if they impose very unintuitive constraints on the pose of the child BodyNode.

The thing that gives me pause is that I'm not totally sure if a single BallJoint coordinate can really be cyclic if the other two are limited. Maybe it can, it's just not immediately obvious to me. With Revolute or Euler joints, it's obvious that the individual coordinates will just wrap around, but I don't know if that works for log maps. Log maps definitely do wrap, but I think all the coordinates wrap in sync with each other. I would have to look more into the geometry behind it to be sure, though.

mkoval:
I added `isPositionLimitEnforced`/`setPositionLimitEnforced` and deprecated `isPositionLimited`/`setPositionLimited` in e5fc7c1. I can't rename `mPositionLimited` in `Properties` because it's a member variable, so we're stuck that name until the next major revision.

I see your point about `BallJoint`. I'm not sure either..

mxgrey:
I've opened an issue for it as a reminder once 6.0 rolls around. I'm still not sure about BallJoint and FreeJoint, but I think the API is good, and we can patch the behavior in the future if it turns out that our geometric assumptions about the independence of the coordinates are ill-founded. For now, the pull request looks good to me :+1:

mkoval:
@mxgrey and I discussed this offline.

We definitely can't represent the topology of `BallJoint` and the rotational components of `FreeJoint` with this scheme because they fundamentally live in SO(3), not SO(2). We must these specially, at the `Joint` rather than the `DegreeOfFreedom` level, when constructing a state space for planning.

We still think it is useful to add `hasPositionLimit` and, possibly, `isCyclic`, perhaps with a few tweaks, to the DART API. We just have to be careful about how we define "cyclic".

mkoval:
I modified `BallJoint` and `FreeJoint` in ede1cbc4ef85d248d380184bfb28c0f8cfee7e60 to only have `isCyclic` return `true` if all DOFs are unbounded. I also changed the description of `isCyclic` in b117075 to:

> Gets whether this DOF is cyclic. Returns true if and only if an infinite number of DOF positions produce the same local transform. If this DOF is part of a multi-DOF joint, then producing a cycle may require altering the position of the Joint's other DOFs.

@mxgrey @psigen What do you think?

mxgrey:
I'm in agreement :+1:

dart/dynamics/SingleDofJoint.cpp
```diff
@@ -474,6 +474,19 @@ double SingleDofJoint::getPositionUpperLimit(size_t _index) const
...
+bool SingleDofJoint::hasPositionLimit(size_t _index) const
+{
+  if (_index != 0)
+  {
+    SINGLEDOFJOINT_REPORT_OUT_OF_RANGE( hasPositionLimit, _index );
+    return true;
+  }
+
+  return !(std::isinf(mSingleDofP.mPositionLowerLimit)
+        && std::isinf(mSingleDofP.mPositionUpperLimit));
```
jslee02: How about this:
```cpp
return (std::isfinite(mSingleDofP.mPositionLowerLimit)
        || std::isfinite(mSingleDofP.mPositionUpperLimit))
```
Less use of `not` operator seems more clear to understand to me. `std::isfinite` is opposite function to `std::isinf` except for `NAN` value case.  Both of them return `false` for `NAN` but I think `NAN` value is meaningless anyways.

dart/dynamics/Joint.h
```diff
@@ -347,6 +354,16 @@ class Joint : public virtual common::Subject
   /// Get upper limit for position
   virtual double getPositionUpperLimit(size_t _index) const = 0;
 
+  /// Gets whether a generalized coordinate is cyclic. Returns true if and only
```
jslee02: We haven't been putting `s` after verb when the comment begin with verb, and I've seen many other projects putting no `s`. Is this matter of personal preference or there is recommended way grammatically?

mxgrey: I would favor not having the ``s``. Grammatically speaking, the situation we have is that the subject of the sentence is implied, and most of the time when a sentence's subject is implied, the subject is "You". In this case, I interpret it as  saying "You get whether a generalized coordinate is cyclic", which is why I prefer not having the ``s``.

On the other hand, if the implied subject is "This function" then having the ``s`` would make sense, because you'd be saying "This function gets whether a generalized coordinate is cyclic".

But I prefer the former over the latter.

psigen: This level of detail is not consistently covered in coding standards, but here are some anecdotal guidelines:

https://www.drupal.org/coding-standards/docs
> For most functions (see exceptions below), summary lines must start with a third person singular present tense verb, and they must explain what the function does. Example: "Calculates the maximum weight for a list."

http://babbage.cs.qc.edu/courses/cs101/Coding_Guidelines.html
> The first thing inside a javadoc block has to be a sentence (capitalize the first word and end with a period) that tells what the class, interface, or method does. The style for this sentence is to start with a present-tense verb with the name of the class, interface, or method as the implicit subject. For example, instead of saying "Interface Abc defines the methods that must ...", say "Defines the methods that must ...". There can be more than one sentence in the description, but early versions of javadoc limited the description to just one, and it's better to make your description brief rather than wordy.

http://docs.oracle.com/javase/7/docs/api/
Convention in Javadoc is similarly to have the `s`:

jslee02:
Does OMPL have another classes to represent geometric configuration space other than SO(2) such as SE(2), SO(3) and SE(3)? I'm considering adding more base classes for joins that has geometric configuration spaces.

dart/dynamics/Joint.h
jslee02: If it depends on personal preference implicating the subject as "You" or "This function", I'd like to vote on "You" and recommend to have no `s` for consistency for now. If we all agree to have `s` with some strong reason later then let's follow the convention at that time.

Also, I'm aware of that some comments in DART begin with "Return". Let's fix them when we set the convention with documentation.

mkoval:
Yes, OMPL has real, SO(2), SE(2), SO(3), and SE(3) state spaces built in.
They also provide a class that represents the Cartaesian product of other
state spaces. In fact, this is how the SE(2) and SE(3) state spaces are
implemented internally. If none of these fit, you can also implement a
completely custom state space by inheriting from an interface.

When possible, I'd like to do the conversion from DART DegreeOfFreedoms to
single-DOF OMPL state spaces. When this is not possible, e.g. for BallJoint
and the rotational axes of FreeJoint, we will convert the entire joint to a
multi-DOF state space as a single unit.

In that case, we will have to throw an error if you try to plan with a
subset of the coupled DOFs.

jslee02:
Great. I should take a look at the built-in state spaces of OMPL. Thanks.

This PR looks good to be merged.

dart/dynamics/Joint.h
mkoval: @mxgrey's argument makes sense for documenting functions, like getters and setters, but it doesn't work for other types of documentation. For example, like in @psigen's example, a class may have the comment that says something like "represents a rigid body transformation" or "attaches a `Node` to a `Skeleton`."

It doesn't make sense for the developer to the subject of the sentence in those cases. I suppose we could make a separate rule for function documentation (which it sounds like Python does). Or we could always use `s`.

mxgrey: Personally, for classes I always explicitly state the class name, so the subject isn't left to be inferred. But yes, if the subject is being implied by the sentence, and the subject is a class or struct, then having the ``s`` would be necessary.

mkoval:
I fixed the comment style and switched from `isinf` to `isfinite`. Are we good to merge once the Travis and AppVeyor builds finish?

jslee02:
Sure, I'm merging.

PR #444, opened 2015-07-08
More robust SDF parsing -- Patch for Release 5.0

mxgrey:
The SDF parser currently assumes there is at least one Joint in the model, so it fails to correctly parse a model that has no Joints (i.e. all bodies are root nodes).

This pull request fixes that issue, and should be more robust in general because it iterates through bodies instead of joints when constructing a Skeleton.


mkoval:
:+1: This fixed the issue I was having. Thanks @mxgrey!

jslee02:
I added regression test for this. Will wait for build tests and merge if they are happy.

Issue #451, opened 2015-07-15
Assimp is missing symbols on Debian and Ubuntu


mkoval:
There is a major bug ([Debian issue tracker](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775890%3E), [duplicate on Launchpad](https://bugs.launchpad.net/ubuntu/+source/assimp/+bug/1473503)) in the version of Assimp released on Debian and Ubuntu. The packaging incorrectly strips several C++ symbols from the shared library, including the `aiScene` constructor and destructor. This explains the [this Assimp issue](https://github.com/assimp/assimp/issues/203) and why DART has to [define these methods](https://github.com/dartsim/dart/blob/master/dart/dynamics/MeshShape.cpp#L51) itself.
### Impact

DART programmatically constructs an `aiScene` to represent deformable geometry in the `MeshShape` associated with each `SoftBodyNode`. There are a few potential dangers in defining our own constructor and destructor:
1. We cannot [initialize the `mPrivate` pointer](https://github.com/assimp/assimp/blob/f3d418a199cfb7864c826665016e11c65ddd7aa9/code/Version.cpp#L91) in the `aiScene` object because the necessary `ScenePrivateData` class is not exposed in Assimp's public headers. This is assumed to exist in [the exporter](https://github.com/assimp/assimp/blob/f3d418a199cfb7864c826665016e11c65ddd7aa9/code/Exporter.cpp#L319) and will likely lead to SEGFAULT if it is not present.
2. There will be duplicate symbols for these functions if the user upgrades Assimp to a version without the packaging bug (e.g. `3.0~dfsg-4`). It's unclear how catastrophic the implications of this are: I suspect it largely depends on what linker you are using.
3. The default behavior of Assimp's constructors and destructors may change, e.g. if a new member is added to `aiScene`, which could lead to a memory leak.
### Long-Term Fix

This is fixed in the `3.0~dfsg-4` release of the `libassimp` package, which is available for Debian `unstable` and Ubuntu Wiley. I started the [stable release process](https://wiki.ubuntu.com/StableReleaseUpdates) (SRU) on Ubuntu to make this change available for the currently supported distributions (Utopic, Trusty, and Vivid). An `#ubuntu-bugs` team member sponsored the SRU, so the next step is to prepare the updated packages for those repositories. I don't have the background knowledge necessary to do so, this may take a while.
### Changes to DART

[...]

PR #455, opened 2015-07-17
Require Assimp < 3.0~dfsg-4

mkoval:
This pull request makes DART depend on the version of Assimp that is missing C++ symbols; see #451 for details. This should stop DART from breaking when Assimp `3.0~dfsg-4` is released on Ubuntu. It would be good if @jslee02 could release a new version of DART with this change to stave off any potential problems.

Next, I will build the Debian packages for the updated version of Assimp and work with @jslee02 to get them in the DART PPA. Once that is done, we can modify the `debian/control` file to depend on `libassimp-dev (>= 3.0~dfsg-4)` and release a version of DART without the custom `aiScene` methods.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -6,6 +6,7 @@ Build-Depends: debhelper (>= 9),
                libeigen3-dev,
                libfcl-dev (>= 0.2.7),
                libassimp-dev (>= 3),
+               libassimp-dev (<< 3.0~dfsg-4),
                freeglut3-dev,
                libxi-dev,
                libxmu-dev,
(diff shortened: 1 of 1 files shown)


jslee02:
> It would be good if @jslee02 could release a new version of DART with this change to stave off any potential problems.

We have not discussed about the next release schedule yet but I'm thinking to release DART 5.0.1 around Aug 3 that would just include patches. If you want to ship this change in DART 5.0.1, please change the target branch to `release-5.0`. Current `master` branch is for DART 5.1.0 (or 6.0.0).

Note that all the patches of DART 5.0.1 will be applied to the next version of DART as well by merging `release-5.0` into `master`.

Beside the target branch :+1:

mkoval:
Yes, that makes sense. This should be included in 5.0.1. I will make a new pull request against that target.

PR #456, opened 2015-07-17
Removed outdated packaging scripts

mkoval:
This pull request removes the `CPack` commands from `CMakeLists.txt` and the `auto.sh` shell script. From talking to @mxgrey and @jslee02, it sounds like these are outdated and are no longer used to build Debian packages.

I thought it would be best to remove them since they led me on a bit of a wild goose chase while trying to build DART into a `.deb`. If anyone is using them, then please feel free to close this pull request.


jslee02:
I'm not using cpack to build Debian packages since Launchpad requires to build source using `debuild`, and I would like to remove outdated and no one using code in general. However, I'm not sure keeping codes for cpack is totally useless (under assumption we update them). Jose (@j-rivero), who is the build engineer of OSRF, might have some idea on this.

j-rivero:
For using `debbuild` we don't need the CPack support. I don't know the history of these scripts in the DART code base but sounds to me like they were using some time ago to generate .deb packages manually in the same box. I agree that if they are outdated, it is better to remove the support than keep them floating around unsupported. +1.

jslee02:
It makes sense. Then let's remove the outdated scripts and consider to restore back when we find it's necessary. Thanks @j-rivero ! :+1:

PR #457, opened 2015-07-17
Fix typo in Signal

### LINKED ISSUE #442: Typo in Signal::cleanupConnections
`clenaupConnections` should be `cleanupConnections` in `common/Signal.h`.

jslee02:
clenaupConnections --> cleanupConnections

Fixes #442, thanks to @mkoval

### DIFF
--- a/dart/common/Signal.h
+++ b/dart/common/Signal.h
@@ -142,7 +142,7 @@ public:
   void disconnectAll();
 
   /// Cleanup all the disconnected connections
-  void clenaupConnections();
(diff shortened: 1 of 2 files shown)


mkoval:
Technically, since this is slated for 5.1.0, you should add the new method, deprecate `clenaupConnections`, and modify it to call the correct method. We can remove the misspelled version in 6.0.0.

jslee02:
You're right. It was misinterpreted as typo in commet. :( Closing.

PR #458, opened 2015-07-17
Add virtual destructors to Properties in Entity and [Soft]BodyNode

jslee02:


mxgrey:
:+1: 

Any idea why Travis seems to be using an outdated version of Boost?

jslee02:
It seems to fail to download boost 1.55. I'm testing without `--quiet` option to see if it's the case.

jslee02:
The boost download issue looks resolved. Merging.

PR #464, opened 2015-07-22
Support for generic URIs

mkoval:
This pull request adds full-fledged URI support to DART. This generalizes the rudimentary support for `package://` URIs in `DartLoader` to the other file loaders and for loading meshes. I believe that this pull request is fully backwards compatible (with one minor exception), so I'd like to get this into DART 5.1.

The key idea is that external resources should be specified by **URIs, not paths**, nested resources should **resolved relative to their parent URI**, and it should be **possible for the user to add support for custom URIs** without modifying DART.

This pull request adds the following classes:

1. `common::ResourceRetriever` provides a standard interface for opening am URI and returning a...
2. `common::Resource` provides file-like access to a resource using an `fread`-like API
3. `common::LocalResourceRetriever` provides access to local paths and `file://` URIs using the standard library
4. `common::Uri` class that provides URI parsing and merging functionality with no dependencies (painstakingly implemented from RFC 3986)
4. `utils::PackageResourceRetriever` resolves `package://` URIs to `file://` URis using a user-specified mapping. This mimics the behavior of `addPackageDirectory` on `DartLoader`
5. `utils::SchemaResourceRetriever` adds support for using multiple `ResourceRetrievers` at the same time; e.g. to support both `file://` and `package://` URIs.
6. `dynamics::AssimpResourceInputResourceAdaptor` maps between the DART's `Resource` and `ResourceRetriever` classes and the corresponding Assimp `IOFile` and `IOSystem` classes. This allows Assimp to transparently load meshes that consist of multiple files (e.g. external textures).

It also makes the following API changes:

1. `MeshShape` optionally takes a `ResourceRetriever` argument. If specified, this is made available through a `getResourceRetriever` function, e.g. to resolve textures using the same `ResourceRetriever` as was used to load the mesh.
2. `MeshShape` now supports arbitrary URIs, which are accessible using the `getMeshUri` method. If the URI is a `file://` URI, then it is also available (for backwards compatibility) via the `getMeshPath` method.
3. `SdfParser` and `SkelParser` methods now take an optional `ResourceRetriever` argument and use it resolve both the input URI and any embedded URIs. If not specified, the argument defaults to a `LocalResourceRetriever`.
4. `SkelParser` now resolves meshes relative to the `.skel` file instead of assuming that they are in `DART_DATA_PATH`. This technically *breaks backwards compatibility*. However, it makes the class so much more useful that I think it's worth it.
4. `DartLoader` now optionally accepts a `ResourceRetriever` argument (see above). If not specified, it defaults to a `SchemaResourceRetriever` that uses a `LocalResourceRetriever` to resolve `file://` URIs and a `PackageResourceRetriever` to resolve `package://` URIs.
5. `DartLoader`'s `addPackageDirectory` function is deprecated in favor of passing a custom `ResourceRetriever`.
6. `DartLoader` is refactored to use primarily static methods. In DART 6.0, we can remove `addPackageDirectory` and provide an entirely static API.

These features **add no dependencies** if the compiler has full C++11 support. Unfortunately, the version of `libstdc++` shipped with GCC 4.8 (this is the version distributed on Ubuntu 14.04) and below have [a broken implementation of `std::regex`](http://stackoverflow.com/a/12665408/111426). [There isn't a reliable way of testing the version of `libstdc++`](http://stackoverflow.com/q/31506594/111426) in both GCC and Clang, so I fall back on Boost.Regex if I detect `libstdc++` being used at all.

This is a big pull request, so I am sure I missed something. Please post if you have any questions or see anything that may break backwards compatibility.


jslee02:
Please merge master branch into this branch so that Travis-CI is happy.

mkoval:
@jslee02 I merged `master` into the branch. I did some cleanup that I missed in my initial pull request, so you may want to take another look at the diff.

jslee02:
Thanks! The purpose of this pull request looks nice. I tried to implement similar thing but this seems more comprehensive than mine. I am currently interpreting your implementation, and will leave feedback soon.

dart/common/LocalResource.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+LocalResource::~LocalResource()
+{
+  if (!mFile)
+    return;
+
+  if (std::fclose(mFile) == EOF)
+  {
+    dtwarn << "[LocalResource::destructor] Failed closing file.\n";
```
mxgrey: We should probably print ``std::strerror(errno)`` here. According to [this reference](http://linux.die.net/man/3/fclose) ``errno`` should get set when an error occurs.

dart/common/LocalResource.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return 0;
+  }
+
+  // The SEEK_END option is not required by the C standard. However, it is
+  // required by POSIX.
+  // TODO: Does this work on Windows?
+  if(std::fseek(mFile, 0, SEEK_END) || std::ferror(mFile))
+  {
+    dtwarn << "[LocalResource::getSize] Unable to compute file size: Failed"
+             " seeking to the end of the file.\n";
```
mxgrey: I think ``fseek`` also sets ``errno``.

dart/common/LocalResource.cpp
```diff
@@ -0,0 +1,139 @@
...
+size_t LocalResource::tell()
+{
+  if(!mFile)
+    return 0;
+  
+  const long offset = std::ftell(mFile);
+  if (offset == -1L)
+  {
+    dtwarn << "[LocalResource::tell] Failed"
+             " seeking to the current position.\n";
```
mxgrey: Could use a ``std::strerror(errno)`` here as well, I think.

dart/common/LocalResource.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+size_t LocalResource::read(void *_buffer, size_t _size, size_t _count)
+{
+  if (!mFile)
+    return 0;
+
+  const size_t result = std::fread(_buffer, _size, _count, mFile);
+  if (std::ferror(mFile)) 
+  {
+    dtwarn << "[LocalResource::tell] Failed reading file.\n";
```
mxgrey: This should be ``[LocalResource::read]``.

dart/common/LocalResourceRetriever.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace dart {
+namespace common {
+
+bool LocalResourceRetriever::exists(const std::string& _uri)
+{
+  common::Uri uri;
+  if(!uri.fromString(_uri))
+  {
+    dtwarn << "[exists] Failed parsing URI: " << _uri << "\n";
```
mxgrey: ``[LocalResourceRetriever::exists]`` would follow our usual pattern.

dart/common/LocalResourceRetriever.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  return std::ifstream(*uri.mPath, std::ios::binary).good();
+}
+
+common::ResourcePtr LocalResourceRetriever::retrieve(const std::string& _uri)
+{
+  common::Uri uri;
+  if(!uri.fromString(_uri))
+  {
+    dtwarn << "[exists] Failed parsing URI: " << _uri << "\n";
```
mxgrey: This should be ``[LocalResourceRetriever::retrieve]``

dart/common/Uri.cpp
```diff
@@ -0,0 +1,435 @@
...
+      input = input.substr(3);
+    }
+    else if(startsWith(input, "./"))
+    {
+      input = input.substr(2);
+    }
+    // B.  if the input buffer begins with a prefix of "/./" or "/.",
+    //     where "." is a complete path segment, then replace that
+    //     prefix with "/" in the input buffer; otherwise,
+    else if(input == "/.")
```
mxgrey: I'm a little bit confused about this. What if a relative file is specified like ``../.hiddenFile.txt`` where I want to load a file whose name starts with a ``.`` (because I want to keep it hidden for some reason)? Wouldn't this clobber the file name and look for ``../hiddenFile.txt`` instead?

psigen: This should only affect a full path string of `/.`.  It's a strict equality check.  `"/.hiddenfile.txt" != "/."`, right?

mxgrey: Yep, I realized that just before you replied.

mkoval: This first `if` statement implements the "or "/.", where "." is a complete path segment" clause. I believe this can only happen if the path ends with `/.`. In the case you described, the path segment is `.hiddenFile.txt`.

These comments are copied and pasted from [Section 5.2.4 of RFC-3986](https://tools.ietf.org/html/rfc3986#section-5.2.4), so we'd have to chat with Tim Berners-Lee if it's incorrect. :smile:

dart/common/LocalResource.cpp
mkoval: Interesting. The [documentation I found](http://www.cplusplus.com/reference/cstdio/fclose/) just says that the function returns `EOF`. It can't hurt to check `errno` and, if it's set, print the `std::strerror(errno)` anyway.

mxgrey: The reference I linked to is basically an internet copy of the GNU Linux Programmer's manual, so it should be legit.

dart/common/LocalResource.cpp
mkoval: The [documentation I found](http://en.cppreference.com/w/cpp/io/c/fseek) says only that:

> If a read or write error occurs, the error indicator for the stream (`std::ferror`) is set and the file position is unaffected.

I don't think there is any guarantee that the value returned by `ferror` has the same meaning as `errno`. 

Again, the documentation [that you found](http://linux.die.net/man/3/fseek) says that `fseek` sets `errno`. I'm not sure what the deal is.

dart/common/LocalResource.cpp
mkoval: It might be true on GNU/Linux, but I don't know if it's guaranteed by the C++ standard.

dart/common/Uri.cpp
```diff
@@ -0,0 +1,435 @@
...
+    return "";
+}
+
+std::string Uri::getRelativeUri(
+  const std::string& _base, const std::string& _relative, bool _strict)
+{
+  Uri baseUri;
+  if(!baseUri.fromString(_base))
+  {
+    dtwarn << "[getRelativeUri] Failed parsing base URI '"
```
mxgrey: Since this file contains two classes (``Uri`` and ``UriComponent``) it would probably be good to be explicit about the function ``[Uri::getRelativeUri]`` so the user knows more exactly where the warning is coming from.

mkoval: I'd actually prefer to replace `UriComponent` with `std::optional`. Unfortunately, [`std::optional` is still experimental](http://en.cppreference.com/w/cpp/experimental/optional) and isn't slated for inclusion until C++17. We could either keep `UriComponent` around or switch to `boost::optional`. Thoughts?

mxgrey: Personally, I'd rather keep ``UriComponent``, at least until ``std::optional`` is available. In my perfect world, the core DART functionality would not have any ``boost`` dependencies, so I'd like to keep any usage of ``boost`` to a minimum in the vain hope that one day the core of DART won't require ``boost`` at all.

I don't have anything against ``boost`` itself, but it's one of those packages that can be a major pain to manage correctly, especially on Windows. ``Boost`` put me through so much suffering the first time I tried to compile DART on my Windows machine.

dart/common/Uri.h
```diff
@@ -0,0 +1,136 @@
...
+  reference_type get_value_or(reference_type _default);
+  reference_const_type get_value_or(reference_const_type _default) const;
+
+private:
+  bool mExists;
+  std::string mValue;
+};
+
+
+class Uri {
```
mxgrey: I wonder if we should mark this as ``final``. It doesn't have a virtual destructor, so I'm thinking it's not meant to be inherited.

dart/dynamics/AssimpInputResourceAdaptor.cpp
```diff
@@ -0,0 +1,207 @@
...
+namespace dart {
+namespace dynamics {
+
+/*
+ * AssimpInputResourceRetrieverWrapper
+ */
+AssimpInputResourceRetrieverAdaptor::AssimpInputResourceRetrieverAdaptor(
+      const common::ResourceRetrieverPtr& _resourceRetriever)
+  : mResourceRetriever(_resourceRetriever)
+{
```
mxgrey: I typically put a ``// Do nothing`` in places like this to make it clear that the programmer didn't simply forget to implement it.

dart/utils/PackageResourceRetriever.cpp
```diff
@@ -0,0 +1,109 @@
...
+{
+  static const std::vector<std::string> empty_placeholder;
+
+  // Lookup the corresponding package path.
+  const auto it = mPackageMap.find(_packageName);
+  if(it != std::end(mPackageMap))
+    return it->second;
+  else
+  {
+    dtwarn << "[PackageResourceResolver::retrieve] Unable to resolve path to"
```
mxgrey: This should be ``[PackageResourceResolver::getPackagePaths]`` since it can be called from both ``retrieve`` and ``exists``.

dart/common/Uri.cpp
psigen: If we already have `UriComponent`, I don't think it's a net win to swap it out for an extra `boost` library right now for the reasons @mxgrey cited.

dart/utils/SchemaResourceRetriever.h
```diff
@@ -0,0 +1,92 @@
...
+#include "dart/common/ResourceRetriever.h"
+
+namespace dart {
+namespace utils {
+
+/// SchemaResourceRetriever allows multiple \ref ResourceRetriever to be used
+/// interchangably by: (1) associating each \ref ResourceRetriever with a
+/// particular URI schema and/or (2) providing a precedence order for trying
+/// multiple retrievers.
+class SchemaResourceRetriever : public virtual common::ResourceRetriever
```
mxgrey: I wonder if something like ``CompositeResourceRetriever`` might be a better name for this. Calling it ``SchemaResourceRetriever`` makes it sound like it's retrieving schemas.

mxgrey:
Overall, awesome pull request, A+. I almost want to pull this stuff out into its own library, maybe make it a submodule of DART, because I can imagine it being super useful for a lot of different projects unrelated to DART. But I won't worry about that for the time being.

@jslee02 had an awesome idea of using this to store the Skeleton models outside of DART's repo. We could have a specialized resource retriever that checks if the model is on disk; then if it's not on disk, it will retrieve the model from the ``http://`` and save it to disk. This way, you'd only download the models that you actually use, and we can cut back the size of the repo by 73%.

jslee02:
> @jslee02 had an awesome idea of using this to store the Skeleton models outside of DART's repo. We could have a specialized resource retriever that checks if the model is on disk; then if it's not on disk, it will retrieve the model from the http:// and save it to disk.

Actually, I just borrowed the idea of what Gazebo does. I would like to call the specialized resource retriever as `CloudResourceRetriever` (or `WebResourceRetriever`).

`CloudResourceRetriever` would have two locations: (a) local location to save the downloaded resource and (b) remote location the original models are stored (another repo outside of DART's repo). The URI scheme for cloud resource would be `dart_model://` or `model://`. `CloudResourceRetriever` then should be able to retrieve Skeleton models from the local location like `PackageResourceRetriever` does now, and download Skeleton models from the remote repo. The locations will be provided as system environment variables when DART is installed.

Also, I'd like to introduce Skeleton file versioning. Then `CloudResourceRetriever` would be able to update the Skeleton model if the downloaded model is outdated.

Compounding several retrievers into `SchemaResourceRetriever` we would be able to load Skeleton models with mesh files through following ways:
```xml
                <visualization_shape>
                    <geometry>
                        <mesh>
                            <!--1. deprecated way-->
                            <!--file_name>obj/foot.obj</file_name-->

                            <!--2. model scheme-->
                            <uri>model://model/mesh.dae</uri>

                            <!--3. file scheme with absolute path-->
                            <uri>file://my/local/location/model/mesh.dae</uri>

                            <!--4. just absolute path-->
                            <uri>/my/local/location/model/mesh.dae</uri>

                            <!--5. relative path-->
                            <uri>./mesh.dae</uri>
                        </mesh>
                    </geometry>
                </visualization_shape>
```

dart/common/LocalResource.cpp
```diff
@@ -0,0 +1,139 @@
+#include <cstring>
```
jslee02: License block is missing.

dart/common/LocalResource.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <iostream>
+#include "dart/common/Console.h"
+#include "LocalResource.h"
+
+namespace dart {
+namespace common {
+
```
jslee02: We usually place horizontal separator before function definition as:
```cpp
LocalResource::LocalResource(const std::string& _path)
{
  // ...
}
```

dart/common/LocalResource.cpp
```diff
@@ -0,0 +1,139 @@
...
+LocalResource::LocalResource(const std::string& _path)
+  : mFile(std::fopen(_path.c_str(), "rb"))
+{
+  if(!mFile)
+  {
+    dtwarn << "[LocalResource::constructor] Failed opening file '" << _path
+          << "' for reading: " << std::strerror(errno) << "\n";
```
jslee02: vertical alignment

dart/common/LocalResource.h
```diff
@@ -0,0 +1,74 @@
...
+ *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
```
jslee02: please add empty line after license block

dart/common/LocalResource.h
```diff
@@ -0,0 +1,74 @@
...
+#ifndef DART_COMMON_LOCALRESOURCE_H_
+#define DART_COMMON_LOCALRESOURCE_H_
```
jslee02: please add empty line after header guard

dart/common/LocalResource.h
```diff
@@ -0,0 +1,74 @@
...
+#include "dart/common/Resource.h"
+
+namespace dart {
+namespace common {
+
+class LocalResource : public virtual Resource
+{
+public:
+  LocalResource(const std::string& _path);
```
jslee02:
```cpp
explicit LocalResource(const std::string& _path);
```
to prevent unintended implicit type conversion

dart/common/LocalResource.h
```diff
@@ -0,0 +1,74 @@
...
+  size_t getSize() override;
+
+  // Documentation inherited.
+  size_t tell() override;
+
+  // Documentation inherited.
+  bool seek(ptrdiff_t _origin, SeekType _mode) override;
+
+  // Documentation inherited.
+  size_t read(void *_buffer, size_t _size, size_t _count) override;
```
jslee02: please put * right after type as:
```cpp
size_t read(void* _buffer, size_t _size, size_t _count) override;
```

dart/common/LocalResource.h
```diff
@@ -0,0 +1,74 @@
...
+
+private:
+  std::FILE* mFile;
+};
+
+} // namespace common
+} // namespace dart
+
+#endif // ifndef DART_COMMON_LOCALRESOURCERETRIEVER_H_
```
jslee02:
```cpp
#endif // ifndef DART_COMMON_LOCALRESOURCE_H_
```

dart/common/LocalResourceRetriever.cpp
```diff
@@ -0,0 +1,48 @@
+#include <iostream>
```
jslee02: license block missing

dart/common/LocalResourceRetriever.h
```diff
@@ -0,0 +1,62 @@
...
+ *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
```
jslee02: please add empty line

dart/common/LocalResourceRetriever.h
```diff
@@ -0,0 +1,62 @@
...
+#ifndef DART_COMMON_LOCALRESOURCERETRIEVER_H_
+#define DART_COMMON_LOCALRESOURCERETRIEVER_H_
```
jslee02: please add empty line

dart/common/LocalResourceRetriever.h
```diff
@@ -0,0 +1,62 @@
...
+  virtual ~LocalResourceRetriever() = default;
+
+  // Documentation inherited.
+  bool exists(const std::string& _uri) override;
+
+  // Documentation inherited.
+  ResourcePtr retrieve(const std::string& _uri) override;
+};
+
+typedef std::shared_ptr<LocalResourceRetriever> LocalResourceRetrieverPtr;
```
jslee02: How about using `using` rather than `typedef` for type alias as:
```cpp
using LocalResourceRetrieverPtr = std::shared_ptr<LocalResourceRetriever>;
```

They are same but [it seems `using` is more preferred](http://stackoverflow.com/questions/24916435/why-do-type-aliases-in-c-use-using-instead-of-typedef-in-their-syntax).

dart/common/Resource.h
```diff
@@ -0,0 +1,88 @@
...
+ *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
```
jslee02: empty line

dart/common/Resource.h
```diff
@@ -0,0 +1,88 @@
...
+#ifndef DART_COMMON_RESOURCE_H_
+#define DART_COMMON_RESOURCE_H_
```
jslee02: empty line

dart/common/Resource.h
```diff
@@ -0,0 +1,88 @@
...
+  /// \note This method has the same API as the standard fread function.
+  virtual size_t read(void *_buffer, size_t _size, size_t _count) = 0; 
+};
+
+typedef std::shared_ptr<Resource> ResourcePtr;
+
+} // namespace common
+} // namespace dart
+
+#endif // ifndef DART_UTILS_RESOURCERETRIEVER_H_
```
jslee02:
```cpp
#endif // ifndef DART_COMMON_RESOURCE_H_
```

dart/common/ResourceRetriever.h
```diff
@@ -0,0 +1,63 @@
...
+ *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
```
jslee02: empty line

dart/common/ResourceRetriever.h
```diff
@@ -0,0 +1,63 @@
...
+#include <memory>
+#include <string>
+#include "Resource.h"
+
+namespace dart {
+namespace common {
+
+/// ResourceRetriever provides methods for testing for the existance of and
+/// accessing the content of a resource specified by URI.
+struct ResourceRetriever {
```
jslee02: Is there a special reason struct is used rather than class? I know struct is fine since it contains only public members, but this has derived classes so it would be more consistent with class here.

dart/common/ResourceRetriever.h
```diff
@@ -0,0 +1,63 @@
...
+  /// \brief Return the resource specified by a URI or nullptr on failure.
+  virtual ResourcePtr retrieve(const std::string& _uri) = 0;
+};
+
+typedef std::shared_ptr<ResourceRetriever> ResourceRetrieverPtr;
+
+} // namespace common
+} // namespace dart
+
+#endif // ifndef DART_UTILS_RESOURCERETRIEVER_H_
```
jslee02:
```cpp
#endif // ifndef DART_COMMON_RESOURCERETRIEVER_H_
```

dart/common/Uri.cpp
```diff
@@ -0,0 +1,435 @@
+#include <cassert>
```
jslee02: license block missing

dart/common/Uri.cpp
jslee02: +1

dart/common/Uri.h
```diff
@@ -0,0 +1,136 @@
...
+ *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
```
jslee02: empty line

dart/common/Uri.h
```diff
@@ -0,0 +1,136 @@
...
+#ifndef DART_COMMON_URI_H_
+#define DART_COMMON_URI_H_
+
+namespace dart {
+namespace common {
+
+class UriComponent {
```
jslee02:
```cpp
class UriComponent 
{
```

dart/common/Uri.h
jslee02:
```cpp
class Uri
{
```

dart/common/Uri.h
```diff
@@ -0,0 +1,136 @@
...
+private:
+  bool mExists;
+  std::string mValue;
+};
+
+
+class Uri {
+public:
+  /// Scheme, e.g. 'http', 'file', 'package'
+  UriComponent mScheme;
```
jslee02: We should consider to encapsulate these public member variables.

dart/dynamics/AssimpInputResourceAdaptor.cpp
jslee02: +1

dart/dynamics/AssimpInputResourceAdaptor.h
```diff
@@ -0,0 +1,130 @@
...
+ *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
```
jslee02: empty line

dart/dynamics/AssimpInputResourceAdaptor.h
```diff
@@ -0,0 +1,130 @@
...
+#ifndef DART_DYNAMICS_ASSIMPINPUTRESOURCEADAPTOR_H_
+#define DART_DYNAMICS_ASSIMPINPUTRESOURCEADAPTOR_H_
```
jslee02: empty line

dart/dynamics/AssimpInputResourceAdaptor.h
```diff
@@ -0,0 +1,130 @@
...
+#include "dart/common/Resource.h"
+#include "dart/common/ResourceRetriever.h"
+
+namespace dart {
+namespace dynamics {
+
+class AssimpInputResourceRetrieverAdaptor : public Assimp::IOSystem
+{
+public:
+  AssimpInputResourceRetrieverAdaptor(
```
jslee02:
```cpp
explicit AssimpInputResourceRetrieverAdaptor(
```

dart/dynamics/AssimpInputResourceAdaptor.h
```diff
@@ -0,0 +1,130 @@
...
+  void Close(Assimp::IOStream* pFile) override;
+
+private:
+  common::ResourceRetrieverPtr mResourceRetriever;
+};
+
+class AssimpInputResourceAdaptor : public Assimp::IOStream
+{
+public:
+  AssimpInputResourceAdaptor(const common::ResourcePtr& _resource);
```
jslee02:
```cpp
explicit AssimpInputResourceAdaptor(const common::ResourcePtr& _resource);
```

dart/dynamics/AssimpInputResourceAdaptor.h
```diff
@@ -0,0 +1,130 @@
...
+  /// \brief Not implemented. This is a read-only stream.
+  void Flush() override;
+
+private:
+  common::ResourcePtr mResource;
+};
+
+
+aiFileIO createFileIO(Assimp::IOSystem* adaptor);
+aiFile createFile(Assimp::IOStream* adaptor);
```
jslee02: Could we make these functions as static member functions of `AssimpInputResourceRetrieverAdaptor` and `AssimpInputResourceAdaptor` to make it clear these functions are related to them?

dart/dynamics/MeshShape.h
```diff
@@ -51,8 +58,11 @@ namespace dynamics {
 class MeshShape : public Shape {
 public:
   /// \brief Constructor.
-  MeshShape(const Eigen::Vector3d& _scale, const aiScene* _mesh,
-            const std::string &path = std::string());
+  MeshShape(
+    const Eigen::Vector3d& _scale,
+    const aiScene* _mesh,
+    const std::string &_path = "",
```
jslee02:
```cpp
const std::string& _path = "",
```

dart/dynamics/MeshShape.h
```diff
@@ -61,7 +71,12 @@ class MeshShape : public Shape {
   const aiScene* getMesh() const;
 
   /// \brief
-  void setMesh(const aiScene* _mesh, const std::string &path = std::string());
+  void setMesh(
+    const aiScene* _mesh, const std::string &path = "",
```
jslee02:
```cpp
    const aiScene* _mesh, const std::string& path = "",
```

dart/utils/PackageResourceRetriever.cpp
```diff
@@ -0,0 +1,109 @@
+#include <cassert>
```
jslee02: license block missing

dart/utils/PackageResourceRetriever.cpp
```diff
@@ -0,0 +1,109 @@
...
+  if (_uri.find(schema) != 0)
+    return false;
+
+  // Split the URI into package and relative path components.
+  const size_t packageIndex = schema.size();
+  const size_t pathIndex = _uri.find_first_of('/', packageIndex);
+
+  if(pathIndex == std::string::npos)
+  {
+    dtwarn << "[PackageResourceRetriever::retrieve] Unable to find package"
```
jslee02: This should be [PackageResourceRetriever::resolvePackageUri] since it can be called from both `retrieve` and `exists`.

dart/utils/PackageResourceRetriever.h
```diff
@@ -0,0 +1,111 @@
...
+ *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
```
jslee02: empty line

dart/utils/PackageResourceRetriever.h
```diff
@@ -0,0 +1,111 @@
...
+#ifndef DART_UTILS_PACKAGERESOURCERETRIEVER_H_
+#define DART_UTILS_PACKAGERESOURCERETRIEVER_H_
```
jslee02: empty line

dart/utils/SchemaResourceRetriever.cpp
```diff
@@ -0,0 +1,106 @@
+#include <iostream>
```
jslee02: license block missing

dart/utils/SchemaResourceRetriever.h
```diff
@@ -0,0 +1,92 @@
...
+ *   DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ *   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ *   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
```
jslee02: empty line

dart/utils/SchemaResourceRetriever.h
```diff
@@ -0,0 +1,92 @@
...
+#ifndef DART_UTILS_SCHEMARESOURCERETRIEVER_H_
+#define DART_UTILS_SCHEMARESOURCERETRIEVER_H_
```
jslee02: empty line

dart/utils/SchemaResourceRetriever.h
jslee02: +1

dart/utils/urdf/DartLoader.cpp
```diff
@@ -27,36 +23,72 @@
 #include "dart/simulation/World.h"
 #include "dart/utils/urdf/urdf_world_parser.h"
 
+using ModelInterfacePtr = boost::shared_ptr<urdf::ModelInterface>;
+
+static std::string getBasePath(const std::string& _uri)
```
jslee02: I wonder this function is necessary since it's not used anywhere.

dart/utils/urdf/DartLoader.cpp
```diff
@@ -27,36 +23,72 @@
...
+
 namespace dart {
 namespace utils {
 
+DartLoader::DartLoader()
+  : mLocalRetriever(new common::LocalResourceRetriever)
+  , mPackageRetriever(new utils::PackageResourceRetriever(mLocalRetriever))
+  , mRetriever(new utils::SchemaResourceRetriever)
+{
+  using namespace std::placeholders;
```
jslee02: Do we need this line?

dart/utils/urdf/DartLoader.cpp
```diff
@@ -27,36 +23,72 @@
...
+    return _uri.substr(0, index);
+  else
+    return "";
+}
+
 namespace dart {
 namespace utils {
 
+DartLoader::DartLoader()
+  : mLocalRetriever(new common::LocalResourceRetriever)
```
jslee02:
```cpp
  : mLocalRetriever(new common::LocalResourceRetriever()),
    mPackageRetriever(new utils::PackageResourceRetriever(mLocalRetriever)),
    mRetriever(new utils::SchemaResourceRetriever())
```

dart/utils/Parser.cpp
```diff
@@ -585,25 +586,42 @@ tinyxml2::XMLElement* getElement(tinyxml2::XMLElement* _parentElement,
...
+      dtwarn << "[openXMLFile] Failed opening URI '"
+             << filename << "'.\n";
+      throw std::runtime_error("Failed opening URI.");
+    }
+
+    // C++11 guarantees that std::string has contiguous storage.
+    const size_t size = resource->getSize();
+    std::string content;
+    content.resize(size);
+    if(resource->read(&content.front(), size, 1) != 1)
```
jslee02: I found similar code in `DartLoader::readFileToString()`.

It would be useful and safer if we provide `Resource::read(std::string& content, size_t size, size_t count)` that stores that file contents into std::string rather than writing similar code in different places.

jslee02:
Most comments I made in the code are about code style. Strictly speaking, we don't have written code convention and furthermore I've followed different code styles myself so far. I tried to comment for something settled styles but you can skip those style issue for now. We might begin to write up our code convention in not too far future and check code style more strictly from that time.

Also, note that `CloudResourceRetriever` doesn't need to be part of this pull request.

dart/dynamics/AssimpInputResourceAdaptor.h
mkoval: This function can take *any* `Assimp::IOSystem` and wrap it in the C API. It is not specific to `AssimpInputResourceRetrieverAdaptor`. Therefore, I think it should remain a free function.

dart/utils/Parser.cpp
mkoval: Adding that method would you let you omit the `resize()` line here. I don't think this justified adding another method to the `Resource` interface.

If we find that we need this function somewhere else, then I propose that we add a helper function somewhere in `utils`.

mkoval:
I think I addressed all of your comments. Please take another look and let me know if I missed anything :sweat:

dart/common/ResourceRetriever.h
```diff
@@ -0,0 +1,65 @@
...
+#include <memory>
+#include <string>
+#include "Resource.h"
+
+namespace dart {
+namespace common {
+
+/// ResourceRetriever provides methods for testing for the existance of and
+/// accessing the content of a resource specified by URI.
+class ResourceRetriever {
```
jslee02:
```cpp
class ResourceRetriever
{
```

dart/utils/CompositeResourceRetriever.h
```diff
@@ -0,0 +1,94 @@
...
+    const common::ResourceRetrieverPtr& _resourceRetriever);
+
+  // Documentation inherited.
+  bool exists(const std::string& _uri) override;
+
+  // Documentation inherited.
+  common::ResourcePtr retrieve(const std::string& _uri) override;
+
+private:
+  std::vector<common::ResourceRetrieverPtr> getRetrievers(const std::string& _uri);
```
jslee02: Exceeded 80 columns. Please wrap this line as:
```cpp
  std::vector<common::ResourceRetrieverPtr> getRetrievers(
      const std::string& _uri);
```

dart/utils/CompositeResourceRetriever.h
```diff
@@ -0,0 +1,94 @@
...
+
+  std::string getSchema(const std::string& _uri);
+
+  std::unordered_map<std::string,
+    std::vector<common::ResourceRetrieverPtr> > mResourceRetrievers;
+  std::vector<common::ResourceRetrieverPtr> mDefaultResourceRetrievers;
+};
+
+typedef std::shared_ptr<CompositeResourceRetriever> CompositeResourceRetrieverPtr;
```
jslee02: Exceeded 80 columns. Please wrap this line as:
```cpp
using CompositeResourceRetrieverPtr
    = std::shared_ptr<CompositeResourceRetriever>;
```

dart/utils/PackageResourceRetriever.h
```diff
@@ -0,0 +1,113 @@
...
+  common::ResourceRetrieverPtr mLocalRetriever;
+  std::unordered_map<std::string, std::vector<std::string> > mPackageMap;
+
+  const std::vector<std::string>& getPackagePaths(
+    const std::string& _packageName);
+  bool resolvePackageUri(const std::string& _uri,
+    std::string& _packageName, std::string& _relativePath);
+};
+
+typedef std::shared_ptr<PackageResourceRetriever> PackageResourceRetrieverPtr;
```
jslee02:
```cpp
using PackageResourceRetrieverPtr = std::shared_ptr<PackageResourceRetriever>;
```

jslee02:
Thanks @mkoval ! I found more minor formatting issues. Once these are modified then we could merge this PR if there is no other feedback.

mxgrey:
Looks good to me :+1:

mkoval:
I think I fixed the remaining formatting issues.

jslee02:
:+1: Nice work!

Issue #411, opened 2015-06-17
Windows Issues with current Master build


jturner65:
I've mentioned these in an email to JS but I will list them here as well.  Didn't think these warranted their own issues, but if so, I apologize.

1) NOMINMAX :
could all the #include <windows.h> be replaced with something along the following?
(from http://stackoverflow.com/questions/4913922/possible-problems-with-nominmax-on-visual-c/4914108#4914108 )
	#ifdef NOMINMAX
	#include <windows.h>
	#else
	#define NOMINMAX
	#include <windows.h>
	#undef NOMINMAX
	#endif

2) "small" variable name :
another <windows.h> issue (sigh)

http://stackoverflow.com/questions/27793470/why-does-smallgive-an-error-about-char
could the variable named "small" in Win3D.cpp line 77 be changed to something else?  windows.h includes a file which in turn #define's "small"

3) GL_MULTISAMPLE
http://stackoverflow.com/questions/4207506/where-is-gl-multisample-defined
this macro is not defined in windows, and is apparently outdated.

jslee02:
1) and 2) are resolved in [windows_build_issues branch](https://github.com/dartsim/dart/tree/windows_build_issues).

I learned that GL_MULTISAMPLE is supported since OpenGL 1.3 from [here](https://www.opengl.org/wiki/History_of_OpenGL#OpenGL_1.3_.282001.29) but I couldn't find any document says it's outdated. Can you point out any reference that says which version of OpenGL is used in windows and if GL_MULTISAMPLE is supported? It seems there are many changes since OpenGL 3.0 but I still stay in older version. :S

jturner65:
sure.  here’s a SO link that discusses it :
http://stackoverflow.com/questions/4207506/where-is-gl-multisample-defined

i guess a conditional redefinition of it could solve the problem.

mkoval:
@jturner65 That post actually says exactly the opposite (in one of the comments):

> Umm... That should read "used to be an extension", multisampling's been a part of OpenGL core for a long time now.

In any case, there will need to be a workaround to make it work on Windows. Regardless of what the standard says. :smile:

jturner65:
i think you meant to quote JS.  :)

mxgrey:
It looks like the SO thread is saying that GL_MULTISAMPLE *is* part of the OpenGL specifications (not outdated), but was not included in Windows for mysterious reasons.

mkoval:
[...]

PR #482, opened 2015-08-01
Fix Windows build issues: NOMINMAX and "small" variable name

jslee02:
This pull request partially resolves Windows build issues reported in #411: NOMINMAX and "small" variable name.


mxgrey:
:+1: 

I ran into a similar issue with the ``min`` macro on Windows in pull request #461. My lazy solution was to simply ``#define NOMINMAX`` all over the place, but I should probably do something closer to this if possible.

jslee02:
I got the solution from #411 thanks to @jturner65. I just wish Windows is compatible with standard C++. :)

jturner65:
yeah, me too @jslee02 .I hope it doesn't take M$ so long in the future to adopt subsequent standards upgrades.  I am glad I could help you guys a bit. I will be glad to check on any windows-specific issues in the future, if you guys need any assistance.

PR #484, opened 2015-08-04
Add Disqus comment box and Facebook "like" and "share" buttons to the tutorial pages

jslee02:
This pull request adds [Disqus](https://disqus.com/) comment box and [Facebook](https://www.facebook.com/) "like" and "share" buttons to the tutorial pages. The main reason Disqus is selected is that it allows anonymous comments. These components can be added to new pages by adding following script:
```html
<div id="fb-root"></div>
<script>(function(d, s, id) {
  var js, fjs = d.getElementsByTagName(s)[0];
  if (d.getElementById(id)) return;
  js = d.createElement(s); js.id = id;
  js.src = "//connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.4";
  fjs.parentNode.insertBefore(js, fjs);
}(document, 'script', 'facebook-jssdk'));</script>

<div class="fb-like" data-href="http://dart.readthedocs.org/en/release-5.1/tutorials/introduction/" data-layout="button_count" data-action="like" data-show-faces="true" data-share="true"></div>

<div id="disqus_thread"></div>
<script type="text/javascript">
    /* * * CONFIGURATION VARIABLES * * */
    var disqus_shortname = 'dartsim';
    
    /* * * DON'T EDIT BELOW THIS LINE * * */
    (function() {
        var dsq = document.createElement('script'); dsq.type = 'text/javascript'; dsq.async = true;
        dsq.src = '//' + disqus_shortname + '.disqus.com/embed.js';
        (document.getElementsByTagName('head')[0] || document.getElementsByTagName('body')[0]).appendChild(dsq);
    })();
</script>
<noscript>Please enable JavaScript to view the <a href="https://disqus.com/?ref_noscript" rel="nofollow">comments powered by Disqus.</a></noscript>
```
Note that ```/release-5.1/tutorials/introduction/``` should be updated for the new pages.

Here is a screenshot:
image
and an [example page](http://dart.readthedocs.org/en/tutorials_latex_test/tutorials/introduction/).

Project settings can be modified at https://dartsim.disqus.com if you have the privilege.


mxgrey:
Hah, that's pretty awesome. :+1:

psigen:
@mkoval and I are very skeptical about the need for facebook integration on a documentation page...
:no_mobile_phones:

mxgrey:
It's meant to allow people to comment on the pages, for Q&A purposes.

I guess if this is mandatorily tied into Facebook accounts, then that would not be ideal. At least having an option for anonymity would probably be better.

psigen:
:+1: for `Feedback` mechanism, :-1: for Facebook login, which I believe this solution does require.

jslee02:
It seems Facebook comment systems doesn't allow anonymous comments. There is an alternative comment system [Disqus](https://disqus.com/) which allows anonymous comments (called guest comment in Disqus). I chose Facebook over Disqus because Facebook looks more neat in terms of design to me, but anonymity is more important.

Replacing Facebook comment boxes with Disqus's ones.

mxgrey:
Do you happen to know if there's a way to get notifications for new comments with Disqus (or with any other service)? It would be unfortunate if someone left a question there and none of us ever knew to answer it, because expert users don't check the tutorial pages regularly... or ever.

jslee02:
Yes, there is a way. In order to get notifications, you need to sign-up Disqus and be a moderator of [DART community](https://dartsim.disqus.com). I can give you the privilege.

By the way, we still can keep the "like" and "share" buttons even though we don't use Facebook comment box. I think it could be one way to let people hear about DART through Facebook community. Thought?

mxgrey:
Keeping the ``like`` and the ``share`` sounds totally fine to me. That way people can easily promote DART if they want to. As long as the commenting system allows for anonymity, I don't think there should be any problems.

mkoval:
I can't think of a situation where I'd "Like" or "Share" a particular documentation page. It might make more sense to put one module (?), for DART as a whole, on each page. It makes a lot more sense to "Like" DART than to like the "DART 5.0 Installation Instructions". :smile:

mxgrey:
But... what if the doc page is like.... *really* good? You would want all your friends and relatives knowing about it? :smirk:

psigen:
Just to muddy the waters a bit, here is a list of blog commenting systems that I compiled:

* Facebook
* Google+
* Disqus
* Jetpack
* Livefyre
* Intense Debate
* Vicomi

And some discussions about the pros/cons of each:
* http://dustn.tv/blog-commenting-systems/
* http://coschedule.com/blog/comment-systems/
* http://www.hongkiat.com/blog/3rdparty-comment-discuss-systems-reviewed/

mkoval:
@psigen All of those links are generally positive about Disqus. It seems like a reasonable option to me.

psigen:
Seems so.  Well I'm satisfied that Disqus is a reasonable option to try.

jslee02:
Yeah, Disqus looks reasonable. Merging.

PR #485, opened 2015-08-05
Initialize mRestPosition if zero is not in range.

mkoval:
This pull request fixes the bug that I fixed in #438. This was intended to be fixed with the addition of `mInitialPositions` in #449. However, that pull request only sets `mInitialPositions`, not `mRestPositions`, in the URDF and SDF loaders.

This pull request restores the same logic to `mRestPositions`.


psigen:
@mxgrey: What is mRestPosition used for?

mxgrey:
I suppose this is alright. I wouldn't normally assume that the rest position should be the same as the initial position, but since we don't have a way of reading a rest position from SDF or URDF, we may as well make the assumption for lack of a better alternative. :+1:

jslee02:
:+1:

PR #490, opened 2015-08-16
Fixed segfault in Linkage::Criteria -- Patch for 5.0

mxgrey:


dart/dynamics/Linkage.cpp
```diff
@@ -319,7 +319,7 @@ void Linkage::Criteria::expandToTarget(
   }
 
   // If we have successfully reached the target, expand from there
-  if(newBns.back() == _target.mNode.lock())
+  if(newBns.size() > 0 && newBns.back() == _target.mNode.lock())
```
jslee02: As I known, `!empty()` is preferred over `size() > 0` for emptiness check because, depending on the implementation,
`empty()` has possibly less overhead and probably at worst it's the same.

mxgrey: Good call, :+1:

jslee02:
Changed the emptiness check code using `empty()`

PR #495, opened 2015-08-27
Added missing target_link_libraries to osgDart

mkoval:
I needed to add these to get `osgDart` to build on OS X.

*Note:* This is a pull request into the `grey/osg` branch, not `master`.


mxgrey:
Appveyor is failing when it tries to build the full version of DART, but that's obviously unrelated to this pull request, so I think it's fine to merge.

PR #507, opened 2015-09-29
Give default Frame::World parameters to setTransform functions

mxgrey:
I thought it would be good to default the ``relativeTo`` parameters for the ``setTransform`` functions of ``SimpleFrame`` and ``FreeJoint``.


jslee02:
:+1:

PR #508, opened 2015-09-29
Bring back declaration of urdf::parsePose

mxgrey:
It seems that the Windows versions of the urdfdom headers are not declaring the function ``urdf::parsePose``, as reported by @jturner65 and as seen in the Appveyor CI tests.

This pull request brings back the declaration.

### DIFF
--- a/dart/utils/urdf/urdf_world_parser.cpp
+++ b/dart/utils/urdf/urdf_world_parser.cpp
@@ -51,6 +51,12 @@
 
 const bool debug = false;
 
+namespace urdf {
+
+bool parsePose(urdf::Pose &pose, TiXmlElement* xml);
+
+} // namespace urdf
+
 namespace dart {
 namespace utils {
 namespace urdf_parsing {


jslee02:
The Windows version of `urdfdom` seems an old version like 2.8. I'm not sure the exact version but it must be less than 2.9 since [`parsePose()` was presented since 2.9](https://github.com/ros/urdfdom/blob/0.2.9/urdf_parser/include/urdf_parser/urdf_parser.h#L64).

I will update the package to 3.0, which is [newest official version for ubuntu distributes (utopic, vivid, wily)](http://packages.ubuntu.com/search?keywords=urdf&searchon=names&suite=all&section=all).

jslee02:
The DART prerequisites installation file was updated (it includes urdfdom 0.3.0). Please find the like to the file at [wiki page](https://github.com/dartsim/dart/wiki/DART%205.0%20Installation%20for%20Windows).

mxgrey:
Then I'll withdraw this pull request.

PR #513, opened 2015-10-02
Resolve Eigen alignment issues of recent changes (Fix #512)

### LINKED ISSUE #512: Eigen alignment issue, line 1415 skeleton.cpp on windows
on line 1415 in skeleton.cpp the following is thrown : 
"Assertion failed: (reinterpret_cast<size_t>(array) & 0xf) == 0 && "this assertion is explained here: " "http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html" " **** READ THIS WEB PAGE !!! ****", file c:\golems\include\eigen\src/Core/DenseStorage.h, line 86"

it is happening when DataCache() is attempting to be pushed onto mTreeCache.  DataCache struct does not have EIGEN_MAKE_ALIGNED_OPERATOR_NEW declared, (which it needs since it contains vectorizable constructs and VS2013 MSVC does not have alignment support, which was added to VS2015 MSVC).  also, mTreeCache is defined as a vector of DataCache's so it needs its allocator alignment specified as well.

so, i added the EIGEN_MAKE_ALIGNED_OPERATOR_NEW  macro to struct DataCache and i declared mTreeCache as follows : 
 ```
  mutable std::vector<DataCache, Eigen::aligned_allocator<DataCache>> mTreeCache; 
```
and that seems to have addressed the problem.

jslee02:
This PR is a patch for #512 and other similar errors.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,6 +14,7 @@ environment:
   BOOST_ROOT: C:\Libraries\boost
   BOOST_LIBRARYDIR: C:\Libraries\boost\stage\lib
   BUILD_EXAMPLES: OFF  # don't build examples to not exceed allowed build time (40 min)
+  BUILD_TUTORIALS: OFF  # don't build tutorials to not exceed allowed build time (40 min)
   matrix:
   - BUILD_CORE_ONLY: ON
   - BUILD_CORE_ONLY: OFF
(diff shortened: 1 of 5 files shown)


mxgrey:
:+1:

PR #514, opened 2015-10-05
Enable compiler optimization by using 'final' keyword for Jacobian functions

mxgrey:
We can devirtualize the Jacobian function calls for ``BodyNode`` and ``EndEffector`` by using the ``final`` keyword on their Jacobian function implementations.


jslee02:
:+1:

PR #517, opened 2015-10-11
Improving URI to be compatible with Windows

jslee02:
This PR makes URI to work on Windows and resolve issues reported in #488 and #497. 

Most changes are related to file URI (file://your/path) to make it compatible on Unix and Windows. Here is a summary of the changes:
* Every function that takes in a filename as a string (e.g., skel/urdf/sdf parsers) now takes `Uri`.
* In order to pass filename to those functions as before, you can use `Uri::createFromPath(filename)`.
* A `Uri` can be created from file URI string (e.g., file:///foo/bar.txt and file:///C:/foo/bar.txt) or raw file path (e.g., /foo/bar.txt and C:/foo/bar.txt).
* Every path should be absolute path.

You can find more details at the discussions in #497.

Note that this PR also makes the AppVeyor CI tests to pass on Windows so afterward PRs of this should pass AppVeyor CI tests as well.


dart/common/Uri.cpp
```diff
@@ -196,12 +206,125 @@ void Uri::clear()
...
+    mPath = pathMatch;
+
+#ifdef _WIN32
+    if(mScheme && *mScheme == "file")
+    {
+      if(!mPath->empty() && mPath->at(0) == '/')
+        mPath = mPath->substr(1, mPath->size() - 1);
+    }
+#endif
+  }
```
mkoval: I don't know if it's a good idea to do this here. `mPath` is supposed to be the "path" component of the URI as defined by the RFC standard. Putting this logic here makes that no longer true.

dart/common/Uri.cpp
```diff
@@ -196,12 +206,125 @@ void Uri::clear()
...
+  static const std::string fileScheme("file://");
+
+  if (_path.compare(0, fileScheme.length(), fileScheme) != 0)
+  {
+#ifdef _WIN32
+    return fromString(fileScheme + "/" + _path);
+#else
+    return fromString(fileScheme + _path);
+#endif
+  }
```
mkoval: This will break on relative paths. You should only append `/` to the start of absolute paths. Windows [does not allow `:` in file names](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx#naming_conventions), so it should be safe to simply check if the string `_path` contains a `:`.

dart/common/Uri.cpp
```diff
@@ -196,12 +206,125 @@ void Uri::clear()
...
+}
+
+//==============================================================================
+bool Uri::fromPath(const std::string& _path)
+{
+  // TODO(JS): We might want to check validity of _path.
+
+  static const std::string fileScheme("file://");
+
+  if (_path.compare(0, fileScheme.length(), fileScheme) != 0)
```
mkoval: I don't know if it's a good idea to allow `fromPath` to accept `file://` URIs. That re-introduces the ambiguity that we were trying to avoid by having separate `fromPath` and `fromString` methods. It also makes it harder to detect relative vs. absolute paths on Windows (see below).

dart/common/Uri.cpp
```diff
@@ -196,12 +206,125 @@ void Uri::clear()
...
+  if (!fromString(_input))
+    return false;
+
+  // Assume that any URI without a scheme is a path.
+  if (!mScheme && mPath)
+  {
+    mScheme = "file";
+
+    // Replace backslashes (from Windows paths) with forward slashes.
+    std::replace(std::begin(*mPath), std::end(*mPath), '\\', '/');
```
mkoval: Can we wrap this in a `#ifdef _WIN32`?

dart/common/Uri.cpp
```diff
@@ -333,62 +416,76 @@ std::string Uri::toString() const
...
 //==============================================================================
-std::string Uri::getUri(const std::string& _input)
+Uri Uri::createFromPath(const std::string& _path)
 {
   Uri uri;
-  if(uri.fromStringOrPath(_input))
-    return uri.toString();
-  else
-    return "";
+  uri.fromPath(_path);
```
mkoval: This ignores the return value of `fromPath`.

dart/common/Uri.cpp
```diff
@@ -333,62 +416,76 @@ std::string Uri::toString() const
...
 
-  return mergedUri.toString();
+  return mergedUri;
+}
+
+//==============================================================================
+std::string Uri::getUri(const std::string& _input)
+{
+  return createFromString(_input).toString();
+}
```
mkoval: It seems like this is the identity function. What am I missing here?

dart/common/Uri.h
```diff
@@ -97,16 +100,34 @@ class Uri final
   /// Fragment, e.g. the part of the URI after the #
   UriComponent mFragment;
 
+  /// Constructor
+  Uri() = default;
+
+  /// Constructor that takes a URI string. Internally, this is equivalent to
+  /// calling fromString(_input) after default constructor.
+  Uri(const std::string& _input);
```
mkoval: This should be marked `explicit` unless you indent this to be an implicit conversion.

unittests/testLocalResourceRetriever.cpp
```diff
@@ -38,78 +38,91 @@
...
 
 TEST(LocalResourceRetriever, exists_FileUriDoesExists_ReturnsTrue)
 {
   LocalResourceRetriever retriever;
-  EXPECT_TRUE(retriever.exists("file://" DART_DATA_PATH "skel/cube.skel"));
+#ifdef _WIN32
+  EXPECT_TRUE(retriever.exists(Uri::createFromPath("file:///" DART_DATA_PATH "skel/cube.skel")));
+#else
+  EXPECT_TRUE(retriever.exists(Uri::createFromPath("file://" DART_DATA_PATH "skel/cube.skel")));
+#endif
```
mkoval: `createFromPath` takes a path (not a URI), so these tests should not add the `file://` prefix.

unittests/testLocalResourceRetriever.cpp
```diff
@@ -38,78 +38,91 @@
...
 {
   LocalResourceRetriever retriever;
-  auto resource = retriever.retrieve("file://" DART_DATA_PATH "test/hello_world.txt");
+#ifdef _WIN32
+  auto resource = retriever.retrieve(Uri::createFromPath("file:///" DART_DATA_PATH "test/hello_world.txt"));
+#else
+  auto resource = retriever.retrieve(Uri::createFromPath("file://" DART_DATA_PATH "test/hello_world.txt"));
+#endif
   ASSERT_TRUE(resource != nullptr);
 }
```
mkoval: See above. There shouldn't be the `file://` prefix here.

unittests/testUri.cpp
```diff
@@ -211,24 +201,37 @@ TEST(UriHelpers, getUri_InputIsUri_DoesNotChange)
...
   {
-    const std::string testUri = "file://" + testUri;
-    EXPECT_EQ(testUri, Uri::getUri(testUri));
+#ifdef _WIN32
+    // On Windows, an absolute path does not begin with forward slash but a 
+    // file URI needs it to represent an empty authority,
+    const std::string testUri = "file:///" + testPath;
+#else
+    // wherease on Unix systems the additional forward slash is not required
+    // since an absolute paht already has it.
```
mkoval: Typo: `paht` -> `path`.

dart/common/Uri.cpp
jslee02: On Windows, a "path" component would include the first slash like `/C:/foo/bar.txt`, and it wouldn't work if we pass it to some other functions that takes a raw path.

I couldn't think a better way to handle this problem. Do you have some idea on this?

dart/common/Uri.cpp
mxgrey: If someone wants to use DART as a dynamics and simulation library, I don't want them to have to understand how URIs work in order to load things from local files. In particular, I don't want them to have to worry about whether the string that they're using to specify a local file is a URI or a path string. I feel like the ``fromPath`` function should at least be smart enough to recognize when a string has a ``file://`` scheme. Is there a situation where recognizing and truncating the ``file://`` scheme from the incoming string would break something?

dart/common/Uri.cpp
jslee02: As I known, a path in file URI should be absolute, and I haven't seen cases that allow relative path. 

This function adds `file` scheme if there is none or it has other scheme so we might need to fully check whether `_path` has `file://` at the front.

dart/common/Uri.cpp
jslee02: :+1:

unittests/testLocalResourceRetriever.cpp
jslee02: Then this test would be the same with `exists_FileUriDoesNotExists_ReturnsTrue` (this name should be changed to `exists_PathDoesExists_ReturnsTrue`).

I think this test is checking if file URIs work so we might need to use `Uri::createFromString` instead of removing `file://`.

unittests/testUri.cpp
jslee02: :+1:

unittests/testLocalResourceRetriever.cpp
jslee02: Same as above.

dart/common/Uri.h
jslee02: As @mxgrey suggested:

> Have an implicit constructor for the Uri class that takes in a string an converts it to a Uri (equivalent to the existing fromString(~) function). This should allow backwards compatibility.

I intended it expecting following code would work:
```cpp
auto world = SkelParser::readWorld("/my/world.skel");  // error
```
but didn't, and following worked:
```cpp 
auto world = SkelParser::readWorld(std::string("/my/world.skel"));  // ok
```

How can I make the first code to work?

Also, I wonder if this would allow backward compatibility even the first code works because `fromString(~)` only correctly handles URI strings but not raw paths (especially Windows style paths). Instead, I left old functions that take filename as deprecated and made them to call the corresponding new functions that take `Uri`.

Here is some cases that should work I think:
```cpp
auto world = SkelParser::readWorld(Uri::createFromPath("/my/world.skel"));  // ok
auto world = SkelParser::readWorld(Uri::createFromString("file:///my/world.skel"));  // ok
auto world = SkelParser::readWorld(Uri::createFromString("file:///C:/my/world.skel"));  // ok
auto world = SkelParser::readWorld(Uri("file:///my/world.skel"));  // ok
auto world = SkelParser::readWorld("file:///my/world.skel");  // possible?
```

mxgrey: You can also create a constructor that accepts a ``const char*``.

mxgrey: Also, I suppose backwards compatibility would require the constructor to call ``fromPath`` instead of ``fromString``. I guess this was the original motivation for having ``fromPathOrString``.

dart/common/Uri.cpp
mkoval: I would prefer to add a getter method that handles this conversion. Naming it `getPath` would be confusing, since it sounds like it would just return `mPath`. How about `getFilesystemPath`?

The key issue is that the word "path" could refer to two different concepts: (1) "the path component of a URI" or (2) "a path in the local filesystem". The `mPath` field on `Uri` specifically refers to definition (1), but definition (2) is more useful if you are trying to access a local file. These are interchangable on Linux, but differ by the leading `/` on Windows.

dart/common/Uri.cpp
mkoval: > Is there a situation where recognizing and truncating the file:// scheme from the incoming string would break something?

I can't think of such a situation off the top of my head. My concern is that it makes the API more confusing. With two separate methods, the rule is simple: call `fromPath` if you have a path and `fromString` if you have a URI. With this change, you now have to worry about what scheme URI you have, too.

> I feel like the `fromPath` function should at least be smart enough to recognize when a string has a file:// scheme.

This is already handled by the `fromPathOrString` function which is, in my opinion, a more accurate name for what this function does. It also keeps these kinds of special cases contained to one function, so users that care about the difference between paths and URIs can bypass these types of checks.

In that case, I am comfortable modifying `fromStringOrPath` to use a heuristic to differentiate between absolute paths on Windows and general purpose URIs. That function could then serve as the "sane default" for users that don't care about the difference between paths and URIs.

mxgrey: That sound reasonable to me. ``fromString`` and ``fromPath`` will have very straightforward standard-conforming functionality while ``fromStringOrPath`` will try to be a bit smarter and perhaps have some special case handling.

mxgrey: That case, the ``Uri`` constructor that takes in a ``std::string``/``const char*`` should probably call the fromStringOrPath function.

dart/common/Uri.cpp
mkoval: I didn't remember seeing anything in the standard that expressly disallowed this, but it appears that [I am incorrect](http://stackoverflow.com/a/19204256/111426). It seems like this is safe after all.

I still think this belongs in a separate function (e.g. `fromPathOrString`), though.

dart/common/Uri.cpp
mkoval: :+1: I like the idea of keeping `fromString` and `fromPath` as simple wrapper functions and put the special-case handling into `fromStringOrPath`. It makes sense to make the implicit conversions call `fromStringOrPath` for backwards compatibility.

jslee02:
@mkoval I have a question about warning policy of `Uri` in the original implementation. When `fromString(~)` return false, it doesn't report any warning, instead other functions that call it report warning with more information of what the URI string means. For example, when `fromRelativeUri(~)` fails to parse the base URI, it reports as "Failed parsing base URI.". So it seems you intended to report warning whenever `fromString(~)` is called. Am I right?

dart/common/Uri.cpp
jslee02: :+1: to the ideas on `fromString`, `fromPath`, and `fromStringOrPath`.

I added `Uri(const char*/const std::string&)` as implicit constructor and now following code works:
```cpp
auto world = SkelParser::readWorld("file:///my/world.skel");
```
but it causes other compiler error; `fromRelativeUri(baseUri, "http:g")` is now an ambiguous call between `fromRelativeUri(Uri, std::string)` and `fromRelativeUri(Uri, Uri)`.

We could just remove `fromRelativeUri(Uri, std::string)` but it would have different behavior. Any idea?

mxgrey: Well, the simplest thing I can think of is to *also* explicitly overload ``fromRelativeUri`` to *also* have a ``fromRelativeUri(Uri, const char*)`` version which will just call ``fromRelativeUri(Uri, std::string)``. Hopefully this issue doesn't keep cascading, though...

dart/common/Uri.cpp
jslee02: I like the idea we have two distinct functions. 

Btw, don't we need to encapsulate other member variables as well? If we do so, we might need to consider whether the getter methods return UriComponent or their strings.

We could have separate getter methods to make it clear like:
* UriComponent getSchemeComponent() 
* std::string getScheme()
or
* UriComponent getScheme() 
* std::string getSchemeString()

dart/common/Uri.cpp
jslee02: The functionality of file scheme check is moved to `fromStringOrPath(~)` as we discussed in other thread.

dart/common/Uri.cpp
jslee02: Added `fromRelativeUri(Uri, const char*)` and `fromRelativeUri(const char*, const char*)` for consistency. I hope so too. :)

dart/common/Uri.h
jslee02: :+1:

unittests/testLocalResourceRetriever.cpp
jslee02: Updated the tests using updated `Uri` constructors.

dart/common/Uri.cpp
mkoval: I was hoping to keep `Uri` as close to a mutable struct as possible. Many `ResourceRetreiver`s rewrite URIs to other types of URIs (e.g. resolve `package://` URIs to `file://` URIs), which is easier to implement if you have direct access to the URI components.

For example, a simple assignment like `*old_uri.schema = "file"` would turn into a verbose (and error-prone) constructor call that looks like this:

```c++
Uri new_uri("file", old_uri.getAuthority(), old_uri.getPath(), old_uri.getQuery(), old_uri.getFragment()
```

We also can't replace `UriComponent` with an `std::string`. The RFC explicitly states that it's important to differentiate between "missing" components and "empty" components, because they may serialize to very different meanings.

jslee02: It makes sense. I'll add some comment on Uri class about the rationale of public members.

mxgrey: Also, to follow the convention throughout DART, we should probably use ``struct`` instead of ``class`` for Uri.

jslee02: :+1:

dart/common/Uri.cpp
jslee02: Fixed in 78051d0

dart/common/Uri.cpp
jslee02: Fixed in  78051d0

jslee02:
I think I made all the changes of what we discussed above. Passed all the test on Ubuntu and testing on Windows.

One thing I would like to make it clear is if we should clear `Uri` when `fromRelativeUri` is failed to parse. Other `from[~]` functions does.

mxgrey:
I think it makes sense to clear the contents of a ``Uri`` object when a parsing operation that's called on it fails.

jslee02:
I thought that too, but in general it would also make sense to keep the original value instead of set it to the default value on failure. For `Uri`, empty component is the default.

I actually don't have strong preference on this. If anyone have more strong idea on this I would make them to clear the contents on failure.

mxgrey:
I feel pretty strongly that when the user calls a function that is meant to alter the components of the Uri but that function fails to do so, it should clear all the Uri components. The reasoning here is that if a string fails to parse and the user fails to catch the failure, we would want that error to propagate forward, because if the ``Uri`` object was being recycled, then there's no telling what it might have been containing before the user attempted to change it, and it will be impossible for later code to recognize that the Uri it's being given is *not* actually the Uri that the user intended.

Alternatively, we could have a second argument in each of the ``from...(std::string)`` functions which would be a boolean for whether or not the Uri should be cleared upon failure. I would still want that argument to default to true (i.e. do clear upon failure) though.

psigen:
The parsing operation returns a success right?

I would think it would be somewhat useful to _not_ clear the `Uri` on a parse failure for the case where one might want to implement "first-available" logic.  You can simply run load commands until failure, and use whatever is in the URI.  Might be useful for implementing local caches and so forth.

Not strongly opposed, just don't see a strong reason to clear it if you already have a return value to indicate success.

mxgrey:
If there are reasonable use cases for not clearing, then I would strongly support the idea of having a boolean flag that defaults to having the Uri cleared.

mkoval:
I am fine with leaving a `Uri` undefined after a failure. If you are against undefined behavior here, then I think clearing the `Uri` is a reasonable compromise.

Some parsing operations could fail with the `Uri` in a partially invalid state; i.e. some of the components are set and others are not. Restoring them to their original state would require introducing temporary variables.

psigen:
If there is no option to leave the Uri in an unchanged state following a failed parse, then I agree with @mxgrey that it should be cleared.  There's no real point in leaving it in an undefined state as far as I can tell.

mxgrey:
The only advantage to leaving it in an undefined state is to avoid the need for temporary variables, which is a valid performance consideration since we're using ``std::string`` which allocates to the heap.

I think my preference would be having a flag in the ``Uri::from...`` functions that will either (1) clear the ``Uri`` if the flag is true or (2) leave the ``Uri`` in an undefined state if the flag is false. And we would have the flag default to being true.

mkoval:
I don't think it's worth adding a flag for this. Parsing a URI requires evaluating a regex and allocating string objects, which is already a quite expensive operation. Loading a resource from the URI, which typically shortly follows constructing it, is even more expensive.

I'm in favor of **always** clearing the `Uri` on failure.

jslee02:
I don't prefer to have a function with undefined behavior. My preference is having a flag with options to clearing the content or keeping the original content. But if we don't want to have a flag then I would like to go with the option of clearing on failure.

It seems a possible compromise is clearing the contents on failure without a flag, which meets our (minimum) preferences. I will change the functions so if we don't have further idea.

I think we still can restore the original contents on failure using a temporary uri.

dart/common/Uri.cpp
```diff
@@ -196,17 +220,129 @@ void Uri::clear()
...
+#else
+  return fromString(fileScheme + _path);
+#endif
+}
+
+//==============================================================================
+bool Uri::fromStringOrPath(const std::string& _input)
+{
+  // Assume that any URI without a scheme is a path.
+  static regex uriSchemeRegex(R"END(^(([^:/?#]+)://))END");
```
mkoval: This regex is not technically correct. URIs only have the `//` delimiter if they have an "authority" component. Here are some example URIs that don't have an authority that I pulled directly out of RFC 3986:

```
mailto:John.Doe@example.com
news:comp.infosystems.www.servers.unix
tel:+1-816-555-1212
urn:oasis:names:specification:docbook:dtd:xml:4.1.2
```

I don't think it's possible to unambiguously differentiate between URIs and files on Windows, where `:` is used to delimit the drive letter. So, begrudgingly, this may be the best we can do.

@psigen Any ideas?

dart/common/Uri.h
```diff
@@ -78,10 +81,15 @@ class UriComponent final
...
 };
 
-
-class Uri final
+/// The Uri struct provides URI parsing and merging functionality based on RFC
+/// 3986.
+///
+/// We have Uri as a struct rather than class to expose member variables. Many
+/// ResourceRetreiver classes rewrite URIs to other types of URIs (e.g, resolve
+/// 'package://' URIs to 'file://' URIs), which is easier to implemet if you
```
mkoval: Typo: `implemet` -> `implement`.

dart/utils/PackageResourceRetriever.cpp
```diff
@@ -71,32 +71,37 @@ void PackageResourceRetriever::addPackageDirectory(
...
   if (!resolvePackageUri(_uri, packageName, relativePath))
     return false;
 
-  for(const std::string &packagePath : getPackagePaths(packageName))
+  for(const std::string& packagePath : getPackagePaths(packageName))
   {
-    const std::string localUri = "file://" + packagePath + relativePath;
-    if (mLocalRetriever->exists(localUri))
+    common::Uri fileUri;
+    fileUri.fromPath(packagePath + relativePath);
```
mkoval: This should use the `fromRelativeUri` method to join `packagePath` and `relativePath`, not direct string concatenation.

dart/utils/PackageResourceRetriever.cpp
```diff
@@ -71,32 +71,37 @@ void PackageResourceRetriever::addPackageDirectory(
...
   if (!resolvePackageUri(_uri, packageName, relativePath))
     return nullptr;
 
-  for(const std::string &packagePath : getPackagePaths(packageName))
+  for(const std::string& packagePath : getPackagePaths(packageName))
   {
-    const std::string localUri = "file://" + packagePath + relativePath;
-    if(const auto resource = mLocalRetriever->retrieve(localUri))
+    common::Uri fileUri;
+    fileUri.fromPath(packagePath + relativePath);
```
mkoval: See above. This should use `fromRelativeUri`.

unittests/testCompositeResourceRetriever.cpp
```diff
@@ -46,7 +47,7 @@ using dart::utils::CompositeResourceRetriever;
 TEST(CompositeResourceRetriever, exists_NothingRegistered_ReturnsFalse)
 {
   CompositeResourceRetriever retriever;
-  EXPECT_FALSE(retriever.exists("package://test/foo"));
+  EXPECT_FALSE(retriever.exists(Uri("package://test/foo")));
```
mkoval: I would prefer to call `fromString` here, instead of using the constructor that calls `fromStringOrPath`, because we know this is a URI. This applies to all of the assertions in this file.

dart/utils/PackageResourceRetriever.cpp
jslee02: If we use `fromRelativeUri` unlike the original implementation, then this should be like:
```cpp
common::Uri fileUri;
fileUri.fromRelativeUri(common::Uri::createFromPath(packagePath), relativePath);
```

But I realized that there is an issue. Since `relativePath` always have `/` at the front by the nature of path component of URI, `fromRelativeUri` regards it as an absolute path. So the resolved path always set to the `relativePath` without the base URI's path.

unittests/testCompositeResourceRetriever.cpp
jslee02: :+1: It seems this applies to testPackageResourceRetriever, right?

dart/common/Uri.cpp
jslee02: One possible solution I can think now is:
* For Unix systems, we use corrected regex (```R"END(^(([^:/?#]+):))END")```) so it really checks existence of scheme
* For Windows, we use different regex that checks Windows style path as ```R"END([a-zA-Z]:[/|\\])END"```. We might don't need scheme check here since we assume that incoming path is an absolute path.

jslee02: I added a test to see if the regex works for URIs without authority component, and didn't work as expected. Then I tested with the above solution and it worked.

So now we assume a string without a scheme component is a path for Unix systems, and a string begin with Windows style path (e.g., C:/ or c:\\).

jslee02:
Sorry for rush but I'm planning to release 5.1 as soon as possible. We still have points to discuss for better implementation but it seems we can merge this since current implementation is at least passing all the tests. So if anyone have something to comment then please post them soon.

mkoval:
I'm happy with the current state of the pull request. :smile: 

@psigen Anything to add?

Issue #424, opened 2015-06-26
Problem with angular momentum conservation in dart 4.3.4


scpeters:
Initially reported as [gazebo issue 1487](https://bitbucket.org/osrf/gazebo/issue/1487/dart-has-lower-accuracy-on-angular). We added a [GetWorldAngularMomentum test](https://bitbucket.org/osrf/gazebo/src/3e1e86e456efae91240b61a42238e437ab07bdad/test/integration/physics_link.cc#cl-241), but we had to [relax the tolerances for dart](https://bitbucket.org/osrf/gazebo/src/3e1e86e456efae91240b61a42238e437ab07bdad/test/integration/physics_link.cc#cl-310). This may be related to #229; I'm not sure. Also, I haven't run the test with dart 5.

jslee02:
This issue should be resolved by #518. I will check the gazebo test again and also add an test to DART to make sure this is kept being resolved.

scpeters:
I believe #518 improved angular momentum conservation, but the linear velocity and position seem to be much less accurate now.

scpeters:
My last comment is based on a test using https://github.com/scpeters/benchmark

acxz:
I was just going through DART issues and was wondering if we could close this issue, since it is on 4.3.4 (an outdated version).
The issue could be kept if the problem persists in the current stable branch of DART tho.

mxgrey:
We should probably have a regression test for the issue and see that the regression test passes before we consider closing it.

jslee02:
As a note, probably this is related to https://github.com/dartsim/dart/issues/1193.

jslee02:
Closing this issue in favor of #1193, which I'm still working on.

PR #518, opened 2015-10-12
Fix incorrect rotational motion of BallJoint and FreeJoint

jslee02:
This PR fixes the incorrect rotational motion of BallJoint and FreeJoint near +/- 180 degrees.


dart/dynamics/FreeJoint.cpp
```diff
@@ -508,25 +508,9 @@ void FreeJoint::setAngularAcceleration(
 
 //==============================================================================
 Eigen::Matrix6d FreeJoint::getLocalJacobianStatic(
-    const Eigen::Vector6d& _positions) const
+    const Eigen::Vector6d& /*positions*/) const
```
mxgrey: I think if ``positions`` isn't going to be used at all here, then we should either

1. Remove the ``positions`` argument entirely or
2.  Save the current positions in a temporary ``Eigen::Vector6d`` set the current positions to ``positions`` and then reset the positions to their original values at the end. If the incoming ``positions`` vector matches the current positions, then this shouldn't be very much overhead, because the bookkeeping should all shortcircuit.

mxgrey: Oh wait, I take that back. Does this function imply that the Jacobian is now completely agnostic to the current position values?

jslee02: Yes. Previously a Jacobian of FreeJoint was dependent on joint positions like other joints but it's only dependent on the transform from child BodyNode to the joint.

dart/dynamics/FreeJoint.cpp
```diff
@@ -575,6 +556,16 @@ const std::string& FreeJoint::getStaticType()
 }
 
 //==============================================================================
+void FreeJoint::setTransformFromChildBodyNode(const Eigen::Isometry3d& T)
+{
+  Joint::setTransformFromChildBodyNode(T);
+
+  // We update Jacobian here since the Jacobian of FreeJoint is a function of
+  // "T" but not of joint positions.
+  mJacobian = math::getAdTMatrix(T);
```
mxgrey: I think we should move this to ``updateLocalJacobian()`` and call ``updateLocalJacobian()`` here. Then in the ``updateLocalJacobian()`` function we would wrap this up in the ``_mandatory`` flag so the computation only gets performed when something demands it.

jslee02: :+1:  Then, I think we don't need this overriding function since `updateLocalJacobian()` is called in `Joint::setTransformFromChildBodyNode()`.

dart/dynamics/BallJoint.cpp
```diff
@@ -71,6 +71,16 @@ const std::string& BallJoint::getStaticType()
 }
 
 //==============================================================================
+void BallJoint::setTransformFromChildBodyNode(const Eigen::Isometry3d& T)
+{
+  Joint::setTransformFromChildBodyNode(T);
+
+  // We update Jacobian here since the Jacobian of BallJoint is a function of
+  // "T" but not of joint positions.
+  mJacobian = math::getAdTMatrix(T).leftCols<3>();
```
mxgrey: Same thing here: I think this should be moved to the ``updateLocalJacobian()`` function and wrapped in the ``_mandatory`` flag.

dart/dynamics/FreeJoint.cpp
jslee02: By the way, don't we need _mandatory flag for `updateLocalJacobianDeriv()`?

dart/dynamics/BallJoint.cpp
jslee02: :+1:

dart/dynamics/FreeJoint.cpp
mxgrey: The motivation of the ``_mandatory`` flag was because the function could be called either (1) when a property that the Jacobian depends on has changed or (2) when a state that the Jacobian *might* depend on has changed.

My impression with ``updateLocalJacobianDeriv()`` is that it's either always zero or always dependent on state. There wouldn't be a situation where it would depend on a property but not depend on a state. If I'm wrong about that, then we should definitely introduce a ``_mandatory`` flag for it. Although even if my assumption is correct, I would be okay with introducing a ``_mandatory`` flag for the sake of consistency if nothing else.

jslee02: You are correct. I scanned all the `Joint`s for double check. I don't think we need to introduce the flag since it seems the consistency is not that important here.

jslee02:
@mxgrey I also removed the Jacobian computation of BallJoint/FreeJoint in the construction time since it will be updated whenever it gets called through the auto-updating kinematics mechanism. I would like to merge this soon so it would be good if you make comments you might have on this PR.

mxgrey:
It looks good to me now :+1: 

At some point, I think we should consider organizing and standardizing the unit tests a bit more rigorously, but that's not an immediate concern.

jslee02:
Yeah, we can first consider to categorize the tests into something like file-wise tests, subject-wise tests, and regression tests.

scpeters:
Does this help with #424? Or is that already fixed in dart5?

jslee02:
Yes, I believe this helps with #424, but for sure I need to double check by rerunning the test with dart 5.1.

Issue #477, opened 2015-07-29
CMake package conventions and conformity


mxgrey:
For the sake of pull request #461 I started digging through some of CMake's packaging documentation and also reviewing DART's current cmake setup. I thought I would discuss some of my findings and my thoughts on how we might improve DART's packaging. Some of this also comes from discussions with @mkoval.

* I've noticed that according to [CMake documentation](http://www.cmake.org/cmake/help/v3.0/command/link_directories.html), the libraries that are returned by ``find_package`` are supposed to be absolute paths, but DART currently just returns library names. It may be a good idea to follow the convention that's stated by the doc.

* The split between ``dart-core`` and ``dart`` is understandable, but I think it could be handled more gracefully. What I would propose is renaming ``dart-core`` to just be ``dart``, and then the extra features that are currently in ``dart`` should each be split out into its own ``COMPONENT`` based on their dependencies. For example, the contents of ``dart/gui`` which depend on OpenGL should go into a ``dart-gui`` library. The contents of ``dart/planning`` which depend on FLANN should go into a ``dart-planning`` library.

It gets a bit trickier when it comes to components that are nested within namespaces. For instance, ``dart/optimizer/nlopt`` could go into a library named ``dart-nlopt`` or it could go into ``dart-optimizer-nlopt``. The first option offers a shorter name, which is nice, but the second option is more precise and consistent. The second option has the added benefit of avoiding possible future collisions, for example if we end up with a ``dart/planner/nlopt`` extension which is independent of the ``dart/optimizer/nlopt``. After giving it some thought, I think I would prefer using the full name, but I'm not necessarily opposed to the shortened name.

* We would use the ``find_package(DART COMPONENTS xxx)`` feature of CMake to allow the user to decide what they want to include in the ``${DART_LIBRARIES}`` variable. For example, ``find_package(DART COMPONENTS gui planning utils-sdf)`` would set ``${DART_LIBRARIES}`` to include ``dart`` along with the GUI library, planning library, and SDF parsing library.

* Each of the libraries would be available on the apt repo independently of each other, giving the user more control over which dependencies they want to deal with.

I think this overall approach would conform better to how CMake packages are usually handled. It may also make DART easier for end users to install, because they'll have a finer level of control over the dependencies.

mxgrey:
[...]

PR #523, opened 2015-10-15
package.xml: depend on bullet

scpeters:
DART is using bullet for collision detection, but bullet wasn't listed in the package.xml file.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -22,6 +22,7 @@
   <buildtool_depend>cmake</buildtool_depend>
   <buildtool_depend>pkg-config</buildtool_depend>
   <depend>assimp</depend>
+  <depend>bullet</depend>
   <depend>eigen</depend>
   <depend>fcl</depend>
   <depend>glut</depend>


scpeters:
[One of the test instances failed](https://travis-ci.org/dartsim/dart/jobs/85494776), but I don't know why.

~~~
In file included from /home/travis/build/dartsim/dart/dart/collision/fcl/FCLCollisionNode.cpp:37:
In file included from /home/travis/build/dartsim/dart/dart/collision/fcl/FCLCollisionNode.h:44:
In file included from /usr/include/fcl/collision.h:42:
In file included from /usr/include/fcl/collision_object.h:43:
In file included from /usr/include/fcl/ccd/motion_base.h:45:
In file included from /usr/include/fcl/BV/RSS.h:43:
In file included from /usr/local/include/boost/math/constants/constants.hpp:18:
In file included from /usr/local/include/boost/lexical_cast.hpp:159:
In file included from /usr/local/include/boost/numeric/conversion/cast.hpp:34:
In file included from /usr/local/include/boost/numeric/conversion/numeric_cast_traits.hpp:28:
/usr/local/include/boost/numeric/conversion/detail/numeric_cast_traits.hpp:12:14: fatal error: 
      'boost/numeric/conversion/detail/preprocessed/numeric_cast_traits_common.hpp'
      file not found
    #include <boost/numeric/conversion/detail/preprocessed/numeric_cast_...
             ^
1 error generated.

~~~

mxgrey:
Travis-CI seems to have a recurring issue where it attempts to use the wrong version of boost once in a while. I've restarted the tests; they should come out fine now.

jslee02:
Yes, it keep repeating this failure. Usually rebuilding resolves it.

Beside on the Travis-CI issue, isn't there a way to mark bullet as a optional dependency?

mkoval:
Catkin has [no concept of optional dependencies](http://answers.ros.org/question/83726/catkin-optional-build-dependency/). Instead, the convention is to split the code that has each dependency into a separate Catkin package. This is one reason why many projects consist of many small packages, instead of one large package with optional dependencies.

This is, in my opinion, not a great convention. But there's not much we can do it about it.

scpeters:
I noticed this because I have a system install of bullet with double precision in `/usr/local` and I was also building bullet in a catkin workspace with dart using single precision. Without the package.xml change, they both built at the same time, so that dart found the double precision version in `/usr/local` with pkg-config, but tried to link against the one in the catkin workspace. It's definitely an edge case, but hopefully you don't mind the change.

mxgrey:
I think in the future we have plans to split DART's homogeneous library setup into a set of smaller libraries where each of the smaller libraries will be split out based on their external dependencies. We'll be keeping everything in a single repo with the same source tree setup that we have now, but we'll have more granularity in the build system. At that point, we may want to make a separate Catkin package manifest for the dart-bullet extension library. That being said, I vaguely remember Catkin being strict about keeping separate packages in separate directories, so I don't know for sure if this would work.

mkoval:
> That being said, I vaguely remember Catkin being strict about keeping separate packages in separate directories, so I don't know for sure if this would work.

It may be possible to work around this with [inventive use of `add_subdirectory` in CMake](http://stackoverflow.com/a/8349410/111426) and "stub" Catkin packages that just invoke the main `CMakeLists.txt` file with different options. However, I'm not sure if this is worth the trouble.

It's critical that no dependencies are missing in the `package.xml` file because this could, like @scpeters reported, can lead to packages being built in the wrong order (e.g. DART before Bullet). However, there is not much harm in exhaustively listing all optional dependencies in the `package.xml` file because Catkin ignores unknown dependencies when determining the build order.

There are only two downsides:

1. the ROS buildfarm would add those dependencies to the Debian package (this is not an issue for DART, which uses it's own Travis:CI instance)
2. `rosdep install` will install those dependencies by default

Both of these are pretty minor issues, in my opinion.

jslee02:
:+1:  I don't have objection to add bullet dependency to Catkin package manifest for DART 5.1.  Once we reorganize DART libraries as proposed in #477, we can consider how to set up separate Catkin packages. But it seems fine to me for now.

jslee02:
It seems to be okay to merge this. Just waiting another day for more comments.

Issue #524, opened 2015-10-15
Joint has the default copy constructor


mkoval:
I noticed that `Joint` does not `= delete` its copy constructor. I believe that all of the fields in `Joint` are copyable, so this means that the default copy (and, possibly, move) constructor is being implicitly defined by the compiler.

I suspect that is not intentional (and potentially dangerous!) because every `Joint` is supposed to be owned by it's child `BodyNode`.

mxgrey:
It's definitely not intentional, and it's definitely dangerous. I'll submit a pull request to correct this later today.

mxgrey:
It's also worth checking if this is an issue for BodyNode and Skeleton, since none of those should permit a copy construction.

mkoval:
@mxgrey fixed this in #525. Thanks!

mkoval:
`MultiDofJoint` and `ZeroDofJoint` still have copy constructors. :smile:

mxgrey:
I'm pretty sure they'd technically be deleted by default since they inherit from a class whose copy constructor is deleted. Still, it would be good to be explicit about it.

mkoval:
They're not - I discovered this because I'm using `libclang` to parse the DART source code and they're listed in the output. :laughing:

mkoval:
`DegreeOfFreedom` also has a copy constructor.

jslee02:
This seems to be resolved by #539.

PR #525, opened 2015-10-15
Fix Copy Safety

mxgrey:
As pointed out by @mkoval in issue #524, we neglected to delete the copy constructors of various classes that are not copy-safe. This pull request deletes their copy constructors.

### DIFF
--- a/dart/dynamics/BodyNode.h
+++ b/dart/dynamics/BodyNode.h
@@ -146,6 +146,8 @@ public:
     virtual ~Properties() = default;
   };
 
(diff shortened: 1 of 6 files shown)


mkoval:
:+1: I was just about to point out that this also applies to `Skeleton`, `MetaSkeleton`, and `Joint`. You beat me to it. :smile:

jslee02:
:+1:  Looks good to me. Could you create another PR targeting `release-5.1`? I would like to release 5.1.1 (or possibly just 5.1.0) with some minor fixes soon.

PR #530, opened 2015-10-19
API for utilizing Analytical IK

mxgrey:
This pull request introduces an API for creating and utilizing analytical inverse kinematics modules into the IK framework that was created in #461 .

It should be noted that implementing an analytical IK solver for a particular robot is up to the user; the role of this API is to allow user-created analytical solvers to be plugged into the framework in an efficient and semantically meaningful way.


mkoval:
I am in the process of integrating HERB's `IKFast` module into DART to use in some of my research code. I'll try using the new `InverseKinematics::Analytical` interface and see how it goes. :smile:

mxgrey:
Looks like AppVeyor had a server issue.

jslee02:
@mxgrey Could you resolve conflict with master branch? It seems they were introduced by #535.

osgDart/examples/osgHuboPuppet.cpp
```diff
@@ -0,0 +1,1434 @@
...
+      if(mMoveComponents[MOVE_E])
+        new_tf.rotate(Eigen::AngleAxisd(-rotationalStep, up));
+
+      new_tf.pretranslate(old_tf.translation());
+      new_tf.rotate(old_tf.rotation());
+
+      mHubo->getJoint(0)->setPositions(FreeJoint::convertToPositions(new_tf));
+    }
+
+    bool solved = mHubo->getIK(true)->solve();
```
jslee02: Unused variable warning here. I guess you used this variable for debug.

dart/math/Geometry.h
```diff
@@ -428,6 +428,22 @@ bool verifyRotation(const Eigen::Matrix3d& _R);
 /// all the elements are not NaN values.
 bool verifyTransform(const Eigen::Isometry3d& _T);
 
+/// Get the remainder of dividing x by y
+inline double mod(double x, double y)
+{
```
jslee02: Any reason we don't use [`std::fmod(double, double)`](http://en.cppreference.com/w/cpp/numeric/math/fmod)?

dart/dynamics/detail/InverseKinematics.h
```diff
@@ -69,4 +79,7 @@ void InverseKinematics::setDofs(const std::vector<DegreeOfFreedomT*>& _dofs)
   setDofs(indices);
 }
 
+} // namespace dart
+} // namespace dynamics
```
jslee02: Need to be switched the comments.

dart/math/Geometry.h
mxgrey: Mostly because I did not know that function existed. Good catch!

dart/dynamics/InverseKinematics.h
```diff
@@ -478,7 +559,7 @@ class InverseKinematics : public common::Subject
 
     /// Constructor
     explicit JacobianTranspose(InverseKinematics* _ik,
-                               double _clamp = DefaultIKGradientComponentClamp);
+                               const Properties& properties = Properties());
```
jslee02: This will break API compatibility. Is this pull request for the next major version up? If so, please ignore this comment.

dart/dynamics/InverseKinematics.h
```diff
@@ -492,6 +573,250 @@ class InverseKinematics : public common::Subject
...
 
+  /// Analytical is a base class that should be inherited by methods that are
+  /// made to solve the IK analytically instead of iteratively. This provides an
+  /// extended API that is relevant to Analytical solvers but not iterative
+  /// solvers.
+  ///
+  /// Creating an Analytical solver will have the side effect of removing the
+  /// error clamp and error weights from your ErrorMethod. If you still want
+  /// your error computations to be clamped and weighted, you should set it
+  /// again after creating the Analytical solver.
```
jslee02: I think it would be good to mention the reason of removing the error clamp and error weights since we have a good reason. The reason seems to be well described in `InverseKinematics::clone(JacobianNode*)`. :smile:

dart/dynamics/InverseKinematics.cpp
```diff
@@ -695,6 +819,420 @@ void InverseKinematics::JacobianTranspose::computeGradient(
...
+      DegreeOfFreedom* dof = mIK->getNode()->getDependentDof(depIndex);
+
+      const size_t gradIndex = gradMap[depIndex];
+      dof->setPosition(dof->getPosition() - _grad[gradIndex]);
+    }
+  }
+
+  getSolutions(desiredTf);
+
+  if(mSolutions.size() == 0)
```
jslee02: `if(mSolutions.empty())` would be a preferred way to check the emptiness.

dart/dynamics/InverseKinematics.cpp
```diff
@@ -695,6 +819,420 @@ void InverseKinematics::JacobianTranspose::computeGradient(
...
+    Solution& s = mSolutions[i];
+    const Eigen::VectorXd& q = s.mConfig;
+
+    for(size_t j=0; j < dofs.size(); ++j)
+    {
+      DegreeOfFreedom* dof = mIK->getNode()->getSkeleton()->getDof(dofs[j]);
+      if(q[j] < dof->getPositionLowerLimit()
+         || dof->getPositionUpperLimit() < q[j])
+      {
+        s.mValidity |= LIMIT_VIOLATED;
```
jslee02: Don't we need to worry about a case that `mValidity` is accidentally marked as LIMIT_VIOLATED somewhere but the solution is actually valid? In that case, this routine should reset the limit validity bit before the double for-loop.

dart/dynamics/InverseKinematics.h
mxgrey: You're right. One option would be to bring back the constructor that I've removed and mark it as deprecated. If we plan on having a 5.2 release, then I definitely think the ``Analytical`` content should go in it.

dart/dynamics/InverseKinematics.cpp
mxgrey: The validity will be set to ``VALID`` by default when the ``Solution`` object is constructed. I guess the question you're asking is whether the user might accidentally set ``mValidity`` to ``LIMIT_VIOLATED`` in their own code when a limit is not actually being violated. This is certainly something that could happen in the sense that the user has the power to do this and we don't have the power to prevent it.

My thought is that this function is really a convenience function for the user to call, and it will cycle through and mark any solutions which violate joint limits. If a user is incorrectly marking solutions on their own, I don't know whether or not we *should* overwrite or correct that; it's plausible that they have some motive for incorrectly marking certain solutions as invalid.

I'd be okay with it either way, but I'm not totally sure which would be preferable behavior.

jslee02: I agree with that the user might not want DART to overwrite it in some cases. We can consider to pass a flag that controls the behavior, and let the user to select the behavior. 

I don't have strong preference here so I would be fine with it as it is now, but, for clarification, I would like to make a note about the behavior of whichever we take.

dart/dynamics/InverseKinematics.h
jslee02: Since we agreed on that the next release is major version up, we don't need to worry about the API compatibility here. Once we have strong reason for minor version up release in the future, we can consider back-port for this.

jslee02:
:+1: Looks good to me.

PR #531, opened 2015-10-19
[Preliminary] Nodes and Addons

mxgrey:
### This is a preliminary pull request for facilitating review and discussion

*Current Status: Everything is implemented except the SoftBodyNode Addon. More unit tests should be created to ensure that everything works as expected.*

This pull request brings a tremendous amount of additional functionality to the ``Node`` class and introduces the ``Addon`` class.

The ``Node`` class exists primarily to allow users to embed custom classes into the kinematic/dynamic structure of a Skeleton. ``Node``s are objects that attach to ``BodyNode``s, and they are allowed to contain a ``State`` structure and a ``Properties`` structure. ``Node``s are also cloneable. The State of a Node will be saved when the State of its Skeleton is saved, and same for the Node's Properties. There is no limit to how many instances of a single Node type you can attach to a single BodyNode.

The ``Addon`` class allows the user to embed extra data into any class that inherits ``common::AddonManager``. An ``Addon`` can contain ``State`` and/or ``Properties``, and Addons are cloneable. Unlike Nodes, you can only add a single instance of a particular Addon type to an AddonManager.

In order to deal with the huge amount of boilerplate code that is needed to produce Nodes and Addons, I took some creative liberties with CRTP and macro patterns. I think my CRTP implementations should be fairly good (even if they don't look very pretty), but my use of macros could probably use some heavy critiquing.

I try not to use macros too much since I consider them a rather gross legacy of C, but they were the only way I could think of to keep the code reasonably DRY. I'm pretty sure we can't avoid using macros if we want DRYness, but I would be very open to any suggestions as far as improving the naming conventions or style that the macros are written with.

Note: This pull request already includes the changes of #530


dart/common/Addon.h
```diff
@@ -0,0 +1,305 @@
...
+
+  /// Clone this Addon into a new manager
+  virtual std::unique_ptr<Addon> cloneAddon(AddonManager* newManager) const = 0;
+
+  /// Set the State of this Addon. By default, this does nothing.
+  virtual void setAddonState(const std::unique_ptr<State>& otherState);
+
+  /// Get the State of this Addon. By default, this returns a nullptr which
+  /// implies that the Addon is stateless.
+  virtual const State* getAddonState() const;
```
mkoval: Does this allocate and return new memory? Or return a pointer to a pre-allocated struct?

In the former case, this should return a `std::unique_ptr<State>` to make it clear that the caller receives ownership of the `State`. It also makes the signature symmetric with `setAddonState`.

dart/common/Addon.h
```diff
@@ -0,0 +1,305 @@
...
+
+  /// Set the Properties of this Addon. By default, this does nothing.
+  virtual void setAddonProperties(
+      const std::unique_ptr<Properties>& someProperties);
+
+  /// Get the Properties of this Addon. By default, this returns a nullptr
+  /// which implies that the Addon has no properties.
+  virtual const Properties* getAddonProperties() const;
```
mkoval: Same comment as `getAddonState`. This should return an `std::unique_ptr` if the caller owns the return value.

dart/common/Addon.h
```diff
@@ -0,0 +1,305 @@
...
+  virtual void changeManager(AddonManager* newManager);
+};
+
+//==============================================================================
+/// AddonWithProtectedState generates implementations of the State managing
+/// functions for an Addon class.
+template <class Base, typename StateData,
+          class ManagerType = AddonManager,
+          void (*updateState)(Base*) = &detail::NoOp<Base*> >
+class AddonWithProtectedState : public Addon
```
mkoval: It took me a while to figure out the purpose of this class (and `AddonWithProtectedProperties`). From my understanding, these let you write an `Addon` that forwards `getState`/`setState` calls to a member variable without having to write the boilerplate myself. Is that correct? Or is the idea that I would inherit from this as a base class?

In either case, it would be helpful to explain this in the class-level Doxygen comment.

dart/common/Addon.h
```diff
@@ -0,0 +1,305 @@
...
+
+//==============================================================================
+/// AddonWithProtectedStateAndProperties combines the
+/// AddonWithProtectedState and AddonWithProtectedProperties classes into a
+/// single templated class
+template <class Base, typename StateData, typename PropertiesData,
+          class ManagerType = AddonManager,
+          void (*updateState)(Base*) = &detail::NoOp<Base*>,
+          void (*updateProperties)(Base*) = updateState>
+class AddonWithProtectedStateAndProperties : public Addon
```
mkoval: I expected this to inherit from both `AddonWithProtectedState` and `AddonWithProtectedProperties`. Is it implemented this way because of some quirk of C++ multiple inheritance?

dart/common/Addon.h
mxgrey: The point of this function is to allow you to peek at the state data structure without needing to allocate new memory to do so.

When you call ``clone()`` on ``Addon::State`` like ``addon->getAddonState()->clone()`` it will return a ``std::unique_ptr<Addon::State>``, and that is what makes it symmetric with ``setAddonState(std::unique_ptr)``.

dart/common/Addon.h
```diff
@@ -0,0 +1,305 @@
...
+  using PropertiesMixer = ExtensibleMixer<Properties, Mixin>;
+
+  /// Virtual destructor
+  virtual ~Addon() = default;
+
+  /// Clone this Addon into a new manager
+  virtual std::unique_ptr<Addon> cloneAddon(AddonManager* newManager) const = 0;
+
+  /// Set the State of this Addon. By default, this does nothing.
+  virtual void setAddonState(const std::unique_ptr<State>& otherState);
```
mkoval: This should take a const reference, not a `std::unique_ptr`. The current signature prevents me from passing a stack-allocated `State` or a state managed by `shared_ptr` as an argument. In general, [it's *never* necessary to pass a `std::unique_ptr` by const reference](http://www.codingstandard.com/rule/8-2-4-do-not-pass-stdunique_ptr-by-const-reference/) because you can always pass the contents by reference, instead.

This comment also applies to `setAddonProperties`.

dart/common/Addon.h
mxgrey: Yes, I originally tried to implement it as you would expect, but the diamond of death problem was too strong in this case. Normally we could have gotten around the diamond of death by having ``AddonWithProtectedState`` and ``AddonWithProtectedProperties`` both *virtually* inherit ``Addon``, but then it would not be possible to ``static_cast`` from an ``Addon`` to either of those classes (or to any class that inherits them), because the casting would require dynamic dispatch. So avoiding multiple inheritance allows us to avoid doing dynamic dispatch every time we want to grab an ``Addon`` from its manager.

dart/common/AddonManager.h
```diff
@@ -0,0 +1,176 @@
...
+/// AddonManager is a base class that should be virtually inherited by any class
+/// that wants to be able to manage Addons.
+///
+/// The base AddonManager class is completely agnostic to what kind of Addons it
+/// is given. Addons are stored in a std::map, so access to its Addons happens
+/// on average in log(N) time. Most often, a class that accepts Addons will have
+/// certain Addon types that it will need to access frequently, and it would be
+/// beneficial to have constant-time access to those Addon types. To get
+/// constant-time access to specific Addon types, there are FOUR macros that you
+/// should use in your derived class:
```
mkoval: Is there a strong reason to prefer `std::map` over `std::unordered_map` here? From what I've read, [`std::unordered_map` is significantly faster than `std::map`](http://supercomputingblog.com/windows/ordered-map-vs-unordered-map-a-performance-study/) for all numbers of elements. The only reason to use `std::map` if you need to iterate over the map in sorted order or if comparison is much cheaper than hashing.

`std::type_info` is hashable, so I see no reason not to use `std::unordered_map` here.

dart/common/Extensible.h
```diff
@@ -0,0 +1,217 @@
...
+  Extensible() = default;
+
+  /// Virtual destructor
+  virtual ~Extensible() = default;
+
+  /// Do not copy this class directly, use clone() or copy() instead
+  Extensible(const Extensible& doNotCopy) = delete;
+
+  /// Do not copy this class directly, use clone() or copy() instead
+  Extensible& operator=(const Extensible& doNotCopy) = delete;
```
mkoval: Why do you explicitly disallow the `operator=` here? I would expect it to be equivalent to:
```c++
other.copy(*this)
```
Is the goal to prevent the user from accidentally making a (potentially expensive) copy?

dart/common/Addon.h
mxgrey: Yep, this is a CRTP class that takes care of boilerplate code for managing the State and cloning. You have your actual Addon class inherit this, and it takes care of setting everything up. Inheriting this class also enables you to use macros for setting and getting each field in your State/Properties (although it just occurred to me that the macros have only been implemented for Properties fields so far; I still need to implement the State versions).

dart/common/Extensible.h
```diff
@@ -0,0 +1,217 @@
...
+
+/// Extensible is a CRTP base class that provides an interface for easily
+/// creating data structures that are meant to be extended. Ordinary copying
+/// does not work with extended data structures, because information from the
+/// derived classes will typically be lost during a copy. The Extensible class
+/// provides an interface for creating a new copy of an extended data structure,
+/// as well as copying information between two extended data structures of the
+/// same type.
+template <class T>
+class Extensible
```
mkoval: I don't know if `Extensible` is a great name for this. `Copyable` or `Cloneable` (ala Java) seem to be more descriptive.

dart/common/Extensible.h
```diff
@@ -0,0 +1,217 @@
...
+};
+
+/// The ExtensibleMixer class is used to easily create an Extensible (such as
+/// Node::State) which simply takes an existing class (Mixin) and creates an
+/// Extensible that wraps it. This creates all the appropriate copy, move, and
+/// clone members, allowing you to follow the Rule Of Zero. You can also
+/// construct an instance in the exact same way that you would construct a Mixin
+/// instance.
+template <class T, class Mixin>
+class ExtensibleMixer : public T, public Mixin
```
mkoval: I'm not sure how this is intended to be used. I see that it's used in the rest of the `AddonManager` code. However, that code is quite abstract, so I'm not sure what exactly it's doing. :sweat: 

Is the idea that this can be used to make a POD data structure satisfy the `Extensible` interface? If so, why are there two template arguments?

dart/common/Extensible.h
```diff
@@ -0,0 +1,217 @@
...
+  // Documentation inherited
+  void copy(const T& other) override final;
+
+};
+
+/// MapHolder is a templated wrapper class that is used to allow maps of
+/// Addon::State and Addon::Properties to be handled in a semantically
+/// palatable way.
+template <typename MapType>
+class ExtensibleMapHolder final
```
mkoval: I am also not sure why this class is necessary. `std::map` and `std::unordered_map` already satisfy the Rule of Five.

dart/common/Addon.h
mxgrey: This is a good point. Also, now I realize that the incorrect comment that @jslee02 pointed out was meant to apply to this. The current setup does not prevent the Addon from altering the contents of ``otherState``, which is a major flaw.

My motivation for passing in a ``unique_ptr`` here was to ensure that ``addon1->setAddonState(addon2->getAddonState()->clone());`` would always work. By default, the State of an Addon is a ``nullptr`` which is meant to indicate that it is stateless. This means that if pass in a dereferenced const reference, then ``addon1->setAddonState(*addon2->getAddonState()->clone())`` could segfault since we'd sometimes be dereferencing a ``nullptr``.

However, end users are **not** generally expected to be using this abstracted functionality. Instead, it should get handled automatically when they set/get the States or Properties of the class that has inherited ``AddonManager``. So it's not hard for us to put in a ``nullptr`` check before attempting to send in a ``State`` or ``Properties`` with the code which is actually meant to be using these functions.

dart/common/detail/Addon.h
```diff
@@ -0,0 +1,338 @@
...
+          << "this Addon is only designed to be attached to a ["\
+          << typeid(ManagerType).name() << "] type manager. This may cause "\
+          << "undefined behavior!\n";\
+    assert(false);\
+  }
+
+#define DART_COMMON_CAST_NEW_MANAGER_TYPE_AND_RETURN_NULL_IF_BAD(\
+  Base, ManagerType, newManager, castedManager, func)\
+  DART_COMMON_CAST_NEW_MANAGER_TYPE(Base, ManagerType, newManager, castedManager, func)\
+  if(nullptr == castedManager) return nullptr;
```
mkoval: Why can't these be implemented as template functions? The only downside is that we may not be able to get the stringified function name `#func`.

mkoval: I'm also peeved that this declares a variable in the current scope. I wonder if we could do something like this, instead:
```c++
#define DART_COMMON_CAST_NEW_MANAGER_TYPE(Base, ManagerType, newManager,\
                                          castedManager, func)\
([]()\
{\
  ManagerType* castedManager = dynamic_cast<ManagerType*>(newManager);\
  if(nullptr == castedManager)\
  {\
    dterr << "[" << typeid(Base).name() << "::" << #func << "] Attempting to "\
          << "use a [" << typeid(newManager).name() << "] type manager, but "\
          << "this Addon is only designed to be attached to a ["\
          << typeid(ManagerType).name() << "] type manager. This may cause "\
          << "undefined behavior!\n";\
    assert(false);\
  }\
  return castedManager;\
})()

```

dart/common/detail/Addon.h
```diff
@@ -0,0 +1,338 @@
...
+template <class Base, typename PropertiesData,
+          class ManagerType, void (*updateProperties)(Base*)>
+std::unique_ptr<Addon>
+AddonWithProtectedProperties<Base, PropertiesData,
+                             ManagerType, updateProperties>::
+cloneAddon(AddonManager* newManager) const
+{
+  DART_COMMON_CAST_NEW_MANAGER_TYPE_AND_RETURN_NULL_IF_BAD(
+        Base, ManagerType, newManager, castedManager, clone);
+  return std::unique_ptr<Base>(new Base(newManager, mProperties));
```
mkoval: This is potentially dangers. You should [*always* parenthesis macro arguments that could be an expression](http://stackoverflow.com/a/23889851/111426):
```c++
return std::unique_ptr<Base>(new Base((newManager), mProperties));
```

dart/common/AddonManager.h
```diff
@@ -0,0 +1,176 @@
...
+
+  /// Remove an Addon from this AddonManager.
+  template <class T>
+  void erase();
+
+  /// Remove an Addon from this AddonManager, but return its unique_ptr instead
+  /// of letting it be deleted. This allows you to safely use move semantics to
+  /// transfer an Addon between two AddonManagers.
+  template <class T>
+  std::unique_ptr<T> release();
```
mkoval: It would be useful to have a `replace` function that does a combination of `set` and `remove`.

In fact, I think we can modify `set` to do this by default. We already have an iterator to the existing element, if it exists, so there is minimal overhead in returning that `std::unique_ptr`.

dart/common/detail/AddonManager.h
```diff
@@ -0,0 +1,240 @@
...
+  template <class T> const T* get() const { return AddonManager::get<T>(); }\
+  template <class T> void set(const T* addon) { AddonManager::set<T>(addon); }\
+  template <class T> void set(std::unique_ptr<T>&& addon) { AddonManager::set<T>(std::move(addon)); }\
+  template <class T> void erase() { AddonManager::erase<T>(); }\
+  template <class T> std::unique_ptr<T> release() { return AddonManager::release<T>(); }\
+  template <class T, typename ...Args> T* create(Args&&... args)\
+  {\
+    T* addon = new T(this, std::forward<Args>(args)...);\
+    mAddonMap[typeid(T)] = std::unique_ptr<T>(addon); return addon;\
+  }
```
mkoval: Could these be implemented in a CRTP base class?

dart/common/Addon.h
```diff
@@ -0,0 +1,305 @@
...
+  /// handling when its AddonManager gets changed.
+  virtual void changeManager(AddonManager* newManager);
+};
+
+//==============================================================================
+/// AddonWithProtectedState generates implementations of the State managing
+/// functions for an Addon class.
+template <class Base, typename StateData,
+          class ManagerType = AddonManager,
+          void (*updateState)(Base*) = &detail::NoOp<Base*> >
```
mkoval: Why use `NoOp` here, instead of passing `nullptr` as a sentinel value?

dart/common/Addon.h
mkoval: I see. Can you add a comment to `getAddonState` that explains that the caller **does not** own the return value?

dart/common/Addon.h
mkoval: :+1: That makes sense. I figured that was the case.

dart/common/AddonManager.h
mxgrey: Yes, there are actually several strong reasons.

1. When we save the States of all the Addons in an AddonManager, they get saved into an ordered map (ordered based on the typeid of the Addon they belong to). Then when we want to copy those states over to the Addons of another AddonManager, we can iterate through the list of States versus the list of Addons in an ordered fashion, allowing exactly linear performance for copying all of the States over. If we used unordered_map, the performance would be linear multiplied by the constant time factor of a hash table lookup, so it would be marginally slower.

2. More importantly, we to be able to specialize certain Addons that get used extremely often so that we get instant (pointer-style) lookup for them (instead of the constant-time hash lookup). We accomplish this by holding onto an iterator to the location in the map where that Addon type is located. In a ``std::map`` an iterator is always valid (as long as the entry it points to is not deleted) whereas in a ``std::unordered_map``, inserting a new object can invalidate previously acquired iterators. So in order to have direct pointer-like lookup for commonly used Addons, we need to use an ordered map.

dart/common/Extensible.h
mxgrey: It's mostly a semantics thing. I don't want the user to think that an Extensible instance can be treated as a POD. The Extensible class is used to erase type, so if someone is naively saying ``extensible1 = extensible2;``, that is most likely a bug IMO since there's no guarantee that ``extensible1`` and ``extensible2`` will be compatible types. Granted, that can't be guaranteed for the ``copy(~)`` function either, but at least ``copy(~)`` requires the user to think a bit about what they're doing, whereas the assignment operator could be mindlessly used out of habit. But really, the user shouldn't generally be interacting with the Extensible base class directly in the first place.

In the implementations of the ``AddonWithProtected...`` classes, I reintroduce the assignment operator, because for those functions there's no question of whether another ``AddonWithProtected...`` object is compatible.

dart/common/Extensible.h
mxgrey: Yes, you're exactly right: this implements the ``Extensible`` interface for POD structures.

The second argument is used as you would expect: it passes in the POD that ``ExtensibleMixer`` is supposed to wrap.

The first argument is understandably less obvious. There are many different independent components of the code base that use the ``Extensible`` class. Specifically, ``Node::State``, ``Node::Properties``, ``Addon::State``, and ``Addon::Properties`` all inherit ``Extensible``. However, I think it's important to ensure that it's not possible to accidentally input an ``Addon::Properties`` object into something that expects a ``Node::State``, because that would be silly.

So ``Extensible`` defines this interface of extensibility, but we do **not** explicitly deal in terms of ``Extensible`` objects; instead, we deal in terms of object types that have inherited ``Extensible``. For example, we would not want to have a ``std::vector<Extensible>`` because this is too abstract and we wouldn't know if we're looking at a vector of ``Addon::State``, ``Node::Properties``, or something else. Instead we would want to have a ``std::vector<Addon::State>`` or a ``std::vector<Node::Properties>`` so there's no question of what specific kind of ``Extensible`` it is.

For the ``ExtensibleMixer`` class to correctly implement the specific type of Extensible that you intend to use, you must pass in that Extensible type as the first argument. For example:

```
using CustomAddonState = common::ExtensibleMixer<Addon::State, CustomData>;
```

That being said, you don't have to do it like this, because I provide ``Addon::StateMixer``, ``Addon::PropertiesMixer``, ``Node::StateMixer``, and ``Node::PropertiesMixer`` so you can do this instead:

```
using CustomAddonState = common::Addon::StateMixer<CustomData>
```

That way you don't need to worry about that first template argument.

I should probably put a note of this in the doxygen comments for ``ExtensibleMixer``.

dart/common/Extensible.h
mxgrey: Because there's no way to make a **deep** copy of a whole ``std::map`` or a ``std::vector`` when it's containing pointers to a base class.

mxgrey: Also, I do some clever things to avoid allocating memory whenever possible. Suppose you already have two ``ExtensibleVector``s both filled up with ``Extensible`` objects and you want to copy the data from the objects in one of those vectors to the objects in the other vector without wiping clean and reallocating all the objects in the receiving vector. ``ExtensibleVector`` will do that automatically when you use the assignment operator. It will only allocate an object if a corresponding object isn't allocated already.

mxgrey:
[Because of some questionable restrictions that are being imposed by the current C++11 standard](http://stackoverflow.com/questions/15867399/static-constexpr-pointer-to-function-difference-between-compilers) it seems that Clang refuses to compile my current approach at CRTP implementations. GCC does not adhere to this restriction, and there are (rightly, I believe) proposals to loosen the restriction, but in the mean time I'll have to somehow mend the CRTP implementation that ensures auto-updating is done correctly for Addons in Skeletons.

dart/common/AddonManager.h
mxgrey: If I understand what you're saying, the behavior of ``set`` does what you describe; it will replace the current Addon with a clone of the Addon you pass in (or by the Addon itself if you use the move-semantics version of ``set``) which replaces the Addon that was previously there (if one already existed).

Are you instead suggesting that we have a ``swap`` function which would exchange the Addons of a specified type between two AddonManagers?

mxgrey: Or do you want to combine ``set`` and ``release`` so that it sets the ``Addon`` to whatever you pass in, and then releases the one that was already there with the return value?

dart/common/detail/Addon.h
mxgrey: I definitely felt gross doing it, and found myself lamenting the current archaic state of macros in C++. I'll gladly invest some time in thinking about a less gross way of implementing these macros.

It looks like this macro is creating a lambda object and then immediately calling it the lambda, so that it will return the value it produces. Is that accurate? So this effectively turns the macro into a function with a return value? That does seem much better than my secretly creating a local variable approach.

dart/common/Extensible.h
mxgrey: I'm okay with renaming it. I'd suggest we aggregate a list of suitable names and then see which ones are most favorable.

dart/common/detail/Addon.h
psigen: Is the concern that this local variable is hanging around in the local scope for an arbitrary amount of time, or if someone has the bad luck of naming something `castedManager` in the same scope as this macro, weird things will happen?

dart/common/AddonManager.h
mkoval: 1. I don't entirely understand the efficiency argument here. Are you currently iterating through the two `std::map`'s in parallel while merging them?

2. This makes sense. I forgot that `std::map` had such nice semantics for iterator invalidation.

dart/common/Extensible.h
mkoval: > But really, the user shouldn't generally be interacting with the Extensible base class directly in the first place.

Where is `Extensible` typically used? How does it insure that you don't try to `clone` into the wrong type of `Extensible`?

dart/common/Extensible.h
mkoval: :+1: That makes sense to me.

dart/common/AddonManager.h
mkoval: I want to modify `set` to return the `unique_ptr` that was replaced. This is identical to the existing behavior if the user ignores the return value because it will be immediately destructed.

mxgrey: If ``set`` were to return anything, I would want it to return a pointer to the newly created Addon instance. Keep in mind that there are two versions of set, and the main version is actually cloning the Addon that gets passed in, so it's not as if the AddonManager is adopting the pointer that you pass to it.

However, I would be perfectly happy to make a ``replace`` function that combines the functionality of ``set`` and ``release``. In fact, then ``release()`` can be implemented as ``return replace(nullptr);``. I would also be open to creating a ``swap`` function that allows you to swap the Addon instances of two AddonManagers.

dart/common/AddonManager.h
mxgrey: > Are you currently iterating through the two std::map's in parallel while merging them?

Yep, pretty much that. I iterate between two std::map objects in parallel. If the "receiving" map has an entry that matches the current entry of the "sending" map, then I can use the ``copy`` function to copy the content of the "sender" over to the "receiver". If the "receiving" map does **not** have an entry that matches the current entry of the "sending" map, then I can insert a new (cloned) entry in the **exact** spot on the receiving map where the entry belongs without needing to have ``std::map`` look for the correct spot to place it (or, depending on the semantics of the operation that I'm performing, I can just skip this entry).

The point of this is to be able to save a set of ``Addon::State``s to a map that can then be loaded back in later, or loaded into another copy of a Skeleton. This ordered approach to saving and loading the States allows us to minimize overhead and to be robust to situations where there isn't a perfect match between the types of ``Addon::State``s that you've saved and the types of ``Addon``s present in the Skeleton that you're trying to load the States into.

dart/common/Addon.h
mxgrey: [According to stackoverflow](http://stackoverflow.com/a/15090036), using a nullptr-valued function pointer is undefined behavior, so if we accept a nullptr template argument, then we would always have to check whether the value is a nullptr before using it. I'd much rather have a no-op function which (1) doesn't require special checking, and which (2) the compiler will optimize away entirely instead of having a branching if-statement (although since it's a ``constexpr``, the compiler might even be able to compile away any branching, so this could be a moot point).

dart/common/detail/AddonManager.h
mxgrey: This is actually a matter of C++ syntax. In order for a class that inherits ``AddonManager`` to specialize Addons for itself, it needs to **shadow** these templated functions before it can specialize these templated functions. Therefore, these templated functions need to be declared and defined in any derived class that wants specialized Addons.

One concern that just occurred to me is whether a derived class can benefit from the Addon specializations of a class that it inherits. This shadowing might prevent that from happening, I'm not sure. This is something I should test.

dart/common/detail/Addon.h
mxgrey: I think the biggest concern is that macros are ugly and can do weird things, and it's not obvious at first look that the macro is declaring a local variable which will continue to exist after the macro is completed. Essentially, there's a non-obvious side effect of this macro, and non-obvious side effects can be dangerous as well as difficult to debug.

dart/common/Extensible.h
mxgrey: ``Extensible`` gets used as a foundation in a higher level framework. As an example, ``Addon::State`` is an ``Extensible``, and so ``AddonManager::State`` is an ``ExtensibleMapHolder`` for an ``AddonManager::StateMap``.

The ``AddonManager::StateMap`` is the key to keeping everything sane. ``AddonManager::StateMap`` is a typedef for ``std::map< std::type_index, std::unique_ptr<Addon::State> >``. This means that ``AddonManager::StateMap`` is mapping a set of ``Addon::State`` pointers to the ``Addon`` types that they correspond to.

``ExtensibleMapHolder`` then wraps up the ``Addon::StateMap`` to (1) obey the rule of five for deep copying and (2) be more efficient about copying, as discussed in another thread.

So the basic pipeline is:

1. Get an ``AddonManager::State`` object by calling ``AddonManager::getAddonStates()``.
2. Store the ``AddonManager::State`` object however you please, since it follows the rule of five (**but** you can't manipulate its contents, because the entries are protected).
3. Pass the ``AddonManager::State`` object back into the ``AddonManager`` it came from, or into any other ``AddonManager`` (any mismatches in the Addons will be handled by ignoring them).

So the users should never actually be touching raw ``Extensible`` objects themselves. Instead, they should just be passing around these higher level wrapper classes, like ``AddonManager::State``.

dart/common/detail/AddonManager.h
mxgrey: Although I guess one thing that *might* be possible is to have a templated class whose job is to specialize an Addon type for the class that inherits it. I could put together some tests to see if this would work. Anything that reduces the amount of macros that we need is good as far as I'm concerned.

mxgrey:
I think the primary concern about this pull request at the moment is the name of the ``Extensible`` class. In retrospect, it's clear to me that ``Extensible`` is not a very good name for this class. The purpose of the ``Extensible`` class is to provide an interface for performing **deep** clones and **deep** copies of objects whose types have been momentarily erased. I think the current front-runners for a preferable name would be ``Cloneable`` and ``Copyable``. Indeed, the ``Extensible`` class has both a ``clone`` function and a ``copy`` function, so either of these would be perfectly valid names.

However, since the class is for *both* copying and cloning, I think I would prefer a name that can encapsulate both of those concepts. Perhaps something like ``Replicable`` or ``Duplicable``. I'm open to any suggestions that anyone has.

mxgrey:
It seems that Visual Studio 2013 can't compile this branch, at least partially due to incomplete support for ``constexpr``. I could *maybe* find a workaround where I don't need to use ``constexpr``, but if I can get this branch to compile in Visual Studio **2015**, then I'd rather we bump up the Windows support requirement to VS 2015 starting with release 5.2.

mkoval:
All of the individual lines of code look reasonable in this pull request, but I'm having a very hard time wrapping my head around what this will look like in use.

@mxgrey Are there any examples of using `AddOn`s either as a user (i.e. someone who is reading/writing values) or as a developer (i.e. someone who is defining a custom `AddOn`) in the DART source code?

mxgrey:
Yes, there are lots of Addons currently implemented in the codebase. The ``Support`` class that is used by ``EndEffector`` is an Addon. Also, every class that derives ``Joint`` which has its own unique properties is now storing those properties within an ``Addon`` so that their properties can automatically be saved whenever the properties of the whole Skeleton are saved.

I will also be modifying the ``SoftBodyNode`` class so that its properties and states are stored within an Addon, but this is not implemented yet.

mxgrey:
On another note, I think there's a lot that can and should be cleaned up with the current CRTP & macro implementation. I'm going to try really hard to move away from macros and towards templates as much as possible, but it's going to depend largely on what exactly the C++ standard allows me to do (and what the current VS2015 support of the standard [or lack thereof] allows me to do).

dart/common/detail/AddonManager.h
```diff
@@ -0,0 +1,240 @@
...
+
+#include "dart/common/AddonManager.h"
+
+#define DART_COMMON_CHECK_ILLEGAL_ADDON_ERASE( Func, it, ReturnType )\
+  if(it ->second && !it ->second->isOptional(this))\
+  {\
+    dterr << "[AddonManager::" #Func << "] Illegal request to remove Addon!\n";\
+    assert(false);\
+    return ReturnType ;\
+  }
```
mkoval: It seems like this could be easily converted to a template function. If not, you should wrap the macro in a `do { ... } while (0)` block to avoid any surprises.

mxgrey: I believe I made it a macro so that it has the ability to decide whether or not to have the calling function return. If it were a function, I'd have to wrap it in an ``if`` statement and return when true or something like that. This wouldn't be a terrible thing, but this makes its use slightly simpler.

I'm not clear on what kind of surprises are likely to occur in this case, or what would be accomplished by a ``do { ... } while(0)`` statement. Would adding in the ``do { } while()`` syntax allow the compiler to throw an error if the macro gets misused? Is that the motivation?

mkoval: [This is a good explanation](http://stackoverflow.com/a/1067238/111426). If you don't use the `do {  } while (0)` trick, then the Bad Things can happen if you mix macros with `if`/`while`/`for` statements.

mxgrey: Now I get it. Using a ``do { } while(0)`` **without the semicolon** allows the macro to be treated exactly like a normal expression whereas currently it does not need to be treated as a normal expression.

mkoval: Yup. That's why I did'nt include a semi colon in my examples. :smile: Perhaps I should have been more clear. :laughing:

mxgrey:
I've been researching the idea of using templated classes (instead of macros) to implement the Addon specializations, and I've been conducting experiments to see how that might be implemented. I'm somewhat convinced at this point that there is no nice way of doing it. The main issue I'm encountering is that you are [not allowed to create an explicit specialization of a member function of a templated class](http://stackoverflow.com/questions/5512910/explicit-specialization-of-template-class-member-function), which means I **cannot** have a class like this:

```
template <typename SpecializedAddon>
class SpecializedManager : public virtual AddonManager
{
public:
  template <class T> T* get() { return AddonManager::get<T>(); }

protected:
  AddonMap::iterator mSpecializedIterator;
}

template<typename SpecializedAddon>
template <>
SpecializedAddon* SpecializedManager<SpecializedAddon>::get<SpecializedAddon>()
{
  return static_cast<SpecializedAddon*>(mSpecializedIterator->second);
}
```

This does not compile because you are **not** allowed to specialize a member function of a templated class without fully specializing the class itself. And if we need to fully specialize the class, then we accomplish absolutely nothing in terms of reusability.

I'm going to keep investigating for a little while still. If anyone has any ideas or suggestions, I'd be happy to explore them, even if you're not sure whether the idea/suggestion would work. I'd **really** truly like to avoid (or at least reduce) how much we're using macros for all of this, but I'm seriously struggling to find a viable alternative.

mxgrey:
After some more experimentation, it seems [the suggestion in the first reply to the post I linked to](http://stackoverflow.com/a/5513109) might actually be viable for our purposes. I'll draft a prototype of an implementation that uses templates and then show what its equivalent would be that uses macros, and we can discuss which one seems preferable.

jslee02:
There is an alternative way which is very similar to the second reply to the post you linked to but more generic. Here is an example code:
```cpp
template <typename SpecializedAddon>
class SpecializedManager
{
public:
  template <class T>
  void get()
  {
    GetImpl<T>::impl();
  }

private:
  template <class T, typename Enable = void>
  struct GetImpl
  {
    static void impl()
    {
      std::cout << "called generic." << std::endl;
    }
  };

  template <class T>
  struct GetImpl<T, typename std::enable_if<std::is_same<T, double>::value>::type>
  {
    static void impl()
    {
      std::cout << "specialized for double." << std::endl;
    }
  };

  template <class T>
  struct GetImpl<T, typename std::enable_if<std::is_same<T, int>::value>::type>
  {
    static void impl()
    {
      std::cout << "specialized for int." << std::endl;
    }
  };

};
```

mxgrey:
I'm not totally clear on how the ``GetImpl`` approach would work when the functions need to access data members of the SpecializedManager class, although I suppose we could make ``GetImpl`` a friend class and pass a ``SpecializedManager*`` to the static functions as an argument.

I've managed to produce some prototype code using the approach suggested by the first reply. It's fairly elegant, and I've even managed to ensure that it can handle diamonds of death. I'll do a write-up of what its API would look like, and I'll show how it compares to the current macro-based implementation. If people like it, I'll start switching out the implementations.

jslee02:
In order to access member of `SpecializedManager`, a simple way would be passing the pointer of `SpecializedManager` to `GetImpl::impl` as `GetImpl::impl(SpecializedManager<SpecializedAddon>*`.

I agree that the first approach is elegant enough for our case. One advantage of using nested struct approach is that we can specialize the template function not only for a specific type per a specialization but also for multiple types that satisfy some conditions per a specialization. For example, we can specialize it for floating types (double, float, and etc) using type traits `std::is_floating_point`.

mxgrey:
After much ado, I've finished the changes that allow us to use templates instead of macros in order to specialize ``Addon``s and ``Node``s. Now a class that wants to have specialized ``Addon``s can simply inherit ``SpecializedAddonManager``. Or a class that derives from ``BodyNode`` or ``Skeleton`` can have its own specialized ``Node`` types by inheriting from ``SpecializedNodeManagerForBodyNode`` or ``SpecializedNodeManagerForSkeleton``. No more macros necessary.

There are still a few things left to do before the Addons and Nodes are completely deployed for the next major release, but I'm going to save those changes for later pull requests since this one is already outrageously big.

The major content of this pull request is now officially finished, and I think all we should do now is review and make small modifications, like adding unit tests or renaming the ``Extensible`` class.

dart/common/Addon.cpp
```diff
@@ -0,0 +1,95 @@
...
+//==============================================================================
+const Addon::Properties* Addon::getAddonProperties() const
+{
+  return nullptr;
+}
+
+//==============================================================================
+bool Addon::isOptional(AddonManager* /*oldManager*/)
+{
+  return true;
```
jslee02: If this function is for the legality check of removing from the manager, I think it should to be determined by its manager rather than by itself. If the decision is tied to the addon, we cannot create addons of the addon type from two managers, where one is general manager (or specialized manager but not for the addon) and the other one is a specialized manager for the addon, expecting one is specialized and the other is not specialized.

dart/common/Addon.h
```diff
@@ -0,0 +1,304 @@
...
+  /// By default, this simply returns true.
+  virtual bool isOptional(AddonManager* oldManager);
+
+protected:
+
+  /// Constructor
+  ///
+  // We require the AddonManager argument in this constructor to make it clear
+  // to extensions that they must have an AddonManager argument in their
+  // constructors.
```
jslee02: Is this non-doxygen comment intended?

dart/common/detail/Addon.h
jslee02: Beside the discussion above, it seems ln 64 will not be reached as `DART_COMMON_CAST_NEW_MANAGER_TYPE` will raise segfault for `nullptr == castedManager`.

dart/common/Addon.cpp
mxgrey: That was my original plan, but I had a very hard time thinking of a clean way to implement that. I found it much easier to have that information embedded in the ``Addon`` rather than trying to get it into the ``SpecializedAddonManager``.

It should be noted that this function provides the ``Addon`` with a pointer to its manager, so the ``Addon`` can determine whether it's required for this particular type of manager. So if we have a ``FooAddon`` that's "required" only for a ``FooManager`` but we attach it to a ``GeneralManager``, then the function ``FooAddon::isOptional(AddonManager* mgr)`` can dynamic_cast the ``mgr`` argument to a ``FooManager*``. If the dynamic_cast fails, then ``FooAddon::isOptional`` will return ``true``, if the cast succeeds then it will return ``false``.

dart/common/detail/Addon.h
jslee02: Also, it seems it should be:
```cpp
return std::unique_ptr<Base>(new Base((castedManager), mProperties));
```

dart/common/Addon.h
mxgrey: I intended that to be a message to the developers, although I suppose it's useful information for users as well, so perhaps it's worth having it in the doxygen.

dart/common/detail/Addon.h
```diff
@@ -0,0 +1,338 @@
...
+//==============================================================================
+template <class Base, typename StateData,
+          class ManagerType, void (*updateState)(Base*)>
+std::unique_ptr<Addon>
+AddonWithProtectedState<Base, StateData, ManagerType, updateState>::
+    cloneAddon(AddonManager* newManager) const
+{
+  DART_COMMON_CAST_NEW_MANAGER_TYPE_AND_RETURN_NULL_IF_BAD(
+        Base, ManagerType, newManager, castedManager, clone);
+  return std::unique_ptr<Base>(new Base(newManager, mState));
```
jslee02: Here also:
```cpp
return std::unique_ptr<Base>(new Base((castedManager), mProperties));
```

dart/common/detail/AddonManager.h
```diff
@@ -0,0 +1,177 @@
...
+    extraction = std::unique_ptr<T>(static_cast<T*>(it->second.release()));
+
+  return extraction;
+}
+
+//==============================================================================
+template <class T>
+constexpr bool AddonManager::isSpecializedFor()
+{
+  return _isSpecializedFor(type<T>());
```
jslee02: Do we need this? `AddonManager` wouldn't have any specialized addon anyways. If this makes sense, this function should just return false.

dart/common/detail/Addon.h
mxgrey: I don't see where a segfault would be created. Are you referring to the abort caused by ``assert(false)`` on line 58? I believe most (probably all?) compilers will wash away calls to ``assert()`` when compiling in release mode, so it's still useful to make sure we return a ``nullptr`` in case this issue arises on code compiled in release mode.

dart/common/AddonManager.cpp
```diff
@@ -0,0 +1,261 @@
...
+void AddonManager::duplicateAddons(const AddonManager* otherManager)
+{
+  if(nullptr == otherManager)
+  {
+    dterr << "[AddonManager::duplicateAddons] You have asked to duplicate the "
+          << "Addons of a nullptr, which is not allowed!\n";
+    assert(false);
+    return;
+  }
+
```
jslee02: Probably we want early return here if `otherManager == this`.

dart/common/detail/Addon.h
jslee02: Ah, right. I forgot how `assert()` works in release mode.

dart/common/detail/Addon.h
mxgrey: The constructor is supposed to expect a generic ``AddonManager*``, so that's what we give it. If we gave it the casted version, there is a risk that it will be a plain ``nullptr``, which is not a scenario that I think the derived Addon type should have to handle.

dart/common/AddonManager.cpp
mxgrey: :+1:

dart/common/detail/AddonManager.h
mxgrey: This was required for an earlier implementation of the ``SpecializedAddonJoiner``, but it looks like that requirement doesn't exist anymore, so I agree :+1:

dart/common/Addon.cpp
jslee02: Actually, I'm not sure if we really need this function. (If there wouldn't be any additional use case,) It's currently used in AddonManager and SpecializedManager to check the legality of removing from the managers. In AddonManager, we might don't need it since all the addons in AddonManager are not specialized. For SpecializedManager, we can check the legality in more simple way. For example, `erase()` function can be implemented as:
```cpp
template <class SpecAddon>
template <class T>
void SpecializedAddonManager<SpecAddon>::erase()
{
  if (typeid( T ).name() == mSpecAddonIterator->first.name())
  {
    std::cout << "Attempting to erase specialized addon, which is not allowed.\n";
    return;
  }

  AddonManager::erase<T>();
//  _erase(type<T>());
}
```

mxgrey: The thing is, not all specialized addons are required. For example, for ``ShapeNode``, we will have a ``VisualData`` and a ``CollisionData`` specialized addon, but neither of them are **required** to be there, because their absence just means that the ShapeNode doesn't want to visualize or doesn't want to collide. On the other hand, ``RevoluteJoint`` has a ``RevoluteJointAddon`` which contains axis properties, and that addon **is** required because the dynamics and kinematics computations can't work without it.

A specialized addon just allows us to have constant-time access to it, and does not imply that the addon is required.

jslee02: It makes sense, but I wonder if there is a use case that the optionality should be checked from addon instead of from the specific manager we want to check. We can check it from `FooManager` and `GeneralManager` for `FooAddon` for  the above case.

Even though there are use cases, `isOptional()` doesn't seem to work as we expect. For example, `AddonWithProtectedPropertiesInSkeleton::isOptional()` would always return true regardless the passed in manager type if true is passed in for the template parameter `OptionalT`.

dart/common/Addon.h
jslee02: Yeah, the message is useful for the developers, but, as you mentioned, the users would like to know the information if they want to create their own addons.

dart/common/detail/Addon.h
jslee02: That makes sense to me. Then other lines should use `newManager` as well, right? I found three lines passing in the cased version in `common/detail/Addon.h` and `dynamics/detail/Addon.h`.

dart/common/Addon.cpp
mxgrey: Regarding the first point: I couldn't figure out a way of embedding the "optional" information into ``FooManager`` or ``GeneralManager`` such that it could be accessed in constant-time.

Regarding the second point, I think the existing behavior makes sense (although if you disagree, I'm open to discussing it). The ``OptionalT`` template parameter defaults to ``true``, meaning we do not generally expect an Addon to be required. However, if the Addon **is** meant to be required for its particular type of manager (given as the ``ManagerT`` template parameter), then ``AddonWithProtectedPropertiesInSkeleton::isOptional`` will check the manager type to see if it matches ``ManagerT``. If the manager type is some other manager type, then it will return ``true`` for being optional, otherwise it returns ``false`` to indicate that it's a required Addon for that type of manager.

So for example, a ``VisualData`` or ``CollisionData`` addon would have ``OptionalT = true``, indicating that they are optional for any type of manager that they get attached to. Conversely, ``RevoluteJointAddon`` would have ``OptionalT = false`` and ``ManagerT = RevoluteJoint`` indicating that it is required for any manager type that has ``RevoluteJoint`` in its inheritance hierarchy, but it is optional for any other manager type.

dart/common/detail/AddonManager.h
```diff
@@ -0,0 +1,177 @@
...
+
+//==============================================================================
+template <class T>
+std::unique_ptr<T> AddonManager::release()
+{
+  std::unique_ptr<T> extraction = nullptr;
+  AddonMap::iterator it = mAddonMap.find( typeid(T) );
+  DART_COMMON_CHECK_ILLEGAL_ADDON_ERASE(release, it, nullptr)
+  if(mAddonMap.end() != it)
+    extraction = std::unique_ptr<T>(static_cast<T*>(it->second.release()));
```
jslee02: This is a question rather than a suggestion. Is `static_cast` necessary here? Wouldn't it be implicitly casted? If it was to make it explicit then please ignore me.

dart/common/detail/Addon.h
mxgrey: I think it'll depend on the context. I'd have to see each case to know which one should be used. I vaguely remember places where I needed to use the casted version, but it's possible that those reasons are now outdated.

dart/common/detail/AddonManager.h
```diff
@@ -0,0 +1,177 @@
...
+}
+
+//==============================================================================
+template <class T>
+void AddonManager::erase()
+{
+  AddonMap::iterator it = mAddonMap.find( typeid(T) );
+  DART_COMMON_CHECK_ILLEGAL_ADDON_ERASE(erase, it, DART_BLANK)
+  if(mAddonMap.end() != it)
+    it->second = nullptr;
```
jslee02: Do we not actually remove the erased item from the map on purpose? Of course, we shouldn't remove items of specialized addons but general addon manager seems to fine to remove the erased items.

dart/common/AddonManager.h
jslee02: Adding `replace()` sounds good to me. Do we implement it in this pull request or in a separate one?

dart/common/detail/AddonManager.h
mxgrey: ``mAddonMap`` contains generic ``Addon*`` entries, so to obtain a concrete type ``T`` we need to downcast, which cannot be done implicitly, so ``static_cast`` is required.

dart/common/SpecializedAddonManager.h
```diff
@@ -0,0 +1,195 @@
...
+
+  /// Check if this Manager is specialized for a specific type of Addon
+  template <class T>
+  static constexpr bool isSpecializedFor();
+
+protected:
+
+  /// Redirect to AddonManager::has()
+  template <class T>
+  bool _has(type<T>) const;
```
jslee02: Well, we don't have any documented convention for code styles yet, but I thought unofficially we prefer not to use leading underscores in naming. I would prefer to rename these method names just not to use the underscores. It wouldn't cause confliction with the public version functions since they have different parameters. Or we can consider using other names such as `hasImpl()`, but my preference would be the names just removed the underscores.

dart/common/AddonManager.h
mxgrey: I'm indifferent, although it would probably be cleaner to hold it off for a second pull request.

dart/common/detail/AddonManager.h
mxgrey: Yes, this is on purpose, and it's very important for two reasons:

1. It's possible for the user to access a specialized manager from a reference to the basic AddonManager. In that case, removing the entry entirely will definitely invalidate the iterator that's being held by the specialized manager.

2. All iterators for a ``std::map`` are guaranteed to remain valid as long as no entries are removed. Removing any entry from a ``std::map`` would risk invalidating any still existing iterators. So even if this entry is not a specialized addon, removing it from the map would risk invalidating any existing pointers to specialized entries.

It would probably be a good idea to put a developer's note about this in the comments of this function.

dart/common/SpecializedAddonManager.h
mxgrey: I did this for a very very very specific reason: to keep these functions out of the public API. From the compiler's point of view, you're right, there's no conflict between the signatures of these functions and the functions that they're implementing. But from the user's point of view, it's potentially confusing for them to know which one they're supposed to use. An IDE like Qt Creator or Visual Studio will show these functions side-by-side with the functions that they're actually supposed to use. Beginning them with an underscore prevents this.

Adding ``Impl`` to the end of the names would remove some of the ambiguity, but in an alphabetical sorting of function names, it would still show up alongside the functions that are actually supposed to be used.

If for some reason the underscore is unacceptable, then I would prefer something like ``implHas(type<T>)`` to keep these functions grouped together and away from the functions that are actually a part of the public API.

dart/common/detail/TemplateJoinerDispatchMacro.h
```diff
@@ -0,0 +1,57 @@
...
+/// provides the implementation for all of them.
+#define DETAIL_DART_COMMON_TEMPLATEJOINERDISPATCH_IMPL(ReturnType, ClassName, Function, Suffix, Args)\
+  template <class Base1, class Base2>\
+  template <class T>\
+  ReturnType ClassName <Base1, Base2>:: Function Suffix\
+  {\
+    if(Base1::template isSpecializedFor<T>())\
+      return Base1::template Function <T> Args ;\
+  \
+    return Base2::template Function <T> Args ;\
```
jslee02: I have one concern on how this macro resolve the request for a joined (specialized) addon manager. Suppose we have a joined addon manager of `FooSpecializedAddonManager` (Base1) and `AnotherFooSpecializedAddonManager` (Base2) where the sub managers are all specialized for `FooAddon`, and `AnotherFooSpecializedAddonManager` has `FooAddon` (containing `FooAddon` instance not `nullptr`) and `FooSpecializedAddonManager` doesn't. Then this macro will return false as it meets the if condition while `AnotherFooSpecializedAddonManager` has the addon.

So I wonder if we can make the logic replaceable by passing the logic object as a template parameter to `AddonManagerJoiner`. We can pass in this logic as a default if we want to.

dart/common/AddonManager.h
jslee02: Either ways would be okay to me as well. I just wonder if it's in the checklist.

dart/common/detail/TemplateJoinerDispatchMacro.h
mxgrey: I actually have a solution to that, which you can see for [SpecializedNodeManager](https://github.com/dartsim/dart/blob/bded5301b75e49760d9fcfc71522b54524becafd/dart/dynamics/SpecializedNodeManager.h#L115).

Unfortunately, as I look at the implementation of [SpecializedAddonManager](https://github.com/dartsim/dart/blob/bded5301b75e49760d9fcfc71522b54524becafd/dart/common/SpecializedAddonManager.h#L185), I'm noticing that I've forgotten to put it in, but I can push a change to fix that very quickly.

But the solution is this: When inheriting a ``SpecializedAddonManager``, it is always virtually inherited, so that only one copy of its contents actually exists. Even if it exists through multiple lines of inheritance, no matter which way down the inheritance chain you go, it will arrive at the same ``SpecializedAddonManager``.

dart/common/detail/AddonManager.h
jslee02: The two reasons definitely make sense. I realized that the modifiers of managers are not virtual functions. 

:+1: to make a note for this.

Now I'm thinking if it would be a good design that a manager has another fixed size map (or even simpler data structure) for special addons. Then we wouldn't need to worry about the points you mentioned when both of general addons and specialized addons are stored in a single map. By the way, this wouldn't impact the API so (if this makes sense) we can discuss this later.

mxgrey: It's not possible to make templated functions virtual, and I imagine we would need these functions to be virtual in order to achieve what I think you're suggesting, because otherwise we would be at risk of having two different versions of the same Addon type within the same manager.

dart/common/detail/TemplateJoinerDispatchMacro.h
mxgrey: The fix has been pushed.

dart/common/detail/AddonManager.h
jslee02: Cool. Once `AddonManage::_isSpecializedFor()` is removed, it would make sense to move `template <class T> struct type` to `SpecializedAddonManager`.

dart/common/SpecializedAddonManager.h
jslee02: I see your point. How about making these methods as private then? Maybe we can move these methods to source file by making them static functions, but I'm not sure if it's worse.

mxgrey: They require access to members of the class, but I guess we could make the implementation functions friends of the class? It wouldn't require an API change, so I'd recommend saving it for a later pull request.

jslee02: For these cases, I would pass the required members to the static functions (or functions within an anonymous namespace). On the other hand, exposing members seems not a preferred way.

Anyways, :+1: to "saving it for a later pull request".

jslee02:
If the current master branch is merged into this, the AppVeyor can successfully build this PR thanks to #595. I created DART dependency binaries for Visual Studio 2015.

dart/dynamics/Addon.h
```diff
@@ -0,0 +1,372 @@
...
+//==============================================================================
+/// AddonWithProtectedPropertiesInSkeleton generates implementations of the
+/// Property managing functions for an Addon class which is embedded within a
+/// Skeleton or a component of a Skeleton (such as a BodyNode, Joint, or custom
+/// Node class). This will increment the version count any time the
+/// Addon::setProperties function is called.
+template <class BaseT, typename PropertiesDataT, class ManagerT = Node,
+          void (*updateProperties)(BaseT*) = common::detail::NoOp<BaseT*>,
+          bool OptionalT = true>
+class AddonWithProtectedPropertiesInSkeleton : public common::Addon
```
jslee02: Is there a special reason why this class is not inherited from `AddonWithProtectedState`?

dart/common/Addon.cpp
jslee02: I basically think current behavior of `AddonWithProtectedPropertiesInSkeleton::isOptional` makes sense. My concern is that there is a pontial danger when the user doesn't implement the function of another customized addon properly. Moreover, if we add a non-optional addon to general manager, the addon cannot be removed from the manager. For this case, it would make more sense if the manager is specialized for the addon.

An expert user may be possible to avoid these situation with a complete understanding of the API. My point is it would be safer if we can prevent these situations in API level.

I think I found a way of embedding the "optional" information into managers that work in constant-time. It should behave as current implementation as well. Here is the [example code](https://gist.github.com/jslee02/d4bd440df1f3c1a64bb9). The idea is we pass a pair of addon type and optionality into a specialized manager as a template parameter, and the manager returns the optionality. Also, general managers shouldn't have non-optional addon, and always return true for isOptional function regardless of addon type.

dart/common/detail/TemplateJoinerDispatchMacro.h
jslee02: The fix looks good :+1:

dart/common/Addon.cpp
mxgrey: There's a very crucial vulnerability in your proposal, though, and I'll illustrate it with an example.

``RevoluteJoint`` has a mandatory ``RevoluteJointAddon`` which is not optional, because the addon is needed in order to compute the kinematics and dynamics of the joint. However, in the vast majority of cases, people interact with the ``Joint`` class, which is not specialized for ``RevoluteJointAddon``. So if we embed the non-optional information at the ``RevoluteJoint`` level, then a user could potentially call ``Joint::remove<RevoluteJointAddon>()``, and the general addon manager of ``Joint`` would allow the addon to be removed.

You could suggest that we specialize the manager of ``Joint`` to be specialized for all the different joint type addons, and make all those addons mandatory, but then if someone gives a ``PrismaticJointAddon`` to a ``RevoluteJoint`` object, there will be no way to remove that addon even though it isn't really meant to be mandatory.

Of course this example is very contrived. I would never expect a sane user to fiddle around with the various joint type addons (... then again, perhaps we can't rely on all users to be sane at all times...). But the fundamental issue remains: in the scheme you've proposed, a base class reference has no way of knowing that any particular addon is mandatory for its fully derived type, because templated member functions can't be virtual.

dart/common/detail/TemplateJoinerDispatchMacro.h
mxgrey: I do have one last concern, though... in the current scheme, if a user inherits the single-argument ``SpecializedAddonManager`` or ``SpecializedNodeManager`` directly (i.e. non-virtually), then it could cause issues. I'm thinking we should actually split out the iterator into a base class (maybe called something like ``detail::SpecializedAddonManagerData``) which will be virtually inherited by ``SpecializedNodeManager``. I think that would **guarantee** under all circumstances that there aren't any multiple-inheritance conflicts.

dart/common/Addon.cpp
jslee02: I one-hundred percent agree that we can't rely on all users to be sane at all time, and the vulnerability in my proposal is true. The current AddonManager::isSpecializedFor() has same issue I think. Also, specializing `Joint` for every specific addons doesn't sound good to me as well.

By the way, I think I found a way to improve my proposal that doesn't have the vulnerability. Here is [the revised code](https://gist.github.com/jslee02/d4bd440df1f3c1a64bb9) (same like but the code was updated). 

The main idea is what I suggested in another comment: using separate map for specialized addons. The type of map is slightly different from the original. The key type of the specialized addon map is `std::pair< unique_ptr< Addon, bool > >`, and we store the optionality as a second value of the pair in the SpecializedAddonManager constructor. Then, the base class, `Joint : AddonManager` in your example case, now has the information that particular addon is mandatory for any derived type.

mxgrey: >  The current AddonManager::isSpecializedFor() has same issue I think

We had already decided that we just won't offer constant-time access when accessing an addon from a reference to a base type which is not specialized for that addon. It's definitely a related issue.

The revised version looks like it should work, but I would suggest we use an unordered_map instead to achieve constant-time lookup. We don't have the same constraints on that map which we do for the map of addons which requires an ordered map. Also, if we're embedding that map at the lowest level instead of in templates, then I wonder if we can figure out a way to separate addon specialization from addon requirement. For example, we could have a templated ``MandatoryAddon`` class similar to the ``SpecializedAddonManager`` which takes a list of Addon types and sets them to be mandatory. Then instead of constructing a bunch of AddonPairs, you would have the class inherit a ``SpecializedAddonManager<Mandatory1, Optional, Mandatory2>`` for the specialized addons and then inherit a ``MandatoryAddon<Mandatory1, Mandatory2>`` to set which addons are mandatory. Addons which are left out of the MandatoryAddon list are defaulted to optional.

jslee02: It seems possible if we can make sure the constructor of `SpecializedAddonManager` is called before `MandatoryAddon`'s one called. What the constructor of `MandatoryAddon` needs to do is to set the optionality of the target addon in the map to false if the target addon exists. `SpecializedAddonManager` will set the value just false.

I'm thinking of order agnostic way by make `MandatoryAddon` inherit `SpecializedAddonManager` so that `MandatoryAddon` adds the target addon when it doesn't exists already.

jslee02: [Here](https://gist.github.com/jslee02/6d642d73cb0023c61243) is another implementation for mandatory addon class.

mxgrey: The current scheme for identifying specialization has zero overhead because specialization is determined at compile-time (notice that my version of ``isSpecializedFor`` is a constexpr). Grabbing a specialized addon is not only constant-time, it is essentially instant. The scheme you're proposing would introduce potentially substantial overhead, especially if the Addon is accessed frequently.

> to using unordered_map for specialized addon map.

That's actually not what I meant to say. I meant to say that an unordered_map should be used to keep track of which addons are **mandatory**, not which ones are specialized. It's important to be clear that specialized addons are not necessarily mandatory, and vice-versa.

MandatoryAddon does not need to be called before SpecializedAddonManager if we store the unordered_map of mandatory addons at the level of the most basic general manager. However, I do agree with having MandatoryAddon inherit SpecializedAddonManager, because there's no reason not to do that.

By the way, I think any of these changes that we're discussing should go into a separate pull request to make it easier to review and evaluate.

jslee02: > The scheme you're proposing would introduce potentially substantial overhead, especially if the Addon is accessed frequently.

You're right, but again I wonder how `Joint : AddonManager` can get the information which addon is specialized by derived specialized addon managers. If we can think of a way to store (concatenate) types into AddonManager, then we may be able to achieve that.

> That's actually not what I meant to say. I meant to say that an unordered_map should be used to keep track of which addons are mandatory, not which ones are specialized. It's important to be clear that specialized addons are not necessarily mandatory, and vice-versa.

I keep being confused specialized addons and mandatory addons..



> MandatoryAddon does not need to be called before SpecializedAddonManager if we store the unordered_map of mandatory addons at the level of the most basic general manager.

My concern was the case that the target addon of MandatoryAddon doesn't exists (hasn't added by SpecialAddonManager) when the constructor of MandatoryAddon is trying to set the optionality to false.




> By the way, I think any of these changes that we're discussing should go into a separate pull request to make it easier to review and evaluate.

Agreed. I would like to merge #598 into this, if it looks good to you, and merge this PR to master soon.

mxgrey: > You're right, but again I wonder how Joint : AddonManager can get the information which addon is specialized by derived specialized addon managers.

It can't without substantial overhead. What we had decided on with respect to addon specialization is that the user should be required to downcast the manager's reference in order to achieve high performance. So if you have a ``Joint*`` but you know it's actually a ``RevoluteJoint*``, then you should use ``dynamic_cast`` (or ``static_cast`` if you're very confident) to convert it to a ``RevoluteJoint*`` and **then** call for its specialized addon.

> My concern was the case that the target addon of MandatoryAddon doesn't exists (hasn't added by SpecialAddonManager) when the constructor of MandatoryAddon is trying to set the optionality to false.

This shouldn't be an issue. If we use a type map which is stored in the basic manager class, then it doesn't even matter whether or not the addon is specialized, we could still set it to be mandatory.

#598 looks good to me :+1:

jslee02: > It can't without substantial overhead. What we had decided on with respect to addon specialization is that the user should be required to downcast the manager's reference in order to achieve high performance. So if you have a Joint* but you know it's actually a RevoluteJoint*, then you should use dynamic_cast (or static_cast if you're very confident) to convert it to a RevoluteJoint* and then call for its specialized addon.

Makes sense to me.


> This shouldn't be an issue. If we use a type map which is stored in the basic manager class, then it doesn't even matter whether or not the addon is specialized, we could still set it to be mandatory.

I'm a bit confused. How MandatoryAddon can set an addon to be mandatory when the addon doesn't exist in the map, which is stored in the basic manager class. If we let MandatoryAddon add the addon in that case, it would add all the addons even for addons that are not passed in SpecializedAddonManager.

Could you tell me which behavior is what you think for `Joint : SpecializedAddonManager<Addon1, Addon3>, MandatoryAddon<Addon1, Addon2>`?
(1) Joint has a Addon1, which is mandatory, and Addon3, which is optional.
(2) Joint has Addon1 and Addon2, which are all mandatory, and Addon3, which is optional.

mxgrey: Yes, the MandatoryAddon class would add an entry to the unordered_map.

In fact, I would suggest a slight change: we use a ``std::unordered_set<type_id> mMandatoryAddons``. If an entry for an Addon exists in ``mMandatoryAddons`` then it's mandatory; otherwise it's optional.

mxgrey: As far as whether I expect the behavior of (1) or (2) it will depend on whether we have ``MandatoryAddon`` also specialize the addon. I think we should, because why not? If an addon is going to be mandatory, we may as well specialize it as well, because clearly we expect it to exist.

jslee02: It sounds like the map can have any addons regardless of each addon is optional or mandatory. Hm, I'm more confused, haha. Here is my understanding. Please let me know which one is different from you.
* Unspecialized addons are always optional.
* Specialized addons can be one of optional and mandatory.

In that sense, it seems the mandatory addons should be a subset of specialized addons. So I stored the optionality within specialized addon map (that is stored in the basic addon manager) in [the code](https://gist.github.com/jslee02/6d642d73cb0023c61243).

mxgrey: I don't really agree with the subset relationship. Here's how I think it should work:

* A manager that is specialized for an addon type will offer zero-overhead access to that addon
* If an addon is mandatory for a certain manager type, then there must be **no** possible way to remove the addon from that manager

These are two completely independent concepts. The only point at which they overlap is that if an addon is mandatory for a certain manager type, then we *may as well* also specialize the manager for that addon.

jslee02: I see your point. What I'm not clear is why we open the case that an addon is mandatory but not specialized. Wouldn't it be too strong restriction if we enforce the user use specialized manager for every mandatory addons so that the user can always take the advantage of the zero-overhaed access to that addon?

mxgrey: Right, I agree. I think the concepts of specialized and mandatory are inherently independent, but if an addon is important enough to be mandatory, then we may as well have the API also specialize it, especially since it would be as easy as having ``MandatoryAddon`` virtually inherit ``SpecializedAddonManager``.

jslee02:
It seems this pull request is ready to merge. There are several items discussed in this pull request that might need some changes in code, but we save them for later pull requests to make it easier to review and evaluate.

DART now have very useful features of flexible data structures, which is another big improvement you made. Nice job @mxgrey ! Merging now.

mkoval:
:+1: Glad this got merged!

I have one follow-up question from the discussion: What is the purpose of a required `Addon` when we could simply put those attributes in the `AddonManager`?

Using @mxgrey's example:

> On the other hand, RevoluteJoint has a `RevoluteJointAddon` which contains axis properties, and that addon is required because the dynamics and kinematics computations can't work without it.

Why not just put these attributes in the `RevoluteJoint` class? Is the idea to use `AddonManager` to generate the `Property` and `State` boilerplate?

mxgrey:
> Is the idea to use AddonManager to generate the Property and State boilerplate?

This is exactly right. And not just generate the boilerplate: By storing joint-type-dependent property data in an Addon, it will naturally get sucked into the ``Skeleton::ExtendedProperties``. This way, we don't need to do any special handling on a per-joint-type basis in order to take a comprehensive snapshot of all the joint property data in a Skeleton. This is especially important for handling the properties and states of user-created custom joint types.

mkoval:
:+1: Very nice!

PR #533, opened 2015-10-20
Workaround for ftell bug in Linux

mkoval:
`ftell` returns `LONG_MAX` as an error code, instead of `-1L` as specified in the documentation, when invoked on a directory. This behavior occurs on EXT4 filesystems, not but NFSv4 file systems, in kernel version 3.16.0 possibly with Ubuntu-specific patches applied (tested in Ubuntu 14.04).

This was not a fun one to debug. :sweat:

### DIFF
--- a/dart/common/LocalResource.cpp
+++ b/dart/common/LocalResource.cpp
@@ -35,6 +35,7 @@
  */
 
 #include <cstring>
+#include <limits>
 #include <iostream>
 #include "dart/common/Console.h"
 #include "LocalResource.h"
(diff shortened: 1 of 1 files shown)


jslee02:
Looks good to me :+1: 

System or compiler dependent behavior is mostly troublesome..

mxgrey:
Looks good. :+1:  AppVeyor might need a restart, though.

jslee02:
AppVeyor keeps failing for release mode tests because they exceeds the limited time, 60 min. But it should have nothing to do with this PR. Merging..

If anyone have an idea to boost AppVeyor buildings, I would like to share. :sweat:

Issue #532, opened 2015-10-19
ResourceRetriever and Resource methods are now const


mkoval:
It looks like #517 `const`-qualified all of the methods on the `ResourceRetriever` and `Resource` classes. I intentionally left these non-`const` when I added these classes in #464 so the user can implement a cache in his or her custom implementations. This is important when retrieving resources over the network or when resolving the URI is computationally expensive.

I just ran into this situation myself and had to make a bunch of my member variables `mutable` as a workaround. I don't see any advantage to making these methods `const`, so I would prefer to remove the `const`-qualifiers added in #517.

jslee02:
I though the methods, `exists` and `retrieve`, should be `const`-qualified by the meaning of them like we generally don't expect that getter functions modify the member variables. As I known, using `mutable` members for caching data (required expensive computation) is well known technique, called logical constness, as we already use it in the auto-updating kinematics mechanism.

Above is my understand so far, and I would prefer to keep the consistency of the use of cost correctness. However, I'm not sure which one is practically preferred way.

mkoval:
I don't think `exists` and `retrieve` are logically const.

`ResourceRetriever`s operate on some sort of remote data store, e.g. the filesystem, the network, or a database. Even if our implementation of those methods are logically const, there is no way we can guarantee that the remote operation is logically const. For example, making an HTTP `GET` request could modify state on the server. The caller can't distinguish between mutation of state in the `ResourceRetriever` and this type of mutation of state on the remote server.

@mxgrey @psigen Thoughts?

psigen:
I don't think that `const` qualification really affects to this case.  Here, the `const` qualifier is indicating that nothing in our data structure is being mutated by the `exists` or `retrieve` call, which is about all we could ever guarantee.  Nothing about our internal logic or state will change as a result of these calls.

We can never make *any* guarantees about remote calls.  Making the same remote call twice might yield different results just because the file on disk changed, or the server drops an HTTP call, or any number of arbitrary conditions, but that has nothing to do with the `const` nature.  As long as we do not change the internal state as a result, and we will follow the exact same logic when we make that call again, I think we can call these `const`.

mxgrey:
I think there's a compelling enough case for them to not be ``const``. I suppose the big question would boil down to: What would be the meaning of a ``const ResourceRetriever``? What would you expect to be able to do and to *not* do with a ``const ResourceRetriever``?

[...]

PR #534, opened 2015-10-21
Remove const qualifier of ResourceRetriever methods

jslee02:
This pull request reverts the const-qualification on the ResourceRetriever methods introduced by #517.

Please see #532 for more details.


mkoval:
:+1: Thanks @jslee02!

PR #535, opened 2015-10-22
Added missing 'override' keywords

mkoval:
This pull request adds several missing `override` keywords that produce warnings in clang 3.7. I also removed a few `virtual` keywords on functions methods that are flagged as `override`. [The `virtual` keyword has no effect if you're overriding a `virtual` function](http://stackoverflow.com/a/29760395/111426), so this is not necessary. It also seems to be the style used in the majority of DART.


jslee02:
:+1:

mxgrey:
:+1:

PR #537, opened 2015-10-25
Delete copy constructors for non-copyable classes (round 2) -- Patch for 5.1

mxgrey:
Some invalid copy constructors were neglected with the last pull request, so this pull request deletes more of them.


mxgrey:
This pull request is targeting the wrong branch.

PR #539, opened 2015-10-25
Delete copy constructors for non-copyable classes (round 2) -- Patch for 5.1

mxgrey:
Same as #538 , but it's a patch for 5.1.


jslee02:
:+1:

PR #542, opened 2015-11-02
Conditional build depending on missing symbols of Assimp package

jslee02:
The Debian packages of Assimp in Ubuntu is missing C++ symbols so DART has our own implementation of them, and we made DART depend on the version of Assimp that is missing symbols by #455. However, it can be problematic when DART is built with Assimp version that contains the missing symbols.

This pull request enables DART to detect whether the installed Assimp has the missing symbols or not in cmake time, and compile our own implementation of them only when they are missing in the Assimp.


mkoval:
:+1: Good idea! This is much nicer than the versioned dependency I added in #455.

jslee02:
Thanks @mkoval. I hope Assimp resolves the missing symbol issues soon.

@mxgrey any comment? If it looks good to you too then I would like to merge.

mxgrey:
Very clever use of cmake. Looks good to me :+1:

jslee02:
Yeah, it's very useful feature of cmake. Thanks!

PR #543, opened 2015-11-03
Add cmake option DART_MSVC_DEFAULT_OPTIONS for Visual Studio build

jslee02:
We have optimized build options of Visual Studio to find best ones for DART. However, the customized options seems to increase the build time so that it can not pass AppVeyor build tests: every build should be done within 1 hour, which is allowed to a testing project.

This pull request adds an cmake option that builds DART with reduced version of custom options so that the build time is reduced. This option is only turned on for AppVeyor tests, and, by default, this option is set to off so the default behavior wouldn't be changed from the previous.


jslee02:
Waiting for AppVeyor to be happy before merging.

PR #544, opened 2015-11-03
Switch caller and callee to suppress warnings

jslee02:


mkoval:
:+1: These warnings always bothered me when building DART, but I couldn't figure out how to fix them. I'm glad you did. :smile:

mxgrey:
This pull request breaks backwards compatibility. If someone extended the ``Function`` class such that they overrided the ``eval(Eigen::Map<Eigen::VectorXd>)`` function signature (which is how it worked up until 5.1), then their derived function will no longer work correctly when passed into a ``Solver`` (it will just behave like a constant 0.0 function instead). The warning was bothersome, but it was a symptom of maintaining backwards compatibility until the next major release.

mxgrey:
It should be noted that the ``Solver`` implementations are calling the ``eval(const Eigen::VectorXd&)`` function signature now, and **not** the ``eval(Eigen::Map<Eigen::VectorXd>)`` function signature.

jslee02:
Yeah, this warnings were annoying. :cry: However, I missed the possibility that @mxgrey pointed out.

It seems we need to keep this as it was for the backward compatibility. Instead, I would like to suppress the warnings for this specific case. I think I found a way to suppress warnings for specific lines but it is not cross platform.

I will close this PR and create another one for it.

psigen:
Is this something that's worth changing the API over on the next major release?  It seems undesirable to accumulate warning suppression cruft over time.

mxgrey:
Yes, my plan was to remove the old method with the next major release.

PR #547, opened 2015-11-03
new bounding box semantics

mklingen:
Replaces old `boundingBoxDim` semantics with a new `BoundingBox` class that has a minimum and maximum. This reduces ambiguity and allows for off-center bounding boxes.


dart/dynamics/Shape.h
```diff
@@ -121,7 +126,7 @@ class Shape : public virtual common::Subject
   //           biased mesh shape. Two Vector3ds might be better; one is for
   //           minimum verterx, and the other is for maximum verterx of the
   //           bounding box.
-  const Eigen::Vector3d& getBoundingBoxDim() const;
+  Eigen::Vector3d getBoundingBoxDim() const;
```
mkoval: This function should be marked as `DEPRECATED` and the `TODO` should probably be made a comment that points the user to `getBoundingBox`.

dart/math/Geometry.h
```diff
@@ -533,6 +533,35 @@ Eigen::Vector2d computeClosestPointOnSupportPolygon(
     const Eigen::Vector2d& _p,
     const SupportPolygon& _support);
 
+
+// Represents a bounding box with minimum and maximum coordinates.
+class BoundingBox {
```
mkoval: I'd prefer if this class had AABB in the name, since some algorithms (e.g. inside FCL) also use OBBs.

mklingen: I think that's very confusing. All bounding boxes (including OBBs) are axis aligned in their own reference frame. In practice these are actually OBBs since they are attached to a moving reference frame (e.g a body node)

dart/dynamics/Shape.h
mklingen: Okay, whats the proper way of deprecating in dart?

mkoval: There is a `DEPRECATED` macro defined in [`Deprecated.h`](https://github.com/dartsim/dart/blob/e8b4f1c40e26140d93ea803bcb4a76547ebe05fb/dart/common/Deprecated.h) that has `#ifdef`s for all of the supported compilers. The argument is the major/minor version number when the deprecation occurred.

DART 5.1.1 was just released, so I suppose it should be `DEPRECATED(5.2)` now?

jslee02: `DEPRECATED(5.2)` would be fine. In general, we might not be sure the deprecation is included in which release at the moment. So I usually increment the minor version with 1 and update the version number if needed when DART is released.

dart/math/Geometry.h
mxgrey: I agree with keeping the name as just ``BoundingBox`` and not saying anything about AABB versus OBB. The class itself is fundamentally just describing the data of a bounding box, and whether it's AABB or OBB will depend on the context in which it's used. In most cases, the Shape that it belongs to will have some arbitrary orientation with respect to the "world" coordinates, so it would really be OBB. In order for it to be strictly AABB, we would need to have a way to compute the bounding box as a function of the shape's orientation.

mxgrey:
This pull request looks reasonable to me. The only question I would have is: Are we currently using the bounding box dimension information anywhere meaningful in the DART source code? All of the changes seem to be internal to the Shape class and its derivatives. Are we using bounding boxes in the collision detection components of DART, or are we just letting the third-party libraries deal with those? I'll check this myself, but if anyone already knows the answer, then that might save me some time.

mxgrey:
It doesn't look like we need to worry about making any other changes to the code, since the bounding boxes are only used internally by shapes right now. :+1:

dart/dynamics/Shape.h
jslee02: This code would work for the deprecating.

```cpp
//
// Shape.h
//
/// [Original comment]
/// \deprecated Please use getBoundingBox() instead
DEPRECATED(5.2)
const Eigen::Vector3d& getBoundingBoxDim() const;

//
// Shape.cpp
//
const Eigen::Vector3d& Shape::getBoundingBoxDim() const
{
  static Eigen::Vector3d boundingBoxDim = mBoundingBox.computeFullExtents();
  return boundingBoxDim;

  // or by not removing mBoundingBoxDim
  mBoundingBoxDim = mBoundingBox.computeFullExtents();
  return mBoundingBoxDim;
}
```

mklingen: Added this in the latest commit.

mklingen:
Okay, assuming the tests pass, can we merge this?

mxgrey:
Looks good to me :+1: 

If @jslee02 approves, I think it's ready to merge.

jslee02:
It looks good to me as well. Thank you for your contribution! Merging now.

PR #550, opened 2015-11-04
Suppress warnings in Function.cpp

jslee02:
For backward compatibility, we have to call deprecated functions in Function.cpp, but the compilers complain for these calls. This pull request suppresses the warnings only for those calls. See also #544.

Note that the suppression code couldn't be made as macros. It needs to use compiler dependent #pragma directives but directives is not allowed to be declared in macros.

### DIFF
--- a/dart/common/Deprecated.h
+++ b/dart/common/Deprecated.h
@@ -34,6 +34,8 @@
  *   POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "dart/config.h"
+
 #ifndef DART_COMMON_DEPRECATED_H_
 #define DART_COMMON_DEPRECATED_H_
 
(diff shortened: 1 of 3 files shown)


mxgrey:
I wonder if we could make this more reusable, for example maybe we could have

```
#define DART_SUPPRESS_DEPRECATED_START // begin suppressing deprecated warnings, based on the compiler
#define DART_SUPPRESS_DEPRECATED_END // stop suppressing deprecated warnings, based on the compiler
```

And then we could wrap those around any function calls that we don't want to see deprecation warnings for.

mkoval:
@mxgrey I don't think this is possible because of what @jslee02 said in his original post:

>  It needs to use compiler dependent #pragma directives but directives is not allowed to be declared in macros.

mxgrey:
Right, I see. I guess the only way to overcome it would be to do some pre-preprocessing, but it's probably not worth it to take that route for this.

mkoval:
Actually, that isn't quite true. [According to StackOverflow](http://stackoverflow.com/questions/3030099/pragma-in-define-macro) you can use `_Pragma("foo")` as an equivalent to `#pragma foo` that can be used in macros.

jslee02:
@mxgrey Yes, I started with similar approach you suggested, but couldn't find a way to contain `#pragma` in macro. However, `_Pragma(string)` seems to be able to do it, and `clang`, `gcc`, and `MSVC` all support this operator. I will try with it now. Thanks @mkoval !

jslee02:
Updated the warning suppression with macros using `_Pragma` operator.



> [Edit] This appears to be an addition in C++11.

Hm, I need to check if MSVC supports this.

mkoval:
As usual, MSVC is a giant pain in the ass. They use `__pragma(foo)` instead of `_Pragma("foo")`: note the two underscores, case, and lack of quotes. I think you can wrap this in a macro like this:

```c++
#ifdef _MSC_VER
#define PRAGMA(x) __pragma(x)
#else
#define PRAGMA(x) _Pragma(#x)
#endif
```

jslee02:
Right, I had to use `__pragma` for MSVC. Many times MSVC requires special treatment.

The wrapper looks good but I would prefer to use it when we found more use cases.

mxgrey:
This looks really good now :+1:

jslee02:
Thanks! Merging now.

PR #553, opened 2015-11-10
Improved wording of the cmake warning messages for ASSIMP

mxgrey:
I felt that the wording of the cmake warning messages for ASSIMP might be difficult to understand, so I attempted to improve their clarity a bit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,9 +170,11 @@ if(ASSIMP_FOUND)
   ASSIMP_AISCENE_CTOR_DTOR_DEFINED)
 
   if(NOT ASSIMP_AISCENE_CTOR_DTOR_DEFINED)
-    message(WARNING "Installed ASSIMP ${ASSIMP_VERSION} is missing symbols "
-            "for constructor or destructor of aiScene. DART will use own "
-            "implementation. Please use ASSIMP that fixed this issue.")
+    message(WARNING "The installed version of ASSIMP (${ASSIMP_VERSION}) is "
+                    "missing symbols for the constructor and/or destructor of "
+                    "aiScene. DART will use its own implementations of these "
+                    "functions. We recommend using a version of ASSIMP that "
+                    "does not have this issue.")
   endif(NOT ASSIMP_AISCENE_CTOR_DTOR_DEFINED)
 
   check_cxx_source_compiles(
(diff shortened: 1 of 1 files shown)


psigen:
:+1: These are a lot clearer.  Do we know which version of Assimp this issue was fixed in?  It may be helpful to specify "Assimp 3.0.0+" or something.

jslee02:
Looks better now :+1: 

There are two classes that Assimp `3.0~dfsg-3` (package in vivid) is missing: `aiMaterial` and `aiScene`. The symbols of `aiMaterial` were included in Assimp `3.0~dfsg-4` (package in wily) but not of `aiScene`. So any released package of Assimp has not completely fixed the issue yet.

mxgrey:
Then maybe we should add "once one becomes available" to the warning message.

jslee02:
Good idea. 

I haven't tested yet for all the platforms again, but it seems the versions of Assimp in the ubuntu package system have these issue. So I would like to recommend to use a version of ASSIMP **package** that does not have this issue once one becomes available or to use a version of ASSIMP built from the source.

jslee02:
This branch needs to merge `master` into this branch in order to be merged into `master` back. No conflicts was sufficient condition to be merged into `master` when `master` was not a [protected branch](https://help.github.com/articles/about-protected-branches/). However, now it seems to require to check if the changes of `master`, that was made after this branch was branched out, pass the status checks with this branch.

Regarding the warning message, just adding "once one becomes available" is also fine to me.

jslee02:
Either of @mxgrey 's suggestion and just current message looks good to me. I think we could merge this pull unless there is more idea.

mxgrey:
I'll tweak the message tonight, and then I think it'll be ready to merge.

jslee02:
Thanks, @mxgrey. Don't need to rush. Just wondered your think on this. Please tweak the message when you have a chance.

mxgrey:
The comment has been updated. I believe this is ready to merge.

mxgrey:
Since ``VskParser`` doesn't have any internal state and its only member function is a static function, we may want to consider making ``VskParser`` a namespace instead of a class. Otherwise having it as a class sort of implies that there may be a situation where you'd want to create an instance of it.

jslee02:
:+1:

Issue #556, opened 2015-11-12
ReferentialSkeletons do not include BodyNodes with zero-dof parent Joints


mkoval:
I expected `Branch::create(node)` to return a branch that includes the sub-tree rooted at `node`, including `node` itself. This is consistent with my interpretation of the Doxygen comment:

> `Branch` is a specialized type of `Linkage` that represents a complete subtree of a `Skeleton`.
> 
> The `Branch` will start at a specific `BodyNode` and will include every `BodyNode` that descends from it, all the way to the leaves.

However, this is not currently the case - the `Branch` does not include `node`. I am not sure if I am misinterpreting the documentation (as in #437 :smile:) or if this is a bug.

mxgrey:
It should definitely include the starting node. I'll look into why that might fail.

mkoval:
Actually, the issue is more subtle than I first thought. `Branch` seems to skip all `BodyNode`s that have a parent `FixedJoint`.

I have a kinematic structure that looks like this:

```
wam6 -[ revolute ]-> wam7 -[ fixed ]-> hand_base -[ revolute]-> ...fingers...
```

The contents of `Branch::create(wam7)` is:
```
- /right/wam7
- /right/finger0_0
- /right/finger0_1
- /right/finger0_2
- /right/finger1_0
- /right/finger1_1
- /right/finger1_2
- /right/finger2_1
- /right/finger2_2
```

The contents of `Branch::create(hand_base)` is:
```
- /right/finger0_0
- /right/finger0_1
- /right/finger0_2
- /right/finger1_0
- /right/finger1_1
- /right/finger1_2
- /right/finger2_1
- /right/finger2_2
```

Note that `hand_base` is missing in *both* outputs.

mxgrey:
That's a less surprising bug. I thought this situation was being handled, but I'm not shocked that it isn't. I'll fix it and write a regression test.

mkoval:
In fact, I think this issue applies to all `ReferentialSkeleton`s. I'm now seeing similar behavior in `Chain` because it stops at the first `FixedJoint` it encounters. For example, `Chain::create(Chain::Criteria(wam1, finger2_2))` stops at `wam7` instead of including `hand_base` and the finger `BodyNode`s.

mxgrey:
[...]

Issue #548, opened 2015-11-03
Group/Metaskeleton semantics are not expressive enough.


mklingen:
Right now, a `Group` or `MetaSkeleton` is fundamentally a collection of DOFs. The `BodyNode`s in a group come along for free by virtue of being attached to the DOFs in the group. But what if I want other body nodes rigidly attached to the group by upstream or downstream fixed DOFs?

For example, suppose I want to get all the bodies in the "arm" of my robot. This should include the base of the arm, all of the joints of the arm, and then any attached bodies on the end effector. If I create a group with just the arm dofs, this excludes the base and the attached bodies.

Here's a diagram with the group selected by "*":

```
arm_base --> *joint_1* --> *link_1* --> ... --> *joint_N* --> *link_N* --> (downstream links/joints)
```

I would like to be able to add additional bodies to the group without adding their associated DOFs.

mkoval:
I've also run into this issue. The underlying problem is, like @mklingen says, is that a `MetaSkeleton` is really only a collection of `DegreeOfFreedom`s. As a result, it can be used to either represent: (1) a collection of `DegreeOfFreedom`s or (2) a collection of `BodyNode`s. It cannot simultaneously represent (1) and (2) at the same time because membership between `BodyNode`s and `DegreeOfFreedom`s are coupled.

In the past, I've worked around this by creating two separate `MetaSkeleton`s: one that I use as a collection of `BodyNode`s and one that I use as a collection of `DegreeOfFreedom`s. This gets the job done, but I agree with @mklingen that it's a bit awkward. It's also error-prone because the former collection contains an arbitrary set of `DegreeOfFreedom`s.

mxgrey:
I'm certainly open to changes in all of the MetaSkeleton related classes. If someone can put together a draft of the kind of API they would like, I could work off of that.

Part of the reasoning behind the current implementation is that it allows you to get dynamics information for subsets of Skeletons. It doesn't make much sense (to me, at least) to be able to decouple degrees of freedom from their bodies (or vice-versa) when talking about dynamics information. Another part of the reasoning is that it makes the dynamics computations fairly performant (the only overhead is copying data; it doesn't need to do any kind of sorting/collating).

[...]

PR #557, opened 2015-11-16
Improvements to ReferentialSkeleton

mxgrey:
This pull request is in response to issues #548 and #556. This fixes the issue where BodyNodes with zero-dof parent Joints will be left out of ReferentialSkeletons, and it allows users to add and remove BodyNodes and DegreesOfFreedom from Groups freely (i.e. without any coupling between BodyNodes and their parent DegreesOfFreedom).

However, in the current implementation there is still a 1-to-1 correspondence between BodyNodes and Joints, even though BodyNodes are decoupled from DegreesOfFreedom. What this means is that ``ReferentialSkeleton::getJoint(i)`` is always equivalent to ``ReferentialSkeleton::getBodyNode(i)->getParentJoint()``. This is despite the fact that there is no coupling between ``ReferentialSkeleton::getBodyNode(~)`` and ``ReferentialSkeleton::getDof(~)``.

So I before this pull request can be considered finished, I'd like to pose the question: Should ``getBodyNode(~)`` be decoupled from ``getJoint(~)`` for ``ReferentialSkeleton``s, or is the current setup okay as it is?

I should also note that this pull request introduces a few changes to the behavior of the API of the ``Group`` class as a consequence of the decoupling. I hope that this does not result in any headaches for the existing users of ``Group``.

### DIFF
--- a/dart/dynamics/Group.cpp
+++ b/dart/dynamics/Group.cpp
@@ -209,7 +209,7 @@ bool Group::removeBodyNode(BodyNode* _bn, bool _warning)
     return false;
   }
 
-  unregisterBodyNode(_bn);
+  unregisterBodyNode(_bn, false);
 
   return true;
 }
--- a/dart/dynamics/Group.cpp
+++ b/dart/dynamics/Group.cpp
@@ -46,9 +46,9 @@ namespace dynamics {
 //==============================================================================
 GroupPtr Group::create(const std::string& _name,
                        const std::vector<BodyNode*>& _bodyNodes,
-                       bool _includeDofs)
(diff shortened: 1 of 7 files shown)


mxgrey:
The latest commit decouples the ``Joint``s in a ``Group`` from the ``BodyNode``s in the ``Group`` as per the discussion in #548.

However, I should write some unit tests to test out these new features.

mxgrey:
I've created unit tests for the new ``Group`` features. I believe this pull request is finished, unless someone spots an issue somewhere.

jslee02:
:+1:  Looks good to me.

Issue #555, opened 2015-11-11
Several functions are declared, but not defined


mkoval:
These functions are declared, but not defined:

- most functions in `render::Camera`
- `dart::constraint::BalanceConstraint::getPseudoInverseDamping()`
- `dart::renderer::Light::GetPosition()`
- `dart::renderer::Light::SetPosition()`
- `dart::renderer::Light::GetSpecular()`
- `dart::utils::convertFloatToDec`
- `dart::utils::minElem(std::vector<double, std::allocator<double> >&, int&)`
- `dart::utils::maxElem(std::vector<double, std::allocator<double> >&, int&)`
- `dart::utils::convertDecToFloat(char*)`
- `dart::constraint::ConstraintSolver::getNumConstraints() const`

mxgrey:
For ``convertFloatToDec``, there *is* a definition given for ``ConvertFloatToDec`` (notice the uppercase "C"). It seems pretty obvious that they belong together. Neither of them appears to be getting used currently. Do we want to simply go with the lowercase version? I believe that would match the existing code style better.

mxgrey:
For ``minElem`` and ``maxElem``, these are currently unused, and there are already ``std`` functions for retrieving the min and max elements of an STL container. I would recommend deleting these. Would anyone object?

mxgrey:
For ``ConstraintSolver::getNumConstraints``, I'm not sure what we would really want this to return. Is there a specific use case for retrieving the number of constraints in a ConstraintSolver? I could imagine it returning either ``mManualConstraints.size()`` or ``mActiveConstraints.size()``, depending on what the purpose of the function is (i.e. whether it's there to tell you how many constraints the user has added, or whether it's telling you how many constraints are currently being imposed by the simulation).

jslee02:
For `convertFloatToDec`, we should go with the lowercase version according to our code style. 

I don't object to removing `minElem` and `maxElem` in C3D.

I intended `ConstraintSolver::getNumConstraints` to return the number of constraints added by the user such as `BallJointConstraint` and `WeldJointConstraint` so that the use can access the constraints recursively within a for-loop with a getter that takes index of the constraint. However, I didn't even add the getter and couldn't find any usecase of it. So we can remove it.

mxgrey:
Regarding the stuff in the ``renderer`` namespace, should we make them all pure virtual, or give them blank definitions? Or maybe give them definitions with a "Not available in your derived class" warnings?

mxgrey:
[...]

PR #558, opened 2015-11-16
Fix missing definitions for various functions

mxgrey:
This pull request fixes the missing definition issue brought up in #555. Most of the fixes are just removing the declarations of the functions whose definitions were missing.


jslee02:
It seems [`OpenGLCamera` also needs to be removed](https://travis-ci.org/dartsim/dart/jobs/91473509#L6010).

jslee02:
:+1:

PR #561, opened 2015-11-28
VSK file parser

jslee02:
This pull request reinstates VSK parser that was obsoleted since DART 3.0. 

The VSK file format, which is developed by Vicon Motion Systems, contains skeleton hierarchy and markers. The specifications of VSK structure can be found [here](https://github.com/jslee02/vsk/blob/master/docs/VSK%20structure.pdf). Note that the VSK file format lacks some information to build dynamic skeleton models such as body node's shape, mass, and inertia. The missing information can be provided through `Options` parameter to the parser.

Also, `Parser.[h/cpp]` files are renamed to more relevant name `XmlHelpers.[h/cpp]` with some improvements on parsing xml attributes.


sehoonha:
Tested on my data, and looks good to me.

jslee02:
I believe this pull request is ready to merge. @mxgrey any comments?

mxgrey:
I'll finish reviewing it tonight.

One thought that comes to mind is that the Marker class should probably be refactored into a Node once PR #531 is finished.

psigen:
@mxgrey: Agreed.  It seems like much of what the `Marker` class was intended for is covered by the proposed functionality of the `Node` w.r.t. render geometry, right?

jslee02:
Agreed to the idea.

mxgrey:
Since ``VskParser`` doesn't have any internal state and its only member function is a static function, we may want to consider making ``VskParser`` a namespace instead of a class. Otherwise having it as a class sort of implies that there may be a situation where you'd want to create an instance of it.

mxgrey:
I've finished looking through the pull request, and it generally looks good to me. I just think we should either change ``VskParser`` to a namespace or explicitly delete its constructor to make it clear that there's no use in creating an instance of the class.

jslee02:
Updated this PR responding the comments. `VskParser` is changed to a namespace because it looks neater than explicitly deleting its constructor.

Will merge this PR once CI's are happy.

Issue #297, opened 2015-01-06
Servo actuator type


jslee02:
A servo actuator (motor) is a actuator that allows for precise control of position, velocity and acceleration. The actual (hardware) servo motor consists of a suitable motor coupled to a sensor for position feedback, and does feedback control. Meanwhile, in the simulation, the joint will be set to dynamic, and the desired position, velocity and acceleration will be achieved by corresponding impulses, which will be computed by the constraint solver.

jslee02:
This is resolved by #566

Issue #317, opened 2015-01-26
Cannot handle joint limit and joint friction together


jslee02:
DART cannot handle joint constraints correctly when there exists joint friction and joint position limit at the same time. This is because the two constraints overwrite the constraint impulse rather than adding up.

scpeters:
Thanks for fixing this and back-porting to 4.3. I've tested with gazebo and seen an improvement, though I had to relax the tolerances slightly (1e-4 to 1e-3).

* https://bitbucket.org/osrf/gazebo/commits/60614007ffa909f01399c3478707a85b7069bb61

Do you have any ideas why to tolerances were slightly violated? Constraint error correction?

jslee02:
The main reason is that DART doesn't correct positional violation directly. Instead, DART uses artificial velocity with ERP (error reduction parameter) so that the position error is corrected by integrating the position with the velocity. If the ERP is 1.0 then position will be completely corrected, and if ERP is 0 then position will not be corrected. The default ERP is 0.01. If we increase this ERP then it might reduce the tolerance.

I picked the low value for the default ERP, which may not correct the error aggressively, because too much ERP would make the system instable.

To handle this more reliably, DART may need to solve LCP twice per simulation for velocity correction and position correction per every simulation step like ODE does unless we change the whole constraint handling framework to other than impulse-based method.

PR #566, opened 2015-12-06
Preliminary implementation for servo motor

jslee02:
This pull request is a fundamental work for servo motor joint.

The current implementation of the servo motor joint works fine for general cases, but there are some corner cases it can't handle properly. For example, a servo joint is allowed to use infinite force to achieve the desired velocity, and, at the same time, the joint is reached the position limit where the desired velocity is toward the position limit. Then the two constraint forces will race against each other to achieve their objectives and will be ended up infinite constraint forces that will drive the constraint solver to undefined behavior. This confliction can happen with Coulomb friction as well.

There are two possible solutions to handle these corner cases I can come up with now:

1. We just don't allow infinite joint forces for Coulomb friction or servo motor.

1. We treat a joint with infinite Coulomb friction or infinite servo motor force as a kinematic joint. For instance, a joint with infinite Coulomb friction theoretically shouldn't be moved at all, or a joint with infinite servo motor force should perfectly follow its desired velocity. Then, we resolve the conflictions by setting their priorities. For example, position limit should be satisfied always, and then Coulomb friction constraint should be satisfied over servo motor constraint.

Also, some user might be interested in getting how much joint force was applied to the joint from the servo motor, but there is no place to store it in the current API. This will be addressed in Actuator class later.


jslee02:
This pull request fixes #297 and #317.

karenliu:
Looking good to me.

PR #573, opened 2015-12-19
add more hint to Pendulum.cpp, revise the typo error in collision.md

JiaYeLi:
There is some code in lesson 2a of Pendulum:
for(size_t i = 0; i < mPendulum->getNumDofs(); ++i)
{
  DegreeOfFreedom* dof = mPendulum->getDof(i);
  double q0 = dof->getRestPosition() + delta;
  dof->setRestPosition(q0);
}
the variable delta may be the parameter of the functions, but the tutorial don't say where the variable 'delta' comes from.
http://dart.readthedocs.org/en/release-5.1/tutorials/multi-pendulum/#lesson-2a-set-joint-spring-rest-position

User may need read the finished file and get hints.

There is a typo error in collision.md

### DIFF
--- a/docs/readthedocs/tutorials/collisions.md
+++ b/docs/readthedocs/tutorials/collisions.md
@@ -575,7 +575,7 @@ And now we can add it to the world without any complaints:
 mWorld->addSkeleton(object);
 ```
 
-### Lesson 3d: Compute collisions
+### Lesson 3c: Compute collisions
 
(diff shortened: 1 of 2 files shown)


jslee02:
:+1: Good catch. If @mxgrey  approves, I think it's good to merge.

JiaYeLi:
wow! so efficient!
Thank you.

I'm glad it is merged.

mxgrey:
This looks good to me :+1:

Issue #576, opened 2015-12-29
Parallel Simulations


costashatz:
Hello,

Merry Christmas to everyone.

I am running in parallel multiple simulations (around 24) and I am getting seg-faults. This is the backtrace:

```
#0  0x00007ffff667c5ab in std::_Rb_tree_rebalance_for_erase(std::_Rb_tree_node_base*, std::_Rb_tree_node_base&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007ffff76b5cce in dart::dynamics::Frame::changeParentFrame(dart::dynamics::Frame*) () from /usr/lib/libdart-core.so.5.1
#2  0x00007ffff76b638d in dart::dynamics::Frame::~Frame() () from /usr/lib/libdart-core.so.5.1
#3  0x00007ffff7769ac9 in dart::dynamics::BodyNode::~BodyNode() () from /usr/lib/libdart-core.so.5.1
#4  0x00007ffff7769ba1 in dart::dynamics::BodyNode::~BodyNode() () from /usr/lib/libdart-core.so.5.1
#5  0x00007ffff77ae8ee in dart::dynamics::Skeleton::~Skeleton() () from /usr/lib/libdart-core.so.5.1
#6  0x00007ffff77ba17d in std::_Sp_counted_ptr<dart::dynamics::Skeleton*, (__gnu_cxx::_Lock_policy)2>::_M_dispose() () from /usr/lib/libdart-core.so.5.1
#7  0x000000000040685b in _M_release (this=0x7fffe00016e0) at /usr/include/c++/4.9/bits/shared_ptr_base.h:149
#8  ~__shared_count (this=<optimized out>, __in_chrg=<optimized out>) at /usr/include/c++/4.9/bits/shared_ptr_base.h:666
#9  ~__shared_ptr (this=<optimized out>, __in_chrg=<optimized out>) at /usr/include/c++/4.9/bits/shared_ptr_base.h:914
#10 ~shared_ptr (this=<optimized out>, __in_chrg=<optimized out>) at /usr/include/c++/4.9/bits/shared_ptr.h:93
#11 ~Hexapod (this=<optimized out>, __in_chrg=<optimized out>) at /home/kchatzil/Workspaces/ResiBots/include/hexapod_dart/hexapod.hpp:8
#12 destroy<hexapod_dart::Hexapod> (this=<optimized out>, __p=<optimized out>) at /usr/include/c++/4.9/ext/new_allocator.h:124
[...]

PR #577, opened 2015-12-29
Fix World Concurrency

mxgrey:
This pull request is meant to fix the segmentation fault reported in issue #576 . I managed to create a unittest which reliably triggered a race condition which I hope corresponds to the race condition reported in #576.


costashatz:
This looks good!

jslee02:
Yeah, it was very prompt fix. Nice job @mxgrey ! Merging now.

PR #580, opened 2016-01-08
Fix Visual Studio 2015 build errors

donnyward:
This pull request fixes all the build errors that I encountered while trying to build DART with Visual Studio 2015.

CMake would fail to configure if you weren't using exactly Visual Studio 2013, so I added a conditional for 2015 as well.


jslee02:
:+1:

Issue #479, opened 2015-07-29
DART needs a new name


karenliu:
Dear DART community,

It's been exciting to see how much DART grow in the past two years thanks to our devoted users and contributors. As DART becomes increasingly popular, we started to receive some requests of name change from other software owners who use similar names. The most similar one is the DARTS rigid body simulator from JPL. To avoid potential confusion down the road, we have made a painful but necessary decision to change the name effective on the next major version release. @jslee02, @mxgrey, and myself have come up with a short list of candidates which can be found [here](https://docs.google.com/forms/d/1FVar29R2TPr9RXi6Tc1Lceu3FVaK8KESZRdEEkAUJng/viewform). We would love to hear your thoughts before we make the final decision. Thank you for your support and interest in DART.

Karen Liu

mxgrey:
[Here's a larger version of the link to the poll, just to make it more clear.](https://docs.google.com/forms/d/1FVar29R2TPr9RXi6Tc1Lceu3FVaK8KESZRdEEkAUJng/viewform)

mklingen:
TIL `DART` is not `DARTS` from JPL!

mkoval:
Was there any conclusion from the poll? How do you plan to roll out the name change?

I recently got burned by this because `DARTConfig.cmake` conflicts with the `DartConfig.cmake` file that ships with recent versions of CMake if you are using a case-insensitive filesystem (the default on OS X).

:sweat:

karenliu:
Since there is no clear winner from the first poll, we came up with a few other names and will do a second poll on November 2. If you have any suggestions, please email @jslee02, @mxgrey, or @ckarenliu.

mklingen:
I have a pretty easy one: `DRAT: Dynamic Robotics and Animation Toolkit`.

scpeters:
I still like rtql8, but I'll wait for the new poll.

mklingen:
`rtql8` ends in a number (which is confusing), is longer than `dart` (by one letter or 3 syllables), is harder to type, and harder to figure out what it says through phonetics. (Rahtwikel-8? Are-tee-kew-el-8?) but at least it has the advantage of being unique!

scpeters:
@mklingen those are excellent points. As someone involved in packaging, I especially appreciate the issues with having a number in the package name. Also, the lower-case `l` could look like the number one, especially when next to the `8`.

I think my main issue with the other options in the survey is that it would be hard to get good search result placement, since they are such common words.

[...]

PR #582, opened 2016-01-12
Renaming to KIDO

jslee02:
As discussed in #479, we are going to release KIDO 0.1.0 first for the initial submitting to Debian by around Feb 18. This pull request is the first step of it renaming DART with KIDO.

Also, `kido-0.1.0` will be used as the release branch for KIDO 0.1.0.


CMakeLists.txt
```diff
@@ -37,39 +37,39 @@ endif()
...
-set(DART_MINOR_VERSION "1")
-set(DART_PATCH_VERSION "1")
-set(DART_VERSION "${DART_MAJOR_VERSION}.${DART_MINOR_VERSION}.${DART_PATCH_VERSION}")
-set(DART_PKG_DESC "Dynamic Animation and Robotics Toolkit.")
-set(DART_PKG_EXTERNAL_DEPS "flann, ccd, fcl")
+set(KIDO_MAJOR_VERSION "0")
+set(KIDO_MINOR_VERSION "1")
+set(KIDO_PATCH_VERSION "0")
+set(KIDO_VERSION "${KIDO_MAJOR_VERSION}.${KIDO_MINOR_VERSION}.${KIDO_PATCH_VERSION}")
+set(KIDO_PKG_DESC "Dynamic Animation and Robotics Toolkit.")
```
traversaro: This description needs also to be changed?

jslee02: Sure, it should be changed to the full name of KIDO, which is "Kinematics, Dynamics and Optimization Library for Animation and Robotics" tentatively. This can be varied slightly. Thanks for the nice catch!

mxgrey:
I think we should also change header names to ``.hpp`` before it goes to the Ubuntu repo.

jslee02:
Agreed. Technically, we can change the header names for KIDO 1.0.0 that will be submitted around Feb 18. But, no reason to wait until then. I will do that in this pull request.

jslee02:
Changed header names to `.hpp` except for ode headers since they are not only for cpp.

mxgrey:
We'll probably have to go through this process at least one or two more times, since there are still features that need to be merged into DART 6.0 / KIDO 1.0. I'm wondering if it would be smart to have a script that can do these replacements automatically.

jslee02:
I used simple script for that, but it needed some manual postprocessing. Here is the script:
```sh
shopt -s globstar
for file in ~/dev/kido/master/**/*.h; do
  mv "$file" "${file%.h}.hpp"
done

```

mxgrey:
I'm thinking we should have it for DART -> KIDO as well if possible. But admittedly, that would be a lot trickier.

jslee02:
DART -> KIDO was done by using IDE(Qt Creator)'s replace functionality, and it was an easier process than the file name changing.

I wish someone provides a magic script that handles all the process I've done 100% automatically. :cry:

mxgrey:
I suspect it could be done without too much difficulty using Python. The next time it's needed, I could give that a try.

jslee02:
That would be good. Thanks!

jslee02:
I believe this pull request is ready to be merged once the version of first KIDO release is decided (see #479).

Issue #583, opened 2016-01-12
Possible memory leaks in constraint solver?


costashatz:
Hello,

After running huge simulations in our cluster, we observed some memory leaks that filled up our RAM and swap (64gb + 64gb!). I was able to create a small program that `valgrind` finds possible memory leaks:

```
==7983== Memcheck, a memory error detector
==7983== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==7983== Using Valgrind-3.10.1 and LibVEX; rerun with -h for copyright info
==7983== Command: ./hexapod
==7983== 
==7983== 
==7983== HEAP SUMMARY:
==7983==     in use at exit: 4,851 bytes in 42 blocks
==7983==   total heap usage: 157,188 allocs, 157,146 frees, 178,053,923 bytes allocated
==7983== 
==7983== 4,720 (1,760 direct, 2,960 indirect) bytes in 10 blocks are definitely lost in loss record 6 of 6
==7983==    at 0x4C2BBA0: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==7983==    by 0x54206E1: dart::constraint::ConstraintSolver::updateConstraints() (in /usr/local/lib/libdart-core.so.5.1.1)
==7983==    by 0x5420D5F: dart::constraint::ConstraintSolver::solve() (in /usr/local/lib/libdart-core.so.5.1.1)
==7983==    by 0x5426522: dart::simulation::World::step(bool) (in /usr/local/lib/libdart-core.so.5.1.1)
==7983==    by 0x43AB1A: main (in /home/kchatzil/ownCloud/documents/backup-2015-12-10/Workspaces/DART/source/dart_test/build/hexapod)
==7983== 
==7983== LEAK SUMMARY:
==7983==    definitely lost: 1,760 bytes in 10 blocks
==7983==    indirectly lost: 2,960 bytes in 30 blocks
==7983==      possibly lost: 0 bytes in 0 blocks
==7983==    still reachable: 131 bytes in 2 blocks
==7983==         suppressed: 0 bytes in 0 blocks
==7983== Reachable blocks (those to which a pointer was found) are not shown.
==7983== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==7983== 
==7983== For counts of detected and suppressed errors, rerun with: -v
==7983== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

[...]

PR #584, opened 2016-01-12
Updating the constraint namespace to C++11

mxgrey:
This change is motivated by issue #583 where it was found that the ``ConstraintSolver`` class is leaking memory. Frequent deletion and creation of ``World`` objects can dramatically exacerbate the issue. By updating to use C++11 smart pointers, we should be able to eliminate these leaks.

### DIFF
--- a/dart/constraint/ConstrainedGroup.cpp
+++ b/dart/constraint/ConstrainedGroup.cpp
@@ -57,7 +57,7 @@ ConstrainedGroup::~ConstrainedGroup()
 }
 
 //==============================================================================
-void ConstrainedGroup::addConstraint(ConstraintBase* _constraint)
+void ConstrainedGroup::addConstraint(const ConstraintBasePtr& _constraint)
 {
   assert(_constraint != nullptr && "Null constraint pointer is now allowed.");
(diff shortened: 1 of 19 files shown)


costashatz:
This seems to solve #583. At least in my case. Merge when ci is good.

dart/constraint/ConstrainedGroup.cpp
```diff
@@ -108,16 +108,20 @@ void ConstrainedGroup::removeAllConstraints()
...
 
 //==============================================================================
-bool ConstrainedGroup::containConstraint(ConstraintBase* _constraint) const
+bool ConstrainedGroup::containConstraint(
+    const ConstConstraintBasePtr& _constraint) const
 {
 //  std::cout << "CommunityTEST::_containConstraint(): Not implemented."
 //            << std::endl;
 
+  // TODO(MXG): Is there any reason these functions are not implemented yet?
```
jslee02: No reason. It seems I just forgot to implement. I'll implement them in a separate pull request soon.

jslee02:
:+1:

PR #585, opened 2016-01-12
Fixed segmentation fault for null meshes

mxgrey:
I found that there is a risk of segmentation fault for meshes that are not immediately initialized. It occurs when the Mesh::_updateBoundingBoxDim function is called. This pull request fixes this issue, and adds some more initialization for ``ArrowShape``.


jslee02:
:+1:

PR #587, opened 2016-01-20
Package reorganizing

jslee02:
This pull request reorganizes packages of KIDO into new composition as discussed in #477.

In the new composition, previous `dart-core` becomes `kido`, and the additional namespaces in `dart` are splited into several components: `kido-planning`, `kido-gui`, `kido-utils`, `kido-optimizer-nlopt`, and `kido-optimizer-ipopt`.

`osgKido` is a separate package rather than a component of `kido`. So `osgKido` can be used by `find_package(osgKido)`. But, it can be merged into `kido` as a component easily if @mxgrey wants.


jslee02:
Waiting to pass build tests on [ppa:dartsim](https://launchpad.net/~dartsim/+archive/ubuntu/ppa/+packages).

mkoval:
It would be nice to add `osgKido` as a component. This would also be a great opportunity to rename `osgKido` to `dart-gui-osg`, to be consistent with the rest of the namespace structure.  @mxgrey did you have a strong reason for putting the code where it is?

mxgrey:
There were a couple reasons for separating osgDart the way we did.

1. The DART libraries were only split into ``dart-core`` and the rest was ``dart``. We didn't want ``dart`` to have an OSG dependency, so we separated it out completely.

2. The naming convention that the OSG maintainers request for extensions to OSG is ``osgLibraryName``.

Since we're totally redoing the package organization, I think it would be a good idea to package ``osgDart`` the same way we're packaging all the other ``dart`` components. And now that we've established a naming convention for ``dart`` extensions, I suppose our own naming conventions should take precedence over the requested convention of the OSG maintainers. And then these changes should obviously carry over to ``kido``.

mxgrey:
Although if we're not concerned about implementing those changes in DART 6.0, then we could just make the changes in ``kido`` and not worry about changing it in ``dart``.

jslee02:
Right, `osgDart` was named so for those reasons. I agree to rename it to `kido-gui-osg` following our new convention. Also, it would be good to change the directory name and location of osgKido to `kido/gui/osg`. How do you think? 

As we are releasing DART 6.0 for seamless changing the name, I would like to make the changes in KIDO 1.0 (and 0.1) but not in DART 6.0.

jslee02:
I created #588 for the change of `osgKido` namespace and directory.

PR #591, opened 2016-01-21
Merge master branch into #531

jslee02:
This pull request merges master branch into `grey/addons` for #531.


mxgrey:
It probably wouldn't make much sense for me to review this pull request in detail since it's coming from master. But were there any significant merge conflicts that I should look at?

jslee02:
There was one conflict on `dart/dynamics/MultiDofJoint.h` and I resolved it simply. You would like to check it for sure.

jslee02:
It seems AppVeyor is failing to build since [`constexpr` is not supported in Visual Studio 2013 (2015 supports it)](https://msdn.microsoft.com/en-us/library/hh567368.aspx). I'm thinking to use a macro to resolve this something like:
```cpp
// constexpr is not supported until Visual Studio 2015 (_MSC_VER == 1900)
#if defined _MSC_VER && _MSC_VER < 1900
  #define DART_CONSTEXPR const
#else
  #define DART_CONSTEXPR constexpr
#endif
```

mxgrey:
There might be some things in the ``Addon``/``Node`` implementations that rely on ``constexpr`` variables for template arguments (I don't know for sure off the top of my head; we'd have to test it), so replacing it with ``const`` might not necessarily work.

I believe VS2015 is just as readily available on Windows 7 as VS2013, so would it not be fair to make VS2015 the minimum requirement? Or does Travis-CI not support VS2015 yet?

jslee02:
You're right. The macros couldn't solve it.

As VS2015 community is free for everyone, it seems to be fair to make VS2015 the minimum requirement. 

AppVeyor supports VS2015, but the problem is that we don't have the DART dependencies built with VS2015 yet. I'm working on this, but it would take more days. So I'd like to increase the minimum requrired VS version to 2015 and don't care the build test until I create the dependencies then.

jslee02:
I made Visual Studio 2015 the minimum requirement in the last two commit. @mxgrey Does this look good to merge?

mxgrey:
Your resolution of the merge conflict looks good to me. If that was the only major one, I think we should be good to merge.

jslee02:
Yes, it was the only conflict. Merging now. Thanks!

PR #598, opened 2016-01-30
Revision of #531

jslee02:
This pull request
* merges master branch into #531 (should pass appveyor tests)
* revises as discussed in #531

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -475,7 +475,7 @@ endif()
 #===============================================================================
 if(MSVC)
   # Visual Studio enables c++11 support by default
-  if(NOT MSVC14)
+  if(MSVC_VERSION VERSION_LESS 1900)
     message(FATAL_ERROR "${PROJECT_NAME} requires VS 2015 or greater.")
   endif()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP4")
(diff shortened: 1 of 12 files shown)


jslee02:
A bug of VS2015 was reported pertain to `constexpr` with functions pointer ([see bug 1378031](http://blogs.msdn.com/b/vcblog/archive/2015/12/02/constexpr-in-vs2015-update-1.aspx)). The document says it's fixed, but [it seems not at least for our case](https://ci.appveyor.com/project/jslee02/dart/build/2397/job/6movxpwjt0cn3e30). Here is the error message:
```
[00:00:50] C:\projects\dart\dart/dynamics/Addon.h(63): error C2131: expression did not evaluate to a constant (compiling source file C:\projects\dart\dart\dynamics\BallJoint.cpp) [C:\projects\dart\build\dart\dart-core.vcxproj]
```

mxgrey:
Is it possible that Appveyor doesn't have the latest updates for VS2015? Or have we found an extra-special case of the ``constexpr`` bug?

jslee02:
The compilation errors are occurring with my VS2015, which is the most recent version with Update 1 (14.0.2478.2), as well. AppVeyor's one is 14.0.24720.0.

I can't say it's a VS2015 bug definitely, but it seems so for now.

jslee02:
Merging even VS2015 can't build this.

PR #603, opened 2016-02-02
Added namespace-scope definitions for constexpr static members

mkoval:
This is necessary to ODR-use those member variables. See [this StackOverflow post](http://stackoverflow.com/a/14396189/111426) for an explanation.

I honestly think this is quite bizarre - I never knew this was necessary for template classes. However, it did resolve my linking problems. :smile:


jslee02:
> I honestly think this is quite bizarre - I never knew this was necessary for template classes.

Yeah, me neither. In that sense, it would be helpful to add simple note why we need this. Beside on note, :+1:

mxgrey:
C++ just really loves throwing those curveballs. :+1: to the changes and :+1: to adding a developer's comment to explain the added declaration/definition (declinition?).

mkoval:
I added the requested comments.

Issue #599, opened 2016-01-31
Build error with Visual Studio 2015 Update 1 and 2


jslee02:
Visual Studio 2015 with Update 1 complains in build of the current master branch. It seems VS2015 fails to evaluate a function pointer to a constant.

Here is the error message.

```
C:\projects\dart\dart/dynamics/Addon.h(63): error C2131: expression did not evaluate to a constant (compiling source file C:\projects\dart\dart\dynamics\BallJoint.cpp) [C:\projects\dart\build\dart\dart-core.vcxproj]
```

jturner65:
yes, i ran into this too.  it seems like the most likely cause of this is the constexpr static declarations in addon.h.  apparently vs2015's support of constexpr is evolving.  notice the link below : 

https://blogs.msdn.microsoft.com/vcblog/2015/12/02/constexpr-in-vs2015-update-1/

particularly the looking forward section, where they say there are still numerous bugs remaining, some of which pertain to pointers-to-members in constant expressions.

jslee02:
So sad VS2015 hasn't completed `constexpr` yet. I hope the next update fix this, and it's before DART 6.0 release.

jturner65:
i know, it's unfortunate how long it's taking them to become compliant with the standard.

jturner65:
are there any plans to address this?

jslee02:
I don't think there is a way to address this unless VS2015 fixes the bug. One possible way could be not using `constexpr`, but it is now required feature for DART 6.0 since #531.

Please note that this is an issue for the current master branch (that will be DART 6.0) but not for DART 5.1 or less.

jturner65:
this isn't an issue with 5.1 so i guess we can close it, although 5.1 has other problems (Eigen alignment assertions).

jslee02:
The Eigen alignment issue should be addressed by #606.

I would like to leave this issue open to inform to other users. Someone who tries to build master branch with VS2015 might want to know about this.

jturner65:
Excellent! Thanks JS.

jslee02:
I hope the `constexpr` issue is addressed in [VS2015 Update 2 CTP](https://blogs.msdn.microsoft.com/vcblog/2016/02/11/compiler-improvements-in-vs-2015-update-2/).

jturner65:
oh, goody.  i'll test it.

jslee02:
The issue seems not resolved yet. I still get the same errors.

jslee02:
[...]

PR #606, opened 2016-02-10
Fix memory alignment in loading meshes

jslee02:
This pull request addresses the memory alignment issue with Eigen object in loading meshes, which was reported by @jturner65 in #599.


mxgrey:
The changes look good :+1:, but is there a particular reason that VS can't compile it in release mode? Is it just timing out?

jslee02:
Hm, I'm not sure. Let me make AppVeyor test it again, and test it myself on my virtual machine in parallel to figure it out.

jslee02:
I got an error as:
```
c:\golems\include\eigen\src\core\assign.h(527): fatal error C1002: compiler is out of heap space in pass 2
```

It seems it needs too much heap memory for 32 bit Windows to build DART. One possible solution would be optimizing the header inclusion as the suggestion of the first answer in [this post](http://stackoverflow.com/questions/1388608/how-to-work-around-visual-studio-compiler-crashes).

mxgrey:
It's very likely that our header organization could be improved dramatically, although I'm afraid it might require some refactoring of classes, such as decomposing them into simpler classes and carefully sorting through the dependencies between them.

I don't know who would have time to actually do that, or when.

jslee02:
Yeah, we should give it a try to improve the header organization at some point, but not now. I'm merging this PR, and creating a new issue for this build failure that might be related to the header organization.

Issue #394, opened 2015-05-25
VisualNode & CollisionNode


mxgrey:
I think it would be appropriate to begin a conversation about the VisualNode/CollisionNode concept which is intended to replace the current method of using purely the Shape class to provide visualization and collision information. I don't anticipate this being finished any time before version 5.1, but I think it's worth opening the conversation now to give us some time to share ideas and flesh out a design.

To review, the current way Shapes are used is:
(i) The Entity class has a `std::vector` of `std::shared_ptr<Shape>` which it uses for visualization
(ii) The BodyNode class has a `std::vector` of `std::shared_ptr<Shape>` which it uses to check collisions

The nice thing about this scheme is that it allows us to share shape information between multiple Entities/BodyNodes which saves on memory. The downside is that Shape has several features, such as Shape::setLocalTransform and Shape::setColor, which could result in destructive interference if two Entities are sharing the Shape but want that shape to have a different local transform or a different color.

Another issue with the current scheme is it does not allow a single BodyNode to have different parts with different physical properties, like coefficients of friction and restitution. A real physical body will have different friction properties on different parts, and different restitution properties on different parts, but DART cannot currently model that. Making friction and restitution properties part of a CollisionNode and allowing a BodyNode to have multiple CollisionNodes would address this issue.
### Design suggestions

I think there is an interest in having the VisualNode and CollisionNode classes inherit the Frame class. This makes sense, because we want these Node classes to have a local transform property, and it would be convenient to have a reference frame tied to the Shapes. However, there is a bit of a conflict here: if we make a VisualNode responsible for visualization and we make VisualNode a Frame, then this means that we might need to remove visualization features from Entities, because classes that are derived directly from Entity (rather than from Frame) are expected to leaf nodes in the kinematic tree, so it would not make sense for a raw Entity to have child VisualNodes (since this would contradict the idea that the Entity is a leaf).

So I have two suggestions to offer for how to resolve this:

(1) **We remove all visualization features from Entity.** This would make the raw Entity class far less valuable, but maybe that's okay. Instead of creating an Entity to render an arbitrary (and non-physical) Shape in a scene, the user would have to create a SimpleFrame (which would inherit VisualNode) and fill it with some Shapes.

[...]

Issue #453, opened 2015-07-15
MeshShape without Assimp


mkoval:
From #451 and #452 I've spent a lot of time looking at the internals of Assimp. I am starting to realize that the entire library, as it's name implies, is designed around **importing** meshes and `aiScene` is very much not intended to represent a mutable structure. As such, I feel think that we should move away from using `aiScene` as the in-memory representation of meshes in DART for mutable geometry and, possibly, all collision geometry.

@psigen, @PyryM, and I had a long discussion about this, so I"ll do my best to summarize. I know this is a large change, so this issue is mostly to start a discussion.

### Multiple Heaps
Much of the Assimp library is designed to guarantee that memory is always freed in the same shared library that allocated it. This is what the Assimp documentation has to say on the issue:

> By design, `aiScene`'s are exclusively maintained, allocated and deallocated by Assimp and no one else. The reasoning behind this is the golden rule that deallocations should always be done by the module that did the original allocation because heaps are not necessarily shared. `GetOrphanedScene()` enforces you to delete the returned scene by yourself, but this will only be fine if and only if you're using the same heap as Assimp. On Windows, it's typically fine provided everything is linked against the multithreaded-dll version of the runtime library. It will work as well for static linkage with Assimp.

I haven't hit this problem before, probably because I rarely use unmanaged memory, so I did a bit of digging on StackOverflow. In general, it's complicated.  This [may be a problem on Windows](http://stackoverflow.com/a/10820167/11142) when using certain versions of the C runtime, mixing libraries with [different debug/release settings](http://stackoverflow.com/a/443192/111426), and different versions of the standard library. This [blog post](http://blogs.msdn.com/b/oldnewthing/archive/2006/09/15/755966.aspx) is a good summary.

### Immutability
To get around this issue, all dynamically allocated memory in an `aiScene` is both allocated and freed by the `Importer` that loaded it. If you are using a built-in importer, like we do in DART, this means that it was allocated on Assimp's heap. *It is not possible to safely delete or re-allocate from within DART.* This is why Assimp always returns `aiScene` objects by `const` pointer.

### Programmatic Construction
[...]

PR #608, opened 2016-02-12
[WIP] ShapeFrame and ShapeNode

jslee02:
This is a work-in-progress pull request for implementing ShapeFrame/ShapeNode as discussed in #394.


dart/dynamics/BodyNode.h
```diff
@@ -613,6 +591,60 @@ class BodyNode :
...
+  /// ShapeNode::Properties argument into this function.
+  template <class ShapeNodeProperties>
+  ShapeNode* createShapeNode(const ShapeNodeProperties& properties);
+
+  /// Create an ShapeNode with the specified name
+  ShapeNode* createShapeNode(
+      const ShapePtr& shape, const std::string& name = "ShapeNode");
+
+  /// Create an ShapeNode with the specified name
+  ShapeNode* createShapeNode(const ShapePtr& shape, const char* name);
```
mkoval: I don't think it is necessary to have both a `const char*` overload and a `const std::string&` overload. Won't the `const char*` implicitly convert to an `std::string`?

mxgrey: I'm not sure about this particular case, but I've seen cases where having an overloaded function where one of the overloads takes a ``std::string``, the compiler won't always implicitly convert it when you pass in a ``const char*``. I've especially seen this for templated functions, which the next one down is. (I don't know about this specific case, since @jslee02 wrote it, but I'm guessing it's a similar situation here.)

jslee02: As @mxgrey said, the implicit converting sometimes didn't work. I'm still not sure exactly when is the case. I just tested this function and `BodyNode::createEndEffector(const char*)`, and the implicit converting worked for both of them. So we can remove these functions.

mxgrey: I think the way it works is that it will attempt to use a templated function instead of performing an implicit conversion, but then the templated function fails because it wasn't meant for the case of a ``const char*``.

jslee02: I see. The process of type deduction for templated function doesn't check if the passed-in parameter type can be implicitly converted to one of possible types. Good to know!

dart/dynamics/Shape.h
```diff
@@ -191,29 +142,19 @@ class Shape : public virtual common::Subject
...
-  /// \brief
-  virtual void draw(renderer::RenderInterface* _ri = nullptr,
-                    const Eigen::Vector4d& _color = Eigen::Vector4d::Ones(),
-                    bool _useDefaultColor = true) const = 0;
-
-  /// Pass in true to prevent this shape from being rendered. Pass in false to
-  /// allow it to render again.
-  void setHidden(bool _hide=true);
+  /// Notify that the color (rgba) of this shape has updated
+  virtual void notifyColorUpdate(const Eigen::Vector4d& color);
```
mxgrey: Do shapes have Color anymore? I thought Shapes were now purely geometric information, and that color information would be in the VisualAddon of the ShapeNode. Or is this just WIP code that's going to be removed later?

jslee02: I thought so too, but `MeshShape` has color properties (colors for the vertices) in it. This function gets called when the color of `VisualAddon` is changed so that `MeshShape` can handle it. By default, it does nothing for other shapes. This function is a compromised solution, so I'm open to any better way for this case.

mxgrey: There has been some discussion about the possibility of MeshShape being changed into something that only holds geometric data. If we do that, then we could have it so that the texture and color data gets automatically put into the VisualAddon as it's parsed with assimp. I think this would be also be nice for moving away from assimp being a core dependency (instead it would only be an ``io`` dependency).

That said, I don't know what the timeline should be for implementing something like that. We could probably hold off on it until version 7.1.

jslee02: Yeah, the change sounds good to me. If I remember correctly, there is ongoing work of `Mesh` class (for moving away from assimp) by @mkoval. Once it's done we can remove this function. Maybe a note for this here would be good.

jslee02: Okay, I didn't remember correctly. :sob: @mklingen was working on it at [here](https://github.com/mklingen/dart/tree/feature/mesh_refactor).

mklingen: I was but abandoned it some time ago, because the assimp dependency ran very deep

mxgrey: I do think it's a worthwhile cause, but it might require some commitment. It would probably be best to tackle alongside the new GLFW-based rendering library that has been discussed.

mklingen: That said, it is useful to have vertex color data in the visual meshes so that you can create dynamically generated meshes with colors for visualization. I know @mkoval at least has found the need for this in his research, and had to make do with publishing custom visualization messages to ROS.

mxgrey: I don't know if this would be silly or not, but we could potentially make VisualAddon be an AddonManager and be able to attach a MeshColor Addon which could contain per-vertex color information.

There might be a better way to approach it than what I've suggested, but I think there would be value in being able to decouple the geometric and visual data. For example, suppose you want to be able to visualize the same geometric data with different color overlays (say you have two copies of the same robot, but you want different per-vertex coloring on each to highlight different contextual data). If the coloring is tied into the geometric data, that's a significant amount of wasted memory, because you'll need a full copy.

mkoval: > There has been some discussion about the possibility of MeshShape being changed into something that only holds geometric data. If we do that, then we could have it so that the texture and color data gets automatically put into the VisualAddon as it's parsed with assimp. I think this would be also be nice for moving away from assimp being a core dependency (instead it would only be an io dependency).

:+1: for moving away from Assimp in the core API. In theory, I like the idea of texture and color data getting stored inside `VisualAddon`. This would work okay for vertex color, but I am concerned about the feasibility of doing this with textures. Assimp has a dozen different types of textures and material properties. My impression is that it's the wild west, especially for meshes loaded from COLLADA.

@mklingen has much more experience here. Hopefully he can provide some illumination about what is actually used.


> It would probably be best to tackle alongside the new GLFW-based rendering library that has been discussed.

What would be the relationship between `SimWindow`, `osgDart`, and the new GLFW-based rendering library? I also hope this allows us to remove the `draw` functions from the `dynamics` API - I really don't think rendering logic belongs there.

> I don't know if this would be silly or not, but we could potentially make `VisualAddon` be an `AddonManager` and be able to attach a `MeshColorAddon` which could contain per-vertex color information.

I am against this for a few reasons:

1. It adds a lot of complexity to an already-complex part of DART.
2. What is the advantage of putting per-vertex colors in separate `MeshColorAddOn` versus directly inside `VisualAddOn`?
3. `MeshColorAddon` only makes sense on `MeshShape`s, but may be added to any type of `ShapeNode`.

dart/dynamics/BodyNode.h
```diff
@@ -613,6 +591,60 @@ class BodyNode :
...
+
+  /// Create an ShapeNode with the specified name and addons
+  template <class... Addons>
+  ShapeNode* createShapeNode(const ShapePtr& shape,
+                             const std::string& name = "ShapeNode");
```
mkoval: I'm not sure if this overload is necessary. It only seems helpful in the case where you want to create a `ShapeNode` with several `AddOn`s, but you don't need to set any properties on them. Any other case would be better handled by adding a `getOrCreate` function to `AddonManager, e.g.:

```c++
ShapeNode* node = bodyNode->createShapeNode();
node->getOrCreate<VisualAddon>()->mColor = Orange();
```

dart/dynamics/Shape.h
mklingen: For rendering complex geometry its typically better to work with the "flyweight" pattern. That is, the mesh data (visual geometry, color, texture data, shaders, etc.) is shared among all instances of that mesh. This saves memory and allows you to render instances in a batch.

 A visual mesh node could be associated with this data in a few ways. One way is merely to associate it with an asset tag, and have it loaded by assimp. But that precludes being able to edit a visual mesh or procedurally generate it. So, here's how I would do it:

1. Visual Nodes are associated weakly with one or more shared "visual assets".
2. A visual asset could be either an assimp mesh, a procedurally generated internal mesh, or some other kind of asset. The only requirement is that it can be drawn by the viewer.

That way, to procedurally generate a mesh, you could create its visual asset, then associate it with the node. The viewer's responsibility would be looking up the visual assets of all the nodes, and then drawing them (with an overloaded `draw` function for each kind of mesh).

mkoval: I believe DART already has that structure. You may have more than one `ShapeNode`s that reference a single `Shape` instance by `std::shared_ptr`. We do not have a central repository of all `Shape` instances, but it seems relatively straightforward for the viewer to construct that itself.

The only exception is `VisualAddOn`, which contains color, is attached to `ShapeNode` instead of `Shape`. One motivation of this is to avoid copying an entire mesh just to change RGBA values. Are you saying that this makes rendering inefficient because you can't render those `Shape`s in a batch?

I'm also still not sure what the API for textures should look like. The [Assimp materials API](http://assimp.sourceforge.net/lib_html/materials.html) is quite complicated.

dart/dynamics/BodyNode.cpp
```diff
@@ -896,6 +819,60 @@ const Joint* BodyNode::getChildJoint(size_t _index) const
...
+std::vector<ShapeNode*> BodyNode::getShapeNodes()
+{
+  std::vector<ShapeNode*> shapeNodes(getNumShapeNodes());
+
+  auto numShapeNode = getNumShapeNodes();
+  for (auto i = 0u; i < numShapeNode; ++i)
+    shapeNodes[i] = getShapeNode(i);
+
+  return shapeNodes;
+}
```
mkoval: Can we return a `const std::vector<ShapeNode*>&` here, like we do in most of the other non-`const` functions that return `vector`s?

dart/dynamics/ShapeFrame.cpp
```diff
@@ -0,0 +1,394 @@
...
+void CollisionShapeFrameAddonUpdate(CollisionAddon* /*addon*/)
+{
+  // Do nothing
+}
+
+//==============================================================================
+void DynamicsShapeFrameAddonUpdate(DynamicsAddon* /*addon*/)
+{
+  // Do nothing
+}
```
mkoval: What are these empty functions for?

dart/dynamics/ShapeFrame.cpp
```diff
@@ -0,0 +1,394 @@
...
+//==============================================================================
+void VisualAddon::hide()
+{
+  setHidden(false);
+}
+
+//==============================================================================
+void VisualAddon::show()
+{
+  setHidden(true);
```
mkoval: `show()` and `hide()` appear to be reversed - unless `setHidden` has very unintuitive behavior. :smile:

dart/dynamics/ShapeFrame.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+  copy(*other);
+}
+
+//==============================================================================
+ShapeFrame& ShapeFrame::operator=(const ShapeFrame& other)
+{
+  copy(other);
+  return *this;
+}
```
mkoval: I'm not sure if it's a good idea to overload the assignment operator on this class. This is a pure virtual class, so all sorts of bad things could happen if you you try to assign two different `ShapeFrame` subclasses.

dart/dynamics/ShapeFrame.cpp
```diff
@@ -0,0 +1,394 @@
...
+  CollisionAddon* collisionData = getCollisionAddon();
+
+  if (createIfNull && nullptr == collisionData)
+    return createCollisionAddon();
+
+  return collisionData;
+}
+
+//==============================================================================
+DynamicsAddon* ShapeFrame::getDynamicsShapeNode(const bool createIfNull)
```
mkoval: This should be named `getDynamicsAddon` to be consistent with the two above methods.

dart/dynamics/ShapeFrame.cpp
```diff
@@ -0,0 +1,394 @@
...
+{
+  DynamicsAddon* dynamicsData = getDynamicsAddon();
+
+  if (createIfNull && nullptr == dynamicsData)
+    return createDynamicsAddon();
+
+  return dynamicsData;
+}
```
mkoval: It would be nicer to replace these three functions with a generic `createOrGet<AddonT>` template function on `AddonManager`. Is there a strong reason to define them here?

dart/dynamics/Shape.h
psigen: While I'm not sure if we will ever have a good full API abstraction for materials and textures, one thing to keep in mind is that a lot of APIs (especially Assimp) are often designed to cater to older rendering systems where lots of things like shadows and transparency were special-cased and needed all sorts of special datatypes.  Much of this gets simplified in newer pipeline-oriented rendering, and if we specifically keep that in mind, we might not need nearly as much cruft.

@mkoval, @mklingen: perhaps we should get an opinion from @PyryM here?

mklingen: why do you want a texture API, @mkoval ? Isn't that what the viewer is supposed to do on its own?

mxgrey: This conversation should probably move into a designated issue, but I would throw this idea out there:

```
VisualAddon
{
  Eigen::Vector4d mDefaultColor;
  std::shared_ptr< std::vector<Eigen::Vector4d> > mVertexColors;
  std::shared_ptr< TextureData > mTexture;

  enum ColorMode {
    DEFAULT_COLOR = 0,
    PER_VERTEX,
    TEXTURE
  };

  ColorMode mColorMode;

  /* ... Other VisualAddon stuff ... */
};
```

This way color and geometry data can be decoupled. If ``mVertexColors`` or ``mTexture`` being nullptr will imply that there is no per-vertex or texture coloring available. If ``mColorMode`` is set to an unavailable color mode, then it will simply use ``mDefaultColor``. If ``mVertexColors`` or ``mTexture`` is incompatible with the ``MeshShape`` of the ``ShapeNode``, then the renderer should print a message (once) that informs the user of this incompatibility and then use ``mDefaultColor`` instead.

I think this approach would satisfy everyone's goals while also being extremely memory efficient. The catch would just be the effort of implementation.

dart/dynamics/ShapeFrame.cpp
mxgrey: Yeah, I'm definitely inclined to agree on this. If anything, we should delete the assignment operator for ``ShapeFrame`` and leave it to ``SimpleFrame`` and ``ShapeNode`` to write it.

dart/dynamics/ShapeFrame.cpp
mxgrey: I'm guilty for starting the trend of allowing a bool to decide whether the ``get<AddonT>(bool createIfNull = false)`` call also creates an instance. I think that's nice for programmatically being able to decide whether or not to automatically generate one.

That said, I do think it would be good to also have a ``getOrCreate<AddonT>()`` function. I would like both to live side-by-side unless there's a strong objection to the ``get<AddonT>(bool createIfNull)`` version.

dart/dynamics/Shape.h
PyryM: I think materials should just be key/value stores that map string keys to [vec4 | texture | string] values. You can try to force everything into an overengineered canonical set of material attributes, but nobody will be happy. There are other vertex attributes you might want to have besides colors, as well. The tricky bit is that for efficiency a renderer should keep the vertex data GPU resident and only update it when it changes, which means that the viewer needs to be notified of mesh vertex data changes. Likewise with textures.

mxgrey: We'll actually have a versioning system that will allow the viewer to know when it needs to update things like color or vertex data.

For the key/value storage, where should that container live? Who should own it? I feel that this should be the job of the end user (DART is not really intended to be a comprehensive rendering engine), and we should design the ShapeFrame so that any customized information that a user might want to attach to their rendering information can be done easily (I would suggest through Addons).

PyryM: The VisualAddon would more or less become the key/value store. The advantage is that then an asset loader can just directly store whatever values it wants, rather than having to figure out how to map "fresnelTint" and "roughness" into "diffuse" and "specular" for example.

dart/dynamics/ShapeFrame.cpp
mkoval: I am fine with `get<Addon>(bool)` being the only API. I am actually suggesting that we remove these three methods entirely, because they are identical to `get<T>(true)`. What's the advantage of `getVisualAddon()` over `get<VisualAddon>(true)`?

mxgrey: For me, the motivation of having these functions baked in is to use the API to make it perfectly clear to users what kind of Addons are specialized for the object.

That said, I think the functions should do nothing but wrap the templated versions. I made a macro for creating such functions, so I'd suggest using that here.

dart/dynamics/Shape.h
mkoval: I very much *do not* want DART in the business of defining a canonical set of material properties. I am cautiously optimistic about @PyryM's suggestion:

> I think materials should just be key/value stores that map string keys to [vec4 | texture | string] values. 

It is straightforward to allow the user to store arbitrary vertex properties. We could simply allow the user to store a `std::vector<T>` with `size()` equal to the number of vertices. I am less sure about providing an API to store textures. Hopefully @PyryM can clarify a few points:

- Assimp supports spherical, cylindrical, and uv mapping. Are all three used in practice?
- I also found some references to uvw-mapping. How does this differ from uv mapping?
- Are uv-coordinates stored as separate vertex properties or attached to the image asset?
- How are uv-coordinates associated with the corresponding assert?
- Are all textures 24-bit RGBA? Is it common to use a different bit depth or number of channels?

My only other concern is that `VisualAddon` is attached to `ShapeFrame`, not `Shape`. These properties only apply to `MeshShape` and are intimately related to the mesh, e.g. arrays of vertex properties must have the same length as the number of vertices in the mesh.

Alternatively, we could take the simpler route I suggested in #453:

> `MeshNode` should contain the bare minimum amount of information necessary to perform collision checking and a dynamical simulation. Additionally, DART should provide an `Uri` to the original resource that can be loaded for visualization. The user is responsible for loading any additional information, e.g. textures and material properties, that is necessary only for visualization.

This has two downsides. First, every mesh is loaded from disk twice: once for collision detection and once for visualization. Second, we cannot programmatically construct a textured or vertex-colored mesh. This latter limitation bothers me more than the former - @mklingen and I independently ran into this in our research when rendering a color map.

psigen: Regarding the method in #453:

> The user is responsible for loading any additional information, e.g. textures and material properties, that is necessary only for visualization.

Consider loading an object from a file URI but then changing the object on disk (I was trying to adjust some objects in a scene while I had it open in `rviz`).  There is zero guarantee that each loading operation will have the same result, especially if assets are lazily-loaded, which can lead to out-of-sync visuals or even more pathological consistency issues.

This might even specifically be intentional in cases where an object is being dynamically updated, such as serving a generated model that results from perception or some other process at an `http` URI.

We also currently have to run an HTTP proxy server and rewrite URIs to implement `dart_rviz`, because we have the problem that URIs that are resolvable in the originating DART process aren't necessarily resolvable in the scope of the visualization (because it can be run over a network).

So, while it is very semantically clean, it does have some significant downsides.

PyryM: > Assimp supports spherical, cylindrical, and uv mapping. Are all three used in practice?

Spherical and cylindrical are ways of automatically assigning uv coordinates. I wouldn't bother trying to directly support them since any 3d editing program can compute these mappings and export them as regular uv coordinates.

> Are uv-coordinates stored as separate vertex properties or attached to the image asset?
> I also found some references to uvw-mapping. How does this differ from uv mapping?

Texture coordinates are vertex attributes, they're part of the mesh data like vertex positions and normals. Typical texture mapping, to simplify a lot, works like `color = my_texture[u,v]`. You can have 3d textures (a 3d array compared to a 2d array for a normal texture) in which case `color = my_texture[u,v,w]`. But the names and meanings of all the vertex attributes are just conventions, and shaders are free to interpret the attributes however they want.

> Are all textures 24-bit RGBA? Is it common to use a different bit depth or number of channels?

32 bit RGBA8/BGRA8 is the most common, and probably the only format that's likely to be encountered in this application. It might be worth supporting float textures (R32F or RGBA32F) to make it convenient to visualize procedurally generated data without having to squash or encode it into RGBA8.

dart/dynamics/ShapeFrame.cpp
jslee02: :+1:

dart/dynamics/ShapeFrame.cpp
jslee02: :+1:

dart/dynamics/ShapeFrame.cpp
jslee02: `setHidden` seemed a very complicated function to me. :joy:

dart/dynamics/ShapeFrame.cpp
jslee02: These functions get called when the addon properties are changed. I created them as placeholders expecting to add something there, but haven't found the use yet. I'll remove these and add back when necessary.

dart/dynamics/BodyNode.cpp
jslee02: :+1: Also, I'm adding const version that returns `std::vector<ShapeNode*>&`.

dart/dynamics/BodyNode.h
jslee02: This function was added exactly for the case you mentioned. One might want to create a `ShapeNode` with several addons as soon as it gets created. [Here is one example case](https://github.com/dartsim/dart/blob/shapenode/tutorials/tutorialBiped-Finished.cpp#L439). Without this function, we need multiple lines for this like:
```cpp
auto shapeNode = bodyNode->createShapeNode(shape);
shapeNode->createVisualAddon();
shapeNode->createCollisionAddon();
shapeNode->createDynamicsAddon();
```
I don't object to remove this function, but this would be useful for the above use case.

dart/common/AddonManager.h
```diff
@@ -86,10 +68,21 @@ class AddonManager
...
 
+  /// Default constructor
+  AddonManager() = default;
+
+  /// It is currently unsafe to copy an AddonManager
+  // TODO(MXG): Consider making this safe by cloning Addons into the new copy
+  AddonManager(const AddonManager&) = delete;
+
+  /// It is currently unsafe to move an AddonManager
+  AddonManager(AddonManager&&) = delete;
```
mkoval: I think you should also delete the copy assignment and move assignment operators. I doubt they are safe if the copy and move constructors are not.

dart/dynamics/BodyNode.cpp
mkoval: This implementation is dangerous. One `std::vector<ShapeNode*>` be used for all `BodyNode` instances.

I also don't see the point of this change. Returning the `vector` by value should be efficient because of return value optimization or, in the worst case, move semantics. The only advantage of returning a const reference is to avoid creating a new `vector` in the first place.

If that is not possible, I suggest going back to returning by value.

dart/dynamics/BodyNode.cpp
```diff
@@ -896,6 +824,77 @@ const Joint* BodyNode::getChildJoint(size_t _index) const
...
+  auto numShapeNode = getNumShapeNodes();
+
+  static std::vector<const ShapeNode*> shapeNodes;
+  shapeNodes.resize(numShapeNode);
+
+  for (auto i = 0u; i < numShapeNode; ++i)
+    shapeNodes[i] = getShapeNode(i);
+
+  return shapeNodes;
+}
```
mkoval: See above. This is dangerous.

dart/dynamics/Shape.h
mkoval: This discussion moved to #612. In the meantime, should we remove this function?

dart/dynamics/ShapeFrame.cpp
mkoval: :+1: to use a macro to create this functions. It avoids silly typos in duplicate code.

dart/dynamics/ShapeFrame.h
```diff
@@ -0,0 +1,372 @@
...
+{
+public:
+
+  CollisionAddon(const CollisionAddon &) = delete;
+  CollisionAddon(dart::common::AddonManager* mgr,
+                      const PropertiesData& properties = PropertiesData());
+
+  DART_COMMON_SET_GET_ADDON_PROPERTY( bool, Collidable )
+  // void setCollisionMode(const bool& value);
+  // const bool& getCollisionMode() const;
```
mkoval: These comments should say `get/setCollidable`. I suspect this is left over from a rename.

dart/dynamics/Shape.h
jslee02: I think we should keep this function as far as `MeshShape` contains color data and we want to use and update it.

dart/dynamics/BodyNode.cpp
jslee02: > This implementation is dangerous. One std::vector<ShapeNode*> be used for all BodyNode instances.

True. 



> The only advantage of returning a const reference is to avoid creating a new vector in the first place.

I thought returning the `vector` by value would create the `vector` not only the first time but also whenever it gets called. Is it incorrect?

It seems we need an efficient way to get the list of nodes from a node manager.  For now, we should return it by value or have the `vector` for particular nodes in a derived class to return it as a const reference. Or, we just use getNumNode/getNode(index) fashion. I created an issue about this #613, but that wouldn't for this pull request at least.

I think making this function to return `vector` by value would be fine for now.

mxgrey: > I thought returning the vector by value would create the vector not only the first time but also whenever it gets called. Is it incorrect?

If it returns by value, it will potentially have to allocate memory once each time it gets called, unless it can avoid that by using return value optimization. But one memory allocation is the worst possible case thanks to move semantics.

The danger of this static variable is that users are likely to catch the return of this function in a const-reference, but the contents of the vector that it's referring to would change as soon as the function gets called on a different ``BodyNode`` instance.

dart/dynamics/BodyNode.h
mxgrey: I like the function, although I wonder if we can have it accept ``Properties`` arguments for each Addon which will use their default constructor if one is not provided. I can't think of a way to do that at the moment, but I suspect it may be possible with enough cleverness.

dart/dynamics/BodyNode.cpp
jslee02: My concern was this case:
```cpp
auto shapeNodes1 = bodyNode->getShapeNodes();  // vector gets created
// change number of shape nodes of bodyNode
auto shapeNodes2 = bodyNode->getShapeNodes();  // another vector gets created
```

If we have the vector as a member variable of BodyNode, then the vector is created once when BodyNode instance is created no matter whether the num of shape nodes is changed after. Probably we would be able to avoid memory allocation if the new number of shape nodes is less than the allocated memory size (of course it depends on the implementation of vector).

mxgrey: Right, so far I've addressed this by having a cache of ``std::vector<const T>`` to complement the ordinarily used ``std::vector<T>``. This was the best solution I was able to come up with at the time, but admittedly it's doubling memory consumption in some places. I'll be very interested to look over your wrapper class idea as soon as I get the chance.

dart/dynamics/BodyNode.h
mkoval: This is only useful if you want to create the `Addon`, but not set any properties on it. Is this something that we expect to happen frequently?

dart/common/AddonManager.h
jslee02: I think this makes sense. @mxgrey how do you think?

mxgrey: Agreed. Eventually I'd like all of these functions to work by doing a deep copy using ``cloneAddon()``, but for now I think we should just delete them.

jslee02:
@mxgrey osgDart is not working. The essential change was replacing `EntityNode` and `FrameNode` with `ShapeFrameNode`. There were some tweaks that might not the best solution. I'd like to improve it in a separate pull request when we have a chance as this pull request is big enough.

It seems `EntityNode` and `FrameNode` are not used in osgDart. Would it be fine to delete them?

dart/collision/bullet/BulletCollisionNode.cpp
```diff
@@ -55,9 +55,11 @@ namespace collision {
 BulletCollisionNode::BulletCollisionNode(dynamics::BodyNode* _bodyNode)
   : CollisionNode(_bodyNode)
 {
-  for (size_t i = 0; i < _bodyNode->getNumCollisionShapes(); i++)
+  auto collShapeNodes = _bodyNode->getShapeNodes<dynamics::CollisionAddon>();
```
mxgrey: I wonder if we should name this function ``getShapeNodesWith<T>()`` or something like that to make it more clear that it's a filter.

dart/collision/bullet/BulletCollisionNode.h
```diff
@@ -59,11 +60,11 @@ namespace collision {
...
 class BulletCollisionDetector;
 
-struct BulletUserData {
-    dynamics::BodyNode* bodyNode;
-    dynamics::ConstShapePtr shape;
-    BulletCollisionNode* btCollNode;
-    BulletCollisionDetector* btCollDet;
+struct BulletUserData
+{
+  dynamics::ShapeNode* shapeNode;
```
mxgrey: Should we maybe use a ``WeakNodePtr`` here, or is this something that's strictly internal and will automatically be destroyed if the ``ShapeNode`` is destroyed?

dart/collision/bullet/BulletCollisionNode.cpp
mkoval: :+1: This is more clear.

dart/collision/fcl/FCLCollisionNode.h
```diff
@@ -62,12 +63,10 @@ class FCLCollisionNode;
 struct FCLUserData
```
mxgrey: Does this struct need to be exposed to the user? Maybe we should have it hidden in ``FCLCollisionNode.h``.

dart/collision/fcl/FCLCollisionNode.h
```diff
@@ -62,12 +63,10 @@ class FCLCollisionNode;
...
 {
   FCLCollisionNode* fclCollNode;
-  dynamics::BodyNode* bodyNode;
-  dynamics::Shape* shape;
+  dynamics::ShapeNode* shapeNode;
```
mxgrey: Same as for the Bullet user data: I wonder if we should make this a ``WeakNodePtr`` instead. That way if the ShapeNode gets deleted, it can see that and update itself accordingly.

dart/collision/bullet/BulletCollisionNode.cpp
jslee02: `getShapeNodesWith<T>()` looks good to me.

dart/collision/bullet/BulletCollisionNode.h
jslee02: It's not guaranteed. `WeakNodePtr` sounds good.

dart/dynamics/ShapeFrame.h
```diff
@@ -0,0 +1,370 @@
...
+  void show();
+
+  /// True iff the ShapeNode is set to be hidden. Use hide(bool) to change this
+  /// setting
+  bool isHidden() const;
+
+protected:
+
+  /// The parent shape frame
+  ShapeFrame* mShapeFrame;
```
mxgrey: This is already available to this class as the ``mManager`` variable. ``mManager`` is even of type ``ShapeFrame*``, because ``ShapeFrame`` was given as the third template argument to this class. ``mManager`` will even be a ``nullptr`` if this instance's manager is not a ``ShapeFrame``. Since this is member variable is redundant, I'd say we should remove it and replace its uses with ``mManager``.

dart/collision/fcl/FCLCollisionNode.h
jslee02: :+1:

dart/dynamics/SimpleFrame.cpp
```diff
@@ -120,6 +132,13 @@ SimpleFrame& SimpleFrame::operator=(const SimpleFrame& _otherFrame)
 }
 
 //==============================================================================
+std::shared_ptr<SimpleFrame> SimpleFrame::createChildSimpleFrame(
```
mxgrey: I'm slightly concerned about this function, because I'm afraid it might imply that the child you create will be held by this ``SimpleFrame`` instance. In other words, the user might create a bunch of children for this SimpleFrame and assume that it's sufficient to hold onto a reference to just the parent SimpleFrame in order to keep all the created children alive. I think we should either:

1. Make it extremely clear in the documentation that the user is responsible for hanging onto the created SimpleFrame. We should also use a different word besides ``create``, because in all other cases where you call ``createXXX()`` on an object instance, that object instance will hang onto the created instance for you. Maybe change the name to ``spawnChildSimpleFrame`` just to make it clear that the behavior is different from other ``createXXX()`` functions.

2. Allow SimpleFrames to hold onto a ``std::vector<SimpleFramePtr>`` for child frames. This would allow a user to hold onto an entire tree of SimpleFrames just by holding onto a reference to the root ``SimpleFrame``. The drawback to this is a user might think that the SimpleFrame tree will behave like a Skeleton where holding onto any member of it is sufficient to hold onto all of them, but that would not be the case here.

dart/dynamics/ShapeFrame.h
jslee02: Yeah, this was added before the `ManagerTrackingAddon` is implemented and not needed anymore. :+1:

dart/dynamics/SimpleFrame.cpp
mkoval: I am strongly against (2). If you want to create a kinematic tree, then you should create a `Skeleton`. I see no point in duplicating that functionality here.

mxgrey: I'm very much inclined to agree with your opposition to (2), although I'm not against it quite as strongly :wink:

dart/dynamics/ShapeFrame.cpp
mxgrey: Should we replace these with macros before or after the pull request goes through?

dart/dynamics/ShapeFrame.cpp
```diff
@@ -0,0 +1,380 @@
...
+                       const ShapePtr& shape)
+  : common::AddonManager(),
+    Entity(ConstructFrame),
+    Frame(parent, name),
+    mShapeUpdatedSignal(),
+    mRelativeTransformUpdatedSignal(),
+    onShapeUpdated(mShapeUpdatedSignal),
+    onRelativeTransformUpdated(mRelativeTransformUpdatedSignal)
+{
+  setName(name);
```
mxgrey: I'm somewhat concerned about calling this here since ``setName`` is a virtual function, and this is a pure abstract class wherein the most-derived implementer of ``setName`` is not likely to be constructed yet.

I actually intend for ``Entity`` to become a completely abstract interface class, so in the future there will be no resolution at all for ``setName`` in this constructor.

dart/dynamics/ShapeFrame.cpp
jslee02: Let's do it in this pull request.

Besides on that, would it be fine with you to change the macro style adding more lines like:
```cpp
#define DART_BAKE_SPECIALIZED_ADDON_IRREGULAR( TypeName, AddonName ) \
  inline bool has ## AddonName () const                              \
  {                                                                  \
    return this->template has<TypeName>();                           \
  }                                                                  \
```
?

dart/dynamics/ShapeFrame.h
```diff
@@ -0,0 +1,365 @@
...
+  /// Constructor
+  VisualDataProperties(
+      const Eigen::Vector4d& color = Eigen::Vector4d(0.5, 0.5, 1.0, 1.0),
+      const bool hidden = false);
+
+  /// Destructor
+  virtual ~VisualDataProperties() = default;
+};
+
+struct CollisionDataProperties
```
mxgrey: I wonder if we should be able to embed information about which bodies to ignore collisions against. For example, maybe there are two bodies in a Skeleton which you always want to ignore collisions between even though they don't technically qualify as adjacent. You could have a ``std::set<std::string>`` containing the names of those bodies in here. Using the body names (and not something like their pointers) allows this information to be trivially cloned and copied. Body names within a Skeleton are guaranteed to be unique, so there's no need to worry about ambiguity. Same thing if we want to specify other ``ShapeNode``s to ignore collisions against.

It gets much trickier if we want to ignore collisions against bodies in other Skeletons. We could have a ``std::map<std::string, std::set<std::string> >`` where the map key is the name of the Skeleton and the set of strings are names of the bodies within that Skeleton. But this would make it difficult to specify a pattern like "I want to ignore all bodies with name ``x`` in Skeletons whose names begin with ``y``".

I know @jslee02 is working on Collision Groups now. I wonder if we could have a ``std::string`` (or something similar) in here that tags the collision group name for this ``ShapeFrame``, and then we could have some way of defining the properties of the collision group elsewhere (in a way that's cloneable).

dart/dynamics/ShapeFrame.h
```diff
@@ -0,0 +1,365 @@
...
+  DynamicsDataProperties(const double frictionCoeff = 1.0,
+                         const double restitutionCoeff = 0.0);
+
+  /// Destructor
+  virtual ~DynamicsDataProperties() = default;
+};
+
+} // namespace detail
+
+class VisualAddon final :
```
mxgrey: I'm not sure if I like that we have Visual**Data**Properties but then Visual**Addon**. I would prefer Visual**Data**Properties and Visual**Data**. Or I'd be okay with Visual**Addon**Properties and Visual**Addon**. But I don't think it makes sense to mix them.

dart/dynamics/SimpleFrame.cpp
mxgrey: Are there any opinions about changing this function's name to ``spawnChildSimpleFrame``? Is anyone in favor of it or against it?

dart/dynamics/SoftBodyNode.h
```diff
@@ -361,7 +356,7 @@ class SoftBodyNode : public BodyNode
   UniqueProperties mSoftP;
 
   /// \brief Soft mesh shape belonging to this node.
-  std::shared_ptr<SoftMeshShape> mSoftShape;
+  ShapeNode* mSoftShapeNode;
```
mxgrey: I think we should use a ``NodePtr<ShapeNode>`` here instead to make sure it doesn't get deleted due to the user fiddling around with the ShapeNodes. I also think we need a more robust way of handling Soft Shapes, especially when it comes to things like cloning, but we can defer that issue for now.

dart/dynamics/SimpleFrame.cpp
jslee02: Documentation and using different name sound reasonable solution to me as well.

dart/dynamics/ShapeFrame.cpp
jslee02: Agree. I'll leave a reminder note here for the future change of `Entity`.

dart/dynamics/ShapeFrame.h
jslee02: I would more prefer the pair of `VisualAddon` and `VisualAddonProperties` since `Data` sounds more general than `Addon` especially when we have `Addon` mechanics.

dart/dynamics/SoftBodyNode.h
jslee02: :+1: I guess you meant `TemplateNodePtr<ShapeNode>` or its typedef `ShapeNodePtr`.

Yeah, one day we need refactoring of soft body dynamics including Soft Shapes.

dart/dynamics/ShapeFrame.h
jslee02: I think it's a good idea to embed the collision filter information in `CollisionAddon` (more precisely `CollsionAddonProperties`. In the new collision API, the collision filtering will be handled by passing filter object, which is a functor class, to the collision detector. The filter function decides the collidability of the two collision objects (`ShapeFrameCollisionObject` in dynamics namespace) based on the information from the two collision objects passed in. Here we can exploit the embedded collision filter information.

My concern is how we encode the information. Using `std::set<std::string>` and `std::map<std::string, std::set<std::string>>` seems a considerable approach. We can also use *absolute name concept* like `skeleton_name.bodynode_name` and simple regex technique like `*.bodynode_name` (to specify all the body nodes with `nodynode_name` in any skeleton) for example. 

Another concern is the level of ignoring collision object. We might want to ignore specific ShapeFrame rather than all the ShapeFrame in the BodyNode. In some cases, we would like to have multiple collision shapes within a BodyNode for the different uses. For example, we'd like primitive decomposition used for low fidelity collision checking and high-fidelity collision geometry used for physics simulation.

Also, I would like to save this issue for Collision Group implementation or another separate pull request.

dart/dynamics/SoftBodyNode.h
mkoval: Won't this introduce a circular reference?! `NodePtr<>` stores a pointer to the parent `BodyNode`, which would ourselves.

dart/dynamics/ShapeFrame.h
mkoval: We intended to split the collision checking improvements into three pull requests: (1) the `ShapeNode` refactor, (2) the collision group API, and (3) methods for easily constructing collision groups from "tag" metadata attached to the `Skeleton`.

I'd prefer to hold off on this API discussion until after (1) and (2) are done, since (3) is mostly a convenience method for constructing collision groups and mask functions. Thoughts?

mxgrey: That plan sounds good. I brought it up here only because it was a thought I had while reviewing, and I wanted to make sure I got the thought out of my head before I forgot it.

dart/dynamics/SoftBodyNode.h
mxgrey: Good catch. We'd want to use the ``Weak`` version.

dart/dynamics/BodyNode.h
```diff
@@ -613,6 +526,63 @@ class BodyNode :
...
+
+  /// Remove all ShapeNodes from this BodyNode
+  void removeAllShapeNodes();
+
+  /// Remove all ShapeNodes containing given Addon from this BodyNode
+  template <class Addon>
+  void removeAllShapeNodes();
+
+  /// Add a collision Shape into the BodyNode
+  void addCollisionShapeNode(const ShapePtr& shape);
```
mxgrey: It looks like this function has been left undefined. I'm not exactly sure how we should approach defining it. Should the ``ShapeNode`` be given default properties? I'm inclined to say that we don't really need this function, since we already have the templated version of ``createShapeNode`` that allows the collision addon to be created simultaneously.

dart/dynamics/BodyNode.h
```diff
@@ -613,6 +526,63 @@ class BodyNode :
...
+
+  /// Remove a collision Shape from this BodyNode
+  void removeCollisionShapeNode(const ShapePtr& shape);
```
mxgrey: Like above, this function is currently undefined. Would it remove all ``ShapeNode``s that contain this particular shape? I'm inclined to think that this function shouldn't exist either.

mxgrey:
I've finished reviewing. All of my comments which are relevant to this pull request have been addressed. I've also gone ahead and made some changes to the API for accessing and creating ShapeNodes. In particular, I've added ``With`` to all functions that operate on a subset of ShapeNodes that contain a particular Addon type. I've also removed the ``AddonAdder`` struct and replaced it with two ``void``-returning variadic functions which are now located in ``common/AddonManager.h``.

I only have one remaining concern: It seems that soft bodies are very very broken right now. If we're okay with soft bodies being broken on ``master``, then I'm fine with merging in this pull request. I do think we should try to get them at least looking believable before the major release, though.

jslee02:
Could you take a look at osgDart as well? I just made it work (I believe ) but there are some stopgaps. For example, `EntityNode` is replaced by `ShapeFrameNode` but WorldNode is still using `EntityNode`.

I'm looking at soft bodies now.

mxgrey:
The drag and drop features seem to be working fine. Unfortunately, the diff for those files isn't showing up for me on Github. If anyone can recommend a good git diff viewing desktop application, I would appreciate it. Otherwise, I can comb through the raw code some time soon.

psigen:
On Ubuntu, I generally use `meld`.

mxgrey:
I use ``meld`` for merging, but I don't like looking back and forth from side to side when it comes to comparing two files. I really like the linear way that Github lays it out. The standard console tool lays it out similarly, but isn't very pretty. I probably won't be able to find what I want, so I'll just have to settle for ``meld`` or the console. It would be nice if Github offered an offline version of their viewer.

jslee02:
> It would be nice if Github offered an offline version of their viewer.
:+1: 

I created two dummy branches (with and without osgDart changes) to help reviewing osgDart [here](https://github.com/dartsim/dart/compare/shapenode_without_osgDart...shapenode_with_osgDart?expand=1). It doesn't include commit-wise changes but you can see the diff.

Also, I resolved the instability of the soft bodies in the last commits.

dart/dynamics/BodyNode.h
jslee02: Yeah, I also intended to remove this function but forgot to remove the declaration. Removing.

psigen:
@mxgrey you can possibly try the vertical diff mode in `vimdiff`?
http://stackoverflow.com/a/244766

I actually find it hard to use the unified vertical diff, because it makes it really tricky to compare large block diffs, but I think several difftools have flags to go into that mode.

mxgrey:
I've pushed some fixes to the way ``Addon``s are cloned, and I've restructured ``osgDart``. Soft bodies seem to be working again after @jslee02 patched it.

I think this pull request is good to merge now :+1:

jslee02:
:+1: looks god to me too. Will merge this once #607 is merged.

PR #616, opened 2016-02-22
Made ConstraintSolver obey C++11 ownership conventions

mkoval:
`constraint::ConstraintSolver` takes ownership of the `setCollisionDetector` passed in to `setCollisionDetector`. This is not clear from the API because the function accepts a raw pointer. This leads to a SEGFAULT if you do not expect `ConstraintSolver` to delete your `CollisionDetector`.

I modified `setCollisionDetector` to take an `unique_ptr` to make the ownership transfer explicit. I also switched to storing the `LCPSolver` in a `unique_ptr` to remove the other manual `delete` from the `ConstraintSolver` destructor.


jslee02:
:+1:

PR #617, opened 2016-02-22
Added missing override specifiers on draw() methods

mkoval:
This fixes a bunch of `-Winconsistent-missing-override` warnings in Clang 3.7.


jslee02:
:+1:

PR #621, opened 2016-02-23
Fixed initialization of position and velocity

mkoval:
The position and velocity of freshly-constructed `Joint`s was being set to zero instead of, respectively,`mInitialPosition` and `mInitialVelocity`. These values would only be applied if the user manually calls `resetPosition()` and `resetVelocity()`.

This pull request modifies the `SingleDofJoint` and `MultiDofJoint` constructors to use the correct values for initialization. I believe all other joints (except `WeldJoint`, which has no DOFs) inherit from these base classes, so we don't need to make any other changes.


jslee02:
:+1:

PR #623, opened 2016-03-02
Create 'frames' directory if not exists for screenshot

### LINKED ISSUE #395: GlutWindow::screenshot() doesn't work

jslee02:
This fixes #395 thanks to @olzhas.


mxgrey:
I'm a little bit concerned about the ambiguity of the term ``frames`` here, since we now have a class called ``Frames``. Same with the name of the variable ``World::mFrame`` and function ``World::getSimFrames()``. I've been thinking this for a while, but this pull request reminded me of the matter. I think it would be good to rename these things for the next major release.

jslee02:
Yeah, I recently noticed that too. It seems one of them needs to be renamed (probably SimFrame). How about creating an issue to track this?

Issue #613, opened 2016-02-17
Consistent interface of container class for element access


jslee02:
DART has many container classes. For example, `Skeleton` is a container of `BodyNode` and `Marker`, and `Joint` is a container of `DegreeOfFreedom`. To access the elements of the containers, there are two styles of interface: (1) getNum(Object)() / getObject(index) and (2) const vector<Object> getObjects(). Some container provide only (1), others provide (2), and some provide both, which is not consistent.

I would like to propose to have a consistent interface or, at least, have an interface that returns an iterable object like `std::vector<Object*>`. With the iterable object, we can utilize the advantage of ranged-based for loop of C++11, which provides a neater way of iterating the elements.

Also, in most cases, container classes return const reference of the iterable object that contains const/non-const pointers like `const std::vector<BodyNode*>&` and `const std::vector<const BodyNode*>&`. For this, two ways are used: (1) having two iterable objects in it, (2) return a const_cast-ed copy of the iterable object. Both of them have downsides: (1) requires unnecessary memory, and (2) requires unnecessary computation for creating the iterable object every time the accessor is called.

To resolve this, I'd like to propose a wrapper class of pointer container that provides a customized iterator for const_cast when the element is accessed instead of holding the copy of const_cast-ed elements inside it. I thought there must be others having a similar issue and reasonable solution, but couldn't find. Please let me know if it's not.

[Here](https://gist.github.com/jslee02/37da0df5c954d7539888) is a brief sketch of the wrapper class with tests.

jslee02:
Resolved by #1644

PR #631, opened 2016-03-11
CollisionGroup class and refactoring of CollisionDetector

jslee02:
This pull request introduces collision group feature with CollisionDetector refactoring. 

An overview of the changes is:

* `CollisionGroup` class is added. Collision detector used to containe collision objects in it and only could check collisions for the objects. So we couldn't be able to check collisions for arbitrary (distict) collision objects. `CollisionGroup` now plays the roll of collision object container, and collision detector can check collisions for object-object, object-group, and group-group.

* `collision` is now only dependent on `Shape` class. So once we move `Shape` and the derived classes to `common` or `math` namespaces then `collision` would be completely independent on 'dynamics' namespace.

* `FCLCollisionDetector` now takes the advantage of broad phase collision detectorion of FCL to avoid checking all pairs of objects.

* Collision detector manages Shapes so that the collision detector doesn't create multiple collision detector specific collision object (e.g., btGImpactMeshShape, fcl::BVHModel<OBBRSS>) per Shape.

This pull request is not completed yet. The major issue remained is the ownership between `CollisionDetector` and `CollisionObject`. I'm considering apply the same mechanism of the customized shared pointer `BodyNodePtr` by making it general to be able to be used for `CollisionObject`.

Note that this pull request includes the changes of #608 .


mxgrey:
Out of curiosity, is there a particular motivation for making the ``collision`` namespace independent of the ``dynamics`` namespace? Are we trying to provide the general public with a generic API that will allow other (non-DART) projects to arbitrarily swap between the different collision engines that we support? If so, then I can understand the motivation, but otherwise I don't see why we'd make that separation.

mxgrey:
> The major issue remained is the ownership between ``CollisionDetector`` and ``CollisionObject``. I'm considering apply the same mechanism of the customized shared pointer ``BodyNodePtr`` by making it general to be able to be used for ``CollisionObject``.

I'm not clear on what the purpose of this would be. Does a ``CollisionObject`` have any meaning outside of its ``CollisionDetector``? Do we expect users to be handling, manipulating, and modifying ``CollisionObject``s independently of the ``CollisionDetector`` that manages them? What would be the scenario where the user has a reference to a ``CollisionObject`` while its ``CollisionDetector`` has no references left?

But if we do decide that we need a special type of smart pointer for ``CollisionObject``s, here's something to take into account: The main factor that makes ``BodyNodePtr`` so complex is that a ``BodyNode`` might get swapped between ``Skeleton``s. Because of this, we need to do some extra weirdness. Assuming a ``CollisionObject`` cannot be transferred between two ``CollisionDetector``s, we should really make its smart pointer more like ``TemplateNodePtr``. All it would have to do is hold onto a reference to a ``CollisionDetectorPtr``. We wouldn't need to do any extra reference counting inside of the ``CollisionObject`` the way we need to for ``BodyNode``.

jslee02:
> Out of curiosity, is there a particular motivation for making the collision namespace independent of the dynamics namespace?

The motivation behind this pull request is to resolve the circular dependencies of `collision` and `dynamics` unless we want to eventually merge the two namespaces.

> Are we trying to provide the general public with a generic API that will allow other (non-DART) projects to arbitrarily swap between the different collision engines that we support?

This would be the natural result of the separation, but this is not the main reason.

mxgrey:
> The motivation behind this pull request is to resolve the circular dependencies of ``collision`` and ``dynamics`` unless we want to eventually merge the two namespaces.

I can understand why ``constraint`` would depend on both ``collision`` and ``dynamics`` since that's where collision constraints are resolved against the dynamics, but I'm not clear on why ``dynamics`` would have to depend on ``collision``. To me it would seem more natural for ``collision`` to depend on ``dynamics``, especially since I feel it's much more likely that a user would want to use the dynamics features without the collision features rather than vice-versa.

mkoval:
It is only valid to use a `CollisionObject` with the `CollisionDetector` that created it. The only reason we are considering adding new smart pointer types for `CollisionGroup` and `CollisionObject` is to avoid creating a circular `smart_ptr` reference.

I do remember concluding that the ownership needs to be coupled, but now I can't recall why. Hopefully @jslee02 can remind us. :sweat:

mkoval:
> To me it would seem more natural for `collision` to depend on `dynamics`, especially since I feel it's much more likely that a user would want to use the `dynamics` features without the `collision` features rather than vice-versa.

I don't see a problem with having `collision` depend on `dynamics`, as long as we don't introduce a circular dependency. @jslee02 can you explain again why this was not possible?

jslee02:
> I can understand why constraint would depend on both collision and dynamics since that's where collision constraints are resolved against the dynamics, but I'm not clear on why dynamics would have to depend on collision. To me it would seem more natural for collision to depend on dynamics, especially since I feel it's much more likely that a user would want to use the dynamics features without the collision features rather than vice-versa.

I think `constraints` and `dynamics` should be merged first unless we want to distinct unconstrained dynamics and constrained dynamics. Also, the two namespaces are already dependent on each other. That means that `dynamics` should depend on `collision` for resolving contact constraints.

I agree that a user would want to use the dynamics features without the collision features like to compute mass matrix and compute inverse dynamics. I think this is for the case of using DART for controller design (as we expect the users to use DART for controller design and kinematics/dynamics simulation). However, in the sense of simulation, I also imagine a user would want to check collision in kinematics simulation without dynamics (e.g., motion planning).

mxgrey:
I think it would make sense to separate implicitly solved constrains (such as what's done with the Articulated Body Algorithm in the ``dynamics`` namespace) and explicitly solved constraints (such as collision constraints in the ``constraint`` namespace). If ``dynamics`` is currently dependent on ``constraint``, then I would prefer if that were changed somehow.

mkoval:
> However, in the sense of simulation, I also imagine a user would want to check collision in kinematics simulation without dynamics (e.g., kinematic motion planning).

Even kinematic motion planning requires the forward (and, often, inverse kinematics) found in the `dynamics` namespace. I don't see a strong use-case for using `collision` without `dynamics` unless, like @mxgrey said, we are trying to provide the general public with a collision detection API.

I don't think that is an immediate goal for any of us, so I don't see an issue with making `collision` depend on `dynamics. This assumes that we resolve the circular dependency issue so `dynamics` does not also depend on `collision`.

jslee02:
This pull request is ready for review. The collision API has been changed reflecting the above and offline discussions. Here is summary of the changes:

* `collision` is dependent on `dynamics`. The major motivation behind this is that `dynamics` should be lightweight as much as possible (either from other namespaces and 3rd party libraries such as collision detectors) for the case it's used in controllers (e.g., computing mass matrix, inverse kinematics, and unconstrained forward dynamics).

* Merged `FCLMeshCollisionDetector` into `FCLCollisionDetector`. Whether mesh is used for primitive shape can be selected by `FCLCollisionDetector::setPrimitiveShapeType`. The default is using mesh for primitive shapes. Primitive shape support of FCL is still buggy (at least FCL 0.4.0) so we should stick with using mesh more.

* Bumped the major version as we changed the API

* Disabled AppVeyor build test (until Visual Studio 2015 can build `constexpr` for function pointer)

I also found that `testWorld` fails with FCL 0.4.0 (fine with FCL 0.3.0). I'll investigate this but not in this pull request.

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,210 @@
...
+
+  /// Return number of ShapeFrames added to this CollisionGroup
+  size_t getNumShapeFrames() const;
+
+  /// Return index-th ShapeFrames registered to this CollisionGroup. Return
+  /// nullptr if the index is out of the range.
+  const dynamics::ShapeFrame* getShapeFrame(size_t index) const;
+
+  /// Return all the ShapeFrames registered to this CollisionGroup
+  const std::vector<const dynamics::ShapeFrame*> getShapeFrames() const;
```
mkoval: Is this supposed to return a `const &`? If not, I see no purpose of having a `const` here.

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,210 @@
...
+  void registerShapeFrame(const dynamics::ShapeFrame* shapeFrame);
+
+  /// Register ShapeFrames to this CollisionGroup
+  void registerShapeFrames(
+      const std::vector<const dynamics::ShapeFrame*>& shapeFrames);
+
+  /// Register all the ShapeFrames in the other CollisionGroups to this
+  /// CollisionGroup
+  template <typename First, typename... Others>
+  void registerShapeFrames(const First* first, const Others*... others);
```
mkoval: Either the name or the comment of this method is misleading:

- the name suggests that it takes a variable number of `ShapeFrame *`s
- the comment suggests that it takes a variable number of `CollisionGroup *`s

At a higher level, I am in favor of moving this method entirely. I believe we can achieve the same functionality by calling the `std::vector` overload:
```c++
registerShapeFrames({shapeFrame1, shapeFrame2, ...});
```

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,210 @@
...
+  CollisionGroup(const CollisionDetectorPtr& collisionDetector);
+
+  /// Destructor
+  virtual ~CollisionGroup();
+
+  /// Return collision detection engine associated with this CollisionGroup
+  CollisionDetector* getCollisionDetector() const;
+
+  /// Register a ShapeFrame to this CollisionGroup
+  void registerShapeFrame(const dynamics::ShapeFrame* shapeFrame);
```
mkoval: `CollisionGroup` sounds like a collection, so I am in favor of renaming this to `addShapeFrame`. Is there a strong reason to use the verb `register` here, instead?

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,210 @@
...
+  /// Register all the ShapeFrames in the other CollisionGroups to this
+  /// CollisionGroup
+  template <typename First, typename... Others>
+  void registerShapeFrames(const First* first, const Others*... others);
+
+  /// Do nothing. This function is for terminating the variadic template
+  /// function template<typename...> registerShapeFrames().
+  void registerShapeFrames();
+
+  /// As as registerShapeFrame()
```
mkoval: Typo: `As as`.

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,210 @@
...
+  void registerShapeFramesOf(const dynamics::ShapeFrame* shapeFrame);
+
+  /// As as registerShapeFrames()
```
mkoval: Typo: `As as`.

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,210 @@
...
+  void registerShapeFramesOf(
+      const std::vector<const dynamics::ShapeFrame*>& shapeFrames);
+
+  /// Register all the ShapeFrames in the other CollisionGroup to this
+  /// CollisionGroup
+  void registerShapeFramesOf(const CollisionGroup* other);
+
+  /// Register all the BodyNode's ShapeFrames having CollisionAddon to this
+  /// CollisionGroup
+  void registerShapeFramesOf(const dynamics::BodyNode* bodyNode);
```
mkoval: I am confused about the semantic of these helper functions: If I call this function on a `BodyNode *`, then add a `ShapeFrame` to it, will that change be propagated to this group?

I don't believe that is the case. If so, we should clearly document this in the comment.

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,210 @@
...
+  /// Return index-th ShapeFrames registered to this CollisionGroup. Return
+  /// nullptr if the index is out of the range.
+  const dynamics::ShapeFrame* getShapeFrame(size_t index) const;
+
+  /// Return all the ShapeFrames registered to this CollisionGroup
+  const std::vector<const dynamics::ShapeFrame*> getShapeFrames() const;
+
+  /// Update engine data. This function should be called before the collision
+  /// detection is performed by the engine in most cases.
+  void update();
```
mkoval: What's the motivation behind making this `public`, instead of `protected` like the other internal methods used by the `CollisionDetector`? Is there any case where I would call this as a user?

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,210 @@
...
+
+  /// Notify that all the CollisionObjects are remove so that the collision
+  /// detection engine do some relevant work
+  virtual void notifyAllCollisionObjectsRemoved() = 0;
+
+  /// Update the collision detection engine data such as broadphase algorithm.
+  /// This function will be called ahead of every collision checking.
+  virtual void updateEngineData() = 0;
+
+protected:
```
mkoval: Is it safe for the user to modify these? My understanding is that these fields should *only* be modified by the non-`virtual` `public` methods on this class. If so, we should make them `private`.

dart/collision/Contact.h
```diff
@@ -0,0 +1,90 @@
...
+#include "dart/dynamics/SmartPointer.h"
+
+namespace dart {
+namespace collision {
+
+/// Contact information
+struct Contact
+{
+  // To get byte-aligned Eigen vectors
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
mkoval: I don't think this is strictly necessary here. `Vector3d` is not a [fixed-size vectorizable type](http://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html) because it is not a multiple of 16 bytes.

dart/collision/Option.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include <cstddef>
+#include <memory>
+
+namespace dart {
+namespace collision {
+
+class CollisionFilter;
+
+struct Option
```
mkoval: The member variables in this `struct` do not follow the same naming convention, e.g. `mEnableContact`, as the rest of DART. I'm not thrilled about the `m` prefix, but consistency rules all. :sweat:

dart/collision/fcl/FCLCollisionDetector.h
```diff
@@ -38,50 +38,104 @@
...
-      const fcl::CollisionObject* _fclCollObj) const;
+  // Documentation inherited
+  bool detect(CollisionGroup* group1, CollisionGroup* group2,
+              const Option& option, Result& result) override;
+
+  /// FCL's primitive shape support is still not complete. FCL 0.4.0 fixed lots
+  /// of the bugs, but it still returns single contact point per shape pair
+  /// except for box-box collision. For this reason, we recommend using mesh for
+  /// primitive shapes until FCL fully support primitive shapes.
+  enum PrimitiveShape_t
```
mkoval: You should make it clear in this comment that primitive shape support is only buggy for *contact point* computation. It works fine for binary checks.

Also, I suggest making this an `enum class` and renaming it to `PrimitiveShape`. Why does this have the `_t` suffix?

dart/collision/CollisionDetector.h
```diff
@@ -43,169 +43,158 @@
...
+  /// Constructor
+  CollisionDetector() = default;
 
-  // TODO(JS): userData is an experimental variable.
-  /// \brief User data.
-  void* userData;
-};
+  /// Claim CollisionObject associated with shapeFrame. New CollisionObject
+  /// will be created if it hasn't created yet for shapeFrame.
+  CollisionObject* claimCollisionObject(const dynamics::ShapeFrame* shapeFrame);
```
mkoval: The `claim`/`unclaim` logic looks *very much* like manual memory management. Could we replace this logic with RAII by using an `std::shared_ptr` to manage `CollisionObject` instances?

Here's a sketch:
- `CollisionDetector` maintains a `std::map<const dynamics::ShapeFrame *, std::weak_ptr<CollisionObject>> mCollisionObjects` of all known `CollisionObject`s internaly.
- `claimCollisionObject` returns `mCollisionObjects[shapeFrame].lock()` if possible.
- If not, it creates a new `CollisionObject`, returns the `shared_ptr`, and stores it in the `mCollisionObjects` map.
- Optionally, we can use a custom deleter on `std::shared_ptr` to clean up internal state in `CollisionDetector`.

mkoval:
I'm unclear on the ownership semantics of `CollisionDetector`, `CollisionGroup`, `CollisionObject`, and `ShapeFrame`. @jslee02 could you explain who-owns-who? How are these ownership relations managed (`shared_ptr`, `unique_ptr`, raw pointer, etc?

@jslee02 Could you clarify this?

dart/collision/CollisionGroup.h
jslee02: The goal of this helper function is to take multiple objects of various kinds that have multiple `ShapeFrame`s such as `CollisionGroup`, `BodyNode`, and `Skeleton`, which can not be achieve using initializer lists.

For example:
```cpp
group->registerShapeFrames(otherGroup, bodyNode1, bodyNode2, skeleton);
```

The comment needs to be updated, though. 

I agree with that the name could be confusing, but there were two reason I stay with this name.

First, `registerShapeFrames` is so general that it can take single ShapeFrame. For example, following is possible:
```cpp
group->registerShapeFrames(shapeFrame1); // or
group->registerShapeFrames(otherGroup, shapeFrame2, skeleton);
```
I thought `registerShapeFrames` would be enough to represent the generality.

(The background C++ knowledge of the second reason possibly might not correct.) `registerShapeFramesOf` or `registerShapeFrameFrom` sounded better name for this function to me at first. But, this recursive variadic template function name needs to be different from other `registerShapeFramesOf` function that takes a single parameter for the recursion termination.

dart/collision/CollisionGroup.h
jslee02: Yeah, it's supposed to be `const &`.

dart/collision/CollisionGroup.h
jslee02: :+1: `addShapeFrame` sounds good to be consistent with other collections.

dart/collision/CollisionGroup.h
jslee02: Right, it should be `Same as`.

dart/collision/CollisionGroup.h
jslee02: It should be protected. One downside is all the collision detector classes needs to be friend class of CollisionGroup as `update()` function is called by them.

dart/collision/CollisionGroup.h
jslee02: :+1: Good point

dart/collision/Contact.h
jslee02: :+1:

dart/collision/Option.h
jslee02: I'm fine with either using `m` prefix or not, but I don't believe we have one consistent rule on this yet. The original code base and I was on the side of not using `m` prefix, but it seems like using it is dominant.

dart/collision/fcl/FCLCollisionDetector.h
jslee02: Will update the comment.

I thought using `_t` suffix is our convention as the recently added enum have the suffix: [1](https://github.com/dartsim/dart/blob/e8b4f1c40e26140d93ea803bcb4a76547ebe05fb/dart/dynamics/Chain.h#L84), [2](https://github.com/dartsim/dart/blob/ba23a8ae30e4e6148a91a813c18b4f5e566c9839/dart/common/Empty.h#L49), [3](https://github.com/dartsim/dart/blob/e14110c8f94cef2cfd93d3d4dc054ba8a8cb2a6b/dart/constraint/BalanceConstraint.h#L85), [4](https://github.com/dartsim/dart/blob/71f6aa2f8644138e321fe6bdfbffda6bcbf16289/dart/dynamics/InverseKinematics.h#L620).

jslee02:
> How are these ownership relations managed (shared_ptr, unique_ptr, raw pointer, etc?

`CollisionGroup` owns `CollisionDetector` as `shared_ptr` so that the collision detector doesn't get destroyed as long as the collision group is alive. For this, `CollisionGroup` must take `shared_ptr<CollisionDetector` when it gets created.

`CollisionDetector` owns `CollisionObject`. When a shape frame is added to a collision group, the collision group claims a collision object for the shape frame to the collision detector. The collision detector then creates a collision object (and owns it as `unique_ptr`), and hand over the raw pointer of the collision object to the collision group.

For now, when `CollisionGroup` is destroyed it reclaims all the collision objects (holding as raw pointers) to the collision detector so that the collision detector release collision objects that are not held by any collision group.

Additionally, `[FCL/Bullet]CollisionDetector` owns and manages `fclCollisionGeometry/btCollisionShape` in order to not creating duplicate `fclCollisionGeometry/btCollisionShape` per same `dart::dynamics::Shape` for the sake of the efficiency.

dart/collision/CollisionFilter.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+  if (object1 == object2)
+    return false;
+
+  auto shapeNode1 = object1->getShapeFrame()->asShapeNode();
+  auto shapeNode2 = object2->getShapeFrame()->asShapeNode();
+
+  if (!shapeNode1 || !shapeNode2)
+    return true;
+  // Assume non-ShapeNode always collides with others.
```
mkoval: This is a misleading comment. Returning `true` here means that we always *check* collision pairs that include anon-`ShapeNode`. It does not mean that it always *collides*; i.e. is in collision.

dart/collision/CollisionGroup.cpp
```diff
@@ -0,0 +1,209 @@
...
+//==============================================================================
+void CollisionGroup::removeShapeFrame(const dynamics::ShapeFrame* shapeFrame)
+{
+  if (!shapeFrame)
+    return;
+
+  const auto search
+      = std::find_if(mShapeFrames.begin(), mShapeFrames.end(),
+                     [&](const dynamics::ShapeFrame* it)
+                     { return it == shapeFrame; });
```
mkoval: `shapeFrame` exactly matches the `value_type` of `mShapeFrames`, so I am not sure why `std::find_if` is necessary here. Shouldn't `std::find(mShapeFrames.begin(), mShapeFrames.end(), shapeFrame)` work?

dart/collision/CollisionGroup.cpp
```diff
@@ -0,0 +1,209 @@
...
+//==============================================================================
+void CollisionGroup::removeShapeFramesOf()
+{
+  // Do nothing
+}
+
+//==============================================================================
+void CollisionGroup::removeAllShapeFrames()
+{
+  notifyAllCollisionObjectsRemoved();
```
mkoval: I noticed that `notifyAllCollisionObjectsRemoved` and `notifyCollisionObjectRemoved` are called *before* the removal actually happens. That could lead to difficult-to-diagnose bugs if the callback manipulates the `CollisionGroup`. Is that intentional?

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,248 @@
...
+
+  /// Constructor
+  CollisionGroup(const CollisionDetectorPtr& collisionDetector);
+  // CollisionGroup also can be created from CollisionDetector::create()
+
+  /// Destructor
+  virtual ~CollisionGroup();
+
+  /// Return collision detection engine associated with this CollisionGroup
+  CollisionDetector* getCollisionDetector();
```
mkoval: `CollisionDetector` is always managed by `shared_ptr` (via `enable_shared_from_this`). Shouldn't we return the `shared_ptr` here, instead of a raw pointer?

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,248 @@
...
+  // CollisionDetector doesn't get destroyed as long as at least one
+  // CollisionGroup is alive.
+
+  /// ShapeFrames added to this CollisionGroup
+  std::vector<const dynamics::ShapeFrame*> mShapeFrames;
+
+  /// CollisionObjects associated with the added ShapeFrames
+  std::vector<std::shared_ptr<CollisionObject>> mCollisionObjects;
+  // CollisionGroup also shares the ownership of CollisionObjects across other
+  // CollisionGroups for the same reason with above.
```
mkoval: I'd prefer to replace the parallel vectors with a `map` or `unordered_map`. This has a few key advantages: (1) it is more efficient to implement the add/remove `ShapeFrame` methods, (2) it eliminates the verbose `std::find_if` calls, and (3) it eliminates the possibility of the two vectors from getting out of sync.

dart/collision/CollisionObject.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "dart/dynamics/ShapeFrame.h"
+
+namespace dart {
+namespace collision {
+
+//==============================================================================
+CollisionObject::~CollisionObject()
+{
+  // Do nothing
+}
```
mkoval: Why not `= default` this?

dart/collision/SmartPointer.h
```diff
@@ -0,0 +1,60 @@
...
+
+namespace dart {
+namespace collision {
+
+DART_COMMON_MAKE_SHARED_WEAK(CollisionDetector)
+DART_COMMON_MAKE_SHARED_WEAK(FCLCollisionDetector)
+DART_COMMON_MAKE_SHARED_WEAK(DARTCollisionDetector)
+#if HAVE_BULLET_COLLISION
+  DART_COMMON_MAKE_SHARED_WEAK(BulletCollisionDetector)
+#endif
```
mkoval: Shouldn't you have a similar block for `FCLCollisionDetector`? I believe FCL is technically an optional dependency.

dart/collision/fcl/FCLCollisionDetector.cpp
```diff
@@ -37,231 +37,1344 @@
...
+      != FCLCollisionDetector::getTypeStatic())
+  {
+      return false;
+  }
+
+  if (group2->getCollisionDetector()->getType()
+      != FCLCollisionDetector::getTypeStatic())
+  {
+      return false;
+  }
```
mkoval: I was expecting a check for
```c++
group1->getCollisionDetector() == group2->getCollisionDetector()
```
The check that is implemented is less strict because it would allow me to pass in `CollisionGroup`s from two different `FCLCollisionDetector`s. Will that work? Is this intentional?

dart/collision/fcl/FCLCollisionDetector.cpp
```diff
@@ -37,231 +37,1344 @@
...
+  if (group1->getCollisionDetector()->getType()
+      != FCLCollisionDetector::getTypeStatic())
+  {
+      return false;
+  }
+
+  if (group2->getCollisionDetector()->getType()
+      != FCLCollisionDetector::getTypeStatic())
+  {
+      return false;
```
mkoval: I believe this is an error condition. Shouldn't this print an error message?

dart/collision/CollisionGroup.cpp
jslee02: :+1: It works. This code just wasn't updated.

dart/collision/CollisionFilter.cpp
jslee02: Will update the comment.

dart/collision/CollisionObject.cpp
jslee02: :+1:

dart/collision/SmartPointer.h
jslee02: FCL is currently the default collision detector of DART so we have it as a mandatory dependency.

dart/collision/fcl/FCLCollisionDetector.cpp
jslee02: It will work for now when it wasn't intended, though. However, this is prone to error. I think this should be more strict so that all the groups should be created from this collision detector as:

```cpp
  if ((this != group1->getCollisionDetector()) 
      || (this != group2->getCollisionDetector()))
  {
      return false;
  }
```

dart/collision/fcl/FCLCollisionDetector.cpp
jslee02: :+1:

dart/collision/CollisionGroup.h
jslee02: I made this to return raw pointer because I couldn't see the use case to share the ownership of collision detector through collision group. However, there is also no compelling reason not to return `shared_ptr` here. Will change so.

dart/collision/CollisionGroup.h
jslee02: Actually, I used `map` here at first then changed to the parallel vectors to return `vector<ShapeFrame>` directly without creating temporary `vector` from the `map` for performance issue in the old version of collision group. In the current version, however, we don't need to get `vector<ShapeFrame>` so I will change to `map`.

jslee02: We still need to get `vector<CollisionObject>` from `map` by the way. 

I think we should have a consistent way of getting an iterable object from stl containers as I suggested #613. This is similar problem to getting `vector<const Object*>` from `vector<Object*>`. We create a temporary `vector<const Object*>` and return it without performance consideration in this case.

dart/collision/CollisionGroup.cpp
jslee02: The calling order is fine. The purpose of this function is to remove `CollisionObject`s from the collision detection engine before they get destroyed by being removed from this `CollisionGroup`.

I think the names are confusing. How about to change the name to `removeAllCollisionObjectsFromEngine`?

mkoval: :+1: for the rename.

dart/collision/CollisionGroup.h
mkoval: > We still need to get `vector<CollisionObject>` from map by the way.

Where is this used? Do we also need to get the `vector<const ShapeFrame*>`?

In general, unless these functions are used in performance-critical code, I suggest returning a temporary `vector` for now. We should address this wholesale in #613.

jslee02: It is used by `DARTCollisionDetector` to get the array of `CollisionObject`s from `CollisionGroup`. This is a special case only for `DARTCollisionGroup` because `DARTCollisionGroup` uses `CollisionObjects` rather than custom collision object like `[FCL/Bullet]CollisionGroup` use `fcl::CollisionObject` and `btCollisionObject`, and doesn't store the list of `CollisionObject`s in it.

I [changed](https://github.com/dartsim/dart/pull/631/commits/01c8e418e6e4107a62408de5e07046087e649dfd) `DARTCollisionGroup` to store `CollisionObject`s in it and removed the getters as they are not used anywhere.

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,243 @@
...
+
+namespace dart {
+namespace collision {
+
+class CollisionGroup
+{
+public:
+
+  friend class FCLCollisionDetector;
+  friend class BulletCollisionDetector;
```
mxgrey: I'm actually a bit concerned about the fact that ``CollisionGroup`` needs to declare each collision detector engine as a friend, because this makes it impossible for a user to implement and integrate their own collision detector without modifying the definition of ``CollisionGroup``, and that strikes me as problematic.

I understand the desire to hide engine-related functions from the casual users, but I think we need a way to do that which doesn't prevent advanced users from plugging in their own engines. If I can think of a good way to do this, I'll post an update.

dart/collision/CollisionGroup.h
```diff
@@ -0,0 +1,243 @@
...
+  /// Collision detector
+  CollisionDetectorPtr mCollisionDetector;
+  // CollisionGroup shares the ownership of CollisionDetector with other
+  // CollisionGroups created from the same CollisionDetector so that the
+  // CollisionDetector doesn't get destroyed as long as at least one
+  // CollisionGroup is alive.
+
+  /// ShapeFrames and CollisionOjbects added to this CollisionGroup
+  std::vector<std::pair<const dynamics::ShapeFrame*,
+                        CollisionObjectPtr>> mShapeFrameMap;
```
mxgrey: For the sake of performance, we could also hold onto a ``std::unordered_map<const dynamics::ShapeFrame*, size_t>`` where the ``size_t`` points to the location in ``mShapeFrameMap``. That would allow us to preserve the ordering of the ``ShapeFrame``s in the ``CollisionGroup`` while also giving us good scalability for large groups. The one catch is that we would have two member fields to keep in sync.

But it's probably not an important matter, so we could put it off until v7.1.

dart/collision/CollisionGroup.h
psigen: Would it make any sense to try to put the advanced methods into a member class that CollisionGroup creates as needed?

Something like:
```c++
class CollisionGroup
{
    [...]
    
    public CollisionGroupExtension advanced() {
        return CollisionGroupExtension(this); // has advanced methods on it, is not persistent.
    }
};

CollisionGroupPtr group;
group->advanced().adv_function();
```

You guys are the experts here, but it seems like it might be constructable in a way that it would be compiled out in `Release` mode?

dart/collision/Option.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include <cstddef>
+#include <memory>
+
+namespace dart {
+namespace collision {
+
+class CollisionFilter;
+
+struct CollisionOption
```
mxgrey: Since the name of the struct is ``CollisionOption``, I wonder if we should also name the header ``CollisionOption.h`` rather than ``Option.h``.

dart/collision/Result.h
```diff
@@ -0,0 +1,81 @@
...
+#ifndef DART_COLLISION_RESULT_H_
+#define DART_COLLISION_RESULT_H_
+
+#include <vector>
+#include "dart/collision/Contact.h"
+
+namespace dart {
+namespace collision {
+
+class CollisionResult
```
mxgrey: Same as ``CollisionOption``: I wonder if we should change the header name to ``CollisionResult.h``. Alternatively, we could change the class name to ``Result`` so that it's ``collision::Result``.

jslee02: This class and `CollisionOption` were just `Result` and `Option`, then I renamed them because `Distance[Option/Result]` will be added for distanct (or raycast) query.

One concern is if having four files for `[Collision/Distance][Option/Result]` would be too many.

mxgrey: Ohhh, I see. If there will be new classes with similar names, then it makes sense to give the header a generic name like this. :+1:

dart/collision/CollisionDetector.h
```diff
@@ -43,169 +43,179 @@
...
 
-/// \brief class CollisionDetector
-class CollisionDetector
-{
-public:
-  /// \brief Constructor
-  CollisionDetector();
+  /// Create CollisionObject
+  virtual std::unique_ptr<CollisionObject> createCollisionObject(
+      const dynamics::ShapeFrame* shapeFrame) = 0;
```
mxgrey: Is there a particular reason that we create these as a ``unique_ptr`` when it seems as though we always end up converting them into a ``shared_ptr``? Is this to make it impossible to jumble up ``claimCollisionObject(~)`` with ``createCollisionObject(~)``?

mxgrey:
I'm under the impression that ``NoneSharingCollisionObjectManager`` is supposed to be used by ``CollisionDetector`` implementations in which it is not possible to share their collision objects between groups, so a new ``CollisionObject`` instance needs to be created for each group in which a single ``ShapeFrame`` instance is being referenced. Then ``SharingCollisionObjectManager`` is used by ``CollisionDetector`` implementations which *do* allow multiple groups to share the same collision object instances. Assuming that's the case, I would recommend some name changes:

``NoneSharingCollisionObjectManager`` --> ``UnshareableCollisionObjectManager`` or ``ManagerForUnshareableCollisionObjects``

``SharingCollisionObjectManager`` --> ``ShareableCollisionObjectManager`` or ``ManagerForShareableCollisionObjects``

Mostly I think "NoneSharing" makes for an awkward class name, so I'd like to change that if we can.

dart/collision/CollisionFilter.h
```diff
@@ -35,47 +34,35 @@
...
-  dynamics::BodyNode* mBodyNode;
+struct BodyNodeCollisionFilter : CollisionFilter
+{
+  bool needCollision(const CollisionObject* object1,
+                     const CollisionObject* object2) const override;
 
-  /// \brief
-  size_t mIndex;
+  bool isAdjacentBodies(const dynamics::BodyNode* bodyNode1,
+                        const dynamics::BodyNode* bodyNode2) const;
```
mxgrey: Kind of a minor grammatical thing, but should we name the function ``areAdjacentBodies`` instead?

jslee02: :+1:

dart/collision/CollisionDetector.h
jslee02: Yes. The main motivation behind this is to make `CollisionDetector` create single `CollisionObject` per `ShapeFrame` using RAII.

`claimCollisionObject(~)` returns `std::shared_ptr<CollisionObject>`. The returning pointer is either of one created from `createCollisionObject` if it's not created for the passed in `ShapeFrame` or one already created for the `ShapeFrame`. The management of the uniquess of `CollisionObject` per `ShapeFrame` is done using `std::map<ShapeFrame*, std::weak_ptr<CollisionObject>` in `CollisionDetector` (more precisely, in `CollisionDetector::SharingCollisionObjectManager`). The `CollisionObject` should be removed from the map when it's not shared by any `CollisionObject`. However, the thing is `std::shared_ptr<CollisionObject>` will be automatically destructed when it's not referenced by any `CollisionGroup` (as the map holds `CollisionObject` as weak pointer), but `CollisionDetector` doesn't know when it's happening.

To handle this, a custom deleter of shared pointer is used where it removes `CollisionObject` from the map before it deletes `CollisionObject`. I could make `createCollisionObject` create `shared_ptr` with the custom deleter by passing it from `CollisionDetector` to each concrete collision detector, but also wondered if it's a good practice. To me, it seemed the management of collision object management should be totally the responsibility of `CollisionDetector` and the each collision detector just should create the concrete collision object (e.g., `FCLCollisionObject`).

I'm open to hearing better solution here.

jslee02:
> Assuming that's the case, I would recommend some name changes:

Yes, exactly. I like the new names. To me, the pair of `ManagerFor[Uns/S]hareableCollisionObjects` sounds clearer than the other pair since [Uns/S]harable seems to describe managers rather than collision objects.

dart/collision/CollisionGroup.h
jslee02: I think this is a good point. One simple solution I can come up with now is declaring the protected function of `CollisionGroup` in each concrete collision group class (e.g., `FCLCollisionGroup`). Then `FCLCollisionDetector` could be able to call the function by downcasting collision group to `FCLCollisionGroup`.

dart/collision/CollisionGroup.h
jslee02: Yeah, that approach sounds good. We could implement something similar to [`nlohmann/fifo_map`](https://github.com/nlohmann/fifo_map) for the sync guarantee.

I would like to do this as a separate PR since this PR blocks others.

mxgrey:
Overall, I think this pull request introduces a much needed renovation of the ``collision`` namespace and introduces many wonderful features. I'm happy with its current state and think it's ready to be merged, unless anyone has an objection.

I do have a few comments for moving forward. These certainly don't need to be addressed in this pull request, but it's something to think about (we can open issues for them later):

1. When a user adds a ``BodyNode`` or a ``Skeleton`` to a ``CollisionGroup``, that ``CollisionGroup`` should update automatically whenever a new ``ShapeNode`` is added or removed from the object. I can think of two ways to accommodate this:
  1. Version tracking. Whenever the ``CollisionGroup`` spots a change in a version number, it can check whether the layout of ``ShapeFrames has changed.
  2. Signals and slots. Have ``BodyNode`` and ``Skeleton`` both offer a signal that gets triggered when a ``ShapeNode`` is added, and another one that is triggered when the ``ShapeNode`` is removed. We could even generalize it so that users can register their own ``ShapeFrameCreated`` signals to a ``CollisionGroup``. I think this approach would be preferable over (i), especially given its generalizability.

2. It looks like it's very easy for a ``CollisionGroup`` to be unknowingly invalidated since they hold raw pointers to the ``ShapeFrame``s that they reference. It would be good if they could do automatic cleanup when a ``ShapeFrame`` that they refer to gets destroyed. Of course this is complicated because ``SimpleFrame``s get managed by ``std::shared_ptr`` whereas ``ShapeNode``s get managed by ``ShapeNodePtr`` which are not compatible smart pointer types. An easy (but not thread-safe) solution would be to use the fact that both types have a ``common::Subject`` base, so they can give out a notice when they destruct, but we'd probably want a cleaner solution than this in the long run.

mkoval:
> When a user adds a `BodyNode` or a `Skeleton` to a `CollisionGroup`, that `CollisionGroup` should update automatically whenever a new `ShapeNode` is added or removed from the object.

I am conflicted about this: this adds *a lot* of complexity and may not always have the desired effect. I'd prefer to keep `CollisionGroup` as a "dumb" collection of `ShapeFrame`s and implement automatic management in a helper (or wrapper) class. I am fine with encouraging people to use this "smart" group by default and leaving the "dumb" group for advanced users.

Of the two options, I *slightly* prefer option (i) for two reasons. First, creating a CollisionObject may be expensive. It is undesirable to constantly create and delete CollisionObjects when modifying the kinematic tree. Second, signals and slots makes it much harder to reason about control flow.

> It looks like it's very easy for a `CollisionGroup` to be unknowingly invalidated since they hold raw pointers to the `ShapeFrame`s that they reference. It would be good if they could do automatic cleanup when a `ShapeFrame` that they refer to gets destroyed. Of course this is complicated because `SimpleFrame`s get managed by `std::shared_ptr` whereas `ShapeNode`s get managed by `ShapeNodePtr` which are not compatible smart pointer types.

I agree. This is a problem far beyond the `CollisionGroup` API: I have the same challenge in a lot of the code that I write that uses `Frame`s. In most cases, I end up using `Node` to dodge the question entirely. 

What is the motivation/use case for having `Frame`, `ShapeFrame`, etc outside of a `Skeleton`? If the concern is overhead, perhaps we can add a *very* lightweight base class to `Skeleton` that is solely responsible for managing ownership. This would let us use `NodePtr` universally.

mxgrey:
> I'd prefer to keep CollisionGroup as a "dumb" collection of ShapeFrames and implement automatic management in a helper (or wrapper) class.

This sounds like a good plan to me.

> It is undesirable to constantly create and delete CollisionObjects when modifying the kinematic tree. Second, signals and slots makes it much harder to reason about control flow.

These are very good points. If we're going to implement it with wrapper classes, then we could offer a wrapper for each approach if we can't make a decision between the two. But your points do make me lean more towards (i) now.

> What is the motivation/use case for having Frame, ShapeFrame, etc outside of a Skeleton? 

What it boils down to is the fact that ``Frame`` is an interface class which is agnostic to any concept of a ``Skeleton``. Requiring every derivation of ``Frame`` to be owned and managed by a ``Skeleton`` seemed pointlessly strict to me.

The ``SimpleFrame`` implementation has an important advantage over any feasible ``Node`` type: The parent frame of a ``SimpleFrame`` can be changed to *any* ``Frame`` at *any* time. In contrast, a ``Node`` must currently descend from a ``BodyNode`` and that parent ``BodyNode`` can't change. It's also plausible that users may want to implement their own ``Frame``-based kinematic structures independent of the ``Skeleton`` concept, and I think they should have the freedom to do so.

I'm open to the idea of being able to move ``Node``s between parent ``BodyNode``s, but we would have to reinvent the ``NodePtr`` infrastructure, and I'm having trouble thinking of a better design at the moment.

In any case, we can move this conversation to an issue if everyone is okay with merging this pull request.

PR #633, opened 2016-03-17
Added get/setLCPSolver functions to ConstraintSolver

mkoval:
There is currently no way to use anything other than the default LCP solver in `ConstraintSolver`. This pull request adds functions that parallel the `get/setCollisionDetector` for the `LCPSolver`.

### DIFF
--- a/dart/constraint/ConstraintSolver.h
+++ b/dart/constraint/ConstraintSolver.h
@@ -110,7 +110,7 @@ public:
 
   /// Set collision detector
   void setCollisionDetector(
-    std::unique_ptr<collision::CollisionDetector>&& _collisionDetector);
+    std::unique_ptr<collision::CollisionDetector> _collisionDetector);
 
   /// Get collision detector
   collision::CollisionDetector* getCollisionDetector() const;
--- a/dart/constraint/ConstraintSolver.h
+++ b/dart/constraint/ConstraintSolver.h
@@ -115,6 +115,12 @@ public:
   /// Get collision detector
   collision::CollisionDetector* getCollisionDetector() const;
 
+  /// Set LCP solver
+  void setLCPSolver(std::unique_ptr<LCPSolver> _lcpSolver);
+
+  /// Get LCP solver
+  LCPSolver* getLCPSolver() const;
+
   /// Solve constraint impulses and apply them to the skeletons
   void solve();
 
--- a/dart/constraint/ConstraintSolver.cpp
+++ b/dart/constraint/ConstraintSolver.cpp
@@ -243,7 +243,7 @@ void ConstraintSolver::setCollisionDetector(
 
 //==============================================================================
 void ConstraintSolver::setCollisionDetector(
-  std::unique_ptr<collision::CollisionDetector>&& _collisionDetector)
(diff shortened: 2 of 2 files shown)


dart/constraint/ConstraintSolver.h
```diff
@@ -110,11 +110,17 @@ class ConstraintSolver
...
   /// Set collision detector
   void setCollisionDetector(
-    std::unique_ptr<collision::CollisionDetector>&& _collisionDetector);
+    std::unique_ptr<collision::CollisionDetector> _collisionDetector);
 
   /// Get collision detector
   collision::CollisionDetector* getCollisionDetector() const;
 
+  /// Set LCP solver
+  void setLCPSolver(std::unique_ptr<LCPSolver> _lcpSolver);
```
jslee02: Passing const reference like `std::shared_ptr<T>` case?

mxgrey: That shouldn't be done for ``unique_ptr`` because we want the ownership of the solver to be passed into the ``ConstraintSolver``. Passing by reference wouldn't allow that.

``unique_ptr`` is really weird when it comes to function arguments.

jslee02: I see, thanks for the clarification. Good to know!

Also, this pull request looks good to merge.

mkoval:
I'm not really sure whether this should be managed by `smart_ptr` or `unique_ptr`. I changed `CollisionDetector` from a raw pointer to a `unique_ptr` in #616 to maintain backwards compatibility and used the same convention here for consistency.

There are no backwards compatibility concerns with `LCPSolver`, since this was never publicly accessible before this pull request, so I am open to either option. @mxgrey @jslee02 Which do you prefer?

mxgrey:
I think ``unique_ptr`` makes the most sense since sharing an ``LCPSolver`` could entail some serious multi-threading issues depending on the implementation of the ``LCPSolver``.

Since ``unique_ptr``s are kind of weird, I would suggest also having a function like

```
template <typename T, typename... Args>
void setLCPSolver(Args&&... args)
{
  setLCPSolver(std::unique_ptr<T>(new T(args...)));
}
```

For the user's convenience. This way an ``LCPSolver`` can be set like:

```
constraintSolver->setLCPSolver<DantzigLCPSolver>(timestep);
```

Instead of requiring the user to do confusing things with ``std::move()``.

mkoval:
I am not a huge fan of that paradigm for two reasons. First, the function signature is inscrutable to people that are unfamiliar with template argument packs. Second, we have to define this type of overload for every function that accepts an `std::unique_ptr`.

I would prefer to provide a `make_unique` function in the `common` or `utils` namespace. This [exists in C++14](https://en.wikipedia.org/wiki/C%2B%2B14#Smaller_library_features) and was [omitted from C++11 as "partly an oversight"](http://stackoverflow.com/a/12580468/111426).

mxgrey:
That sounds like a good alternative to me.

jslee02:
`make_unique` sounds good. I would prefer to have it in `common` like [`StlHelpers.h`](https://github.com/dartsim/dart/blob/collision_group/dart/common/StlHelpers.h).

PR #634, opened 2016-03-17
Fixed DantzigLCPSolver when getTotalDimension = 0

mkoval:
The code previously checked for `getNumConstraints() == 0`, but not `getTotalDimension() == 0`. This caused the LCP solver to access invalid memory if constraints were present, but all constraints had `getDimension() == 0`. I ran into this using a `ServoJointConstraint` in a completely empty environment.

Thanks @jslee02 for helping me debug this.

### DIFF
--- a/dart/constraint/DantzigLCPSolver.cpp
+++ b/dart/constraint/DantzigLCPSolver.cpp
@@ -69,7 +69,7 @@ void DantzigLCPSolver::solve(ConstrainedGroup* _group)
   size_t n = _group->getTotalDimension();
 
   // If there is no constraint, then just return.
-  if (n == 0)
+  if (0u == n)
     return;
 
   int nSkip = dPAD(n);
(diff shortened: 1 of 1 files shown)


dart/constraint/DantzigLCPSolver.cpp
```diff
@@ -63,13 +63,15 @@ DantzigLCPSolver::~DantzigLCPSolver()
...
-  size_t numConstraints = _group->getNumConstraints();
-  if (numConstraints == 0)
-    return;
 
   // Build LCP terms by aggregating them from constraints
+  size_t numConstraints = _group->getNumConstraints();
   size_t n = _group->getTotalDimension();
+
+  // If there is no constraint, then just return.
+  if (n == 0)
```
jslee02: nitpick: some compiler would complain to comparing signed and unsigned. `if (0u == n)` would help. also, @mxgrey would like to use [`Yoda conditions`](https://en.wikipedia.org/wiki/Yoda_conditions). :wink:

mxgrey:
Should it not be ``if (0u == n || 0u == numConstraints)``?

mkoval:
I don't see why we have to check `numConstraints`. Shouldn't `getTotalDimension()` return zero if the `ConstraintGroup` is empty?

mxgrey:
You're right, zero dimensions also implies zero constraints. I assumed that the "dimension" referred to the dimension of the problem, not the sum of constraint dimensions. I should have checked the function definition.

:+1:

PR #635, opened 2016-03-17
Use FCLCollisionDetector by default if version >= 0.4 is available

mkoval:
DART currently uses `FCLMeshCollisionDetector` by default because FCL < 0.4 returns incorrect contact points for some primitive-primitive collision checks. This pull request changes the behavior in two ways:

1. Prints a warning about the bug if CMake detects an old version of FCL.
2. Uses `FCLCollisionDetector`, instead of `FCLMeshCollisionDetector`, if a new version if FCL is found.


mxgrey:
I take it ``DARTCollisionDetector.h`` and ``BulletCollisionDetector.h`` weren't doing anything useful in there since the default is to use FCL.

The changes look good to me :+1:

PR #636, opened 2016-03-17
Removed version check for Bullet.

### LINKED ISSUE #625: Decrease minimum required bullet version from 2.82 to 2.81
I set the minimum required bullet version 2.82 in the first place without any investigation. This prevents us from using DART with bullet on trusty (bullet version of trusty is 2.81) unless we install higher bullet version manually. [According to the change log](https://github.com/bulletphysics/bullet3/blob/master/docs/BulletQuickstart.pdf), it seems there is no change in collision module of bullet in 2.82 so we would be able to use 2.81 without any trouble.

mkoval:
This fixes #625 by removing the test for a Bullet version. This is important because Ubuntu 14.04 ships with Bullet 2.81, which fails the version check for Bullet 2.82 in the current DART `master`.

I would prefer to remove the check entirely, instead of simply lowering to Bullet 2.81, because we have no evidence to believe that the interface will not work on earlier versions of Bullet. We can re-introduce a version check if we find a compelling reason to do so (e.g. DART fails to build against very old versions).

I also removed the logic that uses `pkg-config` on `UNIX` systems. The `find_package` command works fine for me on Linux and was unable to find a justification for these gymnastics in the `git` history.. Does anyone know why this check was introduced?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -360,14 +360,8 @@ else()
 endif()
 
 # BulletCollision
-if(UNIX)
-  pkg_check_modules(BULLET bullet>=2.82 QUIET)
-  if(NOT BULLET_FOUND)
-     pkg_check_modules(BULLET bullet2.82>=2.82 QUIET)
-  endif()
-else()
-  find_package(Bullet COMPONENTS BulletMath BulletCollision QUIET)
-endif()
+find_package(Bullet COMPONENTS BulletMath BulletCollision QUIET)
+
 if(BULLET_FOUND)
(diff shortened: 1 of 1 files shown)


mxgrey:
It may have been written during a time when cmake package information wasn't readily available for Bullet on ``UNIX`` systems. I've seen that kind of issue for OpenSceneGraph on OSX.

If these changes work for the platforms that we support, then it looks good to me :+1:

jslee02:
Right, FindBullet was not introduced before CMake 2.8. This looks good to me too. :+1:

PR #637, opened 2016-03-17
Fixed find_package(DART) on optimizer components

mkoval:
Merging #630 uncovered some latent issues in DART's CMake configuration. The optional `dart::optimizer` components (bindings for `nlopt`, `ipopt`, and `snopt`) were not being added to the `DARTTargets` group and, thus, were not included in `DARTTargets.cmake`. This caused finding those components, e.g. via `find_package(DART COMPONENTS nlopt)`, to fail even if they were built.

I thought the same issue would be present with `FCLCollisionDetector` and `BulletCollisionDetector`, but it appears that all of the available collision detectors are being linked into `dart.so`. There is dead code in those `CMakeLists.txt` file that indicates this was not always the case.

I found this surprising because merging all of these into one library seems undesirable. Was this an intentional change?

### DIFF
--- a/dart/optimizer/ipopt/CMakeLists.txt
+++ b/dart/optimizer/ipopt/CMakeLists.txt
@@ -25,4 +25,7 @@ install(
   DESTINATION include/dart/optimizer/ipopt
   COMPONENT headers
 )
-install(TARGETS dart-optimizer-ipopt EXPORT DARTIpoptTargets DESTINATION lib)
+install(TARGETS dart-optimizer-ipopt
+  EXPORT DARTTargets
+  DESTINATION lib
+)
--- a/dart/optimizer/nlopt/CMakeLists.txt
+++ b/dart/optimizer/nlopt/CMakeLists.txt
@@ -25,4 +25,7 @@ install(
   DESTINATION include/dart/optimizer/nlopt
   COMPONENT headers
 )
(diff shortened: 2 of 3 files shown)


jslee02:
We intended to build DART as a single library just for simplicity at the moment. However, as we plan to build DART component wise, collision should follow the same convention of optimizer components.

What I'm not sure is if this should be done from DART 6.0 or KIDO 7.0.

Besides on `collision`, this PR looks good to merge.

mxgrey:
I think we should break things out component-wise with DART 6.0.

I believe pretty strongly that the **only** difference between DART 6.0 and KIDO 7.0 should be naming, while everything about their build systems, APIs, and features should be identical (except obviously that any usage of "DART" would be replaced with "KIDO"). I think that would make for the smoothest possible transition for users and developers alike.

mkoval:
I am fine with putting off the other component changes to a future version.

However, I would still like to merge this PR. It is currently impossible for me to use `find_package(DART COMPONENTS dart-optimizer-nlopt)` in my code that depends on DART. This worked in the previous version, so it is a clear regression.

jslee02:
> I believe pretty strongly that the only difference between DART 6.0 and KIDO 7.0 should be naming, while everything about their build systems, APIs, and features should be identical (except obviously that any usage of "DART" would be replaced with "KIDO"). I think that would make for the smoothest possible transition for users and developers alike.

I like this plan. We can do this in separate pull requests.

:+1: to merge this.

PR #639, opened 2016-03-20
Add make_unique function omitted from C++11

jslee02:
This pull request adds `make_unique` function as suggested in #633 .

### DIFF
--- a/dart/common/StlHelpers.h
+++ b/dart/common/StlHelpers.h
@@ -39,26 +39,21 @@
 
 #include <cassert>
 #include <cstddef>
+#include <memory>
 #include <vector>
 
-namespace std {
+namespace dart {
+namespace common {
 
 //==============================================================================
 template<typename T, typename... Args>
-unique_ptr<T> make_unique(Args&&... args)
+std::unique_ptr<T> make_unique(Args&&... args)
 {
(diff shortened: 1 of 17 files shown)


dart/common/StlHelpers.h
```diff
@@ -41,9 +41,26 @@
...
+template<typename T, typename... Args>
+unique_ptr<T> make_unique(Args&&... args)
+{
+  return unique_ptr<T>(new T(forward<Args>(args)...));
+}
+// TODO(JS): This is a stopgap solution as it was omitted from C++11 as "partly
+// an oversight". This can be replaced by std::make_unique<T> of the standard
+// library when we migrate to using C++14.
+
+} // namespace std
```
mkoval: It is undefined to add declarations or definitions to the `std` namespace:

> `[C++11: 17.6.4.2.1/1]:` The behavior of a C++ program is undefined if it adds declarations or definitions to namespace `std` or to a namespace within namespace `std` unless otherwise specified. A program may add a template specialization for any standard library template to namespace `std` only if the declaration depends on a user-defined type and the specialization meets the standard library requirements for the original template and is not explicitly prohibited.

jslee02: This makes sense. Thanks!

mxgrey:
These changes look good to me :+1:

PR #640, opened 2016-03-20
Fix FCL version check logic in the main CMakeLists.txt

jslee02:
This pull request fixes the FCL version check logic that was introduced by #635.


CMakeLists.txt
```diff
@@ -160,10 +158,14 @@ else()
...
       )
-      set(DART_USE_FCLMESHCOLLISIONDETECTOR 1)
+      set(DART_USE_FCLMESHCOLLISIONDETECTOR TRUE)
     else()
-      set(DART_USE_FCLMESHCOLLISIONDETECTOR 0)
+      set(DART_USE_FCLMESHCOLLISIONDETECTOR FALSE)
     endif()
+  endif()
+else()
+  if(FCL_VERSION)
```
mkoval: This check will never be satisfied. According to the [`find_package` documentation](https://cmake.org/cmake/help/v3.0/command/find_package.html), the output variables are only set if a matching version is found (emphasis mine):

> These variables are checked by the `find_package` command to determine whether the configuration file provides an acceptable version. They are not available after the find_package call returns. **If the version is acceptable** the following variables are set: [...]

jslee02: :+1:

PR #641, opened 2016-03-20
Use cmakedefine01 instead of cmakedefine

jslee02:
`cmakedefine` defines only when the cmake variable is set to true or 1 in cmake script, whereas `cmakedefine01` always defines either `#define VAR 1` or `#define VAR 0` depending on the variable. `cmakedefine` will go along with `#ifdef VAR`, and `cmakedefine01` will go along with `#if VAR`.

This might be arguable, but `#if VAR` seems to be more intuitive to use to me. I would like to know other's thoughts.


mkoval:
:+1: Makes sense to me.

psigen:
If dealing with boolean situations, I tend to prefer `#if defined(VAR)` to `#ifdef VAR` or `#if VAR` for the following reasons:

* `#if defined(VAR)` is categorically more flexible than `#ifdef VAR` because it allows compound conditionals
* `#if VAR` *requires* that `VAR` have some value if `-Wundef` is in use.  This means that the actual compiler flags become polluted with a huge list like ```-DBUILD_TYPE_DEBUG=0 -DBUILD_TYPE_RELEASE=0 -DBUILD_TYPE_RELEASE=0 -DBUILD_TYPE_RELWITHDEBINFO=1``` because it has to include every option that _isn't_ in use.  I generally don't like the idea of having combinatorial amounts of command line flags for each exclusive option.

Willing to hear other opinions though.

mxgrey:
I'm most in favor of @psigen 's proposal. Explicitly defining every command line variable could result in some hideous output if we ever need to investigate build issues.

jslee02:
> If dealing with boolean situations, I tend to prefer #if defined(VAR) to #ifdef VAR or #if VAR for the following reasons:

To clarify, do you mean to prefer to use `#cmakedefine VAR 1` with `#if defined(VAR)`?



> This means that the actual compiler flags become polluted with a huge list like -DBUILD_TYPE_DEBUG=0 -DBUILD_TYPE_RELEASE=0 -DBUILD_TYPE_RELEASE=0 -DBUILD_TYPE_RELWITHDEBINFO=1 because it has to include every option that isn't in use.

Please correct me if something is wrong in the following. AFAIK, [they are cmake variables](https://github.com/dartsim/dart/blob/master/CMakeLists.txt#L98) rather than compiler flags. Those cmake variables are determined in cmake time based on `CMAKE_BUILD_TYPE` passed in with cmake command. `CMAKE_BUILD_TYPE` also can be omitted. In this case, [it will be set to `Release` by default](https://github.com/dartsim/dart/blob/master/CMakeLists.txt#L88). 

The actual preprocessors are then configured in `Config.h.in` in cmake time based on the cmake variables. The only choice to be made by us is which one is preferred among:

```cpp
/* #undef BUILD_TYPE_DEBUG */
#define BUILD_TYPE_RELEASE 1
/* #undef BUILD_TYPE_RELWITHDEBINFO */
/* #undef BUILD_TYPE_MINSIZEREL */

#define HAVE_NLOPT 1
#define HAVE_IPOPT 1
/* #undef HAVE_SNOPT */
#define HAVE_BULLET_COLLISION 1

#define DART_USE_FCLMESHCOLLISIONDETECTOR 0
```
and
```cpp
#define BUILD_TYPE_DEBUG 0
#define BUILD_TYPE_RELEASE 1
#define BUILD_TYPE_RELWITHDEBINFO 0
#define BUILD_TYPE_MINSIZEREL 0

#define HAVE_NLOPT 1
#define HAVE_IPOPT 1
#define HAVE_SNOPT 0
#define HAVE_BULLET_COLLISION 1

#define DART_USE_FCLMESHCOLLISIONDETECTOR 0
```

All these will be automatically configured by cmake without polluting the compiler flags.

jslee02:
As this doesn't pollute the compiler flags, looks good to merge?

psigen:
I think this is fine as these are strictly going into a header directly.

@mkoval and I discussed some use cases where someone might want to `undefine` something when compiling client code, but we concluded that it's a pretty rare case and we couldn't think of valid use-cases related to these definitions.

jslee02:
Thanks! Merging now.

PR #646, opened 2016-03-27
Suppress unused variables warnings in release mode

mkoval:
Building DART in release mode disables assertions. There are a few parameters in DART that are only used in assertions, so this generates `-Wunused-parameter` warnings. This pull request adds a `DART_UNUSED` that suppresses those warnings. See [this StackOverflow answer](http://stackoverflow.com/a/777359/111426) for more explanation.

### DIFF
--- a/dart/common/StlHelpers.h
+++ b/dart/common/StlHelpers.h
@@ -42,6 +42,10 @@
 #include <memory>
 #include <vector>
 
+// Macro to suppress -Wunused-parameter and -Wunused-variable warnings in
+// release mode when a variable is only used in assertions.
+#define DART_UNUSED(x) do { (void)(x); } while (0)
+
 namespace dart {
 namespace common {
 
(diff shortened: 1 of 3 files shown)


jslee02:
:+1:

psigen:
It seems a bit strange to require arguments that are only used in assertions.

Are these parameters required as part of some standardized spec? Is there a reason we force them to be passed when they are apparently not strictly necessary for the function?

jslee02:
This is a reasonable concern. 

In `DartTNLP` case, we should follow the same function prototype as it overrides `Ipopt::TNLP` functions. We could comment out, but I would like to leave `_m` uncommented to be used in the assertion for debugging purpose (checking the dimensionality).

We could remove `_numDir` for `ODELCPSolver` case. One thing we should note is that this class isn't used anywhere (at least in our code base) so I would be fine to remove `_numDir`. Personally, I'm a bit inclined to leave it until we decide whether remove this unused class `ODELCPSolver` or improve to use somewhere. I think at some point we need to refactor the constraint solving and can figure out what we will do with `ODELCPSolver` at that moment.

PR #648, opened 2016-03-28
Needed to change #ifdef to #if

mxgrey:
It seems like this was overlooked from pull request #641. This change was necessary for me to successfully compile.


jslee02:
:+1: Good catch!

Issue #645, opened 2016-03-27
getMultiDofJointAddon() is not defined


mkoval:
Trying to call `getMultiDofJointAddon()` on a `MultiDofJoint<T>` results in the following error message:
```
In file included from /home/mkoval/dartpy-ws/devel/include/dart/common/SpecializedForAddon.h:193:
/home/mkoval/dartpy-ws/devel/include/dart/common/detail/SpecializedForAddon.h:260:26: error: no matching constructor for initialization of 'dart::dynamics::detail::MultiDofJointAddon<3>'
  SpecAddon* addon = new SpecAddon(this, std::forward<Args>(args)...);
                         ^         ~~~~
/home/mkoval/dartpy-ws/devel/include/dart/common/detail/SpecializedForAddon.h:105:10: note: in instantiation of function template specialization 'dart::common::SpecializedForAddon<dart::dynamics::detail::MultiDofJointAddon<3> >::_create<>' requested here
  return _create(type<T>(), std::forward<Args>(args)...);
         ^
/home/mkoval/dartpy-ws/devel/include/dart/dynamics/MultiDofJoint.h:74:3: note: in instantiation of function template specialization 'dart::common::SpecializedForAddon<dart::dynamics::detail::MultiDofJointAddon<3> >::create<dart::dynamics::detail::MultiDofJointAddon<3>>' requested here
  DART_BAKE_SPECIALIZED_ADDON_IRREGULAR( typename MultiDofJoint<DOF>::Addon, MultiDofJointAddon )
  ^
/home/mkoval/dartpy-ws/devel/include/dart/common/detail/AddonManager.h:201:27: note: expanded from macro 'DART_BAKE_SPECIALIZED_ADDON_IRREGULAR'
    return this->template create<TypeName>(std::forward<Args>(args)...); \
                          ^
/home/mkoval/dartpy-ws/devel/include/dart/dynamics/MultiDofJoint.h:74:3: note: in instantiation of function template specialization 'dart::dynamics::MultiDofJoint<3>::createMultiDofJointAddon<>' requested here
  DART_BAKE_SPECIALIZED_ADDON_IRREGULAR( typename MultiDofJoint<DOF>::Addon, MultiDofJointAddon )
  ^
/home/mkoval/dartpy-ws/devel/include/dart/common/detail/AddonManager.h:183:14: note: expanded from macro 'DART_BAKE_SPECIALIZED_ADDON_IRREGULAR'
      return create ## AddonName();                                      \
             ^
[...]

PR #649, opened 2016-03-28
Make a default argument for MultiDofJointAddon

mxgrey:
This should fix issue #645


jslee02:
:+1:

jslee02:
[Clang complains](https://travis-ci.org/dartsim/dart/jobs/119021878#L1165) that the default argument should be added to the function definition. Surprised gcc doesn't complain this.

mxgrey:
Yeah, I was surprised that GCC allowed it when I realized that it was misplaced. I've pushed a fixed version.

PR #651, opened 2016-03-29
Rename osgDart to dart::gui::osg

jslee02:
This is the first round for reorganizing DART packages with components as discussed in #477.

This pull request includes #631. See [this](https://github.com/dartsim/dart/compare/collision_group...packaging/osgdart) to see only the changes made after #631.


mxgrey:
I've moved OpenSceneGraph to be under the banner of ``Optional Dependencies``. I've also added a flag (``DART_BUILD_GUI_OSG``) that will skip over the ``gui-osg`` library if the flag is false.

There were also a few spots where ``kido`` was being used, even though this is still ``dart`` source code, so I changed those back to ``dart`` for now.

If those changes seem reasonable, I think this is ready to merge.

jslee02:
Looks good to me. Let's merge once this passes the ci tests.

jslee02:
It seems the coverage rate has dropped as we added `gui::osg` into the testing files while without tests for it. I don't know what should be tested for code for gui though.

Also, I realized that `examples` for `gui::osg` are under the source directory unlike other app (this should be renamed). We could move `examples` under `apps` directory. Maybe as an separate PR?

mxgrey:
> We could move examples under apps directory. Maybe as an separate PR?

It's a tricky decision. The nice thing about them being nested is that they will be conditionally compiled. They critically depend on OpenSceneGraph, and being nested the way they are makes it trivial for them to be ignored whenever OpenSceneGraph is unavailable.

I suppose we could do this: rename ``apps`` to ``examples`` and then put the ``gui-osg`` examples into a folder called ``examples/osg``. Then there would be a file ``examples/osg/CMakeLists.txt`` that would conditionally compile each OSG example based on whether or not OSG is available.

> I don't know what should be tested for code for gui though.

I genuinely don't know how to design meaningful unit tests for GUIs, so I don't know what to do about this either.

jslee02:
> I suppose we could do this: rename apps to examples and then put the gui-osg examples into a folder called examples/osg. Then there would be a file examples/osg/CMakeLists.txt that would conditionally compile each OSG example based on whether or not OSG is available.

The plan sounds good. 

Speaking of which, we could apply this for tests. In this case, I would prefer to have subdirectories per module (while keeping the core tests where they are) since we would add tests for modules and they would need similar conditional compilation too.

mxgrey:
> ... we could apply this for tests ... I would prefer to have subdirectories per module 

That sounds good to me as well.

Issue #589, opened 2016-01-20
Merge renderer namespace into gui namespace


jslee02:
Please refer to #588 for details.

jslee02:
Done by #652.

PR #652, opened 2016-03-29
Reorganizing DART library composition

jslee02:
Similarly to #587, this RP changes the library composition of DART from {dart-core, dart} to {dart, optional components depending on the 3rd party library dependencies).

This pull request includes #651. See [this](//github.com/dartsim/dart/compare/packaging/osgdart...packaging/components) to see only the changes made after #651.

Componentizing `collision` is saved for later until we figure out how to handle dynamic library linking of `dart` to the components. For example, `dart` should be compiled without `dart-collision-bullet`, but also `dart` should be able to load `dart-collision-bullet` in runtime to use the bullet collision detection in the constraint solver.


costashatz:
How far are you from merging this branch?

Also, you should remove the opengl dependency since as far as I have seen you removed the draw functions from dynamics..

Thanks for all the work..

jslee02:
> How far are you from merging this branch?

Probably this PR will be merged today.

dart/gui/CMakeLists.txt
```diff
@@ -2,19 +2,20 @@
...
   add_subdirectory(osg)
 endif(HAVE_OPENSCENEGRAPH)
 
-set(dart_gui_hdrs ${dart_gui_hdrs} ${hdrs})
-set(dart_gui_srcs ${dart_gui_srcs} ${srcs})
+set(dart_gui_hdrs ${dart_gui_hdrs} ${hdrs} ${dart_renderer_hdrs})
+set(dart_gui_srcs ${dart_gui_srcs} ${srcs} ${dart_renderer_srcs})
 
 # Library
 dart_add_library(dart-gui ${srcs} ${hdrs})
```
costashatz: The errors come from the fact that we need to add the `dart_renderer_hdrs` and `dart_renderer_srcs` in the `dart_add_library` command too..

jslee02:
In the recent commits:

* All the dependencies except for the core library (`dart`) were changed to optional so that cmake doesn't spit errors when they're missing.
* `renderer` namespace was merged into `gui` (see #589)
*  The changes of #635 was reverted since FCL is not fully functional for primitive shapes yet.
* The debian files were updated -- need tests

I believe this PR is ready to merge once all the ci tests are passed and if no one has objection.

dart/gui/CMakeLists.txt
jslee02: You're correct. This was resolved as `renderer` was merged into `gui` [as planned](https://github.com/dartsim/dart/issues/589).

jslee02:
@mxgrey Could you take a look at this and merge if it looks good to you?

I merged @costashatz's pull request and added more changes for setting PointMass's collision status.

mxgrey:
Can we make ``planning`` a component and make ``flann`` an optional dependency? I'll attempt to do this myself and make a pull request targeted at this branch to make it easy to review my changes.

jslee02:
> I don't know if that variable is important or not, but perhaps it should be updated.

It's used for generating `dart.pc`, which is a pkg-config file. In most case, DART will be found by `DARTConfig.cmake` so it wouldn't be used that much. But, yeah, it should be updated. However, I honestly don't know how to set it up for DART components.

dart/dynamics/BodyNode.h
```diff
@@ -1035,7 +1035,7 @@ class BodyNode :
   UniqueProperties mBodyP;
 
   /// Whether the node is currently in collision with another node.
-  DEPRECATED(6.0)
+  /// \deprecated DEPRECATED(6.0) See #670 for more detail.
```
jslee02: Removed the used of `DEPRECATED(6.0)` for member variable because there is no way to suppress warnings for this. Instead, I put `DEPRECATED(6.0)` into comment so that we don't miss this when searching deprecated API using this keyword.

mxgrey:
I'm no build system expert, but everything here looks reasonable to me. If @jslee02 is finished with this, I see no problem with merging. :+1:

Issue #568, opened 2015-12-08
Proposal: Use .hpp for headers when switching to the new package name


mxgrey:
In various discussions, I've heard interest in the idea of having C-bindings for DART since this would enable easier interfacing with other programming languages, and therefore easier integration into other frameworks. If someone were to implement C-bindings, the most natural file organization would be to have C API headers use ``.h`` while C++ API headers use ``.hpp``, but in our current file organization all the C++ headers are already using ``.h``.

Whether or not anyone is currently planning to implement C-bindings, I think it would be good to use the ``.hpp`` convention to allow for a clear delineation between the existing C++ API and a C API in case one is ever created. Since the package will be undergoing a highly interruptive name change in the near future anyway, I think that would be the best possible time to make a change like this.

jslee02:
:+1: It sounds reasonable.

jslee02:
Addressed by #693

Issue #643, opened 2016-03-23
Proposal: Rename the class Addon to Aspect


mxgrey:
First, I want to apologize for revisiting an issue as contentious as naming. I believe I have some very compelling reasons for recommending this change, so I'm going to try to lay those reasons out as clearly as I can.

### What we currently have (on master branch):

We're using a ``State`` and ``Properties`` pattern to represent the parameters of kinematic and dynamic structures (as well as all other concrete objects in DART).

A ``State`` is a set of parameters which is expected to change arbitrarily often. Typically these are variables which are functions of time, although if a user creates a custom stateful object, they are given the liberty to decide which of their parameters fall under the umbrella of ``State``.

``Properties`` are parameters which are not expected to change very often. Ideally, a change to the ``Properties`` of an object will also increment the "version number" of that object, as well as the version number of any objects to which it is attached. Typically the data within a set of ``Properties`` are constant parameters pertaining to the equations of motion of an object, although if a user creates a custom object with custom properties, they are given the liberty to decide which of their parameters fall under the umbrella of ``Properties``.

The base objects of DART have their basic ``Properties`` data embedded in the class definition itself. For example, ``BodyNode::Properties`` has information about its name, inertia, and collidability. ``Skeleton::Properties`` has information about its name, timestep, and mobility. ``Joint::Properties`` has information about name, relative transforms, and actuator type. This information is stored internally in each of these classes as a member variable.

However, derived classes need to be able to extend this ``Properties`` information, and they need to add their own ``State`` information as well. They need to do it in a way that is directly accessible to the base class as well. Luckily, we've already invented the ``Addon`` class which was designed to allow users to arbitrarily attach custom classes that have their own ``State`` and ``Properties``. We can use this machinery to trivially make the ``States`` and ``Properties`` of derived classes accessible to their base class. Problem solved, right??

### Why is this a problem??

[...]

PR #653, opened 2016-03-29
Changing header extension from .h to .hpp

jslee02:
This PR changes head extension from `.h` to `.hpp` as proposed in #568.

Exceptions: ode files and lodepng.h

This pull request includes #652. See [this](https://github.com/dartsim/dart/compare/packaging/components...packaging/h_to_hpp) to see only the changes made after #652.


mxgrey:
In a day or two I'll have a pull request coming in related to #643 where we end up with some new files and remove some old ones. That'll be guaranteed to have merge conflicts with this.

I suggest we hold off on renaming files and renaming "DART" instances until we've 100% finalized DART 6.0. Otherwise we'll probably end up with ugly merge conflicts and needing to duplicate efforts.

Also, I'm not sure if we should rename the file extensions in DART 6.0. I was thinking we might just want to rename the extensions for KIDO to simplify the API incompatibility between DART 5 and DART 6.

jslee02:
> I suggest we hold off on renaming files and renaming "DART" instances until we've 100% finalized DART 6.0. Otherwise we'll probably end up with ugly merge conflicts and needing to duplicate efforts.

Right, I'll try to resolve the merge conflict once your PR for #643 is merged (or created) since I already put some efforts already. :sweat: 

Another concern I have is the confliction between #631 and `grey/aspects`. Do you see #631 can be merged to your branch without ugly merge conflicts? 

> Also, I'm not sure if we should rename the file extensions in DART 6.0. I was thinking we might just want to rename the extensions for KIDO to simplify the API incompatibility between DART 5 and DART 6.

I thought we intend to minimizing the difference in transitioning from DART 6.0 to KIDO 7.0 (just name change). I'm a bit inclined to making this change in DART 6.0 because the incompatibility of file extension can be straightforwardly resolved by including `dart.hpp` and a few additional component headers like `dart/utils/utils.hpp`.

mxgrey:
> I'm a bit inclined to making this change in DART 6.0 because the incompatibility of file extension can be straightforwardly resolved by including dart.hpp and a few additional component headers like dart/utils/utils.hpp.

That's a fair point. An alternative might be to "deprecate" the ``.h`` extension by having ``.h`` files that simply ``#include`` the ``.hpp`` files, but that might make the source tree too messy.

I suppose I'm okay with it either way. No matter what, it'll be like ripping off band-aid; it's painful but needs to be done sooner or later.

> Another concern I have is the confliction between #631 and grey/aspects. Do you see #631 can be merged to your branch without ugly merge conflicts?

It won't be a trivial merge, but I don't believe it will be very painful either. Any merge conflicts there will be the renaming of ``*Addon*`` to ``*Aspect*``, which is pretty easy to resolve with a search-and-replace.

jslee02:
Will recreate another clean PR once the current open pull request are all merged. Closing.

PR #654, opened 2016-03-30
Deprecate draw functions of classes in dynamics namespace

jslee02:


jslee02:
If there is no objection, I'd like to **remove** `draw` functions instead of deprecating so that `dynamics` has no dependency on `renderer` anymore.

mxgrey:
I think removing is fine. I don't really view the draw functions as being part of the public API anyway.

jslee02:
`draw` functions are now removed by @deffdcd.

I believe this PR is ready to merge unless others have more comments on this.

PR #658, opened 2016-04-06
Fix World::clone() doesn't clone the collision detector

jslee02:


mxgrey:
There seems to be a bit of a conflict where this PR has introduced the function ``getStaticType()`` for derived ``CollisionDetector`` implementations while an earlier PR introduced ``getTypeStatic()``.

I think I prefer ``getStaticType()``, so I'll choose that version as I merge this, unless someone has a reason for preferring ``getTypeStatic()``.

jslee02:
To clarify, we use `getStatic[Name]()` for a static member function, and use `get[Name]Static()` for a fixed-size version of `get[Name]()` (especially when the function returns Eigen's Vector objects), right?

mxgrey:
Yes, that has been our convention so far.

I'm also spotting some conflicts in naming conventions for the strings of the different collision detectors. We'll need to choose between:

``"DART"`` or ``"dart"``
``"Bullet"`` or ``"bullet"``
``"FCL"`` or ``"fcl"``

I don't really have any personal preference for any of these, but a decision must be made :laughing:

jslee02:
Yeah, I would slightly prefer lower letter versions since it's been used in `SKEL` format to specify the collision detectors. As these type would be used to print messages, it would be good to use same strings.

mxgrey:
Sounds good.

jslee02:
Merged master branch into this. I think this is ready to merge.

mxgrey:
I was also working on merging this, but I came across something fishy, and I'm seeing it in your merge as well.

``cloneWithoutCollisionObjects()`` creates a ``unique_ptr``, but ``setCollisionDetector(~)`` now expects a ``shared_ptr``. So how is it that ``setCollisionDetector(cd->cloneWithoutCollisionObjects());`` is being permitted? I can only guess it has to do with ``CollisionDetector`` inheriting ``std::enable_shared_from_this``, but I thought the object needed to be managed by a ``shared_ptr`` to begin with, so can we really just implicitly cast from ``unique_ptr`` to ``shared_ptr``?

jslee02:
Oh, sorry for the confliction on work.

Good catch. I think `cloneWithoutCollisionObjects` should be changed to return `shared_ptr` anyways because it should be able to be shared by collision groups.

Hm, I'm not sure how it was accepted the code. I'm trying to figure it out googling..

mxgrey:
I'm happy with this PR. I'll merge it once the CI tests pass :+1:

Issue #627, opened 2016-03-07
Have a class for composing states and properties during inheritance


mxgrey:
We've established a `State` and `Properties` pattern which enables a huge number of features, but it comes at the cost of a significant amount of boilerplate code, even for very simple cases.

I propose we create a templated `Compose` class that can take care of this boilerplate work. Instead of inheriting a class directly, you would inherit it through the `Compose` class, and the `Compose` class would take care of defining the `UniqueProperties` and full `Properties` for your newly derived class.

A quick sketch up of the idea applied to Properties can be seen here: https://gist.github.com/mxgrey/ffd620c76f16ae5a5dd7

psigen:
Can you put together a gist where you define some simple example `State` and `Properties` using the current API and the proposed `Compose` API?  Maybe a toy problem like adding a `State` that holds a floating point value for each BodyNode?

I'm not sure I even fully understand the current API, so it's hard to figure out the change.

mkoval:
> I'm not sure I even fully understand the current API, so it's hard to figure out the change.

:+1:

mxgrey:
This has been finished and merged.

PR #659, opened 2016-04-08
Aspects, States, and Properties

mxgrey:
This pull request addresses #643, #627, and some conversations that were had offline. It's a doozy of a pull request, but I think it provides a lot of valuable improvements and features with minimal impact on the API (besides the renaming of ``Addon`` to ``Aspect``).

Due to the renaming, the diff that's available on Github is going to be virtually unusable. That being the case, I'm going to provide individual links that highlight important parts of this pull request.

### Air-tight (albeit complex) inheritance structures

[First I'd like to start with a simplified diff which illustrates the primary value of this pull request](https://github.com/dartsim/dart/compare/grey/compare/aspects)

The key thing to note is that we eliminate a huge amount of code duplication between these two classes. EndEffector and ShapeNode no longer store their own relative transform property because now FixedFrame stores this information as part of its ``FixedFrame::Aspect``. Moreover, EndEffector and ShapeNode don't need to define ``Node::State`` or ``Node::Properties`` for themselves, because all of their states and properties -- from the beginning to the end of their inheritance structure -- are stored within Aspects. This allows them to each inherit a class called ``CompositeNode`` which will wrap their ``Composite::State`` and ``Composite::Properties`` up into a ``Node::State`` and ``Node::Properties``.

Essentially, every class that has information which could be labeled as "State" or "Properties" will store that information with one or more ``Aspects``. An ``Aspect`` can embed its information within the object that owns it (completely eliminating any overhead for the object to access it), or the Aspect instance may hold onto the information itself (allowing for more flexible memory usage). This decision is up to the developer/user to determine which approach is best suited for any given Aspect.

The bulk of the implementation for these features can be found in [dart/common/EmbeddedAspect.h](https://github.com/dartsim/dart/blob/grey/aspects/dart/common/EmbeddedAspect.h) and [dart/common/detail/EmbeddedAspect.h](https://github.com/dartsim/dart/blob/grey/aspects/dart/common/detail/EmbeddedAspect.h).

### Direct access to members of Composite State/Properties

Up until now, when you have a ``Composite::State`` (formerly known as ``AddonManager::State``) it was completely opaque. You had no way of accessing or modifying its contents; you could only pull it out of one Composite and pass it to another Composite. Now, however, you can do the following:

```
Composite::Properties properties;
properties.getOrCreate<BodyNode>().mName = "Body";
if(SoftBodyNode::AspectProperties* softProperties = properties.get<SoftBodyNode>())
  softProperties->mDampCoeff = 0.3;
```

This allows you to dynamically construct and access individual Aspects of the Composite State. Identical features are available for Composite Properties. ``get<T>()`` will return a pointer which is ``nullptr`` if the Aspect didn't exist, whereas ``getOrCreate<T>()`` will return a reference, since the information will be created if it didn't exist already.

[The implementation for these features can be found in dart/common/detail/CompositeData.h](https://github.com/dartsim/dart/blob/grey/aspects/dart/common/detail/CompositeData.h).

### Seamless integration of both internal Aspects and user-added Aspects

Because of the new system, any Aspects (formerly known as Addons) that user adds to a ``Skeleton``, ``BodyNode``, ``Joint``, or ``Node`` will get sucked into the standard ``State`` and ``Properties`` of the BodyNode/Joint/Node that it's added to, as well as the Skeleton in which that object exists. This is in contrast to the prior approach where the was an ``ExtendedProperties`` class that mangled together the basic Properties and the ``Composite::Properties``.

The new system is completely unified---functionally, semantically, and mechanically---instead of having different information handled in different ways and then getting mangled together at the last second so that it can be delivered to the user as a single object.

### Template Madness

I think the biggest catch to these changes is the ever growing usage of templates. As much as I may adore templates myself, I do recognize that not everyone is very comfortable with them, and they can result in very complex (read: confusing) inheritance structures when CRTP is involved. This template-based implementation is high performance and very flexible, but it will probably be confusing for reviewers and future developers. With that in mind, I encourage people to post any questions they run into while reviewing the implementation. Even if you can manager to answer the question yourself, it would still be good for me to know the question so that I can address it specifically when it comes time for me to get serious about documenting this implementation.

### API Changes

I tried my hardest to keep the API changes minimal (with the exception of renaming ``Addon`` to ``Aspect`` and ``AddonManager`` to ``Composite``). The biggest difference you'll see is in the ``SomeClass::Properties`` **constructors**. Only the **constructors**. The actual member variables of those classes should remain virtually unchanged, but their constructors won't quite work the way they used to, especially ``BodyNode::Properties``. An easy fix for this is to allow it to perform a default construction and then fill in its member variables after construction.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/659)


mxgrey:
To illustrate the purpose of the Aspect pattern, I think it's useful to look at the Joint types. The base ``Joint`` class is a pure abstract interface class which declares the crucial API needed by ``Skeleton`` and ``BodyNode`` in order to perform kinematics and dynamics computations. There are three classes directly derived from ``Joint``: ``ZeroDofJoint``, ``SingleDofJoint``, and ``MultiDofJoint``. Each of these provides implementations that are shared by the most-derived Joints that fall under each of their categories, such as ``RevoluteJoint`` and ``PrismaticJoint`` for ``SingleDofJoint`` or ``FreeJoint`` and ``BallJoint`` for ``MultiDofJoint``.

Let's consider ``RevoluteJoint`` as an example:

```
Joint <-- SingleDofJoint <-- RevoluteJoint
```

There are three stages of inheritance for this class. Each stage has its own properties which are characteristic of its specific role in the inheritance structure. These characteristic properties are called ``AspectProperties`` (formerly known as ``UniqueProperties``).

* **Joint::AspectProperties**
 * ``std::string mName``
 * ``Eigen::Isometry3d mT_ParentBodyToJoint``
 * ``Eigen::Isometry3d mT_ChildBodyToJoint``
 * ``bool mIsPositionLimited``
 * ``ActuatorType mActuatorType``
* **SingleDofJoint::AspectProperties**
 * ``double mPositionLowerLimit``
 * ``double mPositionUpperLimit``
 * ``double mVelocityLowerLimit``
 * ... etc ...
 * ``double mRestPosition``
 * ``double mDampingCoefficient``
 * ``double mFriction``
 * ``double mDofName``
* **RevoluteJoint::AspectProperties**
 * ``Eigen::Vector3d mAxis``

Each of these Aspects is critical for the overall ``RevoluteJoint`` to function correctly. But the way we decompose this data into Aspects allows the various fully-derived Joint classes to share the implementations of all these things, and it cuts out structural redundancy.

The Joint case is a very simple and linear inheritance structure, so it probably falls short of fully motivating the complexity of the Aspect implementation. Now let's consider a more challenging example which should help convey why we have all these crazy templates: ``EndEffector`` and ``ShapeNode``. If we ignore the classes that are used to specialize templated functions, we get roughly the following inheritance structure for EndEffector:

```
Entity <-- Frame <----------- FixedFrame
Node <^------------ EntityNode    ^
   ^-- JacobianNode <^-- FixedJacobianNode <-- EndEffector
```

Additionally, ``EndEffector`` is specialized for an optional Aspect called ``Support`` which is used to define the support polygons for feet. This gives ``EndEffector`` the following ``Properties``:

* **FixedFrame::AspectProperties**
 * ``Eigen::Isometry3d mRelativeTf``
* **NameAspect::Properties**
 * ``std::string mName``
* **EndEffector**
 * ``Eigen::Isometry3d mDefaultTransform``
* (*Optional*) **Support**
 * ``math::SupportGeometry mGeometry``

Here we have a situation where the inheritance structure is no longer linear. This means we can no longer cleanly build up Properties structures as we become more derived, the way we might be able to with Joint types. Without an automated way of collecting all these different Aspects, the EndEffector implementation would have to keep track of every detail of its subclasses’ Properties and take on the burden of collecting and managing these Properties itself. Indeed, that is how it was done prior to this pull request, and the result was ugly and difficult to maintain. Worst of all, we were violating the [Single Source of Truth principle](https://en.wikipedia.org/wiki/Single_source_of_truth) by having data that was redundant between ``EndEffector`` and ``FixedFrame``. We were also violating the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) by repeating a great deal of implementation between ``EndEffector`` and ``ShapeNode``, just so that they’d each have a better grasp on what their Properties were.

To really drive home the purpose of the Aspect pattern, I think ``ShapeNode`` makes a pretty compelling case. In fact, ``ShapeNode`` provided the original motivation for the ``Addon`` structure which has led us to the ``Aspect`` pattern. Here is a rough sketch of ``ShapeNode``’s inheritance structure:

```
Entity <-- Frame <--------------------------- ShapeFrame
               ^-- FixedFrame <----               ^
Node <---------^--- EntityNode    ^               ^
   ^-- JacobianNode <^-- FixedJacobianNode <-- ShapeNode
```

Additionally, ``ShapeFrame`` is specialized for three optional Aspects: ``VisualAspect``, ``CollisionAspect``, and ``DynamicsAspect``. This gives ``ShapeNode`` the following ``Properties``:

* **FixedFrame::AspectProperties**
 * ``Eigen::Isometry3d mRelativeTf``
* **NameAspect::Properties**
 * ``std::string mName``
* **ShapeFrame**
 * ``ShapePtr mShape``
* (*Optional*) **VisualAspect**
 * ``Eigen::Vector4d mRGBA``
 * ``bool mUseDefaultColor``
 * ``bool mHidden``
* (*Optional*) **CollisionAspect**
 * ``bool mCollidable``
* (*Optional*) **DynamicsAspect**
 * ``double mFrictionCoeff``
 * ``double mRestitutionCoeff``

The main driving point of the new Aspect system is that each subclass defines its own Aspect, and each of those Aspects describes its own State and Properties. Then the underlying ``Composite`` class has all the machinery needed to merge the data from all the different Aspects of an object into a single coherent whole. This eliminates the burden of doing so from the derived classes. The fact that all of this is managed automatically has the added benefit of improving overall maintainability.

mkoval:
The final product looks *much simpler* than the previous implementation :+1:

However, I am having some trouble wrapping my head around everything that is happening here: :sweat: 

1. What exactly do I need to do if I’d like to support `Aspect`s on my custom type? Here is my best guess:

  a. Instead of inheriting from `A1, A2, …`, inherit from the CRTP class `CompositeJoiner<A1, A2, …>`
  b. Define `BasicProperties` to `common::Composite::MakeProperties<???>`.
  c. Define `Properties` to `common::Composite::Properties`
  d. Call the `CompositeJoiner` constructor with some arguments (?).

2. Can you provide a brief description of each template class that I would use in (1) and, at a high level, what its purpose is?

3. Do the template arguments to classes like `MakeProperties` accept *aspects* or *classes that contain `Aspect`s* as template arguments? We currently seem to be using a mixture of both (e.g. `NameAspect` and `FixedFrame` on `ShapeNode`).

Here are a few minor follow-up questions. I think these are mostly a result of my lack of understanding of the implementation:

1. What is the difference between `EmbeddedStateAspect` and `EmbedState`?
2. Is the only difference between `EmbedState` and `EmbedProperties` the name of the member variable (`mAspectState` vs. `mAspectProperties`)?
3. What is the precise requirement on `StateDataT`? Does it *really* need to be POD or just have, e.g. copy and move operators defined?
4. How does inheriting from multiple `EmbedStateAspect`s work without getting a member variable name conflict?
5. Why do we call `setProperties` in the `ShapeNode` constructor immediately after passing `properties` to the `CompositeJoiner` constructor?

---

> The actual member variables of those classes should remain virtually unchanged, but their constructors won't quite work the way they used to, especially `BodyNode::Properties`. An easy fix for this is to allow it to perform a default construction and then fill in its member variables after construction.

I think this is fine. It is already much easier to initialize `Properties` classes by filling in the member variables after construction. Filling in a huge number of fields, many of which share the same types (e.g. `Eigen::Vector`), is unwieldy to begin with.

mxgrey:
Great questions! I'll definitely be referring back to this thread when I write the docs for these features so I know where to focus my explanations.

1. ``What exactly do I need to do if I’d like to support Aspects on my custom type?`` You have a few different options depending on how you'd like to make/use your custom ``Aspect``.

 1. If your ``Aspect`` is something that gets appended to any arbitrary existing object instance, you can simply create your own class which inherits ``Aspect`` and override the virtual member functions of the ``Aspect`` class that are relevant to you. For example, if your ``Aspect`` has a State, then override the ``getAspectState/setAspectState`` functions. If it has Properties, then override the ``getAspectProperties/setAspectProperties``. If it has both, then override both. Any ``Aspect`` can be attached to any object, although some objects have "required" ``Aspect``s without which they can't properly function.

 2. If you want to create a new class (``Derived``) that inherits an existing DART class (``Base``), but the new derived class has something to add to the state (``State``) or properties (``Properties``), then you can inherit ``EmbedStateOnTopOf<Derived, State, Base>`` or ``EmbedPropertiesOnTopOf<Derived, Properties, Base>`` or for both: ``EmbedStateAndPropertiesOnTopOf<Derived, State, Properties, Base>``. That last option will wrap up the functionality of the first two. The one requirement that your ``Derived`` class needs to satisfy when inheriting one of these templates is it must define a function named ``setAspectState(const AspectState&)`` and/or ``setAspectProperties(const AspectProperties&)`` depending on which template it wants to use. But you *will* have to use ``CompositeJoiner`` in some way if your inheritance hierarchy is more complex than a linear progression.

 3. You don't technically need to use ``common::Composite::MakeProperties<???>`` at all. Ever. I should probably give it a better name than ``MakeProperties``, because that sounds too similar to ``Aspect::MakeProperties`` which serves a very different purpose. ``common::Composite::MakeProperties`` allows you to pass it a list of ``Aspects``, and it will combine the Properties of all those aspects into a single data structure which can be implicitly converted to a generic ``Composite::Properties`` object. There is an important distinction to be made:

    1. The ``Composite::Properties`` class is totally generic, so you can't know what data fields it might contain at compile time. To manipulate the data in a ``Composite::Properties``, you need to (1) query for a specific type of ``Aspect``, (2) create an entry for that ``Aspect`` if it didn't already exist, and (3) *then* you can manipulate its fields. The first two steps can be taken care of with the ``getOrCreate<T>()`` function, but that's still an annoying step to have to take in the first place, especially if you have to do it every time you want to change or view a specific value.

    2. ``Composite::MakeProperties<A1, A2, ...>`` will create a composition of ``A1::Properties``, ``A2::Properties``, ``A...::Properties``. This composed structure will provide you with direct compile-time access to each member field of ``A1::Properties, A2::Properties, A...::Properties``. Moreover, this composed structure can be implicitly converted into a ``Composite::Properties`` instance. Perhaps we should name the class ``Composite::ComposeProperties`` or ``Composite::ComposedProperties`` depending on whether we want to use verb or noun naming. I'm open to any other ideas for naming as well.

2. ``Can you provide a brief description of each template class that I would use in (1) and, at a high level, what its purpose is?`` Sure.

 1. ``EmbedState<Derived, State>`` will give your ``Derived`` class a member field named ``mAspectState``, and it will take care of creating an "EmbeddedAspect" for it. The ``EmbeddedAspect`` will ensure that the ``mAspectState`` is treated as an ``Aspect::State`` and automatically gets sucked into your object's ``Composite::State``.

 2. ``EmbedStateOnTopOf<Derived, State, Base>`` is the same as ``EmbedState``, but it will also inherit the ``Base`` class via ``CompositeJoiner``.

 3. ``EmbedProperties<Derived, Properties>`` is like ``EmbedState``, but it creates a field named ``mAspectProperties`` which will automatically get sucked into your object's ``Composite::Properties``.

 4. ``EmbedPropertiesOnTopOf<Derived, Properties, Base>`` is the same as ``EmbedProperties`` but it also inherits the ``Base`` class via ``CompositeJoiner``.

 5. ``EmbedStateAndProperties<Derived, State, Properties>`` combines the functionality of ``EmbedState`` and ``EmbedProperties``.

 6. ``EmbedStateAndPropertiesOnTopOf<Derived, State, Properties, Base>`` is the same as ``EmbedStateAndProperties`` but it also inherits the ``Base`` class via ``CompositeJoiner``.

3. ``Do the template arguments to classes like MakeProperties accept aspects or classes that contain Aspects as template arguments?`` I don't know how people will feel about this answer, but: **Both**. If you pass it a class which inherits the ``Aspect`` base class, it will accept that class directly. If you pass it a class (``T``) which does **not** inherit the ``Aspect`` base class, it will try to use ``T::Aspect`` if the class ``T`` contains a definition (typedef or using-declaration) for ``Aspect``. Otherwise, you get a compilation error. I felt like this would be far more convenient than forcing people to only use one or the other.

> Here are a few minor follow-up questions. 

1. ``What is the difference between EmbeddedStateAspect and EmbedState?``

 1. ``EmbeddedStateAspect`` is a template that defines an ``Aspect`` object and implements its functionality. Users will typically have no reason to touch this class directly.

 2. ``EmbedState`` is a class that gets inherited to do the following:

    1. Create a field named ``mAspectState``

    2. Define a ``getAspectState() const`` function

    3. Require an ``EmbeddedStateAspect`` object

    4. Perform a using-declaration to set ``Aspect`` to the appropriate ``EmbeddedStateAspect`` type

 3. We could probably do a better job of hiding ``EmbeddedStateAspect`` since users shouldn't need to use it directly. I'm also open to a better naming convention if anyone has suggestions.

2. ``State`` and ``Properties`` are two different tracks of internal parameter data. ``State`` is data that you expect to change arbitrarily often---perhaps as often as every time step. ``Properties`` are data that you do not expect to change very often---perhaps ever---such as structural information. If you think of a differential equation, the ``State`` would be the time-dependent variables and the ``Properties`` would be the fixed parameters in the equation. Mechanically, the most critical difference is that any change to a Properties field should always increment the version counter of any object that has ownership over that field. Choosing between ``EmbedState``, ``EmbedProperties``, or ``EmbedStateAndProperties`` is a matter of deciding whether the data you want to embed should be treated as part of the object's ``State`` or part of its ``Properties``.

3. ``What is the precise requirement on StateDataT? Does it really need to be POD or just have, e.g. copy and move operators defined?`` POD is guaranteed to work. Beyond that, I *think* the only thing that's required is copy constructor/operator, since it would probably just use a copy operation if a move operation is not defined. However, I have not tested this, so I'm mostly speculating.

4. ``How does inheriting from multiple EmbedStateAspects work without getting a member variable name conflict?`` The member fields shadow each other. To access ``mAspectState`` of the most derived class, you can just say ``mAspectState`` directly. To access the field of a ``Base`` class, you just say ``Base::mAspectState``. It's probably a bit awkward for people who aren't familiar with shadowing, but I found it pretty easy to get used to personally.

5. ``Why do we call setProperties in the ShapeNode constructor immediately after passing properties to the CompositeJoiner constructor?`` Probably not for any good reason, but I'll look into it. Most likely it's leftover cruft from the prior implementation when we were duplicating code all over the place. If I find that it's actually important, then I'll add a comment in the code about why we do it.

dart/common/Composite.h
```diff
@@ -0,0 +1,212 @@
...
+  template <class T>
+  void set(std::unique_ptr<T>&& aspect);
+
+  /// Construct an Aspect inside of this Composite
+  template <class T, typename ...Args>
+  T* createAspect(Args&&... args);
+
+  /// Remove an Aspect from this Composite.
+  template <class T>
+  void eraseAspect();
```
jslee02: I wonder if this function name can be renamed to `removeAspect()` since we've used `remove-` for this kind of function.

dart/common/Composite.h
```diff
@@ -0,0 +1,212 @@
...
+  void duplicateAspects(const Composite* fromComposite);
+
+  /// Make the Aspects of this Composite match the Aspects of otherComposite. Any
+  /// Aspects in this Composite which do not exist in otherComposite will be
+  /// erased.
+  void matchAspects(const Composite* otherComposite);
+
+protected:
+
+  template <class T> struct type { };
```
jslee02: It seems this can be moved to `SpecializedForAspect`.

dart/common/Composite.h
mxgrey: That sounds good to me :+1:

dart/common/Composite.h
```diff
@@ -0,0 +1,212 @@
...
+  /// A set containing type information for Aspects which are not allowed to
+  /// leave this composite.
+  RequiredAspectSet mRequiredAspects;
+};
+
+//==============================================================================
+/// Attach an arbitrary number of Aspects to the specified Composite type.
+// TODO(MXG): Consider putting this functionality into the Composite class
+// itself. Also consider allowing the user to specify arguments for the
+// constructors of the Aspects.
```
jslee02: Maybe we can put this function into `Composite` as:

```cpp
// header
template <class... Aspects>
  void createAspects();

// source
template <class... Aspects>
struct CreateAspectsImpl;

template <class FirstAspect, class... RestAspects>
struct CreateAspectsImpl<FirstAspect, RestAspects...>
{
  static void create(Composite* composite)
  {
    composite->createAspect<FirstAspect>();

    CreateAspectsImpl<RestAspects...>::create(composite);
  }
};

template <>
struct CreateAspectsImpl<>
{
  static void create(Composite* /*composite*/) {}
};

template <class... Aspects>
void Composite::createAspects()
{
  CreateAspectsImpl<Aspects...>::create(this);
}
```

Passing arguments for each `Aspect` might be tricky. One possible solution I can come up with now is using `std::tuple` with [this workaround](http://stackoverflow.com/questions/7858817/unpacking-a-tuple-to-call-a-matching-function-pointer) to unpack `std::tuple` beck.

mxgrey: In C++14 there will be a way to do the workaround built into the standard, called [integer sequence](http://en.cppreference.com/w/cpp/utility/integer_sequence). My preference would be to hold off on it until then since it's not a crucial feature. Although if people consider it desirable enough right now, I would be okay with using the workaround.

dart/common/Composite.h
mxgrey: You're right. An older implementation of Aspect specialization required the base ``Composite`` class to use it, but that is no longer the case.

dart/common/EmbeddedAspect.h
```diff
@@ -0,0 +1,474 @@
...
+///
+/// Your derived class must implement the following functions:
+///
+/// \code{.cpp}
+/// void setAspectState(const AspectState& state);
+/// \endcode
+///
+/// To embed both state and properties information, use EmbedStateAndProperties.
+template <class DerivedT, typename StateDataT>
+class EmbedState : public virtual common::RequiresAspect<
```
jslee02: Question: I wonder why `-s` is not added to `Embed` when `Requires` has `-s`.

mxgrey: ``Embed`` is being used as a verb while ``Requires`` is being used as an adjective.

``EmbedState`` is giving a command: "Embed this State"

``RequiresAspect`` is describing the derived class: "The derived class Requires this Aspect"

dart/common/Composite.h
jslee02: For now, we can add `createAspects()` that doesn't take any parameters as above then consider later adding another function overloading `createAspects(...)` that takes parameters. But I also think this is not a crucial feature so it's totally up to you.

mxgrey: Personally, my feeling would be to hold off on it entirely until the next minor version since I don't see it as an being important for this major release. Let's give ourselves some time to work with Aspects so we can build up an idea of what we would like from a multi-aspect-creator function.

mxgrey:
For the record, I would like to merge this PR in after all the other PRs (besides the .hpp one). I've been maintaining a branch called ``grey/merge/aspects`` which is merging in all the other pull requests as they get created to ease the burden of merging. Once all other PRs are merged, we can merge ``grey/merge/aspects`` into ``grey/aspects`` and then easily merge this PR into master.

mxgrey:
I've merged in all remaining pull requests, so this branch should be able to effortlessly merge into ``master`` after all the other pull requests have been merged.

dart/common/StlHelpers.h
```diff
@@ -70,6 +71,21 @@ static T getVectorObjectIfAvailable(size_t index, const std::vector<T>& vec)
   return nullptr;
 }
 
+//==============================================================================
+/// static_if_else allows the compiler to choose between two different possible
+/// types at runtime based on a runtime boolean.
+template <bool B, typename T_if, typename T_else>
+struct static_if_else
```
jslee02: I think `static_if_else` is identical to [`std::conditional`](http://en.cppreference.com/w/cpp/types/conditional).

dart/common/detail/CompositeData.h
```diff
@@ -0,0 +1,354 @@
...
+//==============================================================================
+template <class CompositeType, template<class> class GetData, class AspectT,
+          typename... Remainder>
+struct ComposeData<CompositeType, GetData, AspectT, Remainder...> :
+    public GetData<AspectT>::Type,
+    public ComposeData<CompositeType, GetData, Remainder...>
+{
+public:
+
+  enum Delegate_t { Delegate };
```
jslee02: What is the motivation of adding the suffix `_t` to enumerator names?

mxgrey: Honestly, it's because I can't think of a better way to distinguish the type name from the value name for enumeration tags. I'm open to alternative naming conventions.

dart/common/StlHelpers.h
mxgrey: Yes, you are exactly right. I'll replace that right away. I was surprised when I couldn't seem to find anything similar to my ``static_if_else`` in the standard, but I guess it didn't occur to me to look for something called ``conditional``.

dart/common/detail/CompositeData.h
psigen: Could we just do a `Tag` suffix?  I think it would be clear to declare
`enum DelegateTag { Delegate };`
`enum NoArgTag { NoArg };`

dart/common/StlHelpers.h
jslee02: Maybe `conditional` is unintuitive or too general name.

dart/common/detail/CompositeData.h
mxgrey: That sounds like a really good convention to me. I'll make the changes now.

jslee02:
Reviewed 82 of 211 files at r1.
Review status: 81 of 186 files reviewed at latest revision, 11 unresolved discussions.

---

*[dart/common/Aspect.h, line 134 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFp5FUXAN20crRw9VUj:-KFp5FUXAN20crRw9VUk:-307687839) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/Aspect.h#L134)):*
Do we really need to pass the old composite parameter? Wouldn't the `Aspect` lose the currently holding composition instead of the passed-in one?

---

*[dart/common/EmbeddedAspect.h, line 63 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFokYsGVupZZH7suMvi:-KFokYsGVupZZH7suMvj:596188862) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/EmbeddedAspect.h#L63)):*
Nitpick: To me, it seems `Impl` is the actual base of this class and `Base` is the base of the base class. I wonder if we can rename `Base` to something like `BaseOfBase` and `Impl` to `Base`. As this class is exposed as public API, some user might get confused witht the type name.

---

*[dart/common/StlHelpers.h, line 44 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFoM7N749Bo6z5XA2-_:-KFoM7N749Bo6z5XA2-a:-105219618) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/StlHelpers.h#L44)):*
This can be removed.

---

*[dart/common/detail/EmbeddedAspect.h, line 79 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpMrM419mx8fg1TrSQ:-KFpMrM419mx8fg1TrSR:-1563586838) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L79)):*
When would be the case that we don't use the default setter and getter?

---

*[dart/common/detail/EmbeddedAspect.h, line 152 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFovHKSssQE3Kq4Etqt:-KFovHKTsVrxJqyA1OQn:1785818135) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L152)):*
Suggestion: We could have `CompositeTrackingAspect::hasComposite()` as a helper function.

---

*[dart/common/detail/EmbeddedAspect.h, line 221 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpJ4AXbnOzogRy1Elt:-KFpJ4AXbnOzogRy1Elu:-1549064452) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L221)):*
It seems we can do early return if `this->getComposite() == newComposite`.

---

*[dart/common/detail/EmbeddedAspect.h, line 240 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpMmSICyabHbiL-TJ5:-KFpMmSJWx0lI6fq2dzv:920311553) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L240)):*
It seems `mTemporaryState` is also used when this Aspect is being created until it's added to a composite. If so, it would be good to say about it in the comment.

---

*[dart/common/detail/EmbeddedAspect.h, line 411 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpLoIuI1ieD0_lgEKS:-KFpLoIuI1ieD0_lgEKT:-861697728) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L411)):*
I'm a bit confused about the use of this variable. Here is my best understanding. Please correct me if I'm wrong.

* 

---

*[dart/common/detail/SpecializedForAspect.h, line 57 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpES1vTnxh8wx9jD4v:-KFpES1wz9MH0SP-M30a:-769323409) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/SpecializedForAspect.h#L57)):*
This is obviously not a performance critical, but using `std::map::insert` for adding an item here would reduce the unnecessary `std::map::find` call since `std::map::insert` returns the iterator.

---

*[dart/dynamics/Skeleton.h, line 72 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFlZuhqQ4ihy7Sa19YE:-KFlZuhqQ4ihy7Sa19YF:-1488220125) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/dynamics/Skeleton.h#L72)):*
Could we apply the enum name convention (not using `_t` suffix) to here? `ConfigFlags` would look good to me.

---

*[dart/dynamics/detail/BodyNodeAspect.h, line 147 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFoHTF-SLcVLEj1u6RP:-KFoHTF0BBG29LWj9qcC:-235224705) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/dynamics/detail/BodyNodeAspect.h#L147)):*
Could we change this type name to `BodyNodeBase` or ​​other names that don't sound like an aspect? I was a bit confused this as an aspect. Other classes inherit `EmbedStateAndPropertiesOnTopOf<...>` use `~Base`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
*[dart/common/detail/EmbeddedAspect.h, line 411 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpLoIuI1ieD0_lgEKS:-KFpWO0eYNXPKSWF0i_D:882249303) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L411)):*
Never mind ​the above comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
The last comment was made through [Reviewable](https://reviewable.io/) to see how good it is comparing to the default review system of Github especially when the diff is too long to Github to display like this PR. @mxgrey Please let me know if you have any trouble in using Reviewable.

mxgrey:
Review status: 81 of 186 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[dart/common/Aspect.h, line 134 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFp5FUXAN20crRw9VUj:-KFtVH06AwMLMa5T1xIs:-207576978) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/Aspect.h#L134)):*
Not all ``Aspect``s keep track of their current ``Composite``, so we pass in the old Composite just in case the Aspect needs that information.

---

*[dart/common/EmbeddedAspect.h, line 63 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFokYsGVupZZH7suMvi:-KFtVwIbmW3Ec7IgpRDS:-455060378) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/EmbeddedAspect.h#L63)):*
Here's my reasoning behind the naming:

It's true that ``detail::EmbeddedStateAspect`` is technically the base class of ``EmbeddedStateAspect``. But practically speaking, ``EmbeddedStateAspect`` is just the user-facing front-end of ``detail::EmbeddedStateAspect``. It's not as if ``EmbeddedStateAspect`` is really a derivation of ``detail::EmbeddedStateAspect``; it's just the user-friendly API.

It happens to be the case that ``detail::EmbeddedStateAspect`` implements the features of ``EmbeddedStateAspect`` via inheritance, but I'd say that ``CompositeTrackingAspect<CompositeT>`` would be the "logical" Base class of the overall ``EmbeddedStateAspect``.

---

*[dart/common/detail/EmbeddedAspect.h, line 79 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpMrM419mx8fg1TrSQ:-KFtZlqUyDOeGlur46T4:394387339) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L79)):*
This gives users the ability to customize the way their ``EmbeddedAspect`` functions. The default setter function assumes that the derived class has a function named ``setAspectState(const AspectState&)``, and the default getter function assumes that the derived class has a function named ``getAspectState() const``. It's possible that a user might not want to follow this convention for their own custom classes.

---

*[dart/common/detail/EmbeddedAspect.h, line 221 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpJ4AXbnOzogRy1Elt:-KFt_nRMJ5VmyfdkNyv0:614242243) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L221)):*
We should call ``Base::setComposite(newComposite)`` no matter what, because we technically don't know what a user might set their base class to be, and their base class might not be okay with short circuiting.

That said, it would most likely be a bug if ``this->getComposite() == newComposite``, because in order for ``setComposite(~)`` to be called, either:

1. The ``Aspect`` should be newly constructed and therefore have no composite yet.
2. ``loseComposite(~)`` should have already been called on the ``Aspect``, so it should have no composite currently.

We could put an ``assert(nullptr == this->getComposite())`` to make sure that this is true, but I think any situations where we might short-circuit would be a bug.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

mxgrey:
*[dart/dynamics/detail/BodyNodeAspect.h, line 147 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFoHTF-SLcVLEj1u6RP:-KFtf2yI6qCVT8OiGyx5:1701840984) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/dynamics/detail/BodyNodeAspect.h#L147)):*
How about ``BodyNodeCompositeBase`` to match the convention that we have for ``ShapeNode`` and ``EndEffector``?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
Reviewed 31 of 211 files at r1.
Review status: 102 of 186 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[dart/common/Aspect.h, line 134 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFp5FUXAN20crRw9VUj:-KFtcZU_t2urJVOh8UBj:637751753) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/Aspect.h#L134)):*
That makes sense to me. Then I'd suggest adding `assert` to `CompositeTrackingAspect::loseComposite()` that checks if the passed-in `oldComposite` is identical to the tracking composite for sure.

---

*[dart/common/Composite.h, line 199 \[r2\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFtc2xTdaIqYGDYlEQA:-KFtc2xTdaIqYGDYlEQB:683505675) ([raw file](https://github.com/dartsim/dart/blob/93436e82b7ba2aca518e98276071d20cc90a1585/dart/common/Composite.h#L199)):*
Nitpick: mgr --> comp?

---

*[dart/common/detail/CompositeData.h, line 54 \[r2\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFtWLuwlMLtt2ivQ2rM:-KFtWLuxhhWwCrtv0e5k:-1664876667) ([raw file](https://github.com/dartsim/dart/blob/93436e82b7ba2aca518e98276071d20cc90a1585/dart/common/detail/CompositeData.h#L54)):*
It seems `AspectT` can be either of an​ aspect type or a composite type. If so, I'd suggest renaming this to `AspectOrCompositeT` to make it more descriptive.

Also, we could add `static_assert` here something like:
```cpp
  static_assert(
      std::is_base_of<Aspect, AspectOrCompositeT>::value
      || std::is_base_of<Composite, AspectOrCompositeT>::value,
      "AspectOrCompositeT should be a derived class from either of Aspect or "
      "Composite.");
```

---

*[dart/common/detail/EmbeddedAspect.h, line 79 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpMrM419mx8fg1TrSQ:-KFte6gQKwxa2zb-bVQM:-1304509575) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L79)):*
That makes sense. I saw the comment on `EmbedState` that says the derived class **must** implement `setAspectState()` so I assumed that's the case always. How about adding a note what you mentioned to the comment then?

---

*[dart/common/detail/EmbeddedAspect.h, line 221 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpJ4AXbnOzogRy1Elt:-KFtf3FLZw_76cAf_8Cw:-1078290192) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L221)):*
Makes sense to me. :+1: to putting `assert(nullptr == this->getComposite())`.

---

*[dart/dynamics/detail/BodyNodeAspect.h, line 147 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFoHTF-SLcVLEj1u6RP:-KFtfwYiI1YOKJCk47-o:-941257294) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/dynamics/detail/BodyNodeAspect.h#L147)):*
:+1: Sounds good to me.

Hm, as a new user of `Reviewable`,  I'm not sure, in this case, if I should push the `Acknowledge​` button to mark this as resolved, or wait until the code changed. I will leave this dis​cuss open to see what will happen once the code is changed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

mxgrey:
*[dart/common/detail/CompositeData.h, line 54 \[r2\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFtWLuwlMLtt2ivQ2rM:-KFthLPOfu-Y_xf8aqMg:-766848369) ([raw file](https://github.com/dartsim/dart/blob/93436e82b7ba2aca518e98276071d20cc90a1585/dart/common/detail/CompositeData.h#L54)):*
This isn't exactly the case, although it's close. The type that gets passed in must either

1. Be an Aspect.

2. Have a typedef of the form ``AspectT::Aspect``.

It is not necessary or sufficient to be a Composite.

This allows us to do things like

```
Composite::Properties properties;
properties.getOrCreate<BodyNode>();
```

in order to add ``BodyNode::Aspect::Properties`` to the ``properties`` object instead of needing to write

```
Composite::Properties properties;
properties.getOrCreate<BodyNode::Aspect>();
```

The latter will also work; it just involves more typing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

mxgrey:
*[dart/common/detail/EmbeddedAspect.h, line 79 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpMrM419mx8fg1TrSQ:-KFtjr0B_xMzDJAaMXti:-1996381037) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L79)):*
The comment is true when using the ``EmbedState``, ``EmbedProperties`` or ``EmbedStateAndProperties`` class, although I suppose I can add a comment saying that there are potential workarounds by using ``detail::EmbeddedState``, etc. directly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

mxgrey:
Review status: 94 of 185 files reviewed at latest revision, 10 unresolved discussions.

---

*[dart/common/Composite.h, line 199 \[r2\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFtc2xTdaIqYGDYlEQA:-KFtgvGskVynF11HXS_5:-499146009) ([raw file](https://github.com/dartsim/dart/blob/93436e82b7ba2aca518e98276071d20cc90a1585/dart/common/Composite.h#L199)):*
Done.

---

*[dart/common/detail/EmbeddedAspect.h, line 79 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpMrM419mx8fg1TrSQ:-KFtl0m6QOYjMhMh8Hdz:-499146009) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L79)):*
Done.

---

*[dart/common/detail/EmbeddedAspect.h, line 152 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFovHKSssQE3Kq4Etqt:-KFtd24tRUkCjq6d7y7p:-499146009) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L152)):*
Done.

---

*[dart/common/detail/EmbeddedAspect.h, line 240 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpMmSICyabHbiL-TJ5:-KFtdaSRUNHBirM8D-KX:-499146009) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L240)):*
Done.

---

*[dart/common/detail/SpecializedForAspect.h, line 57 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFpES1vTnxh8wx9jD4v:-KFteZcd9_TxtdkAU0tw:-499146009) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/SpecializedForAspect.h#L57)):*
Done.

---

*[dart/dynamics/Skeleton.h, line 72 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFlZuhqQ4ihy7Sa19YE:-KFtewCS2halMJ8-bojA:-499146009) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/dynamics/Skeleton.h#L72)):*
Done.

---

*[dart/common/StlHelpers.h, line 44 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFoM7N749Bo6z5XA2-_:-KFtd4EFDhN2ahJe2pkL:-499146009) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/StlHelpers.h#L44)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
Review status: 94 of 185 files reviewed at latest revision, 3 unresolved discussions.

---

*[dart/common/detail/EmbeddedAspect.h, line 152 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFovHKSssQE3Kq4Etqt:-KFtoH_uvSZchjEz_V72:1613610567) ([raw file](https://github.com/dartsim/dart/blob/401bd26c80b51759ffdfb18f519fbd05d99eb01d/dart/common/detail/EmbeddedAspect.h#L152)):*
I expected to use the function here, but it has almost no difference. :+1:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
Reviewed 87 of 211 files at r1, 1 of 1 files at r2, 7 of 11 files at r3, 1 of 1 files at r4.
Review status: 173 of 185 files reviewed at latest revision, 19 unresolved discussions.

---

*[dart/dynamics/FixedFrame.h, line 59 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFtwckpGlSLHjIYRfpT:-KFtwckpGlSLHjIYRfpU:-1638475302) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/FixedFrame.h#L59)):*
Could we move this under `detail` directory?

---

*[dart/dynamics/ScrewJoint.cpp, line 171 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu-Dl5dJOR1cPhODy4:-KFu-Dl5dJOR1cPhODy5:1398419405) ([raw file](https://github.com/dartsim/dart/blob/2eaf7322538731d7cb06ba13d89d51078e889a30/dart/dynamics/ScrewJoint.cpp#L171)):*
Why don't we check the validity here?

---

*[dart/dynamics/ShapeFrame.h, line 106 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu-XDLNIkOki1Pbnbw:-KFu-XDLNIkOki1Pbnbx:-528585246) ([raw file](https://github.com/dartsim/dart/blob/2eaf7322538731d7cb06ba13d89d51078e889a30/dart/dynamics/ShapeFrame.h#L106)):*
It seems the places of the definitions of properties structs are not consistent. What is the convention of them in terms of visibility? Should they​ be in `detail` namespace and under `detail` directory?

---

*[dart/dynamics/ShapeNode.h, line 59 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu0hNNbBUnKEOClnK_:-KFu0hNNbBUnKEOClnKa:-843252306) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/ShapeNode.h#L59)):*
The same. I'm a bit inclined​ to moving all the definitions in `detail` namespace should be under `detail` directory.

---

*[dart/dynamics/SimpleFrame.h, line 163 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu1B4ZjRA6EeBNHO6_:-KFu1B4ZjRA6EeBNHO6a:-1701345687) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/SimpleFrame.h#L163)):*
`virtual` is not necessary with `override`. I'm thinking to remove all the `virtual` keywords when `override` is used together​ as a wholesale, though.

---

*[dart/dynamics/SoftBodyNode.h, line 310 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu1rBYn61mHBMfHtaT:-KFu1rBZZBBuByxEu9rt:1760811064) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/SoftBodyNode.h#L310)):*
Can be removed

---

*[dart/dynamics/SoftMeshShape.cpp, line 94 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu2334KcEIZ7m736zc:-KFu2334KcEIZ7m736zd:1312074174) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/SoftMeshShape.cpp#L94)):*
`common::make_unique<aiMesh>())`?

---

*[dart/dynamics/detail/BodyNodeAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu4MbyuInDKO9EwFN_:-KFu4MbyuInDKO9EwFNa:625595779) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/BodyNodeAspect.h#L37)):*
Please rename `PROPERTIES` to `ASPECT`.

---

*[dart/dynamics/detail/BodyNodeAspect.h, line 59 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu2f0Wj7trkGJzYqUQ:-KFu2f0Wj7trkGJzYqUR:1637189041) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/BodyNodeAspect.h#L59)):*
Could add newline

---

*[dart/dynamics/detail/EulerJointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu4fFE31iA5HAdEZ9S:-KFu4fFE31iA5HAdEZ9T:625595779) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/EulerJointAspect.h#L37)):*
Please rename `PROPERTIES` to `ASPECT`.

---

*[dart/dynamics/detail/JointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu4l4W1MhWvRMQhKAi:-KFu4l4W1MhWvRMQhKAj:1461660921) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/JointAspect.h#L37)):*
Please rename to `DART_DYNAMICS_DETAIL_JOINTASPECT_H_`.

---

*[dart/dynamics/detail/MultiDofJointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu41IMS9BoBLWZxI-a:-KFu41IMS9BoBLWZxI-b:-519193805) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/MultiDofJointAspect.h#L37)):*
Please rename to `DART_DYNAMICS_DETAIL_MULTIDOFJOINTASPECT_H_`

---

*[dart/dynamics/detail/PlanarJointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu51yejK3H6gyhGJSf:-KFu51yejK3H6gyhGJSg:625595779) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/PlanarJointAspect.h#L37)):*
Please rename `PROPERTIES` to `ASPECT`.

---

*[dart/dynamics/detail/PrismaticJointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu58z3ZywXbEkOgu6M:-KFu58z3ZywXbEkOgu6N:625595779) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/PrismaticJointAspect.h#L37)):*
Please rename `PROPERTIES` to `ASPECT`.

---

*[dart/dynamics/detail/RevoluteJointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu5ExRlBA7U64v5wCJ:-KFu5ExRlBA7U64v5wCK:625595779) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/RevoluteJointAspect.h#L37)):*
Please rename `PROPERTIES` to `ASPECT`.

---

*[dart/dynamics/detail/ScrewJointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu5K6oC_V6lOGuhSW6:-KFu5K6oC_V6lOGuhSW7:625595779) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/ScrewJointAspect.h#L37)):*
Please rename `PROPERTIES` to `ASPECT`.

---

*[dart/dynamics/detail/SingleDofJointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu5ObwrNViUXxhAPoo:-KFu5ObwrNViUXxhAPop:625595779) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/SingleDofJointAspect.h#L37)):*
Please rename `PROPERTIES` to `ASPECT`.

---

*[dart/dynamics/detail/SoftBodyNodeAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu5U32-B5pACns3iCl:-KFu5U32-B5pACns3iCm:625595779) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/SoftBodyNodeAspect.h#L37)):*
Please rename `PROPERTIES` to `ASPECT`.

---

*[dart/dynamics/detail/UniversalJointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu5ahcfL0uV1KqJYWA:-KFu5ahcfL0uV1KqJYWB:625595779) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/UniversalJointAspect.h#L37)):*
Please rename `PROPERTIES` to `ASPECT`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
Reviewed 10 of 211 files at r1, 2 of 11 files at r3.
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---

*[dart/dynamics/FixedJacobianNode.cpp, line 215 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu6gMJhFmxmfROVgWT:-KFu6gMJhFmxmfROVgWU:398073493) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/FixedJacobianNode.cpp#L215)):*
Please remove one line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

mxgrey:
*[dart/dynamics/ScrewJoint.cpp, line 171 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu-Dl5dJOR1cPhODy4:-KFuChAnvlLEzX3dsMO1:-1003141804) ([raw file](https://github.com/dartsim/dart/blob/2eaf7322538731d7cb06ba13d89d51078e889a30/dart/dynamics/ScrewJoint.cpp#L171)):*
It seems that the line was lost during a merge conflict resolution. I'll add it back in.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

mxgrey:
*[dart/dynamics/ShapeFrame.h, line 106 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu-XDLNIkOki1Pbnbw:-KFuDBS9XICaS62c__-G:-184694510) ([raw file](https://github.com/dartsim/dart/blob/2eaf7322538731d7cb06ba13d89d51078e889a30/dart/dynamics/ShapeFrame.h#L106)):*
For the sake of consistency, that might be preferable. We can move these ``Properties`` structs to a file named ``detail/ShapeNodeAspect.h``.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
Reviewed 7 of 211 files at r1, 23 of 23 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/dynamics/ShapeNode.h, line 59 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu0hNNbBUnKEOClnK_:-KFuPPQVghvAlKZ-DREc:96459339) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/ShapeNode.h#L59)):*
Has not addressed yet.

---

*[dart/dynamics/detail/JointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu4l4W1MhWvRMQhKAi:-KFuPDwxQGPWkHAcY3yG:96459339) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/JointAspect.h#L37)):*
Has not addressed yet.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
Reviewed 4 of 5 files at r6, 1 of 1 files at r7.

---

*[dart/dynamics/ShapeNode.h, line 59 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu0hNNbBUnKEOClnK_:-KFv0IJdfmgWYrncGNSg:-499146009) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/ShapeNode.h#L59)):*
Done.

---

*[dart/dynamics/detail/JointAspect.h, line 37 \[r4\]](https://reviewable.io:443/reviews/dartsim/dart/659#-KFu4l4W1MhWvRMQhKAi:-KFv0PtpbyElv7_C1VaW:-499146009) ([raw file](https://github.com/dartsim/dart/blob/2e5f422bb4f9c61b23ea24eb0682dea6f8d5c20c/dart/dynamics/detail/JointAspect.h#L37)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659#-:-KFv0zgljLfmXjKJnD_Y:1417088721)*

mxgrey:
I snuck in one last commit that removed some hardcoded file paths that I was using :sweat_smile:

jslee02:
Reviewed 1 of 1 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/659)*

jslee02:
@mkoval Any follow-up comment?

PR #660, opened 2016-04-08
Support for Bullet built with double precision

### LINKED ISSUE #657: Linking with Bullet fails on OS X El Capitan
I am attempting to build DART on OS X 10.11 (El Capitan). Compilation fails with this error:
```
Undefined symbols for architecture x86_64:
  "btStaticPlaneShape::btStaticPlaneShape(btVector3 const&, float)", referenced from:
      dart::collision::BulletCollisionNode::BulletCollisionNode(dart::dynamics::BodyNode*) in BulletCollisionNode.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
I have Bullet 2.83.6 installed from Homebrew.

mkoval:
Bullet requires a pre-processor variable to be set before including any headers to determine whether the library was built with single or double precision. Failing to do so may result in a linking error.

This commit:

1. Forces CMake to use CONFIG mode (`BulletConfig.cmake` shipped with Bullet) instead of MODULE mode (`FindBullet.cmake` shipped with CMake by default) in find_package(Bullet).

2. Updates references to `BULLET_INCLUDE_DIRS` and `BULLET_LIBRARIES` accordingly

3. Test whether Bullet was built using double precision floating point numbers by attempting to compile code that produces a linking error when the necessary preprocessor definition is missing.

4. Set the necessary preprocessor definition in `config.h`.

This fixes #657: it turns out that Homebrew builds Bullet with double point precision by default.


scpeters:
Just for clarification, I'm pretty sure that bullet is single precision by default. The upstream option for `USE_DOUBLE_PRECISION` is [off by default](https://github.com/bulletphysics/bullet3/blob/master/CMakeLists.txt#L23). In the homebrew formula, double precision is an [option](https://github.com/Homebrew/homebrew-core/blob/master/Formula/bullet.rb#L27), which means off by default (recommended means on by default).

It will use double precision if you install gazebo and bullet is not yet installed:

* https://github.com/osrf/homebrew-simulation/blob/master/gazebo7.rb#L26

mkoval:
@scpeters you are likely right that Bullet got installed through Gazebo. I recently used [Mike Purvis' installation script](https://github.com/mikepurvis/ros-install-osx) to reinstall ROS, which does install Gazebo by default.

In any case, I see no reason why DART *shouldn't* work when Bullet is compiled with double precision. :smile:

scpeters:
Absolutely it should be able to use whichever version of bullet is available. I think we get around it in gazebo because the `-DBT_USE_DOUBLE_PRECISION` definition is included in the pkg-config file. This looks like a good pull request for detecting it with cmake.

jslee02:
Yeah, using `CheckCXXSourceCompiles` is an good idea. 

I made minor update for the cmake output for consistency. 

Looks good to me. :+1:

scpeters:
I just ran into this same problem while compiling gazebo with dart 5.1; is it possible to backport this fix to `release-5.1`?

jslee02:
Sure, I'll create an issue for that.

PR #667, opened 2016-04-14
Removing OpenGL dep from main lib

costashatz:
I removed OpenGL (and renderer) libs/includes from main library (libdart) and moved them to gui library (lib-dart-gui). Also, I minor fix for installing properly the gui-osg include files..


jslee02:
Thanks! I'll fix the compilation errors in #652.

PR #669, opened 2016-04-15
toRadian(), toDegree, and mathematical constants

### LINKED ISSUE #314: Use const variable instead of macro definitions for type safety

jslee02:
This PR replaces the macros for mathematical constants with type safe template functions, and addes utility functions, `toRadian()`, `toDegree()`, for degree-radian conversion.

This resolves #314.


mxgrey:
I like the idea of this pull request, although ``constants<T>`` could get somewhat annoying, especially since I think most of the time we will want to use ``constantsd`` anyway.

What if we also provided [user-defined literals](http://en.cppreference.com/w/cpp/language/user_literal) for pi and degrees? For example:

```
namespace dart {
namespace math {

constexpr double operator"" _pi (double x)
{
  return x * constantsd::pi();
}

constexpr double operator"" _deg(double x)
{
  return x * constantsd::pi() / 180.0;
}

} // namespace math
} // namespace dart
```

jslee02:
> I like the idea of this pull request, although constants<T> could get somewhat annoying, especially since I think most of the time we will want to use constantsd anyway.

Right. I also considered providing only double version rather than templating. I don't have a strong preference here so I'm fine with that too.

Here are some other alternatives we could consider:
* If we provide only double version, this template struct can be just namespace and each item would be variable instead of function as:
```cpp
namespace constants {
constexpr double pi = 3.14xxx;
}

auto twoPi = 2.0 * constant::pi;  // not constant::pi()
```
* We could consider shorter name like `const::pi()` or `const::pi`.

> What if we also provided user-defined literals for pi and degrees?

Sounds really good. By the way, `_deg` and `_rad` seem to make sense to me, but I'm not sure if `_pi` is also good practice. 

Sorry @mkoval @psigen, we're going to use another C++11 feature. :smiling_imp:

mxgrey:
I'm happy with this pull request. I think we can merge once the CI tests pass.

PR #670, opened 2016-04-15
Adding collision sets to CollisionResult

mxgrey:
This pull request addresses the discussion in PR #668

### DIFF
--- a/dart/collision/Contact.cpp
+++ b/dart/collision/Contact.cpp
@@ -2,7 +2,7 @@
  * Copyright (c) 2016, Georgia Tech Research Corporation
  * All rights reserved.
  *
- * Author(s): Jeongseok Lee <jslee02@gmail.com>
+ * Author(s): Jeongseok Lee <jslee02@gmail.com>,
  *            Michael X. Grey <mxgrey@gatech.edu>
  *
  * Georgia Tech Graphics Lab and Humanoid Robotics Lab
--- a/dart/collision/Contact.cpp
+++ b/dart/collision/Contact.cpp
@@ -3,6 +3,7 @@
  * All rights reserved.
  *
  * Author(s): Jeongseok Lee <jslee02@gmail.com>
+ *            Michael X. Grey <mxgrey@gatech.edu>
  *
  * Georgia Tech Graphics Lab and Humanoid Robotics Lab
  *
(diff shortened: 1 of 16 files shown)


dart/collision/Contact.cpp
```diff
@@ -3,6 +3,7 @@
  * All rights reserved.
  *
  * Author(s): Jeongseok Lee <jslee02@gmail.com>
+ *            Michael X. Grey <mxgrey@gatech.edu>
```
jslee02: Nitpick: We've added a comma at the end of the line when adding authors.

jslee02:
Overall, looks good. Could we add a simple test for this?

mxgrey:
Yep, I'm working on a test right now.

dart/collision/Result.cpp
```diff
@@ -74,15 +81,73 @@ const std::vector<Contact>& CollisionResult::getContacts() const
 }
 
 //==============================================================================
+const std::unordered_set<const dynamics::BodyNode*>
```
jslee02: Maybe returning const reference instead of just const?

dart/collision/Result.cpp
```diff
@@ -74,15 +81,73 @@ const std::vector<Contact>& CollisionResult::getContacts() const
...
+    return;
+  }
+
+  const dynamics::ShapeFrame* frame = object->getShapeFrame();
+  mCollidingShapeFrames.insert(frame);
+
+  if(frame->isShapeNode())
+  {
+    const dynamics::ShapeNode* node =
+        dynamic_cast<const dynamics::ShapeNode*>(frame);
```
jslee02: We could use `asShapeNode()` instead of dynamic casting.

mxgrey: I considered that. I'm not actually sure which would be cheaper: dynamic_cast or the v-table lookup for a virtual function. The difference is probably negligible, so I'm fine with either.

jslee02: Yeah, actually I also wondered that when adding `asShapeNode()`. But I found some post like [this](http://stackoverflow.com/a/9959645) saying virtual function is cheaper so added `asShapeNode()` for these use cases.

jslee02: Maybe we would like to add comment encouraging `asShapeNode()` rather than dynamic_cast when possible.

PR #691, opened 2016-04-23
Small updates for 6.0

mxgrey:
There were a few ``TODO`` tags related to version 6.0. This takes care of two of them:

1. Now, when we clone a ``Skeleton``, the full ``State`` of the original ``Skeleton`` gets copied into the new ``Skeleton``. (Before now, only ``Properties`` were getting copied)

2. Initial positions and initial velocities are now included in the constructors of ``SingleDofJoint`` and ``MultiDofJoint``.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/691)

PR #692, opened 2016-04-23
Change Markers into Nodes

mxgrey:
Up until now, Markers have been managed separately from Nodes, because Markers predate the Node concept in DART. This PR changes Markers into a Node type so that it shares all the same machinery as Nodes instead of having its own machinery.

Note that this PR builds off of #691

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/692)


jslee02:
Reviewed 20 of 43 files at r1, 1 of 1 files at r2, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/dynamics/BodyNode.h, line 614 \[r3\]](https://reviewable.io:443/reviews/dartsim/dart/692#-KG4qRmSGPLz0EFKCoVl:-KG4qRmSGPLz0EFKCoVm:-722966481) ([raw file](https://github.com/dartsim/dart/blob/2985e15ef442a93e774b7a339b26151d5f3119bd/dart/dynamics/BodyNode.h#L614)):*
This function seems useful. How about having another `createMarker` that takes the marker properties like for `ShapeNode` and `EndEffector` for consistency? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/692)*

mxgrey:
Review status: 2 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/dynamics/BodyNode.h, line 614 \[r3\]](https://reviewable.io:443/reviews/dartsim/dart/692#-KG4qRmSGPLz0EFKCoVl:-KGDNcGCMy__bfd898oO:-1251002210) ([raw file](https://github.com/dartsim/dart/blob/2985e15ef442a93e774b7a339b26151d5f3119bd/dart/dynamics/BodyNode.h#L614)):*
I'll also modify the ``createEndEffector(const Properties&)`` implementation to no longer be templated. It used to be a templated function in order to work around the fact that ``EndEffector.h`` couldn't be visible to ``BodyNode.h`` due to a circular dependency. That circular dependency is gone now, so we no longer need the template there.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/692)*

jslee02:
:lgtm:

---

Reviewed 14 of 19 files at r5, 4 of 4 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/692#-:-KGDPuXm0ijObhJAnQj6:1417088721)*

PR #694, opened 2016-04-24
Improve collision detection API for binary check

jslee02:
In the current API, we always pass collision options and result to `collide()` even though we just want the binary result. This PR adds some convenient functions for binary collision checks that don't take `CollisionResult` or for both of `CollisionOption` and `CollisionResult` so that we don't need to create a dummy `CollisionResult` for nothing. When `CollisionOption` is not passed in, `collide()` interprets it as binary check.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/694)


mxgrey:
I was thinking about this matter during the code review, but I was going to save it for later. I wonder if instead of having the ``CollisionResult`` argument be a reference, we should have it be an optional pointer with a default value of ``nullptr``. This would allow users to programmatically decide whether or not they want the collision result. It would also allow us to avoid having dummy data structures. I think the only catch would be significantly more code changes, because every collision detector implementation would need to check whether the ``CollisionResult*`` argument is a ``nullptr`` before doing anything with it.

mkoval:
:+1: :+1: :+1: for adding these helper functions. It's currently quite verbose to perform a binary check.

I also slightly prefer @mxgrey's proposal here. I can think of a lot of cases where I'd like to decide whether or not I want a `CollisionResult` programmatically. E.g. to allow the user to inspect the `CollisionResult` for debugging purposes in code that wraps a `collide` call.

jslee02:
Optional pointer sounds good to me. But I'm still considering to have the dummy variables because it would prevent the `nullptr` checks in the each collision detector.

If we change to take pointer for `CollisionResult`, I would also like to change `const CollisionOption&` to `const CollisionOption*` as well for consistency.

jslee02:
Hm, one downside would be that it would be still verbose if we call `collide()` function from the each collision detectors directly.

jslee02:
Updated the `collide()` functions as suggested. I didn't change option parameter to pointer in order to enable to pass default parameter.

jslee02:
I believe this PR is ready to merge. @mxgrey , @mkoval  could you guys take a look at quickly?

psigen:
I think making `option` a const reference and `result` a pointer makes sense here, because one is an input param and the other is an optional out param.

jslee02:
It's strange `testCollision` fails on Mac. I'm investigating on this.

jslee02:
The failing tests on Mac were fixed in the last commits.

I'm merging this since there is no objection. Some catches of bullet collision detector can be resolved in the next patch or minor version ups.

PR #708, opened 2016-04-28
merge(~) feature for Composite State and Properties

mxgrey:
Up until now, ``Composite::State`` and ``Composite::Properties`` has had a ``copy(~)`` function which will do a deep copy making the contents of one instance exactly the same as the contents of another instance.

This PR adds a ``merge`` function which will allow the receiving instance to hang onto any entries it has which are not present in the sending instance. So suppose we have:

```
Composite::Properties receiver, sender;
sender.create<Aspect1>();
receiver.create<Aspect2>();
receiver.merge(sender);
```

Then ``receiver`` will get the ``Aspect1`` contents of ``sender``, but it will also get to keep its own ``Aspect2`` contents. On the other hand if we used ``receiver.copy(sender)`` instead, then ``receiver`` would lose its ``Aspect2`` contents in order to be a perfect duplicate of ``sender``.

I've also added a ``has<Aspect>()`` function for ``Composite::State`` and ``Composite::Properties``.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/708)

### DIFF
--- a/dart/common/Cloneable.hpp
+++ b/dart/common/Cloneable.hpp
@@ -248,11 +248,27 @@ public:
   /// Map-based move assignment operator
   CloneableMap& operator=(MapType&& otherMap);
 
-  /// Copy the contents of another cloneable map into this one.
-  void copy(const CloneableMap& otherMap);
+  /// Copy the contents of another cloneable map into this one. If merge is set
+  /// to false, then any fields in this map which are not present in the other
+  /// map will be erased; otherwise they will be kept. Setting merge to false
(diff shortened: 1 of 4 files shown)


jslee02:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/708#-:-KGTP9Obg0ZjjLEK-JP9:1417088721)*

Issue #702, opened 2016-04-25
enableSelfCollision has a misleading argument


mkoval:
`enableSelfCollision` takes a `_enableAdjacentBodyCheck` boolean argument. This leads to the unintuitive behavior that this line actually *enables* self-collision checking:
```c++
skeleton->enableSelfCollision(false);
```

It also leads to a strange asymmetry, where there is no way to change the value of `isEnabledAdjacentBodyCheck` without also enabling self-collision detection. I suggest deprecating `enableSelfCollision` and replacing it with two separate functions: one for enabling/disabling self collision checks and one for enabling/disabling adjacent body checks.

jslee02:
Resolved by #718

PR #718, opened 2016-05-08
Separate enableSelfCollision into two concrete functions

jslee02:
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/718)


mkoval:
👍 This resolves #702.

PR #722, opened 2016-05-12
Add warnings for unsupported shape pairs of DARTCollisionDetector

jslee02:
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/722)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## DART 6
 
+### DART 6.0.1 (2016-XX-XX)
+
+* Collision detection
+
+  * Added warnings for unsupported shape pairs of DARTCollisionDetector: [#722](https://github.com/dartsim/dart/pull/722)
+
 ### DART 6.0.0 (2016-05-10)
 
 * Common data structures
@@ -9,7 +15,7 @@
   * Added `ShapeFrame` and `ShapeNode`: [#608](https://github.com/dartsim/dart/pull/608)
   * Added `BoundingBox`: [#547](https://github.com/dartsim/dart/pull/547), [#546](https://github.com/dartsim/dart/issues/546)
 
-* Kinematics features:
+* Kinematics
 
   * Added convenient functions for setting joint limits: [#703](https://github.com/dartsim/dart/pull/703)
   * Added more description on `InverseKinematics::solve()`: [#624](https://github.com/dartsim/dart/pull/624)
@@ -24,7 +30,7 @@
   * Fixed `InverseKinematics` when it's used with `FreeJoint` and `BallJoint`: [#683](https://github.com/dartsim/dart/pull/683)
   * Fixed ambiguous overload on `MetaSkeleton::getLinearJacobianDeriv`: [#628](https://github.com/dartsim/dart/pull/628), [#626](https://github.com/dartsim/dart/issues/626)
 
-* Dynamics features:
+* Dynamics
 
   * Added `get/setLCPSolver` functions to `ConstraintSolver`: [#633](https://github.com/dartsim/dart/pull/633)
   * Added `ServoMotorConstraint` as a preliminary implementation for `SERVO` actuator type: [#566](https://github.com/dartsim/dart/pull/566)
@@ -34,7 +40,7 @@
   * Fixed incorrect applying of joint constraint impulses: [#317](https://github.com/dartsim/dart/issues/317)
   * Deprecated `draw()` functions of dynamics classes: [#654](https://github.com/dartsim/dart/pull/654)
 
-* Collision detection features
+* Collision detection
 
   * Added `CollisionGroup` and refactored `CollisionDetector` to be more versatile: [#711](https://github.com/dartsim/dart/pull/711), [#704](https://github.com/dartsim/dart/pull/704), [#689](https://github.com/dartsim/dart/pull/689), [#631](https://github.com/dartsim/dart/pull/631), [#642](https://github.com/dartsim/dart/issues/642), [#20](https://github.com/dartsim/dart/issues/20)
   * Improved API for self collision checking options: [#718](https://github.com/dartsim/dart/pull/718), [#702](https://github.com/dartsim/dart/issues/702)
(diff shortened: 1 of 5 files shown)


mxgrey:
Reviewed 3 of 4 files at r1.
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/dart/DARTCollide.cpp, line 53 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/722#-KH_lnQoy-6q3xBr4jRE:-KH_lnQoy-6q3xBr4jRF:-1812831523) ([raw file](https://github.com/dartsim/dart/blob/0bbcfa4677a479ad6743f854c6d5ab2ad7723a04/dart/collision/dart/DARTCollide.cpp#L53)):*
> ```cpp
>         << getShapeName(shape1) << "'-'"\
>         << getShapeName(shape2) << "'. "\
>         << "Returning false.\n";
> ```

Maybe tweak this to:

```
  dterr << '[DARTCollisionDetector] Attempting to check for an "\
        << "unsupported shape pair: [" << typeid(*shape1).name()\
        << "] - [" << typeid(*shape2).name() << ]. Returning false.\n";
```

This way it would accurately report the issue even if a non-standard shape gets used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/722)*

jslee02:
*[dart/collision/dart/DARTCollide.cpp, line 53 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/722#-KH_lnQoy-6q3xBr4jRE:-KH_p73nki_agYM_ZwAC:-866479112) ([raw file](https://github.com/dartsim/dart/blob/0bbcfa4677a479ad6743f854c6d5ab2ad7723a04/dart/collision/dart/DARTCollide.cpp#L53)):*
<details><summary><i>Previously, mxgrey (Michael X. Grey) wrote…</i></summary>
> Maybe tweak this to:
> 
> ```
>   dterr << '[DARTCollisionDetector] Attempting to check for an "\
>         << "unsupported shape pair: [" << typeid(*shape1).name()\
>         << "] - [" << typeid(*shape2).name() << ]. Returning false.\n";
> ```
> 
> This way it would accurately report the issue even if a non-standard shape gets used.</details>
:+1: Good idea! I have a little concern that typeid(~).name() sometimes return a little bit mangled name, but I would prefer it since it's more flexible and the mangled​ name s are still reasonably readable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/722)*

mxgrey:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/dart/DARTCollide.cpp, line 53 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/722#-KH_lnQoy-6q3xBr4jRE:-KH_q4IsfHFz22Wzk-Lw:1120441099) ([raw file](https://github.com/dartsim/dart/blob/0bbcfa4677a479ad6743f854c6d5ab2ad7723a04/dart/collision/dart/DARTCollide.cpp#L53)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary>
> :+1: Good idea! I have a little concern that typeid(~).name() sometimes return a little bit mangled name, but I would prefer it since it's more flexible and the mangled​ name s are still reasonably readable.</details>
Yeah, it would be nice if a de-mangled "pretty print" version were available for typeid(~).name(), but we can always de-mangle the printout while debugging.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/722)*

mkoval:
Review status: 2 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/dart/DARTCollide.cpp, line 53 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/722#-KH_lnQoy-6q3xBr4jRE:-KHaHa_GV4J2PD22iikI:1176165657) ([raw file](https://github.com/dartsim/dart/blob/0bbcfa4677a479ad6743f854c6d5ab2ad7723a04/dart/collision/dart/DARTCollide.cpp#L53)):*
<details><summary><i>Previously, mxgrey (Michael X. Grey) wrote…</i></summary>
> Yeah, it would be nice if a de-mangled "pretty print" version were available for typeid(~).name(), but we can always de-mangle the printout while debugging.</details>
I would prefer to add virtual `getType` and a static getStaticType` functions, like we currently have on `Joint`. That way we get the best of both worlds: (1) no RTTI, which is brittle when using dynamic linking, (2) a readable name, and (3) user-extensibility. Thoughts?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/722)*

jslee02:
*[dart/collision/dart/DARTCollide.cpp, line 53 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/722#-KH_lnQoy-6q3xBr4jRE:-KHb2wrMQXvwBF-yNG_n:1735911922) ([raw file](https://github.com/dartsim/dart/blob/0bbcfa4677a479ad6743f854c6d5ab2ad7723a04/dart/collision/dart/DARTCollide.cpp#L53)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> I would prefer to add virtual `getType` and a static getStaticType` functions, like we currently have on `Joint`. That way we get the best of both worlds: (1) no RTTI, which is brittle when using dynamic linking, (2) a readable name, and (3) user-extensibility. Thoughts?</details>
My only concern is the performance of type checking. The cost would be virtual function call + string comparing. As I known, string comparing is just pointer comparison so virtual function call is the only overhead, right?

I don't have an objection to​ using type functions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/722)*

jslee02:
Review status: 2 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/dart/DARTCollide.cpp, line 53 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/722#-KH_lnQoy-6q3xBr4jRE:-KHbu0K96otD4o1zZ_-R:-188028785) ([raw file](https://github.com/dartsim/dart/blob/0bbcfa4677a479ad6743f854c6d5ab2ad7723a04/dart/collision/dart/DARTCollide.cpp#L53)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary>
> My only concern is the performance of type checking. The cost would be virtual function call + string comparing. As I known, string comparing is just pointer comparison so virtual function call is the only overhead, right?
> 
> I don't have an objection to​ using type functions.</details>
I remember that there was one reason we stayed with enumerate​. The shape type wouldn't be easy to be extended with full support of related functionalities such as rendering and collision detection. Also, in general, `switch` statement with enumerate would be cheaper than `if-else if-else` statement. However, now I'm inclined to having `virtual getType()` because of consistency and that this is not the performance critical.

This can be handled in a separate PR. I'm merging this PR if there is no objection.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/722)*

Issue #724, opened 2016-05-12
Proposal: SphereShape


jslee02:
This might be silly (and was declined before), but I'd like to try one more time. :sweat: 

We have used `EllipsoidShape` to represent sphere by setting the radii to be the same since a sphere is a special case of an ellipsoid. However, we could use the speciality of the sphere to take the advantage of cheaper collision checking algorithms for spheres from the collision detectors. We could still use ellipsoid to create sphere shape in the collision detectors when the radii are the same, but modifying the radii not to be the same would require to change the collision detector's shape to ellipsoid (and vice versa) once we have [collision shape refresh feature](https://github.com/dartsim/dart/issues/368). Still we could handle this using ellipsoid, but I imagine having `SphereShape` would be more straigtforward and clean in implementation.

mxgrey:
From an abstraction standpoint it would seem silly for ``SphereShape`` to be its own class, independent from ``EllipsoidShape``, but I agree that from an implementation standpoint it completely makes sense, because the machinery used to collision check an ``EllipsoidShape`` is so different from the optimized machinery that could get used for ``SphereShape``. :+1:

psigen:
I agree that this is an extremely useful special case to handle independently.

If anything, the `EllipsoidShape` class might do some special casing to handle itself like a `SphereShape` (or even internally construct a `SphereShape`) when radii are equal, but there are many places where we would want to semantically distinguish spheres from ellipsoids.

I guess another alternative would be a lot of boilerplate involving branching on an `isSphere` check on `EllipsoidShape`, but I don't see a huge downside to simply having another Shape.

mxgrey:
> I guess another alternative would be a lot of boilerplate involving branching on an isSphere check on EllipsoidShape

[...]

PR #744, opened 2016-07-20
Distance API

jslee02:
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/744)


mkoval:
This looks good. I posted a few comments on Reviewable. The only two major ones are:

- The current behavior makes it impossible to compute signed distance because it terminates as soon as the signed distance is `<= 0`.
- We should verify which `Shape` pairs work in FCL and either: (1) disable the other pairs, (2) internally convert them to meshes for distance queries, or (3) print an error if we use them in a distance query.

---

Reviewed 23 of 23 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[dart/collision/CMakeLists.txt, line 20 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAGPRr6P85JAavNrmW:-KNAGPRs6KAA2njwHevV:-469590015) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/CMakeLists.txt#L20)):*
> ```CMake
> dart_get_filename_components(header_names "collision headers" ${hdrs})
> # TODO: remove below line once the files are completely removed.
> list(REMOVE_ITEM header_names "Option.hpp" "Result.hpp")
> ```

Why remove the headers from `header_names` while they still exist?

---

*[dart/collision/CollisionDetector.cpp, line 64 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAGZ72VMYDYchi8q3d:-KNAGZ72VMYDYchi8q3e:1257946951) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/CollisionDetector.cpp#L64)):*
> ```cpp
> 
>   return;
> }
> ```

It bothers me that these functions have dummy implementations, but `collide` is left as pure-`virtual`. I don't see any reason for the inconsistency.

I would prefer that both functions be left as pure `virtual` and the stub method be implemented implemented by each collision detector that does not support distance queries.

---

*[dart/collision/CollisionDetector.hpp, line 114 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAHC1zyqv5s6g3OmZa:-KNAHC2-lWMlMrNPrawl:-579142835) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/CollisionDetector.hpp#L114)):*
> ```cpp
>       CollisionResult* result = nullptr) = 0;
> 
>   /// Perform collision check for a single group.
> ```

This comment is inaccurate.. It would be good to specify explicitly that (1) this is _signed distance_ and (2) result reflects the minimum distance between any two `CollisionObjects` in the group.

Please reply if either of my assumptions are incorrect here. It's not entirely obvious what the single-argument `distance` call should do: the distance from a set to itself is always zero.

---

*[dart/collision/CollisionGroup.hpp, line 185 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAHseo5pjlkaonOT6f:-KNAHseo5pjlkaonOT6g:-151012106) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/CollisionGroup.hpp#L185)):*
> ```cpp
>       CollisionResult* result = nullptr);
> 
>   /// Perform distance check within this CollisionGroup.
> ```

It would be good to clarify the docstrings in this file, too.

---

*[dart/collision/DistanceOption.hpp, line 45 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAIHB5Yh3ad8UuLRuM:-KNAIHB5Yh3ad8UuLRuN:1650688485) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/DistanceOption.hpp#L45)):*
> ```cpp
> struct DistanceOption
> {
>   bool enableNearestPoints;
> ```

Comment?

---

*[dart/collision/DistanceOption.hpp, line 48 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAIHtvT44ai-VWZ2SJ:-KNAIHtvT44ai-VWZ2SK:-1097562342) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/DistanceOption.hpp#L48)):*
> ```cpp
> 
>   /// Distance filter
>   std::shared_ptr<DistanceFilter> distanceFilter;
> ```

Comment? Also, what does it mean if this is `nullptr`, like it is by default?

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1175 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJIeniVf1N6p-3t3W:-KNAJIeniVf1N6p-3t3X:1850268203) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1175)):*
> ```cpp
> 
>   // Perform narrow-phase check
>   fcl::distance(o1, o2, fclRequest, fclResult);
> ```

Does this change the `fclResult.result` pointer or only populate its value? I am concerned that you're storing a pointer by value above, so any change in the _pointer_ will not be reflected below.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNAJVChYnPPukB-ROc7:-1013064734) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
> ```cpp
>   // TODO(JS): Do we don't care what's the actual minimum distance when
>   // collision is found? Or should we continue to the narrow check to check
>   // every pairs?
> ```

We definitely care about the penetration distance. I suggest allowing the user to specify a "minimum distance" that they care about. The user pass `-INFINITY` if they want to compute signed distance.

---

*[unittests/testDistance.cpp, line 288 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAK2fsbJchlP_XjvG5:-KNAK2ft3XexGRiQuiyF:-1577776758) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/unittests/testDistance.cpp#L288)):*
> ```cpp
> 
>   auto dart = DARTCollisionDetector::create();
>   testSphereSphere(dart);
> ```

Doesn't only FCL implement this functionality? How do the Bullet and DART tests pass?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744#-:-KNAKZKWoYX_PCJANc51:-1929722112)*

mkoval:

jslee02:
Review status: 10 of 29 files reviewed at latest revision, 9 unresolved discussions.

---

*[dart/collision/CMakeLists.txt, line 20 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAGPRr6P85JAavNrmW:-KNCDc8MYOBsAaXRPo2y:526600029) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/CMakeLists.txt#L20)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Why remove the headers from `header_names` while they still exist?
</details>
They are deprecated by `CollisionOption.hpp` and `CollisionResult.hpp`, and `Option.hpp` and `Result.hpp` emit warning messages when included. People wouldn't like to see the warnings when they just included `collision.hpp` or `dart.hpp`.

---

*[dart/collision/CollisionDetector.hpp, line 114 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAHC1zyqv5s6g3OmZa:-KNEMTXHldqBZJfaFmF8:-499146009) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/CollisionDetector.hpp#L114)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> This comment is inaccurate.. It would be good to specify explicitly that (1) this is _signed distance_ and (2) result reflects the minimum distance between any two `CollisionObjects` in the group.
> 
> Please reply if either of my assumptions are incorrect here. It's not entirely obvious what the single-argument `distance` call should do: the distance from a set to itself is always zero.
</details>
Done.

---

*[dart/collision/CollisionGroup.hpp, line 185 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAHseo5pjlkaonOT6f:-KNEMUeBue1stQobyMuc:-499146009) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/CollisionGroup.hpp#L185)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> It would be good to clarify the docstrings in this file, too.
</details>
Done.

---

*[dart/collision/DistanceOption.hpp, line 45 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAIHB5Yh3ad8UuLRuM:-KNEMVB6aTwyivvY21_U:-499146009) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/DistanceOption.hpp#L45)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Comment?
</details>
Done.

---

*[dart/collision/DistanceOption.hpp, line 48 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAIHtvT44ai-VWZ2SJ:-KNEMVv-N4z_N6Mo5EpK:-499146009) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/DistanceOption.hpp#L48)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Comment? Also, what does it mean if this is `nullptr`, like it is by default?
</details>
Done.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1175 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJIeniVf1N6p-3t3W:-KNCFSfheg3tbMs5no0d:866043438) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1175)):*
> fcl::distance(o1, o2, fclRequest, fclResult);
Which one do you refer to between `fclResult` and `result`? `fclResult` is referenced and the value is changed by `fcl::distance`, and `result` is a pointer but passed in `interpreteDistanceResult` as a reference.

Sorry if I didn't get you right.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNE0oUOooGGDlk4p7cx:1342164317) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> We definitely care about the penetration distance. I suggest allowing the user to specify a "minimum distance" that they care about. The user pass `-INFINITY` if they want to compute signed distance.
</details>
Agree on having "minimum distance" in Option. Also, I think zero would be good for the default value.

---

*[unittests/testDistance.cpp, line 288 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAK2fsbJchlP_XjvG5:-KNCD-Re6KRfn_3w163O:1353034579) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/unittests/testDistance.cpp#L288)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Doesn't only FCL implement this functionality? How do the Bullet and DART tests pass?
</details>
The secret is in the test function:
```cpp
  if (cd->getType() != collision::FCLCollisionDetector::getStaticType())
  {
    dtwarn << "Aborting test: distance check is not supported by "
           << cd->getType() << ".\n";
    return;
  }
```
:smiley: 

---

*[dart/collision/CollisionDetector.cpp, line 64 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAGZ72VMYDYchi8q3d:-KNEMYHM5wdPTjNEmykj:-499146009) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/CollisionDetector.cpp#L64)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> It bothers me that these functions have dummy implementations, but `collide` is left as pure-`virtual`. I don't see any reason for the inconsistency.
> 
> I would prefer that both functions be left as pure `virtual` and the stub method be implemented implemented by each collision detector that does not support distance queries.
> 
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

mkoval:
Reviewed 1 of 5 files at r2, 19 of 19 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[CHANGELOG.md, line 3 \[r3\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNEN7kawR0gUvHYHG0W:-KNEN7kawR0gUvHYHG0X:361481756) ([raw file](https://github.com/dartsim/dart/blob/19d06e159cb5b1f2e17087495cc6e6e729fdbfe4/CHANGELOG.md#L3)):*
> ```Markdown
> ## DART 6
> 
> ### DART 6.1.0 (2016-XX-XX)
> ```

Missing a `CHANGELOG` entry for distance measurements. Or do you do this later in the release process?

---

*[CHANGELOG.md, line 7 \[r3\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNEN4l90VK3WwwHgguJ:-KNEN4l90VK3WwwHgguK:-2080564008) ([raw file](https://github.com/dartsim/dart/blob/19d06e159cb5b1f2e17087495cc6e6e729fdbfe4/CHANGELOG.md#L7)):*
> ```Markdown
> * Misc improvements and bug fixes
> 
>   * Added `virtual Shape::getType()` and deprecated `ShapeType Shpae::getShapeType()`: [#724](https://github.com/dartsim/dart/pull/724)
> ```

Typo: `Shpae`. :smile: 

---

*[dart/collision/DistanceResult.hpp, line 63 \[r3\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNENoQz_L7KuCAo5-x5:-KNENoQz_L7KuCAo5-x6:1878008095) ([raw file](https://github.com/dartsim/dart/blob/19d06e159cb5b1f2e17087495cc6e6e729fdbfe4/dart/collision/DistanceResult.hpp#L63)):*
> ```cpp
>   /// The nearest point on shapeFrame1. The point is calculated only when
>   /// DistanceOption::enableNearestPoints is true, which is false by default.
>   Eigen::Vector3d nearestPoint1;
> ```

Are the coordinates expressed w.r.t. `shapeFrame1` or `WorldFrame`?

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNEORkxqdhYC7i61SyD:-1666217740) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary>
> Agree on having "minimum distance" in Option. Also, I think zero would be good for the default value.
</details>
I would prefer `-INFINITY` as the default value since all of the use-cases I can think of (e.g. anything involving optimization) require signed distance.

However, I am not strongly opposed to `0.0` if you have a good justification for that default. Do you have a particular use case in mind?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

jslee02:
Review status: 27 of 29 files reviewed at latest revision, 4 unresolved discussions.

---

*[CHANGELOG.md, line 3 \[r3\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNEN7kawR0gUvHYHG0W:-KNEPMMloE7sC888ImrI:-196463082) ([raw file](https://github.com/dartsim/dart/blob/19d06e159cb5b1f2e17087495cc6e6e729fdbfe4/CHANGELOG.md#L3)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Missing a `CHANGELOG` entry for distance measurements. Or do you do this later in the release process?
</details>
I'm switching to updating changelog when the change happens. So yeah this should be updated in this PR.

---

*[CHANGELOG.md, line 7 \[r3\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNEN4l90VK3WwwHgguJ:-KNERBHapjzClZYbUYfy:-499146009) ([raw file](https://github.com/dartsim/dart/blob/19d06e159cb5b1f2e17087495cc6e6e729fdbfe4/CHANGELOG.md#L7)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Typo: `Shpae`. :smile: 
</details>
Done.

---

*[dart/collision/DistanceResult.hpp, line 63 \[r3\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNENoQz_L7KuCAo5-x5:-KNERC2a9NQ9tQgb-I32:-83881297) ([raw file](https://github.com/dartsim/dart/blob/19d06e159cb5b1f2e17087495cc6e6e729fdbfe4/dart/collision/DistanceResult.hpp#L63)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Are the coordinates expressed w.r.t. `shapeFrame1` or `WorldFrame`?
</details>
It's the world coordinates. Updated the comment.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNERXeqYxdVH2vqYJzF:762472425) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> I would prefer `-INFINITY` as the default value since all of the use-cases I can think of (e.g. anything involving optimization) require signed distance.
> 
> However, I am not strongly opposed to `0.0` if you have a good justification for that default. Do you have a particular use case in mind?
</details>
I also don't have a strong objection to `-INFINITY`.

I "anticipated" that people would expect positive value because my first impression was so, but you may have a closer feel to what planning people would expect from this function.

For clarification, we could have both functions as `distance` with zero threshold and `signedDistance` with `-INFINITY`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

mkoval:
Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNETuwjaaw_EiAysaAD:-211884535) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary>
> I also don't have a strong objection to `-INFINITY`.
> 
> I "anticipated" that people would expect positive value because my first impression was so, but you may have a closer feel to what planning people would expect from this function.
> 
> For clarification, we could have both functions as `distance` with zero threshold and `signedDistance` with `-INFINITY`.
> 
> 
</details>
I have generally seen signed distance used in planning and perception code. The problem with the "intuitive" definition of distance is that: (1) no gradient information is available when you're in contact and (2) there is a discontinuity in the derivative at zero.

I see some implementation issuers with defining separate `signedDistance` and `distance` functions:

- `distance` can still return negative values in the current implementation, so the behavior is unintuitive (I would expect `0.0` for any objects in collision).
- `distance` would have to ignore the `minimumDistanceThreshold` option
- `signedDistance` would not compute signed distance with the default value of `minimumDistanceThreshold`

Overall, I would prefer keeping _one function_ and change the default value of `minimumDistanceThreshold` to `-INFINITY`. I slightly favor the name `distance` over `signedDistance` because it only computes signed distance when `minimumDistanceThreshold` is `-INFINITY`. I do not feel too strongly about this aspect, though.

Maybe @cdellin, @psigen, or @mklingen have a stronger opinion?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNE_A1i2jNLua71llqY:347117390) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> I have generally seen signed distance used in planning and perception code. The problem with the "intuitive" definition of distance is that: (1) no gradient information is available when you're in contact and (2) there is a discontinuity in the derivative at zero.
> 
> I see some implementation issuers with defining separate `signedDistance` and `distance` functions:
> 
> - `distance` can still return negative values in the current implementation, so the behavior is unintuitive (I would expect `0.0` for any objects in collision).
> - `distance` would have to ignore the `minimumDistanceThreshold` option
> - `signedDistance` would not compute signed distance with the default value of `minimumDistanceThreshold`
> 
> Overall, I would prefer keeping _one function_ and change the default value of `minimumDistanceThreshold` to `-INFINITY`. I slightly favor the name `distance` over `signedDistance` because it only computes signed distance when `minimumDistanceThreshold` is `-INFINITY`. I do not feel too strongly about this aspect, though.
> 
> Maybe @cdellin, @psigen, or @mklingen have a stronger opinion?
> 
> 
</details>
I just chatted with @mkoval, and what I think we arrived at was that the function should be called `distance`, and that it should always return a clamped distance, with the `minimumDistanceThreshold` defaulting `0.0`. (That is, if the actual distance is less than `minimumDistanceThreshold`, the function will return `minimumDistanceThreshold`.)  This is easy to motivate and understand in the documentation.  Separately, the caller may want to know the first distance found that caused the clamp to occur, so there should be some mechanism for the caller to discover this information (similar to how the caller can ask which pair of bodies caused a collision check to shortcut to true).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

mkoval:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNEnx7nSeWwQoLsK3eR:-2052334559) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, cdellin wrote…</i></summary>
> I just chatted with @mkoval, and what I think we arrived at was that the function should be called `distance`, and that it should always return a clamped distance, with the `minimumDistanceThreshold` defaulting `0.0`. (That is, if the actual distance is less than `minimumDistanceThreshold`, the function will return `minimumDistanceThreshold`.)  This is easy to motivate and understand in the documentation.  Separately, the caller may want to know the first distance found that caused the clamp to occur, so there should be some mechanism for the caller to discover this information (similar to how the caller can ask which pair of bodies caused a collision check to shortcut to true).
</details>
I agree with @cdellin. To summarize:

- Have one function called `distance`.
- Clamp the output of the function to `std::max(actualDistance, minimumDistanceThreshold)`.
- Expose the true `actualDistance` value (even if it's less than `minimumDistanceThreshold`) as a field in `DistanceResult`.
- Set the default `minimumDistanceThreshold` to `0.0`.
- Update the documentation accordingly.

@jslee02 What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNExhT5SXDyr-GnVpIf:-507978386) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> I agree with @cdellin. To summarize:
> 
> - Have one function called `distance`.
> - Clamp the output of the function to `std::max(actualDistance, minimumDistanceThreshold)`.
> - Expose the true `actualDistance` value (even if it's less than `minimumDistanceThreshold`) as a field in `DistanceResult`.
> - Set the default `minimumDistanceThreshold` to `0.0`.
> - Update the documentation accordingly.
> 
> @jslee02 What do you think?
</details>
And by `actualDistance`, we mean the first distance found that is lower than `minimumDistanceThreshold`, correct? Not the actual distance. (This would be as expensive to find as with no clamping.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

mkoval:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNEy9jjuv3OydNzOhDt:965377692) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, cdellin wrote…</i></summary>
> And by `actualDistance`, we mean the first distance found that is lower than `minimumDistanceThreshold`, correct? Not the actual distance. (This would be as expensive to find as with no clamping.)
</details>
Yes, that's what I mean. The fact that the best name for it is `firstDistanceLessThanMinimumDistanceThreshold` is a good indication that it's a _bad value to return by default_. :sweat: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

jslee02:
Review status: 24 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNFUjNTbSJWylqB3s99:-474835740) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Yes, that's what I mean. The fact that the best name for it is `firstDistanceLessThanMinimumDistanceThreshold` is a good indication that it's a _bad value to return by default_. :sweat: 
</details>
I ​agree with you but have some questions:

What's the motivation behind of clamping? Would it be worthwhile even if the distance doesn't refer to actual distance? Also, I would concern the corner case of that there is no distance less than `minimumDistanceThreshold`. In this case, `distance` would return `minimumDistanceThreshold` when every two points (from distinct objects) in the group are farther than `minimumDistanceThreshold`.

Do you mean that `distance` returns the clamped dista​nce and also have it in `DistanceResult`?

Why do we prefer `0.0` for `minimumDistanceThreshold` now?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: 24 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNFghrykaowJ2QXdnsI:614591566) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary>
> I ​agree with you but have some questions:
> 
> What's the motivation behind of clamping? Would it be worthwhile even if the distance doesn't refer to actual distance? Also, I would concern the corner case of that there is no distance less than `minimumDistanceThreshold`. In this case, `distance` would return `minimumDistanceThreshold` when every two points (from distinct objects) in the group are farther than `minimumDistanceThreshold`.
> 
> Do you mean that `distance` returns the clamped dista​nce and also have it in `DistanceResult`?
> 
> Why do we prefer `0.0` for `minimumDistanceThreshold` now?
> 
> 
</details>
My understanding is that there are situations where the caller doesn't care what the distance is if it's below a certain value (whether that threshold is negative or positive, e.g. padding). For example, Mike gave the sample of covering a path with free C-space bubbles, where the distance is only useful if it is positive.  This can be represented elegantly by modifying the API to return a clamped value.  In these cases, the distance function can often return more quickly.

I must admit I don't understand your corner case example.  If there is no distance less than `minimumDistanceThreshold`, then the actual distance `d`, which would be returned with no clamping (i.e. if `minimumDistanceThreshold = -INFINITY`), would be greater than `minimumDistanceThreshold`, so the function would return `d`, and no clamping would be performed.

You might have been referring to the opposite bound -- a `maximumDistanceThreshold`.  There, if `d` were larger than this value (e.g. if every two points from distinct objects are farther than `maximumDistanceThreshold`, then the result would clamp to this value.  (In fact, if you squint, you could argue that a binary collision checker is equivalent to a clamped distance call, with `minimumDistanceThreshold = -EPSILON` and `maximumDistanceThreshold = +EPSILON`.)

I'll let @mkoval give his motivation for the extra data that would be available in `DistanceResult` -- I think the idea is that this would be the first discovered unclamped value outside of the clamping range, but not necessarily the minimum distance across all bodies, because the function returned early.

We also felt that a default of `0.0` might make more sense for a function called `distance` (as opposed to `signedDistance`), although I'm personally flexible on that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

jslee02:
Review status: 24 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNJDQICJH-HEBJmgm8d:1262531316) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, cdellin wrote…</i></summary>
> My understanding is that there are situations where the caller doesn't care what the distance is if it's below a certain value (whether that threshold is negative or positive, e.g. padding). For example, Mike gave the sample of covering a path with free C-space bubbles, where the distance is only useful if it is positive.  This can be represented elegantly by modifying the API to return a clamped value.  In these cases, the distance function can often return more quickly.
> 
> I must admit I don't understand your corner case example.  If there is no distance less than `minimumDistanceThreshold`, then the actual distance `d`, which would be returned with no clamping (i.e. if `minimumDistanceThreshold = -INFINITY`), would be greater than `minimumDistanceThreshold`, so the function would return `d`, and no clamping would be performed.
> 
> You might have been referring to the opposite bound -- a `maximumDistanceThreshold`.  There, if `d` were larger than this value (e.g. if every two points from distinct objects are farther than `maximumDistanceThreshold`, then the result would clamp to this value.  (In fact, if you squint, you could argue that a binary collision checker is equivalent to a clamped distance call, with `minimumDistanceThreshold = -EPSILON` and `maximumDistanceThreshold = +EPSILON`.)
> 
> I'll let @mkoval give his motivation for the extra data that would be available in `DistanceResult` -- I think the idea is that this would be the first discovered unclamped value outside of the clamping range, but not necessarily the minimum distance across all bodies, because the function returned early.
> 
> We also felt that a default of `0.0` might make more sense for a function called `distance` (as opposed to `signedDistance`), although I'm personally flexible on that.
> 
</details>
Thank you for the answers. Now I get a ​better understanding.

The corner case example was incorrect. Somehow I thought every distance less than the threshold is rejected.

Before I make changes based on the discussion, I would like to also replace the name `minimumDistanceThreshold` with `minimumDistanceResolution`. The minimumDistance**Threshold** sounds like a minimum distance for a valid distance for the check (like I was confused).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNKcHFc_po162n9PUZh:447397602) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary>
> Thank you for the answers. Now I get a ​better understanding.
> 
> The corner case example was incorrect. Somehow I thought every distance less than the threshold is rejected.
> 
> Before I make changes based on the discussion, I would like to also replace the name `minimumDistanceThreshold` with `minimumDistanceResolution`. The minimumDistance**Threshold** sounds like a minimum distance for a valid distance for the check (like I was confused).
</details>
I'm not sure `minimumDistanceResolution` is a good name.

Resolution often implies a discretization level in the collision check.    For example, OpenRAVE uses the terminology `DOFResolution` to describe the discretization of joint DOF values that are queried when doing collision checking.  It also has similar connotations in voxelized fields, including signed distance fields based on structures like occupancy grids.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

jslee02:
Review status: 24 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNKdHel5jhEDM2Kow1P:667841895) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, psigen (Pras Velagapudi) wrote…</i></summary>
> I'm not sure `minimumDistanceResolution` is a good name.
> 
> Resolution often implies a discretization level in the collision check.    For example, OpenRAVE uses the terminology `DOFResolution` to describe the discretization of joint DOF values that are queried when doing collision checking.  It also has similar connotations in voxelized fields, including signed distance fields based on structures like occupancy grids.
</details>
That makes sense. `minimumDistanceResolution` got two down votes today (and now another). :sweat: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

jslee02:
Review status: 24 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNKdyxycsxNmWz8yoiB:2118431355) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary>
> That makes sense. `minimumDistanceResolution` got two down votes today (and now another). :sweat: 
</details>
Those are good examples the term of "resolution" though. Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: 24 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNKlcm_PWXsC6a153Ab:1324693201) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary>
> Those are good examples the term of "resolution" though. Thanks!
</details>
I agree that `minimumDistanceThreshold` might be confusing -- what about `distanceClampLower`, `distanceClampMinimum`, or something to that effect? (See http://www.boost.org/doc/libs/1_54_0/libs/algorithm/doc/html/algorithm/Misc.html for example terminology.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNNrCStylphTS6Pct-l:1682623713) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, cdellin wrote…</i></summary>
> I agree that `minimumDistanceThreshold` might be confusing -- what about `distanceClampLower`, `distanceClampMinimum`, or something to that effect? (See http://www.boost.org/doc/libs/1_54_0/libs/algorithm/doc/html/algorithm/Misc.html for example terminology.)
</details>
Hmm, `clamp` usually refers to an operation, not a bound.  And it's semantically not quite right because the clamp distance might change the underlying logic of the check (at least according to @cdellin's previous comment).

1. We seem to be conflating two parameters here.  One is the distance within which this function should terminate if a collision is found.  The other is a clamping value for the distance output.

2. It seems like the main reason to have the parameter `minimumDistanceThreshold` is to allow optimization in the first case on the check.  If there was no performance improvement from specifying that parameter, then it seems like just returning the signed distance would be the simplest and clearest option (letting the user clamp the `actualDistance` on their own if they wanted to).

3. As far as I can tell, there does not seem to be any use case in which setting a `maximumDistanceThreshold` would ever be able to help performance, and thus no reason to add this parameter other than syntactic sugar for clamping.

----

I propose:

* We add a single parameter `distanceThreshold` and describe additional detail in the docstring rather than the name.  As far as I can tell, this parameter should be: _the distance below which the function will return immediately if a collision is found_.  The default value of this parameter should be `0.0`.  This allows positive values to represent padding, and negative values to represent acceptable interpenetration (for compliant objects).

* We return only `actualDistance` in the result structure.  If we are computing it anyway, we might as well return the value that matches the body information that we are returning.

Other than people possibly being surprised that `distance < 0.0` is possible, I'm not seeing any cases where returning `distance < 0.0` is actually problematic.  Any use cases that I am missing here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: 24 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNNxTmwsvZUef-VxFZ9:1420188955) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, psigen (Pras Velagapudi) wrote…</i></summary>
> Hmm, `clamp` usually refers to an operation, not a bound.  And it's semantically not quite right because the clamp distance might change the underlying logic of the check (at least according to @cdellin's previous comment).
> 
> 1. We seem to be conflating two parameters here.  One is the distance within which this function should terminate if a collision is found.  The other is a clamping value for the distance output.
> 
> 2. It seems like the main reason to have the parameter `minimumDistanceThreshold` is to allow optimization in the first case on the check.  If there was no performance improvement from specifying that parameter, then it seems like just returning the signed distance would be the simplest and clearest option (letting the user clamp the `actualDistance` on their own if they wanted to).
> 
> 3. As far as I can tell, there does not seem to be any use case in which setting a `maximumDistanceThreshold` would ever be able to help performance, and thus no reason to add this parameter other than syntactic sugar for clamping.
> 
> ----
> 
> I propose:
> 
> * We add a single parameter `distanceThreshold` and describe additional detail in the docstring rather than the name.  As far as I can tell, this parameter should be: _the distance below which the function will return immediately if a collision is found_.  The default value of this parameter should be `0.0`.  This allows positive values to represent padding, and negative values to represent acceptable interpenetration (for compliant objects).
> 
> * We return only `actualDistance` in the result structure.  If we are computing it anyway, we might as well return the value that matches the body information that we are returning.
> 
> Other than people possibly being surprised that `distance < 0.0` is possible, I'm not seeing any cases where returning `distance < 0.0` is actually problematic.  Any use cases that I am missing here?
> 
> 
</details>
Hmm.  I agree entirely with your `#2` ... if there were not opportunity for performance improvement, I would prefer that this function just return the true (signed) distance, and let the caller clamp or test against whatever bound they want.  In fact, I would be open to the default value of this minimum parameter to be `-INFINITY`, so that by default the function does exactly this.

The reason I'm open to adding this threshold/clamp/bound parameter (whatever it's called) is because I believe it is in fact easy to have a much faster call if the implementation knows that the caller only cares about the true distance if it is above some minimum.  While looping through object pairs, it can shortcut out once it finds a distance that is below the clamp value.   I think this conflicts with your `#3` (although I may have misunderstood).

The biggest question I think is what the implementation should do when it finds a inter-object distance below the clamp threshold.  Option A is to return this out-of-bounds distance. Option B is to return the clamp threshold, and store the value of the first violation in an auxiliary location.  I prefer option B because the primary result is repeatable and a property of the scene, not the particular order that an implementation chooses to consider the pairs.  We can also therefore test the result.

I think this parallels very nicely with the boolean `collision` API.  By default, it shortcuts on the first collision found. The primary return value is `true` in that case.  Which pair happened to have provoked the shortcutting behavior is available in a collision report of some kind, usually for debugging purposes, but it shouldn't be treated as a trusted quantity -- it surely isn't guaranteed to be the "worst" collision, after all.

As a consequence, I'm not a big fan of your proposed documentation for the `distanceThreshold` parameter, because I think it enforces too much about the underlying implementation (even the fact that it will iterate through pairs and shortcut).  I could imagine for some checkers that it may actually be preferable to simply compute the actual distance, and then compare it afterwards to the threshold value.  I prefer to only specify in the API that the function will return the distance within a clamped range. This allows the implementation to take advantage of the shortcutting logic internally if it makes sense. I also worry that if the primary return value is that of the first violation that the implementation happened to find, callers will incorrectly think that this is the actual (worst) distance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

mkoval:
Review status: 24 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNNxT997vq113AYz7LD:1289806491) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
> Other than people possibly being surprised that distance < 0.0 is possible, I'm not seeing any cases where returning distance < 0.0 is actually problematic. Any use cases that I am missing here?

It is unintuitive that I can set `distanceThreshold` and still get a result that is less than that threshold. In fact, since `distance = distanceThreshold` is a lower-dimensional set, I will *always* receive a return value less than the threshold. This _seriously_ violates the "principle of least surprise."

@cdellin pointed out that distances greater than `distanceThreshold` are a _property of only the environment_ and are independent of the implementation up to numerical precision (and ignoring implementation bugs). Values less than `distanceThreshold` _are_ implementation-defined and may change at any time.

Consider your example (emphasis mine):

> This allows positive values to represent padding, and _negative values to represent acceptable interpenetration (for compliant objects)_.

It would be dangerous because it depends on undefined behavior. Suppose you upgrade DART (or any of its dependencies) and the behavior of the collision detector changes to always terminate with `actualDistance = distanceThreshold`, which is _a completely valid_ implementation of the API. Suddenly your compliance model stops working.

The _correct_ way to implement this would be to choose `distanceThreshold = -epsilon` and use the values `-epsilon <= distance < 0` to model compliance. Or to run a normal `collide` query and inspect the contact points.

The *only* use-case I can find for returning `actualDistance` is debugging spurious collisions caused by numerical imprecision. This is served just as well by a clearly labeled "here be dragons" option in the `Result` struct.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNO2auHxNwK6wjcBP5t:2066290402) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
@cdellin:

> As a consequence, I'm not a big fan of your proposed documentation for the distanceThreshold parameter, because I think it enforces too much about the underlying implementation [...]

Yeah, this makes sense.  I think the point is really that `distanceThreshold` specifies the distance within which you care about collisions.  You know nothing about the order of checks, or if shortcutting will go on, but you do know that if there are two objects closer than this distance, they will be returned, and anything else may or may not be checked.

> The biggest question I think is what the implementation should do when it finds a inter-object distance below the clamp threshold. Option A is to return this out-of-bounds distance. Option B is to return the clamp threshold, and store the value of the first violation in an auxiliary location. I prefer option B because the primary result is repeatable and a property of the scene, not the particular order that an implementation chooses to consider the pairs. We can also therefore test the result.

I think that Option A is fine if we make the guarantee that the value returned will definitely be `<=` the threshold. We can consistently test the result this way, and detectors that can provide more accurate information are free to do so. The alternative is going to involve floating point equality testing, which isn't exactly great either.

@mkoval:

> It is unintuitive that I can set distanceThreshold and still get a result that is less than that threshold. In fact, since distance = distanceThreshold is a lower-dimensional set, I will always receive a return value less than the threshold. This seriously violates the "principle of least surprise."

I don't see that as being unintuitive at all.  It's the same as setting a frustrum on a ray-cast or bounding parameters on any other check.  You are setting a threshold for the logic of the function here, not the output value:  If there are two objects that are closer than this distance, then this function will return them.  If there are not, then nothing will be returned.

> It would be dangerous because it depends on undefined behavior. Suppose you upgrade DART (or any of its dependencies) and the behavior of the collision detector changes to always terminate with actualDistance = distanceThreshold, which is a completely valid implementation of the API. Suddenly your compliance model stops working.

I'm not sure what this second example is about.  I'm talking about modeling a collision with an object that can be intersected a certain amount during motion, like a foam or rubber surface.  I think you are thinking of using this to model a spring force, which is not my intent.  And I am talking exactly about choosing `distanceThreshold=+epsilon` for padding checks and `distanceThreshold=-epsilon` for compliant collisions, so I'm not sure what the _correct_ part is correcting.

> The only use-case I can find for returning actualDistance is debugging spurious collisions caused by numerical imprecision. This is served just as well by a clearly labeled "here be dragons" option in the Result struct.

Agreed, there are no guarantees on `actualDistance`between collision detectors, so I don't really see the point in clamping it one way or another either.  As @cdellin pointed out, you could have an underlying binary collision detector which might not reveal the distance at all.  Or a detector which shortcuts mesh-mesh collision by returning on _any distance_ below the collision threshold, which would return an `actualDistance` which is between the true minimum distance between the objects and `distanceThreshold`.   I think you can use this field if you _know_ which detector you are using though.

----

I might be missing something here.  You guys seem to both be very interested in clamping this value and I'm not sure why.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1192 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNO3EJalYEg9i3jNHwP:-KNO3EJalYEg9i3jNHwQ:1626928303) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1192)):*
>  If there are not, then nothing will be returned.

I don't actually mean "nothing" here. I meant nothing _special_.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

mkoval:
Review status: 24 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1192 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNO3EJalYEg9i3jNHwP:-KNO8cCGKyF_vyiWjpOc:-1869957444) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1192)):*
> I'm talking about modeling a collision with an object that can be intersected a certain amount during motion, like a foam or rubber surface.

I understand the example. However, I only see two possible use-cases:

1. *Your model requires accurate penetration distance.* You should not use `distanceTolerance = 0` because the result is undefined.

2. *Your model does not require accurate penetration distance.* Then clamping to `distanceTolerance` will not affect the model.

The _only_ valid use-case I can conceive for this is debugging.

>  I might be missing something here. You guys seem to both be very interested in clamping this value and I'm not sure why.

I feel strongly that the return value of `distance`, just like `collide`, should be a _fundamental property of the environment_. Implementation defined values (like `actualDistance`) belong in the `Result` struct for expert use.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: 24 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNO94Oy7__TnSoDELRM:583958547) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
@pkv, can you think of a use-case (besides debugging / printing diagnostics) where a value < `threshold` would be useful?

My principal concern is what the API would look like.  For Option A:
> `double distance(scene, threshold)`: If the minimum distance between object pairs in the scene is at least `threshold`, then return this distance.  Otherwise, return a distance that is less than `threshold`, corresponding to the distance between some pair of objects. Note that this distance is not guaranteed to be the actual minimum distance in the scene, and that for the same scene, the result may vary between implementations or between calls to a given implementation.

For Option B:
> `double distance(scene, threshold)`: Return the minimum distance between object pairs in the scene, clamped to be at least `threshold`.

I'm also concerned that the proposed documentation above for option A is actually too strict.  I would imagine an implementation making use of bounding volumes to shortcut as soon as it determines that a given pair of objects has a distance below the threshold, without actually calculating the true distance between the objects!  If we want to allow this case, the documentation for option A must actually read:
> `double distance(scene, threshold)`: If the minimum distance between object pairs in the scene is at least `threshold`, then return this distance.  Otherwise, return a distance that is less than `threshold`, *which has no actual meaning in the scene*.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: 24 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1192 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNO3EJalYEg9i3jNHwP:-KNOC5baVN8tt_GBjADW:-1267515054) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1192)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> > I'm talking about modeling a collision with an object that can be intersected a certain amount during motion, like a foam or rubber surface.
> 
> I understand the example. However, I only see two possible use-cases:
> 
> 1. *Your model requires accurate penetration distance.* You should not use `distanceTolerance = 0` because the result is undefined.
> 
> 2. *Your model does not require accurate penetration distance.* Then clamping to `distanceTolerance` will not affect the model.
> 
> The _only_ valid use-case I can conceive for this is debugging.
> 
> >  I might be missing something here. You guys seem to both be very interested in clamping this value and I'm not sure why.
> 
> I feel strongly that the return value of `distance`, just like `collide`, should be a _fundamental property of the environment_. Implementation defined values (like `actualDistance`) belong in the `Result` struct for expert use.
</details>
(By the way, I really don't like calling this `actualDistance` -- it's really `actualDistanceFoundSoFarBeforeShortcutting`.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOFVG6ejCS5uxNVhGq:-287039582) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
Ok, I see.

> `double distance(scene, threshold)`: Return the minimum distance between object pairs in the scene, clamped to be at least threshold.

This seems good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1192 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNO3EJalYEg9i3jNHwP:-KNOFiV49Ab-7IDCCnhG:833691278) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1192)):*
<details><summary><i>Previously, cdellin wrote…</i></summary>
> (By the way, I really don't like calling this `actualDistance` -- it's really `actualDistanceFoundSoFarBeforeShortcutting`.)
</details>
I think you were getting confused here. I was talking about setting `distanceTolerance = -0.5` and using values of distance _greater_ than this value.  Which is unaffected by clamping.  But is also unaffected by _not_ clamping.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOGiMM5Iau13yY890_:-129097449) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, psigen (Pras Velagapudi) wrote…</i></summary>
> Ok, I see.
> 
> > `double distance(scene, threshold)`: Return the minimum distance between object pairs in the scene, clamped to be at least threshold.
> 
> This seems good.
> 
</details>
I like @cdellin's API.  Just calling it `distanceThreshold`, `minDistance` or even `threshold` seems good to me.  While people might be unsure what it means at first, we can easily explain it in the docs and comments.  It is also the only behavior for this parameter that really seems to make sense in this context, given our extensive discussion here.

And more importantly, by using fewer words, we won't imply that this parameter does something that it doesn't!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1192 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNO3EJalYEg9i3jNHwP:-KNOHlaE5kyih7QeMtwk:309131650) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1192)):*
<details><summary><i>Previously, psigen (Pras Velagapudi) wrote…</i></summary>
> I think you were getting confused here. I was talking about setting `distanceTolerance = -0.5` and using values of distance _greater_ than this value.  Which is unaffected by clamping.  But is also unaffected by _not_ clamping.
</details>
@cdellin that's a good point, it's usually just some intermediate value in most detectors.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: 24 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1192 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNO3EJalYEg9i3jNHwP:-KNOLoFQ1EvhgjybhvUh:-1942408522) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1192)):*

> I think you were getting confused here. I was talking about setting distanceTolerance = -0.5 and using values of distance greater than this value.

Ah, gotcha. My mistake!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: 24 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOMKMwe3p6NCZx3f4I:-114159760) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, psigen (Pras Velagapudi) wrote…</i></summary>
> I like @cdellin's API.  Just calling it `distanceThreshold`, `minDistance` or even `threshold` seems good to me.  While people might be unsure what it means at first, we can easily explain it in the docs and comments.  It is also the only behavior for this parameter that really seems to make sense in this context, given our extensive discussion here.
> 
> And more importantly, by using fewer words, we won't imply that this parameter does something that it doesn't!
</details>
Sounds good! One other idea might to call it `minDistance` or similar (lower bound, etc), because it allows us to introduce a similar `maxDistance` clamping threshold in the future.  I can imagine cases where callers only care about distances if they're below a threshold, and the same optimizations can be employed by the implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOQMxpQdLV9yVYNtdp:-357488522) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, cdellin wrote…</i></summary>
> Sounds good! One other idea might to call it `minDistance` or similar (lower bound, etc), because it allows us to introduce a similar `maxDistance` clamping threshold in the future.  I can imagine cases where callers only care about distances if they're below a threshold, and the same optimizations can be employed by the implementation.
</details>
**Unrelated to current API discussion:**
@cdellin: One curious thing to note is that clamped `minDistance`/`maxDistance` parameters would not provide any reasonable way to converge to binary collision checking:  setting both to the same value wouldn't actually produce a distinguishable result, _because of the clamping_.  It would simply always return `distance` = `minDistance` = `maxDistance`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: 24 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOS6UIj4un7aUwuDZt:1326762235) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, psigen (Pras Velagapudi) wrote…</i></summary>
> **Unrelated to current API discussion:**
> @cdellin: One curious thing to note is that clamped `minDistance`/`maxDistance` parameters would not provide any reasonable way to converge to binary collision checking:  setting both to the same value wouldn't actually produce a distinguishable result, _because of the clamping_.  It would simply always return `distance` = `minDistance` = `maxDistance`.
</details>
@pkv Well, that's certainly a good point.  (Of course you could hack it by passing `minDistance = -std::numeric_limits<double>::min()` and `maxDistance = std::numeric_limits<double>::min()`.)  Alternatively, I could also imagine providing in the `Result` structure flags for whether clamping occurred.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

mkoval:
Review status: 24 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOYk866nnxL_aQTOaQ:-1475219640) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, cdellin wrote…</i></summary>
> @pkv Well, that's certainly a good point.  (Of course you could hack it by passing `minDistance = -std::numeric_limits<double>::min()` and `maxDistance = std::numeric_limits<double>::min()`.)  Alternatively, I could also imagine providing in the `Result` structure flags for whether clamping occurred.
</details>
Whew. It sounds like we've come to a consensus. :sweat_smile:

This is my understanding:

- `CollisionDetector` has _one_ member function called `distance`
- `distance` returns a `double` that is minimum distance between any pair of `CollisionObject`s being checked clamped to `minDistance`
- `minDistance` defaults to `0.0`
- `DistanceResult` has a `getDistance` member function that returns the clamped distance and a get `getUnclampedDistance` member function that contains the unclamped distance (along with a warning that the value is implementation defined).

@psigen @cdellin @jslee02 Any objections?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: 24 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOaynrkvNIzdbKTmZR:-1554784513) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Whew. It sounds like we've come to a consensus. :sweat_smile:
> 
> This is my understanding:
> 
> - `CollisionDetector` has _one_ member function called `distance`
> - `distance` returns a `double` that is minimum distance between any pair of `CollisionObject`s being checked clamped to `minDistance`
> - `minDistance` defaults to `0.0`
> - `DistanceResult` has a `getDistance` member function that returns the clamped distance and a get `getUnclampedDistance` member function that contains the unclamped distance (along with a warning that the value is implementation defined).
> 
> @psigen @cdellin @jslee02 Any objections?
</details>
I'm `+1`, but could we have `DistanceResult::getUnclampedDistance` return not only the violating distance (or bound on the distance, whatever the implementation calculated), but also the names or IDs of the bodies that caused the shortcut? Similar to the way you can ask the collision detector which pair of bodies caused a collision.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

mkoval:
Reviewed 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOiUo797Y8MaCry6WL:-1167157116) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, cdellin wrote…</i></summary>
> I'm `+1`, but could we have `DistanceResult::getUnclampedDistance` return not only the violating distance (or bound on the distance, whatever the implementation calculated), but also the names or IDs of the bodies that caused the shortcut? Similar to the way you can ask the collision detector which pair of bodies caused a collision.
</details>
That's already available in `DistanceResult`:

https://github.com/dartsim/dart/blob/js/distance/dart/collision/DistanceResult.hpp#L45

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

cdellin:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOjwG4_QMoSLMsY3kx:343070624) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> That's already available in `DistanceResult`:
> 
> https://github.com/dartsim/dart/blob/js/distance/dart/collision/DistanceResult.hpp#L45
</details>
Hmm, I generally think this is fine, but actually calling it `getUnclampedDistance` implies to me that if I were to set the `minDistance` parameter to `-INFINITY`, that would be the distance I would get. What about something like `getClampViolatingDistance` or something? (I'm sure there are better ideas out there!)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

psigen:
*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNOmAMfbVShcEARsn46:593078471) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, cdellin wrote…</i></summary>
> Hmm, I generally think this is fine, but actually calling it `getUnclampedDistance` implies to me that if I were to set the `minDistance` parameter to `-INFINITY`, that would be the distance I would get. What about something like `getClampViolatingDistance` or something? (I'm sure there are better ideas out there!)
</details>
If the value is implementation defined, would it make sense for this not to be in `DistanceResult` at all, but instead be in another structure that is dependent on the CollisionDetector (and possible requires a `dynamic_cast<>`  at some point)?  This would be similar to how `Joint` information works. If the case that @mkoval and @cdellin made before actually true, then this unclamped distance is already an implementation specific value which is meaningless without knowing which detector provided it, and may depend on other _implementation specific values that may need to be made available_.

In short, I think if we add this variable into the interface, we open the door to needing other random fields because we really don't know what this value will mean.  It might require that we return additional geometric information, or boolean flags about how the check ended up being done, etc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

mkoval:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[dart/collision/fcl/FCLCollisionDetector.cpp, line 1189 \[r1\]](https://reviewable.io:443/reviews/dartsim/dart/744#-KNAJVChYnPPukB-ROc6:-KNP27UHmQ2Bp7PgeO1W:1718703830) ([raw file](https://github.com/dartsim/dart/blob/fd53ab64bcf12e017e809c05a68bd64a9df74186/dart/collision/fcl/FCLCollisionDetector.cpp#L1189)):*
<details><summary><i>Previously, psigen (Pras Velagapudi) wrote…</i></summary>
> If the value is implementation defined, would it make sense for this not to be in `DistanceResult` at all, but instead be in another structure that is dependent on the CollisionDetector (and possible requires a `dynamic_cast<>`  at some point)?  This would be similar to how `Joint` information works. If the case that @mkoval and @cdellin made before actually true, then this unclamped distance is already an implementation specific value which is meaningless without knowing which detector provided it, and may depend on other _implementation specific values that may need to be made available_.
> 
> In short, I think if we add this variable into the interface, we open the door to needing other random fields because we really don't know what this value will mean.  It might require that we return additional geometric information, or boolean flags about how the check ended up being done, etc.
> 
</details>
I am fine with removing the unclamped field entirely. We can always add it in a future version of DART if there proves to be a need for it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/744)*

jslee02:
Sorry for the delayed update.

I basically agree with the consensus but have one concern on it. `DistanceResult` provides three kinds (but they are all related) information: distance, objects, nearest points. If we removing the unclamped field completely then it could cause some confusion by having inconsistent distance and the nearest points.

From my perspective, the basic goal of distance query is to find **minimum** distance across all the objects in the collision group, and optionally the nearest points that forms a line segment with the distance. The lower bound option is for perfomance boosting in case the caller doesn't care about distance below a certain value. In that sense, I would prefer having at least one set of API that returns consistent distance, object pair, and nearest points. This could be done by having following fields: `minDistance`, `shapeFrame1/2`, `nearestPoint1/2`. `minDistance` (called unclamped distance or actual distance) is the minimum distance among being checked objects. We could have `getMinDistance` for this field, and `getDistance` returns the clamped distance.

Maybe I have a different view on the distance query so probably I'm not completely convinced to that clamped distance is the default distance. Also, I might have something missed from the above long discussion. Please let me know if there is.

cdellin:
@jslee02 Good point -- those additional return values (colliding objects and nearest points) are only valid (and really should only be populated) when the return value is not clamped. (That is, in the case that `dist == minDistance` and therefore there was a clamp performed, the computation probably shortcutted early, so any objects or nearest points are probably not correct anyway.) So perhaps the `DistanceResult` could include a flag for whether clamping occurred, and only populate those values if `clamped == false`? If the user sets `minDistance = -INFINITY`, then no clamping will ever be performed, and those values can always be trusted.

mkoval:
I like @cdellin's suggestion of adding a flag to the `Result` struct that indicates whether or not clamping occurred. This correctly captures the semantic that the fields you mentioned (nearest points, etc) points are valid as long as the actual distance is greater than `minDistance`.

Like @cdellin said, this condition *is always satisfied* when `minDistance = -infinity`

jslee02:
I updated the distance API according to the discussion in the last commit.

I also like to have an indicator for whether the distance was clamped. For that `Distance::isMinDistanceClamped()` is added that returns true if `minDistance` and `unclampedMinDistance` are equal rather than using a flag. This is because comparing the two distance is a bit simpler to implement when `DistanceResult` could be `nullptr`. Comparing floating values would be  more cost than checking boolean, but don't think this is performance critical.

Thanks for the comments @mkoval , @cdellin , and @psigen . Please take a look at the changes and let me know if it looks good to merge.

cdellin:
Sounds (and looks) good to me, thanks @jslee02!

mkoval:
LGTM. 👍

PR #745, opened 2016-07-20
SphereShape

jslee02:
as proposed in #724.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/745)


mkoval:

jslee02:
#749 is merged.

dart/gui/osg/SupportPolygonVisual.cpp
```diff
@@ -387,8 +386,7 @@ void SupportPolygonVisual::initialize()
         dart::dynamics::Frame::World(), "com");
 
   mComRadius = mCentroidRadius;
-  mCom->setShape(std::make_shared<dart::dynamics::EllipsoidShape>(
-                                mComRadius/2.0*Eigen::Vector3d::Ones()));
```
jslee02: @mxgrey Did you mean `mComRadius*2.0` in the first place? In that case, the below code should be just `mComRadius` rather than `mComRadius/4.0`.

PR #747, opened 2016-07-21
GenericJoint class

jslee02:
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/747)


mkoval:

PR #749, opened 2016-07-21
Update API to support fcl 0.5 and tinyxml 4.0

jslee02:
FCL switched to use the smart pointers of the C++11 standard library since 0.5.0, and removed `XML_NO_ERROR` since 4.0.0.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/749)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Collision detection
 
-  * Added support of FCL 0.5.0: [#749](https://github.com/dartsim/dart/pull/749)
+  * Added support of FCL 0.5 and tinyxml2 4.0: [#749](https://github.com/dartsim/dart/pull/749)
   * Added warnings for unsupported shape pairs of DARTCollisionDetector: [#722](https://github.com/dartsim/dart/pull/722)
 
 * Misc improvements and bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Collision detection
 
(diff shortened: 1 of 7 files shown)


mkoval:
You can eliminate _a lot_ of duplicate `#if` checks by defining `using` alias:
```c++
#if FCL_VERSION_AT_LEAST(0,5,0)
template <class T> using fcl_pointer = std::shared_ptr<T>;
#else
template <class T> using fcl_pointer = boost::shared_ptr<T>;
#endif
```

I prefer this to the current implementation because it doesn't clutter up the implementation with a bunch of `#if/#else/#endif` checks.

jslee02:
> You can eliminate a lot of duplicate #if checks by defining using alias:

:+1:  As there are uses of `weak_ptr` as well, I would prefer `fcl_shared_ptr` and `fcl_weak_ptr` (or `fcl::shared_ptr` and `fcl::weak_ptr`) though.

jslee02:
Also, it seems there are API changes in tinyxml2 as well according to the travis build log. Will create an PR for it.

mkoval:
That sounds good to me. I am concerned about the code duplication, not the name of the alias. :)

jslee02:
It seems tinyxml2 removed `XML_NO_ERROR` in version 4.0. In the last commit, I switched to use `XML_SUCCESS` rather than `XML_NO_ERROR` since they were interchangeable ([they were assigned with the same value 0](https://github.com/leethomason/tinyxml2/blob/2.2.0/tinyxml2.h#L485)).

PR #760, opened 2016-08-25
Remove explicit SSE2 compiler flag

j-rivero:
We add unconditionally the SSE2 compiler flag when using GCC. This is making [non x86 arches to fail](https://buildd.debian.org/status/logs.php?pkg=kido&ver=0.1.0%2Bdfsg-1).

The bug report [comes from Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=833817) and the developers made a couple of good points about it:

 * SSE2 is enable by default if you run 64 bits on x86.
 * We can not assume that i386 supports it since i386 for Debian implies 686 (or greater) and SSE2 was introduced in pentium4.

I think that the impact of removing it is really low.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/760)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,7 @@ elseif(CMAKE_COMPILER_IS_GNUCXX)
   if(DART_TREAT_WARNINGS_AS_ERRORS)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
   endif()
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -msse2 -fPIC")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -fPIC")
   execute_process(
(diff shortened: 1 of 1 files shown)


jslee02:
:+1: This makes sense to me.

We could consider using [`-march=native`](https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/i386-and-x86-64-Options.html) option for gcc and clang that enables all instruction subsets supported by the local machine. This could be done in a separate pull request.

j-rivero:
> We could consider using -march=native option for gcc and clang that enables all instruction subsets supported by the local machine. This could be done in a separate pull request.

If we go with this option, please consider to make it optional via flag. While most of the people use the same machine for compiling and running the applications. there are other cases (cross compilations, binary generation for packaging, etc) where the host used to compile is different than the computer running binaries.

jslee02:
It seems the builds on OS X [fail](https://travis-ci.org/dartsim/dart/jobs/155140111#L2096) due to the recently released [FLANN 1.9.1](https://github.com/mariusmuja/flann/releases/tag/1.9.1), which might have nothing to do with the change of this pull request. We could address the issue separately. Merging this PR.

PR #763, opened 2016-08-26
Fix direction of contact normal of BulletCollisionDetector

jslee02:
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/763)


psigen:
Wow, I'm surprised this was not causing strange problems.

jslee02:
Probably `BulletCollisionDetector` is not that popular. :sweat:

PR #777, opened 2016-10-04
Support for Bullet built with double precision (backport #660)

### LINKED ISSUE #776: Backport #660 for DART 5.1
[requested](https://github.com/dartsim/dart/pull/660#issuecomment-251467758)

scpeters:
I squashed the three commits of #660 and applied them to the `release-5.1` branch. Closes #776.

Bullet requires a pre-processor variable to be set before including any
headers to determine whether the library was built with single or double
precision. Failing to do so may result in a linking error.

This commit:

1. Forces CMake to use CONFIG mode (BulletConfig.cmake shipped with
   bullet) instead of MODULE mode (FindBullet.cmake shipped with CMake
   by default) in find_package(Bullet).

2. Updates references to BULLET_INCLUDE_DIRS and BULLET_LIBRARIES
   accordingly

3. Test whether Bullet was built using double precision floating point
   numbers by attempting to compile code that produces a linking error
   when the necessary preprocessor definition is missing.

4. Set the necessary preprocessor definition in config.h.

combined with commits that
Reverted to using Bullet MODULE mode.
and
Minor output message update for consistency

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/777)


jslee02:
:+1: LGTM.

PR #803, opened 2016-10-17
Lemke fails trivial LCP problems with release-5.1 (tag 5.1.5)

tianyang9310:
Lemke implementation fails trivial LCP problem, where both A and b are 1-dimensional. When b is nonnegative, the solution is trivial. When b is negative and A is positive, the solution should simply be (-b/A), while A is nonpositive, there would be no solution.

Test codes:
```
    Eigen::MatrixXd testA;
    testA.resize(1,1);
    testA<< 1;
    Eigen::VectorXd testb;
    testb.resize(1);
    testb<< -9.8;
    std::cout<<testA<<std::endl;
    std::cout<<testb<<std::endl;
    Eigen::VectorXd* f =  new Eigen::VectorXd(1);
    int err = dart::lcpsolver::Lemke(testA,testb,f);
    std::cout<<err<<std::endl;
    std::cout<<(*f)<<std::endl;
    std::cin.get();
```
P.S. Please ignore previous pull request which includes my local meaningless commits. I have closed it. Please see this PR>

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/803)


jslee02:
I didn't write the code and don't know much about Lemke algorithm. Do you know why it can't handle one-dimensional cases? Is Lemke not supposed to handle 1-d cases by nature or is it the implementation problem?

tianyang9310:
I have updated the commit I made. Right now it works for any general case. As far as I know, Lemke should work for any dimensions.

BTW, you can pick the `Fix Lemke bug` commit. The second commit is my testing for corner cases which the original implementation fails.

tianyang9310:
Sorry for bothering this thread. I will update my all fix and provide a unittest for PR.

Issue #804, opened 2016-10-24
Rendering - white lines for non-osg display with nvidia driver


fredowski:
Hi, 

i compiled the current DART git version "last commit: Update Control - 19.10.2016" and tried some of the examples. I noticed that there is a difference in the rendering result between the osg examples and the non-osg examples. I think there is a difference in the OpenGL configuration. My system setup is debian jessie NVidia GTX 1080 with native NVidia Driver version 367.48. 

Here is the non-osg atlasSimbicon on NVidia GTX1080 with white lines:

![non-osg-whitelines](https://cloud.githubusercontent.com/assets/1280457/19650679/6c7fadc0-9a0a-11e6-9e1c-caa1aefabaa0.JPG)

Here is osgAtlassimbicon on GTX1080  with no white lines:
![osg-no-whitelines](https://cloud.githubusercontent.com/assets/1280457/19650729/92adf3bc-9a0a-11e6-8267-adb4484a5294.JPG)

Jeongseok figured out that the white lines are not visible with a Intel Integrated Graphics Controller:

![intel-no-white-lines](https://cloud.githubusercontent.com/assets/1280457/19650816/e3e3fac4-9a0a-11e6-81de-c40e4e1830a3.png)

fredowski:
Hi Jeongseok,

i think I found the problem regarding the strange white lines. With pullrequest #805 this should be fixed. With the change there are no white lines anymore.

Regards

Friedrich

jslee02:
Awesome! Resolved by #805.

PR #805, opened 2016-10-24
fixed issue #804 - Rendering white lines with nvidia driver

fredowski:
The non-osg display shows strange white lines with an nvidia driver.
The fix is based on this proposal
http://gamedev.stackexchange.com/questions/31941/opengl-behaviour-depending-on-the-graphics-card
that I found by googling around... I tested this with GTX1080 and NVidia
driver version 367.48 on debian linux. The strange white lines disappeared.

Compared to the rendering result as shown in issue #804, the new rendering is as espected. See:

non-osg-fixed

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/805)


jslee02:
This change looks good. Thank you for the fix! I'll merge this once Travis becomes happy, which should be.

PR #806, opened 2016-10-27
Use Codecov for online code coverage report

jslee02:
Coveralls and Codecov are all promising coverage reporting services, but it seems [the layout of Codecov](https://codecov.io/gh/dartsim/dart/branch/codecov) looks more organized and intuitive to me comparing to [Coveralls'](https://coveralls.io/github/dartsim/dart?branch=master).

We could use Codecov with this minimal changes of this PR since they both work based on `gcov` but uploading to different services. There are some CMake scripts could be cleaned up, but I'd like to save it for future PRs.

---

This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/806)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,9 @@ install:
 script:
   - 'ci/script.sh'
 
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
+
 after_failure: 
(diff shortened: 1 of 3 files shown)


mkoval:
👍 The Codecov user interface looks _much_ better than Coveralls.

PR #807, opened 2016-10-27
fixed tutorialDominoes-Finished - this did not work

fredowski:
I tried the tutorial with the dominoes but it did not work. I could never add
a new domino. There were three issues in the code.

a) The new domino was never added to the world.
b) The collision detection was always detecting a collision because
   the new domino is in contact with the floor.
c) The tutorial description on the website is not related to
   the code for the collision detection.

I fixed a) and b) and added some comments with this commit.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/807)

### DIFF
--- a/tutorials/tutorialDominoes-Finished.cpp
+++ b/tutorials/tutorialDominoes-Finished.cpp
@@ -258,24 +258,40 @@ public:
 
     newDomino->setPositions(x);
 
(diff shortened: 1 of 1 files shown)


jslee02:
This looks good to me. Thanks @fredowski !

PR #808, opened 2016-10-28
Lemke fails several LCP problems with release-5.1 (tag 5.1.5) 

tianyang9310:
Hi JS,

I have cleaned up my fix and provide a unit test for it. Please check `testLemke.cpp` in the unittest folder. You will find that the original implementation fails my unittest, while my fix should work.

Yang

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/808)


jslee02:
I haven't investigated the logic of Lemke, but it seems convincing as it passes the tests. I'll wait for 1 or 2 days looking for comments from others before merging this PR. Thanks!

psigen:
@jslee02 can we do a followup cleanup PR?  The fix seems alright, but there are a lot of unnecessary syntax and whitespace changes in here that diverge from DART style a bit.

jslee02:
I agree with you. Let me do that when I have a chance.

I hope we have a guideline for the code style with clang-format setting soon. :)

tianyang9310:
Sorry about the formatting issue. I just used my IDE's auto-reformatting tool which is clearly not DART's convention. Thanks for cleaning up the code. BTW, this PR could also be cherry-picked for release-6.1 since its Lemke implementation is the same as release-5.1.

PR #811, opened 2016-11-03
Setup clang-format target

jslee02:
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/811)


psigen:
Is the `clang-format` check supported by TravisCI or Appveyor?

jslee02:
This PR adds two custom CMake targets: `format` and `check-format`. `format` formats the code while `check-format` checks whether the code complies with the style and returns the success. So TravisCI or Appveyor could use `make check-format` to check the style if `clang-format` is installed.

Issue #827, opened 2016-12-20
Unconventional definition of ellipsoid size


ljklonepiece:
It seems the input to EllipsoidShape is not the radius but it is the 2*radius (diameter) along each dimension, which is a bit confusing to use. (takes me a while to figure out it is the case)

It will be better if convention is followed:)

psigen:
Could you clarify where you have seen the convention of using radius for this?

The convention that I have seen typically is to specify an ellipse by the lengths of its principle axes, e.g.: https://en.wikipedia.org/wiki/Ellipsoid

ljklonepiece:
from the wiki page, it uses semi-principal axes to define an ellipsoid.
but anw, the principle axes shall not be termed as the radius, right? take a sphere as an example, its length of principle axis is 2 times of its radius.

psigen:
Right, the semi-principal axes are the convention I am familiar with.

You are right that when the semi-principle axis lengths are used they should not be termed the `radii` in any of the function parameters or examples.  Thanks for the clarification!

ljklonepiece:
No problem! Thanks for creating and maintaining this great physics simulator for the community:)

jslee02:
I agree. For the API compatibility, we might want to add `get/setRadii()` first and then change the constructor in the next major release.

jslee02:
I believe this is addressed by #829.

PR #829, opened 2016-12-22
Add getters/setters for radii and diameters of EllipsoidShape

jslee02:
DART currently uses diameters for the primary properties of `EllipsoidShape`, which is not the commonly used convention. This PR is preliminary changes for replacing diameters with radii that include:
* deprecate get/setSize
* add get/setDiameter
* add get/setRadii

In the next major release, diameters will be totally replaced by radii.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/829)


psigen:
Wait, I think we might have some confusion here.  I thought the conclusion in #827 was that semi-principal axis lengths (diameters) was the convention, but that the word `radii` was used to describe these lengths in certain places.

Am I missing something?

jslee02:
Well, maybe I'm the one confused here. I understand that the lengths along the semi-principal axes (i.e., a, b, and c) represent the radii and that's the convention. Please correct me if I'm wrong.

psigen:
Ok, I think the confusion stems from the following:
1. The lengths of the semi-principal axes are typically represented in *diameters* (distance from one side to the other, passing through the center of the ellipse.
2. The notation in DART describes ellipses in terms of _radii_, which usually means 1/2 of the diameter.

ClintLiddick:
@psigen: @jslee02 and I talked, looked through wikipedia and the linked issue, and it seems like there was a miscommunication. Below is what we believe is correct.

1. Ellipsoids are most commonly parameterized by their semi-principal axes, which is equivalent to (but not called) the radii (i.e. they go from the center, through a focus, to an edge)
1. Principal axes are equivalent to diameter (from edge to edge, through center and foci)
1. Dart currently uses principle axes, but calls them diameter.
1. Dart wants to transition to using semi-principal axes, and in this PR (technically incorrectly) refers to it as this radii.

Dart *could* use either the principal or semi-principal axes, but should term them as such to avoid confusion.

psigen:
Ah, this makes sense.  I think both JS and I had confusion around various terminology.

Issue #839, opened 2017-01-25
Compile errors in tutorial-demo after fresh brew install


ruffson:
Hi,

I just tried to build with Dart.
It wasn't possible because dart/common/detail/AspectWithVersion.hpp was missing the import of Composite: "#include /dart/common/Composite.hpp"

This problem arose with homebrew on macOS Sierra.

jslee02:
I could successfully install `dartsim6` (v6.1.2) using homebrew on macOS Sierra myself now.

Could you paste or show the whole error messages you got?

r-t-r:
This is the error:

`In file included from bipedtest.cpp:32:
In file included from /usr/local/include/dart/dart.hpp:33:
In file included from /usr/local/include/dart/common/common.hpp:3:
In file included from /usr/local/include/dart/common/Aspect.hpp:37:
In file included from /usr/local/include/dart/common/Cloneable.hpp:324:
In file included from /usr/local/include/dart/common/detail/Cloneable.hpp:1:
In file included from /usr/local/include/dart/dart.hpp:33:
In file included from /usr/local/include/dart/common/common.hpp:4:
In file included from /usr/local/include/dart/common/AspectWithVersion.hpp:35:
/usr/local/include/dart/common/detail/AspectWithVersion.hpp:15:30: error: unknown type name 'Composite'
          class CompositeT = Composite,
                             ^
/usr/local/include/dart/common/detail/AspectWithVersion.hpp:16:45: error: use of undeclared identifier 'NoOp'
          void (*updateState)(DerivedT*) = &NoOp<DerivedT*> >
                                            ^
/usr/local/include/dart/common/detail/AspectWithVersion.hpp:16:50: error: 'DerivedT' does not refer to a value
          void (*updateState)(DerivedT*) = &NoOp<DerivedT*> >
                                                 ^
/usr/local/include/dart/common/detail/AspectWithVersion.hpp:14:30: note: declared here
template <class BaseT, class DerivedT, typename StateDataT,
                             ^
/usr/local/include/dart/common/detail/AspectWithVersion.hpp:16:59: error: expected expression
          void (*updateState)(DerivedT*) = &NoOp<DerivedT*> >
                                                          ^
/usr/local/include/dart/common/detail/AspectWithVersion.hpp:16:61: error: expected unqualified-id
          void (*updateState)(DerivedT*) = &NoOp<DerivedT*> >
                                                            ^
[...]

Issue #752, opened 2016-07-22
Add documentation on how to compile tutorials.


psigen:
While there's a lot of documentation [explaining the source code of the included DART tutorials](http://dart.readthedocs.io/en/latest/tutorials/introduction/), there actually isn't any documentation on how to build the tutorials in the first place.  This can lead to confusion for new users (e.g. #751).

We should add a section to the tutorials documentation that explains how to build just the DART tutorials if someone has already installed a version of DART.  We may also want to consider if it would be useful to document/create a version of the `tutorials` CMakeLists.txt that can build independently from the DART source tree, for users that have a system-installed version of DART (from `deb` or `brew`) but want to compile just these tutorials.

mkoval:
I *strongly* agree with moving the tutorials out of the DART source tree. Most users install DART as a binary package and build code *outside of the source tree* that depends on it, e.g. using `find_package(DART)` in CMake. As far as I can tell, the _only way_ to currently build the tutorials is as part of DART (see below).

I suggest either: (1) moving the tutorials to a separate repository or (2) installing the tutorials to `/usr/share/dart` and providing instructions for how to copy them to the user's working directory. In both cases, it's important that the tutorials find DART as an external dependency.

If we run these tutorials on Travis, this could also partially address #738.

---

I am not a C++ novice by any means and am confused myself about how the tutorials are supposed to work. I just attempted to build them out-of-source and failed on several fronts:

1. Determine which version of DART I have installed (note: figuring out how to do this took quite a bit of Googling):
    
    ```console
    $ dpkg-query --show libdart6
    libdart6	6.0.1.ppa4~trusty
    ```
2. Checkout the corresponding version of DART from source (note: there is no tag for `v6.0.1`, so I installed `6.0.0` and hoped for the best):

    ```console
    $ git clone https://github.com/dartsim/dart.git dart_checkout
    $ cd dart_checkout
    $ git checkout v6.0.1
    error: pathspec '6.0.1' did not match any file(s) known to git.
    $ git checkout v6.0.0
    ```

3. Copy the tutorial files out into a separate directory:

    ```console
    $ cp -r tutorials ../dart_tutorials
    $ cd ../dart_tutorials
    ```

4. Guess which `COMPONENTS` are being used:

    ```console
    $ grep -hR 'using namespace dart::' . | sort | uniq
    using namespace dart::common;
    using namespace dart::dynamics;
    using namespace dart::gui;
    using namespace dart::math;
    using namespace dart::simulation;
[...]

Issue #751, opened 2016-07-22
dominoes tutorial doesn't work on mac


ghost:
I get 20 errors:

```
/usr/local/include/dart/common/detail/Cloneable.hpp:128:55: error: a space is

      required between consecutive right angle brackets (use '> >')

  return common::make_unique<MakeCloneable<Base, Mixin>>(*this);

                                                      ^

/usr/local/include/dart/common/detail/Cloneable.hpp:212:1: error: 'auto' not

      allowed in function return type

auto ProxyCloneable<Base, OwnerT, DataT, setData, getData>::operator =(

^~~~

/usr/local/include/dart/common/detail/Cloneable.hpp:213:22: error: expected ';'

      at end of declaration

    const Data& data) -> ProxyCloneable&

                     ^

/usr/local/include/dart/common/detail/Cloneable.hpp:213:23: error: cannot use

      arrow operator on a type

    const Data& data) -> ProxyCloneable&

                      ^

/usr/local/include/dart/common/detail/Cloneable.hpp:223:1: error: 'auto' not

      allowed in function return type

auto ProxyCloneable<Base, OwnerT, DataT, setData, getData>::operator =(

^~~~

/usr/local/include/dart/common/detail/Cloneable.hpp:224:17: error: expected ';'

      at end of declaration

    Data&& data) -> ProxyCloneable&

                ^

/usr/local/include/dart/common/detail/Cloneable.hpp:224:18: error: cannot use

      arrow operator on a type

    Data&& data) -> ProxyCloneable&

                 ^

/usr/local/include/dart/common/detail/Cloneable.hpp:234:1: error: 'auto' not

      allowed in function return type

auto ProxyCloneable<Base, OwnerT, DataT, setData, getData>::operator =(

^~~~

/usr/local/include/dart/common/detail/Cloneable.hpp:235:33: error: expected ';'

      at end of declaration

    const ProxyCloneable& other) -> ProxyCloneable&

                                ^

/usr/local/include/dart/common/detail/Cloneable.hpp:235:34: error: cannot use

      arrow operator on a type

    const ProxyCloneable& other) -> ProxyCloneable&

                                 ^

/usr/local/include/dart/common/detail/Cloneable.hpp:245:1: error: 'auto' not

      allowed in function return type

auto ProxyCloneable<Base, OwnerT, DataT, setData, getData>::operator =(

^~~~

/usr/local/include/dart/common/detail/Cloneable.hpp:246:28: error: expected ';'

      at end of declaration

[...]

PR #842, opened 2017-01-31
Allow examples and tutorials to be built out of DART source tree

jslee02:
This PR allows the examples and tutorials can be built out of the DART source tree when DART is installed on the system. The examples and tutorials are now installed under system directory (e.g., for Linux, `/usr/share/doc/dart/examples` and `/usr/share/doc/dart/tutorials`) with sample data files (`/usr/share/doc/dart/data`) when DART is installed. You can copy the directories to your workspace to build them. Detailed instruction can be found in the directories.

We are still able to build examples and tutorials through the DART build system. It requires distinct build script, which is in the main `CMakeLists.txt`, because the build scripts in the examples and tutorials assumes that DART is already installed which is not always true when building DART from source. One downside of doing this would be that we should keep maintaining two build scripts for the same projects.

`SampleResourceRetriever` class is introduced to be able to load the sample data files from either of the system directory or in DART source. It first tries to load from the system directory and then from the system directory if failed. A specific URI is also introduced to refer to the sample data (e.g., `file://sample/skel/biped.skel`).

TODOs:
* [x] Add instruction at the root of examples and tutorials
* ~~Add debian packages for examples and tutorials~~ (will be done by other PR)
* ~~Add Travis-CI tests for out-of-source build of examples and tutorials if possible~~ (will be done by other PR)

Note: This PR will be squashed and then merged.

Related issues: #751, #752, #839 

This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/842)


jslee02:
@mxgrey, @mkoval, @psigen Please take a look at this PR if interested. :smile:

mxgrey:
Reviewed 232 of 232 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[dart/utils/sdf/SdfParser.cpp, line 100 at r1](https://reviewable.io:443/reviews/dartsim/dart/842#-KbvCmYJihxYOkh5Yjwf:-KbvCmYJihxYOkh5Yjwg:b-mmi3ur) ([raw file](https://github.com/dartsim/dart/blob/762c693f9befd54ee4bbac15f5ea61d40ed4483d/dart/utils/sdf/SdfParser.cpp#L100)):*
> ```cpp
> using JointMap = std::map<std::string, SDFJoint>;
> 
> simulation::WorldPtr readWorld(tinyxml2::XMLElement* worldElement,
> ```

It looks like this should be bumped down a line for formatting.

---

*[dart/utils/sdf/SdfParser.cpp, line 174 at r1](https://reviewable.io:443/reviews/dartsim/dart/842#-KbvD4UMTPyb85qCp-dQ:-KbvD4UMTPyb85qCp-dR:b-6kkrx8) ([raw file](https://github.com/dartsim/dart/blob/762c693f9befd54ee4bbac15f5ea61d40ed4483d/dart/utils/sdf/SdfParser.cpp#L174)):*
> ```cpp
>     dynamics::ShapeNode* shapeNode);
> 
> void readVisualizationShapeNode(dynamics::BodyNode* bodyNode,
> ```

Also should be bumped down a line.

---

*[dart/utils/sdf/SdfParser.cpp, line 184 at r1](https://reviewable.io:443/reviews/dartsim/dart/842#-KbvD7Lp0XTGwXkR1hi7:-KbvD7Lp0XTGwXkR1hi8:bra7v7u) ([raw file](https://github.com/dartsim/dart/blob/762c693f9befd54ee4bbac15f5ea61d40ed4483d/dart/utils/sdf/SdfParser.cpp#L184)):*
> ```cpp
>     const common::ResourceRetrieverPtr& retriever);
> 
> void readAspects(const dynamics::SkeletonPtr& skeleton,
> ```

Also should be bumped down a line

---

*[examples/CMakeLists.txt, line 19 at r1](https://reviewable.io:443/reviews/dartsim/dart/842#-KbvDw8kBT-Ayc0y6bPr:-KbvDw8kBT-Ayc0y6bPs:bvhc9ct) ([raw file](https://github.com/dartsim/dart/blob/762c693f9befd54ee4bbac15f5ea61d40ed4483d/examples/CMakeLists.txt#L19)):*
> ```CMake
> add_subdirectory(softBodies)
> add_subdirectory(speedTest)
> add_subdirectory(vehicle)
> ```

I know CMake considers it better practice to explicitly list all subdirectories/files, but that's just because doing so ensures that the "make" command will rerun CMake any time a new subdirectory/file gets added. I expect that the DART source tree will remain pretty static (except for DART developers), so I don't really see this as necessary (since I think DART developers can be trusted to know when to rerun CMake manually).

I'm not necessarily opposed to explicitly listing them out, but I do prefer automating the build system when possible, because that means less development overhead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/842)*

mxgrey:
I've published a Reviewable review, but I'll add one more high-level comment:

I very much like these changes. I like the flexibility it offers to users of DART.

I get the impression that the build system for the examples was changed to reflect the best practices according to the developers of CMake rather than the practices that might be most convenient from a development standpoint. If that's the case, I can accept that. I somewhat appreciated the genericity of the CMakeLists.txt files that we were using in the example folders, but admittedly they do fly in the face of CMake's official stance on best practices.

jslee02:
Yes, exactly. I intended that so that the users can build the first DART example without being bothered to understand too many CMake things.

---

Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[dart/utils/sdf/SdfParser.cpp, line 100 at r1](https://reviewable.io:443/reviews/dartsim/dart/842#-KbvCmYJihxYOkh5Yjwf:-KbvIWfD2a6eVTkCQygH:b-896fix) ([raw file](https://github.com/dartsim/dart/blob/762c693f9befd54ee4bbac15f5ea61d40ed4483d/dart/utils/sdf/SdfParser.cpp#L100)):*
<details><summary><i>Previously, mxgrey (Michael X. Grey) wrote…</i></summary><blockquote>

It looks like this should be bumped down a line for formatting.
</blockquote></details>

Done.

---

*[dart/utils/sdf/SdfParser.cpp, line 174 at r1](https://reviewable.io:443/reviews/dartsim/dart/842#-KbvD4UMTPyb85qCp-dQ:-KbvIW85OJMXEWPaLmHK:b-896fix) ([raw file](https://github.com/dartsim/dart/blob/762c693f9befd54ee4bbac15f5ea61d40ed4483d/dart/utils/sdf/SdfParser.cpp#L174)):*
<details><summary><i>Previously, mxgrey (Michael X. Grey) wrote…</i></summary><blockquote>

Also should be bumped down a line.
</blockquote></details>

Done.

---

*[dart/utils/sdf/SdfParser.cpp, line 184 at r1](https://reviewable.io:443/reviews/dartsim/dart/842#-KbvD7Lp0XTGwXkR1hi7:-KbvIVtABWdJNZ7a-_lC:b-896fix) ([raw file](https://github.com/dartsim/dart/blob/762c693f9befd54ee4bbac15f5ea61d40ed4483d/dart/utils/sdf/SdfParser.cpp#L184)):*
<details><summary><i>Previously, mxgrey (Michael X. Grey) wrote…</i></summary><blockquote>

Also should be bumped down a line
</blockquote></details>

Done.

---

*[examples/CMakeLists.txt, line 19 at r1](https://reviewable.io:443/reviews/dartsim/dart/842#-KbvDw8kBT-Ayc0y6bPr:-KbvI_87Y43ogQ71kG-w:b2gj9bh) ([raw file](https://github.com/dartsim/dart/blob/762c693f9befd54ee4bbac15f5ea61d40ed4483d/examples/CMakeLists.txt#L19)):*
<details><summary><i>Previously, mxgrey (Michael X. Grey) wrote…</i></summary><blockquote>

I know CMake considers it better practice to explicitly list all subdirectories/files, but that's just because doing so ensures that the "make" command will rerun CMake any time a new subdirectory/file gets added. I expect that the DART source tree will remain pretty static (except for DART developers), so I don't really see this as necessary (since I think DART developers can be trusted to know when to rerun CMake manually).

I'm not necessarily opposed to explicitly listing them out, but I do prefer automating the build system when possible, because that means less development overhead.
</blockquote></details>

I totally agree with you. But the reason I did this is to leave the CMake scripts in *examples* and *tutorials* as plain as possible. The CMake scripts are for new users to DART and maybe even new ones to building C++ projects using CMake. I would like to lower the entry barrier to get started with DART as much as possible.

On the other hand, the [CMake script](https://github.com/dartsim/dart/blob/building_examples_tutorials/CMakeLists.txt#L268) for building examples and tutorials in DART source tree are much more complex but (hopefully) convenient to the developers (it uses predefined custom macros and other CMake features).

You can see that I still used the listing way even for the in source building. The reason to do so was because of the different target library dependencies. But I'm also open to use `file(GLOB ...)` to gather the list of subdirectories/files with the superset of dependencies.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dartsim/dart/842#-:-KbvOxRnpp29GPcHn_Ym:bz710md)*

psigen:
I think this is a great set of changes. The only question that I have is regarding the use of the `SampleResourceRetriever` in conjunction with a `file://sample/` URI.  This strikes me as a bit strange because this URI is not treated as a typical `file://` URI at all.

It seems like as-is, this could introduce two issues:
1. A new user cuts and pastes one of the URIs into another piece of code using a `LocalResourceRetriever`, which attempts to resolve it and fails.  It's unclear to the new user why this `file://` URI is not working, since the `LocalResourceRetriever` will _think_ it can resolve this scheme.
2. Someone actually has a resolvable `sample` hostname that makes this URI resolution actually valid as a  plain file URI and `LocalResourceRetriever` _can_ resolve it to an unintended host.

Would it make sense to use a custom URI scheme such as `sample://` or even `dart://` or `dart-sample://` for these resources to clearly indicate that they are _not_ simple file URIs?

mkoval:
This is great! I think this is a great use for the `ResourceRetriever` API. It's nice that the examples only need one extra line of code (creating the `SampleResourceRetriever`) to access data.

I agree with @psigen that I'd prefer to use use a custom schema over `file://` for the reasons that he listed above.

jslee02:
I also agree on @psigen's suggestion. [The schema was initially `sample://` at first](https://github.com/dartsim/dart/blame/95b80d434d219692728a8edc83641e5a80d1b062/dart/utils/SampleResourceRetriever.hpp#L52). The reason I changed to `file://` was that I thought it refers to all kinds of *local* resources, which I don't think so now. Maybe this confusion was originated from the name of the class that resolves `file://` schema is `LocalResourceRetriever`. So I'm thinking of changing the name to `FileResourceRetriever`. How do you think?

jslee02:
I've changed:
- `SampleResourceRetriever` to `DartResourceRetriever`
- URI pattern from `file://sample/skel/shapes.skel` to `dart://sample/skel/shapes.skel`

joenorby:
Good timing - I am a new user to DART, CMake, and C++, so perhaps I can test out the accessibility of the tutorials/getting started in general.

jslee02:
Thank you for the comments! We could consider the renaming of `FileResourceRetriever` later if needed.

PR #843, opened 2017-01-31
relative zoom and translation with mouse

nurimbet:
Smooth zooming and translation.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/843)


jslee02:
The camera work looks much better. Thanks!

Issue #813, opened 2016-11-03
Consider having static member function createShared/Unique()


jslee02:
Using smart pointers is a recommended practice in the modern C++ (C++11 and later), and we're already using them for DART. However, we need to carefully choose how to create `std::shared_ptr` depending on the classes. In general, `std::make_shared` is the most efficient way. If the constructor is not public, we should do `std::shared_ptr<MyClass>(new MyClass)`. For the class that needs to be created in an aligned memory by having fixed size of Eigen object, we should use `Eigen::make_aligned_shared`. Knowing which one should be used may not be straightforward to all the users.

Also, some of the existing classes already have similar member functions, but the function name doesn't seem to be consistent. Some classes use `create()` for creating `std::shared_ptr` while some classes explicitly specify the kind of smart pointer such as `createCollisionGroupAsSharedPtr()`.

To address this issue, I'd suggest:
* having static member functions `createShared/Unique` (explicitly specify the type of shared and unique) that create smart pointers accordingly
* applying the createX functions for some important classes at first and then gradually applying it to other classes
* using macros for the createX functions in favor of DRY
* deprecating the existing createX functions that don't follow the name convention
* encouraging the users to use them

If this idea is acceptable, then we could consider `cloneShared/Unique` as well.

jslee02:
Closing as there are no plans for DART 6 and DART 7 will adopt a different approach for memory management.

PR #844, opened 2017-02-07
Add createShared() and createUnique() pattern

jslee02:
This PR adds a set of macros that defines two static functions ( `createShared` and `createUnique`) of creating shared and unique instances of the object. Developers must carefully choose which macro should be used depending on two factors: (1) whether the object requires aligned memory allocation and (2) whether the object's constructor is `public` or `protected`. Whereas, the users are encouraged to use `createShared` and `createUnique` without worrying about those factors if the static creators are defined.

Related issue: #813

**Open Decisions**
* [x] Name for shared object creator: `create()` vs `createShared()`: We use `create` and `createShared` for `std::shared_ptr` and raw pointer, respectively

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/844)


psigen:
I like this change! :+1:

PR #845, opened 2017-02-13
Apply naming convention of createShared/Unique to clone functions

jslee02:
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/845)


mxgrey:
I just had an idea which might be kind of nuts, but I thought I'd throw it out there anyhow.

I'm a little concerned about the verbosity of the createShared/Unique, and I'm also somewhat concerned about the way the container is being hardcoded into those two functions.

What if we approach it from a different angle? We could have a templated ``clone<Container>()`` function which accepts a container as a template argument. We can also make the template variadic to allow users to pass in additional arguments to their container of choice if needed. For example, if they want a specialized deleter, they could pass that as a template argument and/or function argument.

Then we can also provide a non-templated ``clone()`` function which can be used as a default, and it would use the container which is "recommended" for general purpose use by the DART developers.

[Here is a gist depicting the concept with a simple test implementation.](https://gist.github.com/mxgrey/dcc22a5c79b3efd797d06f980c3725c4)

I think the key advantages of this approach are that we get complete flexibility without costing us any verbosity. This would be broadly compatible with ``boost`` containers, custom containers, and any future containers that might not exist yet.

That said, there are some classes where we cannot currently deploy this, specifically when it comes to the ``Skeleton``, ``BodyNode``, and ``Joint`` classes. Speaking as the person who implemented the memory management scheme for those classes, I think it would be worth revisiting the ownership structure for all of them at some point in the future. The scheme that I cobbled together seems to be working, but it could most likely be improved. I think there's a challenging negotiation between user freedoms, user convenience, and maintainability that needs to be resolved in order to design an ideal memory management scheme for those structures.

mxgrey:
Also, I might suggest using the same template-based scheme for the ``create`` functions. I'm sorry that I didn't bring this up when issue #813 was raised, but this was all a spur of the moment idea that I had just now.

Although, as a counter-argument to myself, it might not accomplish much to offer the templated version of the clone function, since the user can probably just manually construct the container of their choosing more easily than figuring out how to correctly use a templated ``clone<Container>()`` function.

jslee02:
Making `create` and `clone` function flexible sounds good to me. However, we might need some additional code to achieve the original motivation behind of #813, which is to make those functions to be agnostic to aligned/non-aligned memory allocation. Simply, I would like to extend your test implementation to use `std::make_shared` and `dart::common::make_unique` in it if applicable.

[Here](https://ideone.com/01cR75) is extended implementation that I believe this is a minimal change.

The reason we need the additional structures is that the creator functions for smart pointers are varied in the smart pointer type (e.g., `std::make_shared`, `dart::common::make_unique`) and even in whether aligned memory allocation is required (e.g., `dart::common::make_aligned_shared`). So we need to template the creator functions using template struct.

Allowing custom deletor with the templated creator functions is rather tricky because [`std::make_shared` doesn't take custom deletor for example](http://stackoverflow.com/a/34243637). We could use `std::allocate_shared` as suggested in the post, but it requires [Alllocator](http://en.cppreference.com/w/cpp/concept/Allocator) rather than allocator/deallocator functions so we might need another template struct to handle this. In the test implementation, I left that part not to use `std::make_shared` for simplicity.

jslee02:
Another thought: If we go with this implementation, then it seems we have two API for the same operation.

For example,
```cpp
auto make_shared_test = make_shared<Object>();
auto create_shared_test = Object::create<std::shared_ptr>(); // same as above but different API

auto make_unique_test = make_unique<Object>();
auto create_unique_test = Object::create<std::unique_ptr>(); // same as above but different API
```

If we choose one API among them, `make_unique` looks more C++ STL way to me, which we might prefer.  Also, if we don't support the static create function, then we can reduce the code complexity by removing the template struct  for this (`CreateImpl` in the test implementation).

I don't have strong preference here, but I feel like it would be good to follow C++ STL way here.

mxgrey:
I think the original motivation for the ``create`` functions came from the ``Skeleton`` class. There were two reasons that I recall for it: (1) Skeleton objects need to carry a ``std::weak_ptr`` reference to itself, which at the time I thought required some extra work outside of the constructor, because I wasn't aware of ``std::enable_shared_from_this``, and (2) Skeleton objects need to be managed as ``std::shared_ptr``, so we can't allow the user to instantiate Skeletons into any other type of container.

The motivation for (1) is easily dismissed because we can just use ``std::enable_shared_from_this``, but motivation (2) is still relevant. We can't allow the user to instantiate a ``Skeleton`` object into a ``unique_ptr`` because it would clobber the memory management system and probably result in a considerable amount of undefined behavior.

I'd be very open to reworking the memory management system if we can think of an alternative implementation that makes motivation (2) irrelevant, but I haven't been able to think of a viable alternative for the past two years (not to say that an alternative is nonexistent).

However, I'm pretty sure these concerns exclusively pertain to the ``Skeleton`` class. Any other classes that have a ``create()`` function were probably just mimicking the API of ``Skeleton`` for the sake of consistency. So I guess the question is whether we want to have (1) one single paradigm for the sake of consistency or (2) two paradigms where one is used for the classes that need it while the other is used everywhere else.

jslee02:
> Any other classes that have a create() function were probably just mimicking the API of Skeleton for the sake of consistency.

Besides on that reason, the another goal is to have API that is agnostic to the memory alignment requirement so that the users don't need to worry about which creator function should be used. Previously, the user had to carefully choose the proper creator function between `std::make_shared()` and `dart::common::make_aligned_shared()` depending on the class when creating a shared object for example. Using wrong creator function can result in causing segfault without a descriptive error message.

It seems the two paradigms are (1) using static member function like `Object::create<ContainerT>()` and (2) using global function like `make_shared<Object>()` or `make_unique<Object>()`. I believe (1) would just call (2) so having (1) means having both of (1) and (2). Also, additional metaprogramming is required for (1) to support generic creator for different smart pointer types.

I would prefer to have one single paradigm so (2) is preferred (removing all the static create function) since looks more the standard way of C++ to me. So I would expect to have following use cases:

```cpp
auto skeleton_shared = dart::common::make_shared<Skeleton>(); // make_shared is specialized for Skeleton to call the protected constructor

// Object -- Aligned memory allocation is NOT required
auto object_shared = dart::common::make_shared<Object>();
auto object_unique = dart::common::make_unique<Object>();

// AlignedObject -- Aligned memory allocation required
auto aligned_object_shared = dart::common::make_shared<AlignedObject>();
auto aligned_object_unique = dart::common::make_unique<AlignedObject>();
```

For clone functions, we could apply your suggestion as:
```cpp
auto clone_shared = object->clone<std::shared_ptr>();
```

mxgrey:
I'm a little bit concerned about having an API that looks *so* much like the C++ std API without actually *being* the C++ std API, because it could lead to user confusion (i.e. not understanding that they should use ``dart::common::make_shared`` rather than ``std::make_shared``) or namespace confusion for the compiler (since there would be ambiguity whenever a user has both ``using namespace std;`` and ``using namespace dart::common;`` in their code).

I don't know if it's just meant to be a funny anecdote, but I've seen videos from Stephan T. Lavavej that the STL API intentionally has unpopular characteristics (like all lowercase and underscores) to avoid collisions with non-STL APIs. From what I understand, it would somewhat go against the spirit of the design of the STL if we were to mimic it.

jslee02:
Okay. I don't have a strong preference here. Do you have a suggestion for the names of them? My goal of this change is providing identical API regardless of the aligned memory requirement so naming wouldn't be a matter to me. I'm also open to having static create member function.

mxgrey:
I've given it some thought, but I can't say that I've arrived at a solid conclusion. I think I have some bias since I'm pretty sure I initially put in the ``create`` static member function pattern.

Using an external function rather than a static member function does seem better to me from a developer viewpoint, because we wouldn't have to define so many member functions for so many different classes. But from a user standpoint, there is considerably more text required for something like
```dart::common::make_shared<dart::dynamics::Skeleton>()``` 
than 
```dart::dynamics::Skeleton::create()```
and I personally find the latter to be more aesthetically appealing and more easy to read.

I would probably recommend seeking out a second opinion about this if you don't have a pattern than you prefer, because I can't really decide which pattern I think would be better.

jslee02:
Let's continue this discussion in #1017.

Closing in favor of #1017

PR #849, opened 2017-02-28
The same bug for SDF Parser

sehoonha:
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/849)


jslee02:
Thanks for the fix!

PR #855, opened 2017-03-15
Fix transform of ScrewJoint with thread pitch

jslee02:
There was a bug in calculating the transform of ScrewJoint with thread pitch. This bug was found in a [pull request of Gazebo](https://bitbucket.org/osrf/gazebo/pull-requests/2547/dart-6/diff#comment-33149101).

This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/855)


scpeters:
Thanks @jslee02

scpeters:
This fixes the gazebo test that was failing. Thanks!

jslee02:
Great! Will merge this once Travis becomes happy.

PR #856, opened 2017-03-18
Use gtest_main

jslee02:
This PR links `gtest_main` to each test target so that we don't need to repeatedly write the same `main()` function.

[Reference](https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#writing-the-main-function)

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/856)


psigen:
:+1: This is exactly what `gtest_main` exists for.  Good find.

mkoval:
👍 LGTM. Great find! 🎉

Issue #434, opened 2015-07-06
TinyXML vs TinyXML2


mkoval:
DART currently depends on both TinyXML (version 2.6.2) and TinyXML2. I initially thought that these two packages were the same, but they actually do correspond to separate Debian packages (`libtinyxml-dev` vs. `libtinyxml2-dev`). Compilation breaks if I remove the dependency on TinyXML2.

What is the difference between these two versions? Can we remove one of these dependencies?

mxgrey:
It looks like the URDF parser (which is the ``urdfdom`` library) depends on tinyxml, while DART's native skel parser depends on tinyxml2.

I'm all in favor of minimizing dependencies, but at the same time I'd rather not use a worse/older version of a library. It doesn't look like we can get rid of tinyxml if we want to be able to use ``urdfdom``, so either we'd have to adopt the old tinyxml library everywhere or we'd have to write our own version of ``urdfdom``, which would also involve a complete rewrite of the urdf parser. Unless someone knows of a urdf parsing alternative to ``urdfdom``.

mkoval:
I see. If TinyXML is a transitive dependency of `urdfdom`, then DART shouldn't have to depend on it explicitly (it should be handled by `find_package(urdfdom)`. Does DART itself use TinyXML (not TinyXML2) anywhere internally?

mxgrey:
I can't say for sure, but if we *are* using both TinyXML and TinyXML2 internally, then we should definitely consolidate on one of them, preferably TinyXML2.

mxgrey:
I did a comprehensive search, and it turns out that our interface to urdfdom does use TinyXML, because urdfdom seems to take in a ``TiXmlDocument`` to parse.

I know our urdf parser implementation is very old, so it might be possible to avoid the TinyXML interface. I'll take a quick look at this. But if the API of urdfdom hasn't changed, then we're stuck with a TinyXML dependency.

mxgrey:
Okay, I need to revise my statement. ``urdfdom`` does **not** take in a ``TiXmlDocument``.

The situation is this: ``urdfdom`` takes in a ``std::string`` and passes back a ``ModelInterface`` which represents a single robot. This means that ``urdfdom`` is not suitable for handling a urdf that contains more than one model/robot/skeleton. We therefore have a layer of translation that goes through an xml document, finds each model, and passes them one by one into ``urdfdom``. Our layer of translation was written using TinyXML, for reasons that were probably lost three years ago.

[...]

Issue #29, opened 2013-02-03
Shape transformations


cerdogan:
Three problems:

1- Shape::mOffset is not used and should be removed. 
2- The shape is offset by mLocalCOM for visualization and not offset at all for collision checking.
3- Shape class should have a transformation wrt the local frame of the node it belongs to.

PR #858, opened 2017-03-21
Drop tinyxml dependency

jslee02:
As discovered in #434, `tinyxml` is only a transitive dependency of `urdfdom` and DART doesn't use it in the public API. This PR drops the explicit dependency of `tinyxml`.

---
This change is [Reviewable](https://reviewable.io/reviews/dartsim/dart/858)


jslee02:
The build error on macOS is because the [urdfdom formula just depends on tinyxml as a build dependency](https://github.com/ros/homebrew-deps/blob/master/urdfdom.rb#L10). I will submit a pull request to the upstream to fix that.

This PR will be pending on the change for sure.

jslee02:
Pull request is created: [homebrew-deps #29](https://github.com/ros/homebrew-deps/pull/29)

jslee02:
The Homebrew formula of urdfdom is fixed as [homebrew-deps #29 is merged](https://github.com/ros/homebrew-deps/pull/29). However, it seems `urdfdom` doesn't convey the dependency of `tinyxml` through `urdfdom_LIBRARIES`. This should be fixed in upstream. I'll submit another pull request for this.

jslee02:
Let me reopen when it's ready

PR #861, opened 2017-03-28
Add ODE collision detector

jslee02:
This PR adds the ODE collision detector support to DART.

The supported shape types are sphere, box, capsule, cylinder plane, and mesh; and the unsupported types are ellipsoid, cone, multi-sphere, and soft mesh.

The ODE collision detector only supports collision detection queries but not distance queries.

The ODE collision detector is only enabled when ODE 0.11 or later is installed on the system.

### DIFF
--- a/ci/install_osx.sh
+++ b/ci/install_osx.sh
@@ -9,6 +9,7 @@ cmake
 assimp
 fcl
 bullet
+ode
 flann
 boost
 eigen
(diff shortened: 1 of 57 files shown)


jslee02:
`test_Collision` fails due to a bug in `BulletCollisionDetector`. Will take care of this in a separate pull request. Merging this PR.

PR #862, opened 2017-03-29
Improve MultiSphereShape rendering

jslee02:
Previously, DART only draws the spheres in a `MultiSphereShape` while the actual shape is the convex hull enclosing all the spheres. This PR still doesn't render the convex hull but improves the visualization by drawing the cylinders that connect the pairs of the spheres.

Before:
image

After:
image

### DIFF
--- a/dart/gui/OpenGLRenderInterface.cpp
+++ b/dart/gui/OpenGLRenderInterface.cpp
@@ -154,33 +154,16 @@ void OpenGLRenderInterface::drawSphere(double radius, int slices, int stacks)
   gluSphere(quadObj, radius, slices, stacks);
 }
 
-void OpenGLRenderInterface::drawMultiSphere(
-    const std::vector<std::pair<double, Eigen::Vector3d>>& spheres,
+void drawOpenCylinderConnectingTwoSpheres(
+    OpenGLRenderInterface* ri,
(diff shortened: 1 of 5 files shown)


mxgrey:
I'm a bit concerned by the name of the class. I don't think the name really conveys what the class is actually representing. This whole time, I've been thinking that the ``MultiSphereShape`` is something more like a point cloud. What if it were called something like ``MultiSphereConvexHullShape``? Or if there's a technical term for this type of shape, that would probably be better. We could keep a ``MultiSphereShape`` alias for backwards compatibility. I can open this as a separate issue if you think it warrants more of a discussion.

jslee02:
I agree with you. I named it inspired by Bullet's `btMultiSphereShape` and [commented](https://github.com/dartsim/dart/pull/770/files#diff-68920f0af6029efc6c7a98aac46f5a60R42) that it's a convex hull shape, but it would be much better to change the name to show that explicitly.

`MultiSphereConvexHullShape` sounds good to me. I also came up with `SphereSweptConvexHullShape`, but it could confuse people that the shape was created by sweeping with a single sphere I think.

By the way, let's address the naming in a separate PR to wait for other feedback on the name.

psigen:
Sounds reasonable to me.

It might be enough to name it `MultiSphereHullShape` too. But adding some
"hull" terminology seems important.

PR #864, opened 2017-03-30
Add Factory class and collision detector factory

jslee02:
This PR adds `Factory`, which is an implementation of the [Abstract Factory Pattern](https://en.wikipedia.org/wiki/Abstract_factory_pattern), and applies it to creating collision detectors from the key of `std::string` type.

By doing so, it decouples the dependencies of bullet collision detector from `SkelParser`. Previously, the bullet collision detector type was hard-coded in `SkelParser` so that `dart-collision-bullet` had to be one of the dependencies of `dart-utils`. Now the optional collision detector (only bullet collision detector for now) can be created through the factory when `dart-collision-bullet` is linked.


dart/utils/SkelParser.cpp
```diff
@@ -695,40 +692,41 @@ simulation::WorldPtr readWorld(
...
+      }
     }
 
     if (!collision_detector)
     {
-      collision_detector = collision::FCLCollisionDetector::create();
+      collision_detector = collision::CollisionDetector::Factory::create("fcl");
       auto cd = std::static_pointer_cast<collision::FCLCollisionDetector>(
             collision_detector);
       cd->setPrimitiveShapeType(collision::FCLCollisionDetector::MESH);
```
mkoval: Can we put the code to create the `fcl_mesh` collision detector into a helper function? I am worried that the two versions of this code (i.e. the one in the `cdType == "fcl_mesh" block and in the default block) will become out of sync.

unittests/unit/test_Factory.cpp
```diff
@@ -0,0 +1,257 @@
...
+class Seattle : public City
+{
+public:
+  Seattle(int year) : City(year) {}
+
+  std::string getName() const override
+  {
+    return "Seattle";
+  }
+};
```
mkoval: I like your choice of city names. 😉

dart/common/Factory.hpp
```diff
@@ -0,0 +1,195 @@
...
+/// auto fclCd = CdFactory::create("fcl");
+/// \endcode
+template <typename KeyT,
+          typename BaseT,
+          template<typename...> class SmartPointerT = std::shared_ptr,
+          typename... Args>
+class Factory final
+{
+public:
+  using This = Factory<KeyT, BaseT, SmartPointerT>;
```
jslee02: That wouldn't work since we cannot use `this` in type aliasing.

dart/common/Factory.hpp
```diff
@@ -0,0 +1,195 @@
...
+  using Creator = std::function<CreatorReturnType(Args...)>;
+  using CreatorMap = std::unordered_map<KeyT, Creator>;
+  using RegisterResult = std::pair<typename CreatorMap::iterator, bool>;
+
+  /// Registers a object creator function with a key.
+  static RegisterResult registerCreator(const KeyT& key, Creator creator);
+
+  /// Registers the default object creator function with a key.
+  template <typename Derived>
+  static RegisterResult registerCreator(const KeyT& key);
```
jslee02: This is a legacy code that was used to support `ScopedFactoryRegistrar`, which is removed. Will remove the return type as well.

> As an added note - why does the first call to registerCreator win? I would expect a later call to win, i.e. use operator[] instead of insert() to perform the insert operation.

:+1:

dart/common/Factory.hpp
```diff
@@ -0,0 +1,191 @@
...
+  ~Factory() = delete;
+
+private:
+  /// Returns the object creator function map. A static map is created when this
+  /// function is firstly called.
+  static CreatorMap& getMap();
+};
+
+/// The default creator. Specialize this template class for smart pointer types
+/// other than std::unique_ptr and std::shared_ptr.
```
jslee02: `DefaultCreator` is now able to take arbitrary arguments.

dart/common/Factory.hpp
```diff
@@ -0,0 +1,191 @@
...
+template <typename KeyT,
+          typename BaseT,
+          typename DerivedT,
+          template<typename...> class SmartPointerT = std::shared_ptr>
+class FactoryRegistrar final
+{
+public:
+  using FactoryType = Factory<KeyT, BaseT, SmartPointerT>;
+  using Creator = typename FactoryType::Creator;
+  using This = FactoryRegistrar<KeyT, BaseT, DerivedT, SmartPointerT>;
```
jslee02: I believe `this` cannot be used here (see above).

jslee02:
@mkoval I believe all your comments are addressed. Could you do a second pass?

jslee02:
Sorry @mkoval , @mxgrey, I'm merging this as this is a blocker for other changes. Let's handle remained issues in the future if there is.

PR #865, opened 2017-03-30
Rename MultiSphereShape to MultiSphereConvexHullShape

jslee02:
This PR renames `MultiSphereShape` to `MultiSphereConvexHullShape` as suggested in this [comment](https://github.com/dartsim/dart/pull/862#issuecomment-290128731).

### DIFF
--- a/dart/dynamics/CMakeLists.txt
+++ b/dart/dynamics/CMakeLists.txt
@@ -8,6 +8,8 @@ dart_add_core_sources(${srcs} ${detail_srcs})
 
 # Generate header for this namespace
 dart_get_filename_components(header_names "dynamics headers" ${hdrs})
+# TODO: remove below line once the files are completely removed.
+list(REMOVE_ITEM header_names "MultiSphereShape.hpp")
 dart_generate_include_header_list(
   dynamics_headers
   "dart/dynamics/"
--- a/dart/dynamics/MultiSphereConvexHullShape.hpp
+++ b/dart/dynamics/MultiSphereConvexHullShape.hpp
@@ -101,7 +101,8 @@ private:
 
 };
 
-using MultiSphereShape = MultiSphereConvexHullShape;
+DART_DEPRECATED(6.2)
+typedef MultiSphereConvexHullShape MultiSphereShape;
 
 }  // namespace dynamics
 }  // namespace dart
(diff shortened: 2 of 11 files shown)


dart/dynamics/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ dart_add_core_sources(${srcs} ${detail_srcs})
 
 # Generate header for this namespace
 dart_get_filename_components(header_names "dynamics headers" ${hdrs})
+# TODO: remove below line once the files are completely removed.
+list(REMOVE_ITEM header_names "MultiSphereShape.hpp")
```
mkoval: Why is this necessary?

jslee02: Yes, this excludes `MultiSphereShape.hpp` from the generated header `dart/dynamics/dynamics.hpp` so that it suppresses the warning of including the deprecated header.

dart/dynamics/MultiSphereConvexHullShape.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+  /// Update bounding box (in the local coordinate frame) of the shape.
+  void updateBoundingBoxDim();
+
+  /// Spheres
+  Spheres mSpheres;
+
+};
+
+using MultiSphereShape = MultiSphereConvexHullShape;
```
mkoval: I _think_ we can mark this `using` alias as deprecated:

That makes sense. I _think_ we can also [mark the `using` alias as deprecated](http://en.cppreference.com/w/cpp/language/attributes):

> Indicates that the use of the name or entity declared with this attribute is allowed, but discouraged for some reason. This attribute is allowed in declarations of classes, **typedef-names**, variables, non-static data members, functions, enumerations, and template specializations. A name declared non-deprecated may be redeclared deprecated. A name declared deprecated cannot be un-deprecated by redeclaring it without this attribute.

jslee02: We probably should stay with `DART_DEPRECATED(...)` until DART uses C++14 because `[[deprecated]]` is available since C++14.

By the way, maybe I'm missing something here, but it seems `DART_DEPRECATED(...)` is not available on typedef using `using`:
```cpp
DART_DEPRECATED(6.2)
using MultiSphereShape = MultiSphereConvexHullShape; // can't compile

DART_DEPRECATED(6.2)
typedef MultiSphereConvexHullShape MultiSphereShape; // okay
```

mkoval: That might be true. That webpage only mentions `typedef`. While I generally favor `using` directives to `typedef`s, I am slightly in favor of using `DART_DEPRECATED(6.2) typedef` here since this is a temporary workaround that will only last until the next major revision.

PR #866, opened 2017-03-31
Use make_unique with full namespace

jslee02:
If DART is built with C++14 enabled, `dart::common::make_unique()` conflicts with `std::make_unique` (introduced in C++14). This PR adds the full namespace where it's missing to avoid the confliction.

### DIFF
--- a/dart/common/detail/Cloneable.hpp
+++ b/dart/common/detail/Cloneable.hpp
@@ -177,7 +177,7 @@ template <typename... Args>
 ProxyCloneable<Base, OwnerT, DataT, setData, getData>::ProxyCloneable(
     Args&&... args)
   : mOwner(nullptr),
-    mData(make_unique<Data>(std::forward<Args>(args)...))
+    mData(dart::common::make_unique<Data>(std::forward<Args>(args)...))
 {
   // Do nothing
 }
@@ -278,7 +278,7 @@ void ProxyCloneable<Base, OwnerT, DataT, setData, getData>::set(
     return;
   }
 
-  mData = make_unique<Data>(std::move(data));
+  mData = dart::common::make_unique<Data>(std::move(data));
 }
 
 //==============================================================================
@@ -338,7 +338,7 @@ template <class Base, class OwnerT, class DataT,
 std::unique_ptr<Base> ProxyCloneable<
     Base, OwnerT, DataT, setData, getData>::clone() const
 {
-  return make_unique<ProxyCloneable>(get());
+  return dart::common::make_unique<ProxyCloneable>(get());
 }
 
 //==============================================================================


mxgrey:
I don't have anything against this change, but I don't see why there would be a conflict, unless someone recklessly places ``using namespace std;`` before an ``#include <dart/...>`` which contains this header. If that particular case is what we're trying to avoid, then this makes sense.

jslee02:
> I don't see why there would be a conflict, unless someone recklessly places using namespace std; before an #include <dart/...> which contains this header.

I believe so, but I got following error when building DART as it is with `-std=c++14`:
```bash
In file included from /dart/ode_collision_detector_2/dart/common/Cloneable.hpp:331:0,
                 from /dart/ode_collision_detector_2/dart/common/Aspect.hpp:37,
                 from /dart/ode_collision_detector_2/dart/common/detail/EmbeddedAspect.hpp:35,
                 from /dart/ode_collision_detector_2/dart/common/EmbeddedAspect.hpp:35,
                 from /dart/ode_collision_detector_2/dart/dynamics/BodyNode.hpp:44,
                 from /dart/ode_collision_detector_2/dart/dynamics/BodyNode.cpp:32:
/dart/ode_collision_detector_2/dart/common/detail/Cloneable.hpp: In instantiation of ‘std::unique_ptr<_Tp> dart::common::ProxyCloneable<Base, OwnerT, DataT, setData, getData>::clone() const [with Base = dart::common::Aspect::Properties; OwnerT = dart::dynamics::BodyNode; DataT = dart::common::CloneableMap<std::map<std::type_index, std::unique_ptr<dart::common::CloneableVector<std::unique_ptr<dart::dynamics::Node::Properties, std::default_delete<dart::dynamics::Node::Properties> > > > > >; void (* setData)(OwnerT*, const DataT&) = dart::dynamics::detail::setAllNodeProperties; DataT (* getData)(const OwnerT*) = dart::dynamics::detail::getAllNodeProperties]’:
/dart/ode_collision_detector_2/dart/dynamics/BodyNode.cpp:2478:1:   required from here
/dart/ode_collision_detector_2/dart/common/detail/Cloneable.hpp:341:37: error: call of overloaded ‘make_unique(dart::common::ProxyCloneable<dart::common::Aspect::Properties, dart::dynamics::BodyNode, dart::common::CloneableMap<std::map<std::type_index, std::unique_ptr<dart::common::CloneableVector<std::unique_ptr<dart::dynamics::Node::Properties, std::default_delete<dart::dynamics::Node::Properties> > > > > >, dart::dynamics::detail::setAllNodeProperties, dart::dynamics::detail::getAllNodeProperties>::Data)’ is ambiguous
   return make_unique<ProxyCloneable>(get());
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
```
There are some places `using namespace std;` is used in example and tests. Maybe it's because of that.

mxgrey:
Based on that error, I can only surmise that the ``using namespace`` directive applies to the entire scope that it exists in, including any lines that precede it. I'm having trouble finding any reference that explicitly states that this is the case, but if it is, then we definitely need this change.

Issue #850, opened 2017-03-02
DART is incompatible with FCL 0.6 (current master of FCL)


mxgrey:
[DART](https://github.com/dartsim/dart/blob/ca61f40764d107c54498d57cee0a8f2d7d915a94/dart/math/MathTypes.hpp#L75) and [FCL 0.6](https://github.com/flexible-collision-library/fcl/blob/2e3d262195fbfd5369e5aa00fb6c6c7c0ae855de/include/fcl/common/types.h#L124) both contain identical definitions of ``Eigen::aligned_vector`` and a few other similar data types. This causes a definition collision during compilation.

While the motivation behind these names is sensible, it highlights the dangers of defining things within external namespaces. My suggestion would be that DART and FCL rename these types as ``dart::aligned_vector`` and ``fcl::aligned_vector`` or something similar. The two will be interoperable, but giving them separate names will allow compilation.

This is low priority since there is no strict need to use FCL 0.6 with DART, but we'll want to fix this eventually.

a-price:
FCL 0.6 has had a pretty significant refactor, so there's a lot more than just the aligned_vector code that needs to be changed. The new FCL code is highly templated, and the name of e.g. the main vector class has changed. I've created a [branch](https://github.com/a-price/dart/tree/fcl-0.6) over in my fork with the necessary changes to get things to work with 0.6, but it's not backwards-compatible with 0.5. Probably what would need to happen would be to have the [FCLTypes](https://github.com/dartsim/dart/blob/master/dart/collision/fcl/FCLTypes.hpp) file typedef a bunch of fcl types according to the version number.

jslee02:
I believe this is resolved by #873 and #936.

PR #873, opened 2017-04-12
Handle FCL 0.6

a-price:
I abstracted out the FCL incompatibilities by typedefing a bunch of the common FCL types in the header, depending on which version was found. This likely needs more thorough testing, as I only had FCL 0.3 and 0.6 to test against.


a-price:
This is in response to #850. Also, it took a few tries to find all the relevant fcl code, so it might be better to squash and rebase this PR rather than doing a standard merge.

dart/collision/fcl/FCLTypes.hpp
```diff
@@ -55,25 +79,97 @@ template <class T> using fcl_shared_ptr = boost::shared_ptr<T>;
...
 
-namespace dart {
-namespace collision {
+namespace dart
+{
+namespace fcl
+{
+#if FCL_VERSION_AT_LEAST(0,6,0)
+// Geometric fundamentals
+typedef ::fcl::Vector3<double> Vector3;
```
jslee02: The current convention with FCL specific types for is using a prefix. Could you remove `fcl` namespace and prepend `FCL` for the type aliasing (e.g., `using FCLVector3 = ::fcl::Vector3<double>` or `using FCLVector3 = ::fcl::Vector3d`)?

dart/collision/fcl/FCLTypes.hpp
```diff
@@ -55,25 +79,97 @@ template <class T> using fcl_shared_ptr = boost::shared_ptr<T>;
...
+#endif
+
+#if FCL_VERSION_AT_LEAST(0,4,0) && !FCL_VERSION_AT_LEAST(0,6,0)
+typedef ::fcl::Ellipsoid Ellipsoid;
+#endif
+
+using ::fcl::GJKSolverType;
+
+template<typename BV>
+using BVHModel = ::fcl::BVHModel<BV>;
```
jslee02: I believe there were no changes for `BVHModel` across 0.5 and 0.6. Could you remove this type aliasing?

dart/collision/fcl/FCLTypes.hpp
```diff
@@ -55,25 +79,97 @@ template <class T> using fcl_shared_ptr = boost::shared_ptr<T>;
...
+
+} // namespace dart::fcl
+
+namespace collision
+{
```
jslee02: Nit: Place `{` on the same line.

dart/collision/fcl/FCLTypes.hpp
```diff
@@ -55,25 +79,97 @@ template <class T> using fcl_shared_ptr = boost::shared_ptr<T>;
...
+typedef ::fcl::DistanceRequest DistanceRequest;
+typedef ::fcl::DistanceResult DistanceResult;
+typedef ::fcl::Contact Contact;
+#endif
+
+#if FCL_VERSION_AT_LEAST(0,4,0) && !FCL_VERSION_AT_LEAST(0,6,0)
+typedef ::fcl::Ellipsoid Ellipsoid;
+#endif
+
+using ::fcl::GJKSolverType;
```
jslee02: I believe we don't need this line. Please remove.

dart/collision/fcl/FCLTypes.hpp
a-price: I saw that that was the existing usage, but wasn't sure how to resolve conflicts like dart::collision::FCLCollisionObject, which is already a Dart class.

a-price: Apparently "collision" triggers an emoji...

jslee02: I see. Then I would like to use `dart::collision::fcl` namespace rather than `dart::fcl` because it's part of the collision component.

@mxgrey @mkoval Any thoughts to add? :sweat:

cmake/FindCCD.cmake
```diff
@@ -17,6 +17,14 @@ find_package(PkgConfig QUIET)
...
 
 # Include directories
+# Give explicit precedence to ${PC_CCD_INCLUDEDIR}
+find_path(CCD_INCLUDE_DIRS
+    NAMES ccd/ccd.h
+    HINTS ${PC_CCD_INCLUDEDIR}
+    NO_DEFAULT_PATH
+    NO_CMAKE_PATH
+    NO_CMAKE_ENVIRONMENT_PATH
+    NO_SYSTEM_ENVIRONMENT_PATH)
```
jslee02: Could you elaborate why we need this?

cmake/FindFCL.cmake
```diff
@@ -17,15 +17,30 @@ find_package(PkgConfig QUIET)
...
+  find_path(FCL_INCLUDE_DIRS
+      NAMES fcl/collision.h
+      HINTS ${PC_FCL_INCLUDEDIR}
+      PATHS "${CMAKE_INSTALL_PREFIX}/include")
+else()
+  find_path(FCL_INCLUDE_DIRS
+      NAMES fcl/narrowphase/collision.h
+      HINTS ${PC_FCL_INCLUDEDIR}
+      PATHS "${CMAKE_INSTALL_PREFIX}/include")
+endif()
```
jslee02: Could we use remove this if-statement by using one header (like `fcl/config.h`) that is in common for all the versions we support?

a-price: Both of these changes were required for installing FCL 0.6 alongside a ROS installation. Basically, it forces cmake to try to use the pkg-config info before trying out any of the default directories (which, according to the documentation, it should do already, but that wasn't the result I was getting; maybe it works in a newer version of cmake?). With the include paths, if you just specify `fcl/config.h` it was happily finding the ROS includes, compiling as FCL 0.5, then linking to FCL 0.6, triggering some linker sadness. I also had to modify my `CPATH` environment when using alongside ROS, so maybe that is sufficient for the include directory finding. I can test this further once the semester's over.

jslee02: Thank you for the clarification! That makes sense. It would be good then to add a note of the summary for a future reference. :smile:

PR #875, opened 2017-04-17
Add readAll() to Resource and ResourceRetriever

jslee02:


dart/common/Resource.cpp
```diff
@@ -0,0 +1,55 @@
...
+//==============================================================================
+std::string Resource::readAll()
+{
+  std::string content;
+  content.resize(getSize());
+  const auto result = read(&content.front(), content.size(), 1);
+  // Safe because std::string is guaranteed to be contiguous in C++11.
+
+  if (result != 1)
+    throw std::runtime_error("Failed reading data from a resource.");
```
mkoval: I am personally fine with this. But I thought DART tried explicitly avoided using exceptions. For consistency, it might be better to make the `std::string` an output parameter and return a `bool` success flag.

dart/common/Resource.hpp
```diff
@@ -75,6 +75,12 @@ class Resource
   /// \param[in] _count Number of elements, each of _size bytes.
   /// \note This method has the same API as the standard fread function.
   virtual std::size_t read(void *_buffer, std::size_t _size, std::size_t _count) = 0;
+
+  /// Reads all remaining data from this resource, and returns it as a string.
```
mkoval: I don't think "read remaining data" is true. I suspect that this function will fail if it is called on a `Resource` that has been partially consumed because `getSize()` returns the size of the _entire_ resource and `read` starts reading from the current cursor position.

dart/common/Resource.hpp
```diff
@@ -75,6 +75,12 @@ class Resource
...
+  ///
+  /// \return The string retrieved from the resource.
+  /// \throw std::runtime_error when failed to read sucessfully.
+  std::string readAll();
```
mkoval: I suggest marking this `virtual`. Some classes may be able to implement this more efficiently than the generic implementation used in the base class.

dart/common/ResourceRetriever.hpp
```diff
@@ -53,11 +53,18 @@ class ResourceRetriever
...
-  // ResourceRetriever will be interacting with external resources that you
-  // don't necessarily have control over so we cannot guarantee that you get the
-  // same result every time with the same input Uri. Indeed, const-qualification
-  // for those functions is pointless.
+  /// Reads all data from the resource of uri, and returns it as a string.
+  ///
+  /// \param[in] uri URI to the resource to be retrieved.
+  /// \return The string retrieved from the resource.
+  /// \throw std::runtime_error when failed to read sucessfully.
+  std::string readAll(const Uri& uri);
```
mkoval: I suggest also making this virtual, for the same reason as I explained above.

dart/common/Resource.cpp
mxgrey: I think there's some leniency with exceptions used during parsing, since someone would have to be pretty self-destructive to be doing any parsing during a critical control loop.

jslee02: > But I thought DART tried explicitly avoided using exceptions.

That is true, but I thought we would like to use exceptions when we handle external resources such as files. For example, we have already used exceptions in the parsers.

psigen:
@mxgrey: Just a side question:  Does DART already make use of `noexcept` to distinguish code paths that never throw exceptions from ones that do?

It seems like that could make it clearer to end-users which DART functions have this guarantee and which don't, since you are right that many loading/visualization/debugging functions don't need it and could have very nice `std::exception`s instead.

I don't know what the pros/cons are here though, just curious.

mxgrey:
I agree that we should consider using ``noexcept`` for the benefit clarity and potential performance improvements. The catch (no pun intended) is that we do allow exceptions (in the form of ``assert``) everywhere when DART is compiled in debug mode. The logic is that unwinding the stack is likely to be problematic when running a critical control loop, but it can be very useful when debugging errors. 

Perhaps even more problematic is that applying the ``noexcept`` specifier to a function which calls any potentially-throwing functions will result in ``std::terminate`` if the ``noexcept`` function does not handle the exception itself. So if anything used inside of a DART ``noexcept`` function does throw an exception, then that DART function **must** handle the exception or else the entire program will be terminated.

I *would* like to be able to offer a guarantee to the user that DART functions will handle all possible internal exceptions in a reasonable way, but I think we would need to take some time to examine where all the potential exceptions could pop up inside of DART before we commit to applying the ``noexcept`` specifier anywhere.

Conclusion: I think if we have an opportunity to review the exception safety of DART in detail and can determine that using the ``noexcept`` specifier is safe, then we could use a macro which expands to ``noexcept`` in release mode while being blank in debug mode.

psigen:
Thanks @mxgrey.  I moved the discussion to #878 so we can continue brainstorming about this.

PR #879, opened 2017-04-19
Add getters for joints/bodies to MetaSkeleton

jslee02:
The newly added APIs are just syntactic sugar to get the joints and body nodes with a name. Probably, some of the implementation are not that efficient.

### DIFF
--- a/dart/dynamics/MetaSkeleton.hpp
+++ b/dart/dynamics/MetaSkeleton.hpp
@@ -95,19 +95,19 @@ public:
   /// Get const BodyNode whose index is _idx
   virtual const BodyNode* getBodyNode(std::size_t _idx) const = 0;
 
-  /// Returns the joint of given name.
+  /// Returns the body node of given name.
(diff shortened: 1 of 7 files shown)


dart/dynamics/MetaSkeleton.hpp
```diff
@@ -95,12 +95,41 @@ class MetaSkeleton : public common::Subject
   /// Get const BodyNode whose index is _idx
   virtual const BodyNode* getBodyNode(std::size_t _idx) const = 0;
 
+  /// Returns the body node of given name.
+  ///
+  /// When there are multiple body nodes with the same name, returns the first
+  /// body node, which is implementation dependent, and prints a warning.
```
mkoval: Nit: Add a note about how this can occur, i.e. when the `MetaSkeleton` contains `BodyNode`s from multiple `Skeleton`s. The same comment applies to the other three functions you added.

PR #887, opened 2017-05-18
Add an analytic IK solver (IkFast)

jslee02:
This pull request introduces an API for [IkFast](http://openrave.org/docs/0.8.2/openravepy/ikfast/), which is an analytic IK solver that is a submodule of OpenRAVE.

Here is an example of Solving IK using IkFast for Barrett WAM Arm:
[IK using IkFast for WAM](https://www.youtube.com/watch?v=JBFiLsv5VHQ)

TODOs
- [ ] Add benchmark (IKFast vs iterative Jacobian based IK solver) of the Barret WAM Arm demo


dart/common/detail/SharedLibraryManager.cpp
```diff
@@ -0,0 +1,69 @@
...
+    const std::string& fileName)
+{
+  const auto iter = mLibraries.find(fileName);
+
+  const auto found = iter != mLibraries.end();
+  if (found)
+  {
+    auto lib = iter->second.lock();
+
+    if (lib)
```
mkoval: Add a comment that this check could fail if all instances to the library go out of scope, since `iter->second` is a `std::weak_ptr`.

dart/common/SharedLibrary.hpp
```diff
@@ -0,0 +1,114 @@
...
+  /// Windows).
+  /// \return Pointer to the created SharedLibrary upon success in loading.
+  /// Otherwise, returns nullptr.
+  explicit SharedLibrary(ProtectedContructionTag, const std::string& fileName);
+
+  /// Destructor
+  virtual ~SharedLibrary();
+
+  /// Returns the file name of the shared library
+  const std::string& getFileName() const;
```
mkoval: Nit: I believe this returns the path, not the filename.

dart/common/SharedLibrary.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+  /// Returns true if the shared library loading was successful.
+  bool isValid() const;
+
+  /// Returns a symbol from the shared library
+  void* getSymbol(const std::string& symbolName) const;
```
mkoval: What does it return if the symbol does not exist? `nullptr`?

Also, it might be worth mentioning that you have to `reinterpret_cast` the return value to the appropriate type to make use of the `void*` returned by this function.

dart/common/detail/SharedLibraryManager.cpp
```diff
@@ -0,0 +1,69 @@
...
+      mLibraries.erase(iter);
+  }
+
+  const auto newLib = std::make_shared<SharedLibrary>(
+      SharedLibrary::ProtectedConstruction, fileName);
+
+  if (!newLib->isValid())
+    return nullptr;
+
+  mLibraries[fileName] = newLib;
```
mkoval: This performs the same lookup on the map twice. I suggest changing the `find()` above to an `insert()` or `emplace()` and checking the `bool` flag that it returns to check whether the key already exists. That lets you replace this lookup with a simple iterator assignment.

dart/common/detail/SharedLibraryManager.hpp
```diff
@@ -0,0 +1,70 @@
...
+namespace common {
+
+class SharedLibrary;
+
+namespace detail {
+
+class SharedLibraryManager final : public Singleton<SharedLibraryManager>
+{
+public:
+  /// Loads a shared library of fileName.
```
mkoval: Nit: Strange wording. "Loads the shared library with the specified filename" might be better.

dart/common/detail/SharedLibraryManager.hpp
```diff
@@ -0,0 +1,70 @@
...
+  ///
+  /// \param[in] The filename of the shared library. If the filename doesn't
+  /// include the extension, this function will use the best guess depending on
+  /// the OS (e.g., '.so' for Linux, '.dylib' for macOS, and '.dll' for
+  /// Windows).
```
mkoval: I am still very confused if this is a filename (that then goes through some search path), an absolute path, a relative path, or any of the above. Please add some documentation about this. Does this add the `lib` prefix to libraries on Linux? What if I don't want it to add an extension, e.g. if I am using a non-standard extension?

I also suggest adding some instructions for cross-platform usage. E.g. should I omit the extension, if possible, so DART can figure out the correct extension for each platform?

dart/common/detail/SharedLibraryManager.hpp
```diff
@@ -0,0 +1,70 @@
...
+  /// \return Pointer to the shared library upon success. Otherwise, returns
+  /// nullptr.
+  std::shared_ptr<SharedLibrary> load(const std::string& fileName);
+
+protected:
+  friend class Singleton<SharedLibraryManager>;
+
+protected:
+  std::unordered_map<std::string, std::weak_ptr<SharedLibrary>> mLibraries;
```
mkoval: Note that `mLibraries` never gets "garbage collected" when an element goes out of scope. A cleaner way of implementing this would be to use a `shared_ptr` with a custom deleter.

dart/dynamics/IkFast.cpp
```diff
@@ -0,0 +1,329 @@
...
+  eerot[2*3+1] = tf.linear()(2, 1);
+  eerot[2*3+2] = tf.linear()(2, 2);
+
+  eetrans[0] = tf.translation()[0];
+  eetrans[1] = tf.translation()[1];
+  eetrans[2] = tf.translation()[2];
+}
+
+//==============================================================================
+bool isFreeJoint(int numFreeParams, const int* freeParams, int index)
```
mkoval: Nit: `numFreeParams` should be a `std::size_t`.

dart/dynamics/IkFast.cpp
```diff
@@ -0,0 +1,329 @@
...
+  ikfastSolution.GetSolution(solutionValues, freeValues);
+
+  bool limitViolated = false;
+
+  auto index = 0u;
+  solution.mConfig.resize(dofIndices.size());
+  for (auto i = 0u; i < solutionValues.size(); ++i)
+  {
+    if (isFreeJoint(numFreeParameters, freeParameters, i))
+      continue;
```
mkoval: You initialize `solution.mConfig` to an array of `dofIndices.size()` zeros. This `continue` skips incrementing `index`, which makes me think that the last `dofIndices.size() - numFreeParameters` elements of `solution.mConfig` will always be zero. This looks like a bug to me.

Also, I think it would be cleaner to implement this by converting `freeParameters` to an `std::vector<bool>` of flags that indicate whether each DOF is a free parameter or not.

dart/dynamics/IkFast.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+  return true;
+}
+
+//==============================================================================
+void IkFast::configure() const
+{
+  const auto ikFastNumJoints = getNumJoints();
+  const auto ikFastNumFreeJoints = getNumFreeParameters();
+  const auto ikFastNumNonFreeJoints = ikFastNumJoints - ikFastNumFreeJoints;
```
mkoval: Check that `ikFastNumJoints >= ikFastNumFreeJoints`.

dart/dynamics/IkFast.cpp
```diff
@@ -0,0 +1,329 @@
...
+      dterr << "[IkFast::configure] Failed to configure. An element of the "
+            << "given " << dofMapName << " '" << dof
+            << "' is not a dependent dofs of Node '" << ik->getNode()->getName()
+            << "'.\n";
+      return false;
+    }
+  }
+
+  return true;
+}
```
mkoval: I think this function is trying to verify that all of the DOFs effect the `EndEffector` frame. However, I am having ar really hard time following the implementation. Can you add a few comments?

dart/dynamics/IkFast.hpp
```diff
@@ -0,0 +1,112 @@
...
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef DART_DYNAMICS_IKFAST_HPP_
+#define DART_DYNAMICS_IKFAST_HPP_
+
+#include <array>
+
+#define IKFAST_HAS_LIBRARY
```
mkoval: What is the purpose of this `#define`?

dart/dynamics/IkFast.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+namespace dart {
+namespace dynamics {
+
+class IkFast : public InverseKinematics::Analytical
+{
+public:
+  /// Constructor
+  IkFast(
+      InverseKinematics* ik,
```
mkoval: What is this argument? Also, this function desperately needs docstrings for its parameters.

dart/dynamics/InverseKinematics.cpp
```diff
@@ -1078,7 +1090,8 @@ void InverseKinematics::Analytical::computeGradient(
       mIK->getErrorMethod().computeDesiredTransform(
         mIK->getNode()->getWorldTransform(), _error);
 
-  if(PRE_ANALYTICAL == mAnalyticalP.mExtraDofUtilization
+  if((PRE_ANALYTICAL == mAnalyticalP.mExtraDofUtilization
+      || PRE_AND_POST_ANALYTICAL == mAnalyticalP.mExtraDofUtilization)
```
mkoval: What does this mean?

dart/dynamics/SharedLibraryIkFast.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+  if (!symbol)
+  {
+    dterr << "Failed to load the symbol '<<" << symbolName
+          << "' from the file '" << fileName << "'.\n";
+    return false;
+  }
+
+  func = reinterpret_cast<FunctionType>(symbol);
+  // TODO(JS): Is there a way to check the compatibility of the symbol?
```
mkoval: Not that I know of. 😞

dart/dynamics/SharedLibraryIkFast.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+//==============================================================================
+int SharedLibraryIkFast::getNumFreeParameters() const
+{
+  if (!mGetNumFreeParameters)
+  {
+    dterr << "[SharedLibraryIkFast::getNumFreeParameters] This SharedLibrary is "
+          << "invalid. Returning 0.\n";
+    return 0;
+  }
```
mkoval: Why are these checks necessary here, when you also check the return value of `loadFunction` in the constructor?

dart/dynamics/SharedLibraryIkFast.hpp
```diff
@@ -0,0 +1,118 @@
...
+class SharedLibraryIkFast : public IkFast
+{
+public:
+  SharedLibraryIkFast(
+      InverseKinematics* ik,
+      const std::string& fileName,
+      const std::vector<std::size_t>& dofMap,
+      const std::vector<std::size_t>& freeDofMap,
+      const std::string& methodName = "IKFast",
+      const Analytical::Properties& properties = Analytical::Properties());
```
mkoval: Missing documentation,particularly about the arguments.

dart/gui/osg/DragAndDrop.cpp
```diff
@@ -814,6 +814,8 @@ void BodyNodeDnD::saveState()
   else
   {
     mIK = bn->createIK();
+    // TODO(JS): I guess this should be changed to:
+    // mIK = bn->getIK(true);
```
mkoval: Is this necessary?

data/urdf/wam/wam.urdf
```diff
@@ -0,0 +1,231 @@
+<?xml version="1.0" ?>
+<!-- =================================================================================== -->
+<!-- |    This document was autogenerated by xacro from /home/js/dev/catkin_workspaces/dev_libherb/src/herb_description/robots/wam_standalone.urdf.xacro | -->
```
mkoval: Nit: Remove the path from this comment.

data/urdf/wam/wam.urdf
```diff
@@ -0,0 +1,231 @@
...
+      </geometry>
+    </collision>
+  </link>
+  <joint name="/j1" type="revolute">
+    <origin rpy="-1.1127E-16 0 0" xyz="0.22 0.14 0.346"/>
+    <parent link="/wam_base"/>
+    <child link="/wam1"/>
+    <axis xyz="0 0 1"/>
+    <!-- HERB's J1 joints are modified to have a +180 degree offset. -->
+    <limit effort="1.8" lower="0.54159265359" upper="5.74159265359" velocity="0.75"/>
```
mkoval: If you are going to distribute this file with DART, then I suggest reverting to the default joint limits of the Barrett WAM arm. HERB's WAMs were specially modified by Barrett to have different joint limits, to improve the workspace when two WAMs are mounted side-by-side.

unittests/unit/SharedLibraryWamIkFast.hpp
```diff
@@ -0,0 +1,81 @@
...
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef DART_TEST_SHAREDLIBRARYWAMIKFAST_HPP_
+#define DART_TEST_SHAREDLIBRARYWAMIKFAST_HPP_
+
+#include "dart/common/SharedLibrary.hpp"
+#include "dart/dynamics/IkFast.hpp"
+
+class SharedLibraryWamIkFast : public dart::dynamics::IkFast
```
mkoval: Is this the code that an end-user would implement to use IKFast on their robot? Why is this necessary versus just passing the path to the `.so` file into one of the `SharedLibrary` functions?

dart/dynamics/IkFast.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+    if (solutionValues[i] > skel->getDof(dofIndex)->getPositionUpperLimit())
+    {
+      limitViolated = true;
+      break;
+    }
+
+    index++;
+  }
+
```
mkoval: This class - or the `Analytic` base class - should support a `bool` flag that toggles error checking. It's inexpensive to verify that running the IK solver's solution through DART's forward kinematics returns the expected `BodyNode` pose.

This is a very important check - to the extent that we may want to enable it by default - because it catches to broad classes of errors: (1) using an analytic IK solution for a slightly different model and (2) bugs in the analytic IK solution. OpenRAVE avoids the former by hashing the model - which is not really possible here. I have also been burned by the latter due to bugs in IKFast.

mxgrey:
> Once I have the IK solver loaded, how does using Analytical differ from using other types of IK? How do I get multiple solutions?

Suppose you have an ``InverseKinematicsPtr`` object called ``ik`` and it's been assigned an Analytical gradient method. You can access the multiple solutions of the Analytical method using:

```
std::vector<InverseKinematics::Analytical::Solution> solutions =
          ik->getAnalytical()->getSolutions();
```

> Additionally, I would like to have optional support for checking the solution returned by IKFast against forward kinematics.

This can already be done in two ways. If you implicitly use the Analytical solver by calling ``ik->solve()``, then the boolean return value will be false if a valid solution was not actually found. Alternatively, if you get the entire set of options returned by the Analytical solver using ``ik->getAnalytical()->getSolutions()``, then each object ``InverseKinematics::Analytical::Solution`` has a bit field``mValidity`` which is 00 when valid, 01 when the target could not be reached, 10 when the solution is violating joint limits, and 11 when the target is not reached *and* the joint limits are violated.

So I believe the functionality that you're looking for is already immediately available. If you have ideas for ways to improve it or make it more accessible, I'd be very open to making changes.

> Finally, I am not sure where this class enumerates the free parameters to generate a set of IK solutions. Can you point me to that code so I can take another look at it?

If you're referring to the ``Analytical`` class, you can use ``InverseKinematics::setDofs(~)`` to affect which DoFs are available for the solver to use, and then the behavior of any DoFs which are not used directly in the analytical solution can be changed using ``Analytical::setExtraDofUtilization(~)``. Currently there is no method built into the base ``Analytical`` class that would sample values for those extra DoFs to produce a set of solutions, but that can certainly be done in the class that derives ``Analytical``, or we could add that as one of the options for ``setExtraDofUtilization(~)``.

That being said, I'm personally not satisfied at all with the current design of the "extra DoF utilization". I mean, it works for my own personal use cases (which is why I made it in the first place), but it's kind of clumsy and not at all extensible, so I'd love to hear ideas for alternative designs.

dart/dynamics/InverseKinematics.cpp
mxgrey: The Analytical base class has an option for how to make use of any DOFs which aren't directly used by the Analytical solver. They'll be used in a gradient descent based on their Jacobian transpose, and this can either be done before the Analytical solution is computed (``PRE_ANALYTICAL``), after the Analytical solution is computed (``POST_ANALYTICAL``) or both (``PRE_AND_POST_ANALYTICAL``).

However, as I mention in my top-level comment, I would very much like to redesign this approach to utilizing the extra DOFs. I don't have any particularly compelling ideas right now (nor do I expect to have any in the immediate future, since I'm tied up with thesis work), but I'd be happy to discuss any ideas that anyone has.

dart/dynamics/IkFast.cpp
mxgrey: I believe I've addressed this in my top-level comment, but let me know if I'm misunderstanding your concern.

mkoval:
> This can already be done in two ways. If you implicitly use the Analytical solver by calling ``ik->solve()``, then the boolean return value will be false if a valid solution was not actually found. Alternatively, if you get the entire set of options returned by the Analytical solver using ``ik->getAnalytical()->getSolutions()``, then each object ``InverseKinematics::Analytical::Solution`` has a bit field ``mValidity`` which is 00 when valid, 01 when the target could not be reached, 10 when the solution is violating joint limits, and 11 when the target is not reached *and* the joint limits are violated.
> 
> So I believe the functionality that you're looking for is already immediately available. If you have ideas for ways to improve it or make it more accessible, I'd be very open to making changes.

I don't think this addresses my concern. Suppose you call `q = InverseKinematics(x)`. I want to check that `IsClose(ForwardKinematics(q), x)`. If that condition is not satisfied, then there is a bug in the analytical solution and we should: (1) ignore the solution and (2) print a warning.

This most often occurs because there is a mismatch between the kinematics used to create the analytical solution and the `Skeleton` in DART. It can also occasionally happen due to bugs in IKFast. This is an insidious bug because it is often is not obvious upon casual inspection.

I am fine with putting this in the `IKFast` class. However, I think it could be useful as a first-class citizen in the `Analytical` base class because this class of bugs can occur for any type of analytical solution.

dart/common/detail/SharedLibraryManager.cpp
jslee02: Your suggestion is generally the right way to go, but here I try to avoid creating duplicate shared library when it already exists. If I use `emplace()` instead of `find()`, then I have to create the shared library regardless of the existence of the corresponding shared library to the given path.

dart/common/detail/SharedLibraryManager.hpp
jslee02: Now, this function takes a whole path and doesn't add extensions.

dart/common/detail/SharedLibraryManager.hpp
jslee02: Added comment. Will be addressed later.

unittests/unit/SharedLibraryWamIkFast.hpp
jslee02: An end-user would juse use this class. This class enables the user to use DART's inverse kinematics framework with IkFast `.so` file without any further modification. All the user needs to do is passing the path to the `.so` file and the list of free/non-free joint indices like:
```cpp
  std::string libName = "libGeneratedWamIkFast.so";
  std::vector<std::size_t> ikFastDofs{0, 1, 3, 4, 5, 6};
  std::vector<std::size_t> ikFastFreeDofs{2};
  ik->setGradientMethod<dynamics::SharedLibraryIkFast>(
        libName, ikFastDofs, ikFastFreeDofs);
```

`SharedLibrary` only loads arbitrary `.so` file and extracts symbols from the file, but doens't have any information how to interact with IkFast nor the inverse kinematics framework.

dart/gui/osg/DragAndDrop.cpp
jslee02: Will open as a separate PR

dart/dynamics/IkFast.hpp
jslee02: According to `ikfast.h`: 
> if defined, will include library-specific functions. by default this is off

It's not built without `IKFAST_HAS_LIBRARY` because all the necessary definitions are defined under this guard in `ikfast.h`.

dart/dynamics/SharedLibraryIkFast.cpp
jslee02: This is because the constructor just tries to load the functions but does nothing even when it fails to load them. So it has to check whether it's actually loaded before being used.

jslee02:
I believe I've addressed all the @mkoval's comments except for the platform specific macros. Let me handle that in a separate PR.

unittests/unit/SharedLibraryWamIkFast.hpp
junhyeokahn: @jslee02 Could you give an example how to generate ```.so``` file with wam robot?

jslee02: Once you have a `.cpp` file generated for WAM, you just compile the file with any compiler you like. Hope [this](http://openrave.org/docs/0.8.2/openravepy/ikfast/#compiling-with-gcc) helps. Note https://github.com/dartsim/dart/blob/master/examples/osgExamples/osgWamIkFast/ikfast/ikfast71.Transform6D.4_6_9_10_11_12_f8.cpp is just an example of IK for WAM. You might want different [free joints](http://openrave.org/docs/0.8.2/openravepy/ikfast/#details) setting for the same robot.

PR #889, opened 2017-05-21
Use CONFIG mode to find DART in examples and tutorials

### LINKED ISSUE #888: Unable to build DART tutorial out of source tree on MacOS
Hello everyone,

I was trying to build a tutorial out of DART source tree following instructions in dart/tutorials/README.md. But I got this error message from Cmake log: Could NOT find Dart (missing:  DART_ROOT).

I am using OSX, and I'm not sure whether this is related to collision with another "Dart testing software": https://github.com/Kitware/CMake/blob/master/Modules/FindDart.cmake 

How should I resolve this?

Thanks,
Yifeng

jslee02:
`find_package(DART)` uses [`FindDart.cmake`](https://github.com/Kitware/CMake/blob/master/Modules/FindDart.cmake) of the [DART programming language](https://www.dartlang.org/) by default. This PR changes the DART examples and tutorials to use `CONFIG` mode instead to find our DART using [`DARTConfig.cmake`](https://github.com/dartsim/dart/blob/master/cmake/DARTConfig.cmake.in).

This PR resolves #888.

### DIFF
--- a/examples/addDeleteSkels/CMakeLists.txt
+++ b/examples/addDeleteSkels/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.12)
 
 project(addDeleteSkels)
 
-find_package(DART 6.1.1 REQUIRED COMPONENTS utils-urdf gui)
+find_package(DART 6.2.0 REQUIRED COMPONENTS utils-urdf gui CONFIG)
 
 add_compile_options(-std=c++11)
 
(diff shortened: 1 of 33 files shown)


psigen:
Perhaps we should consider if we _really_ want to use the exact name
"DART" (as opposed to "DARTSIM" or other minor variation), because
`FindDart.cmake` is now distributed _with_ CMake [1], meaning all users
will have to manually specify `CONFIG` mode to not have this issue in the
future.

[1] https://github.com/Kitware/CMake/blob/master/Modules/FindDart.cmake

jslee02:
This might be fine on Linux because Linux is a case-sensitive system (not that the DART language uses "Dart" while we use "DART"). But on macOS this could be an issue as you pointed out.

Issue #897, opened 2017-08-03
Compilation failure without osg on release-6.2 branch


scpeters:
I was recently testing a build of gazebo against a source build of the `release-6.2` branch of dart in our buildfarm, and dart failed to compile with complaints about osg. It works fine with `release-6.1`.

* [release-6.2 failure](https://build.osrfoundation.org/job/gazebo-any_dart4-devel-xenial-amd64-gpu-nvidia_docker/20/console)
* [release-6.1 success](https://build.osrfoundation.org/job/gazebo-any_dart4-devel-xenial-amd64-gpu-nvidia_docker/21/)

During configuration, it fails to find OpenSceneGraph:

~~~
CMake Warning at dart/gui/osg/CMakeLists.txt:30 (message):
  Could NOT find OpenSceneGraph nor OpenThreads -- we will skip dart-gui-osg

  If you believe you do have both OSG and OpenThreads installed, try setting
  OSG_DIR


~~~

but it still tries to compile osg support anyway:

~~~
Scanning dependencies of target dart-gui-osg
[ 86%] Building CXX object dart/gui/osg/CMakeFiles/dart-gui-osg.dir/ImGuiViewer.cpp.o
In file included from /var/lib/jenkins/workspace/gazebo-any_dart4-devel-xenial-amd64-gpu-nvidia_docker/dart/dart/gui/osg/ImGuiViewer.hpp:37:0,
                 from /var/lib/jenkins/workspace/gazebo-any_dart4-devel-xenial-amd64-gpu-nvidia_docker/dart/dart/gui/osg/ImGuiViewer.cpp:32:
/var/lib/jenkins/workspace/gazebo-any_dart4-devel-xenial-amd64-gpu-nvidia_docker/dart/dart/gui/osg/Viewer.hpp:39:28: fatal error: osgViewer/Viewer: No such file or directory
compilation terminated.
dart/gui/osg/CMakeFiles/dart-gui-osg.dir/build.make:62: recipe for target 'dart/gui/osg/CMakeFiles/dart-gui-osg.dir/ImGuiViewer.cpp.o' failed
~~~

mxgrey:
After some refactoring was done to DART's CMake setup, we're no longer properly guarding `dart-gui-osg` against the absence of OpenSceneGraph. I have a fix coming up.

jslee02:
FYI, this might be an oversight made by #880.

scpeters:
I'll race you @mxgrey

[...]

PR #898, opened 2017-08-04
Protect dart-gui-osg and check more rigorously for the presence of OSG

mxgrey:
This PR has two related (but independent) purposes:

1. Prevent `dart-gui-osg` from being built when OSG is not detected. ~~This is done by wrapping the second half of `dart/gui/osg/CMakeLists.txt` in an if-block.~~ This is done by returning immediately if OpenSceneGraph cmake variables cannot be found. This was not needed in the past because the call to `add_subdirectory(osg)` used to be wrapped in an if-statement.

2. More rigorously check whether OSG exists. In the process of fixing this, I discovered that the version of `FindOpenSceneGraph.cmake` that is provided with `cmake-3.5` will set `OPENSCENEGRAPH_FOUND` to true if it finds libopen**threads**, even if libopen**scenegraph** isn't actually installed. I believe this is a bug on the part of either CMake or OpenSceneGraph, because this behavior contradicts the variable's description provided by `FindOpenSceneGraph.cmake`, but I found a workaround by checking both `OPENSCENEGRAPH_FOUND` and `OSG_FOUND`. The latter seems to correctly return false when OpenSceneGraph is not available.

Note that the vast majority of the diff is due to white space from indenting the content inside the if-statement.

### DIFF
--- a/dart/gui/osg/CMakeLists.txt
+++ b/dart/gui/osg/CMakeLists.txt
@@ -8,6 +8,11 @@ if(DART_BUILD_GUI_OSG)
 
   find_package(OpenSceneGraph 3.0 QUIET
     COMPONENTS osg osgViewer osgManipulator osgGA osgDB)
+
+  # It seems that OPENSCENEGRAPH_FOUND will inadvertently get set to true when
+  # OpenThreads is found, even if OpenSceneGraph is not installed. This is quite
+  # possibly a bug in OSG's cmake configuration file. For now, it seems that
+  # requiring OSG_FOUND to be true as well fixes this.
   if(OPENSCENEGRAPH_FOUND AND OSG_FOUND)
     if(DART_VERBOSE)
       message(STATUS "Looking for OpenSceneGraph - ${OPENSCENEGRAPH_VERSION} found")
--- a/dart/gui/osg/CMakeLists.txt
+++ b/dart/gui/osg/CMakeLists.txt
@@ -12,7 +12,6 @@ if(DART_BUILD_GUI_OSG)
     if(DART_VERBOSE)
       message(STATUS "Looking for OpenSceneGraph - ${OPENSCENEGRAPH_VERSION} found")
     endif()
-    set(HAVE_OPENSCENEGRAPH TRUE)
   else()
     # dart-gui-osg requires both OSG and OpenThreads. This section attempts to
     # identify which of those are missing from the building machine and offer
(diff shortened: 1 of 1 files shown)


mxgrey:
PR #897 by @scpeters made me realize that CMake offers a return() function. That allows us to eliminate the `HAVE_OPENSCENEGRAPH` variable altogether and make the diff much nicer. I'm going to shamelessly steal his approach.

dart/gui/osg/CMakeLists.txt
```diff
@@ -8,11 +8,10 @@ if(DART_BUILD_GUI_OSG)
 
   find_package(OpenSceneGraph 3.0 QUIET
     COMPONENTS osg osgViewer osgManipulator osgGA osgDB)
-  if(OPENSCENEGRAPH_FOUND)
+  if(OPENSCENEGRAPH_FOUND AND OSG_FOUND)
```
jslee02: Are all these variables are set to true when OSG is found? It seems `OPENSCENEGRAPH_FOUND` is defined when [`FindOpenSceneGraph.cmake`](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSceneGraph.cmake) is used while `OSG_FOUND` is defined when [`FindOSG.cmake`](https://github.com/openscenegraph/OpenSceneGraph/blob/1bdd2aceaba5bbc91fe174d943f63b0de754ec76/CMakeModules/FindOSG.cmake) is used. Shouldn't this to be `OPENSCENEGRAPH_FOUND OR OSG_FOUND`? Also, if `FindOSG.cmake` is used, it looks like we need to use `OSG_LIBRARY` instead of `OPENSCENEGRAPH_LIBRARIES`. Am I missing something here?

jslee02: Are all these variables are set to true when OSG is found? It seems `OPENSCENEGRAPH_FOUND` is defined when [`FindOpenSceneGraph.cmake`](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSceneGraph.cmake) is used while `OSG_FOUND` is defined when [`FindOSG.cmake`](https://github.com/openscenegraph/OpenSceneGraph/blob/1bdd2aceaba5bbc91fe174d943f63b0de754ec76/CMakeModules/FindOSG.cmake) is used. Shouldn't this to be `OPENSCENEGRAPH_FOUND OR OSG_FOUND`? Also, if `FindOSG.cmake` is used, it looks like we need to use `OSG_LIBRARY` instead of `OPENSCENEGRAPH_LIBRARIES`. Am I missing something here?

scpeters:
I only noticed the `return()` function because it was already used at the top of that file:

* https://github.com/dartsim/dart/blob/d4d6fbbc7dfc1e105da78f89502c4099e210ba15/dart/gui/osg/CMakeLists.txt#L3

dart/gui/osg/CMakeLists.txt
jslee02: Oh, you already mentioned about this in the post. Sorry, wasn't paying sufficient attention. :sweat: I think it would be much great if we add a simple comment why we check both of them.

Issue #903, opened 2017-08-04
Provide setRelativeTransform() function for WeldJointConstraint


mxgrey:
Per the discussion [here](https://bitbucket.org/osrf/gazebo/pull-requests/2756/dart-kinematic-loops/activity#comment-42226916) it would likely be a useful feature.

pchorak:
Gazebo [issue #2339](https://bitbucket.org/osrf/gazebo/issues/2339) could be solved with this feature.

jslee02:
Resolved by #910

PR #909, opened 2017-08-24
Provide setters and getters for WeldJointConstraint::mRelativeTransform

mxgrey:
I couldn't find any reason to restrict access to setting `mRelativeTransform`. I also couldn't identify any other data members who would be impacted by changing `mRelativeTransform`, so the setter is just a straightforward copy.

This addresses #903.


mxgrey:
Since I branched this off of master, I'll just create a new PR that branches off of 6.3.

PR #912, opened 2017-09-03
Reformat LICENSE to be detected by Github

jslee02:
This PR slightly changes the format of `LICENSE` file to enable Github to detect it correctly.


LICENSE
```diff
@@ -1,28 +1,25 @@
-﻿Copyright (c) 2011-2016, Humanoid Lab, Georgia Tech Research Corporation
+Copyright (c) 2011-2016, Humanoid Lab, Georgia Tech Research Corporation
```
mxgrey: I'm curious: What actually changed here? Some invisible character?

jslee02: Yeah, although it's not visible here there was an unknown (to me) character at the beginning of the line. 🤔

Issue #753, opened 2016-07-22
Build error with Visual Studio 2015 Update 3


iamyaoting:
Hello，I  install the Visual Studio Update3, but I built it with following error:
`2>D:\Projects\Cpp\3rdsource\dart\dart/common/AspectWithVersion.hpp(80): error C2974: 'dart::common::detail::AspectWithState': invalid template argument for 'CompositeT', type expected (compiling source file D:\Projects\Cpp\3rdsource\dart\dart\collision\CollisionGroup.cpp)
2>  D:\Projects\Cpp\3rdsource\dart\dart/common/detail/AspectWithVersion.hpp(48): note: see declaration of 'dart::common::detail::AspectWithState' (compiling source file D:\Projects\Cpp\3rdsource\dart\dart\collision\CollisionGroup.cpp)
2>  D:\Projects\Cpp\3rdsource\dart\dart/dynamics/EndEffector.hpp(57): note: see reference to class template instantiation 'dart::common::AspectWithStateAndVersionedProperties<dart::dynamics::Support,dart::dynamics::detail::SupportStateData,dart::dynamics::detail::SupportPropertiesData,dart::dynamics::EndEffector,dart::dynamics::detail::SupportUpdate,dart::dynamics::detail::SupportUpdate>' being compiled (compiling source file D:\Projects\Cpp\3rdsource\dart\dart\collision\CollisionGroup.cpp)`
What's the problem? Can you help me?
Thank you

jslee02:
The support for Visual Studio 2015 building DART 6.0 has been suspended because of the incomplete `constexpr` support of VS2015. However, it seems VS2015 Update 3 has resolved the issue, and there are other build issues as you reported. I'll look into it when having a chance.

iamyaoting:
Ok, As you say, in update 3, I did not see the constexpr error. 
But the template errors came out. 
Thank you~

mottosso:
I can confirm that the same error occurs with Visual Studio 2017 RC1.

```
C:\Users\marcus\dart\dart\common\AspectWithVersion.hpp(80): error C2974: 'dart::common::detail::AspectWithState': invalid template argument for 'CompositeT', type expected (compiling source file C:\Users\marcus\dart\dart\simulation\Recording.cpp) 
```

Along with these.

```
[...]

Issue #260, opened 2014-11-12
Consider using package manager on Windows


jslee02:
It would be nice to exploit a package manager for windows instead of providing our own installer for the DART dependencies. Currently, I'm considering using [`vcpkg`](https://github.com/Microsoft/vcpkg/) of Microsoft.

Here is the list of supported libraries:
---
#### Required dependencies
- [x] [eigen3](https://github.com/Microsoft/vcpkg/tree/master/ports/eigen3)
- [x] [libccd](https://github.com/Microsoft/vcpkg/tree/master/ports/ccd)
- [x] [fcl](https://github.com/Microsoft/vcpkg/tree/master/ports/fcl)
- [x] [assimp](https://github.com/Microsoft/vcpkg/tree/master/ports/assimp)
- [x] [boost](https://github.com/Microsoft/vcpkg/tree/master/ports/boost)
#### Optional dependencies
- [x] [nlopt](https://github.com/Microsoft/vcpkg/tree/master/ports/nlopt)
- [ ] [ipopt](https://github.com/Microsoft/vcpkg/issues/233)
- [x] [bullet3](https://github.com/Microsoft/vcpkg/tree/master/ports/bullet3)
- [x] [ode](https://github.com/Microsoft/vcpkg/tree/cf54838139aa03e6b7f7d13c6963bd2cb7ba2fa4/ports/ode)
- [x] [flann](https://github.com/Microsoft/vcpkg/tree/master/ports/flann)
- [x] [tinyxml](https://github.com/Microsoft/vcpkg/tree/master/ports/tinyxml)
- [x] [tinyxml2](https://github.com/Microsoft/vcpkg/tree/master/ports/tinyxml2)
- [x] [console-bridge](https://github.com/Microsoft/vcpkg/tree/master/ports/console-bridge) (Note that hyphen is used instead of underscore)
- [x] [urdfdom-headers](https://github.com/Microsoft/vcpkg/tree/master/ports/urdfdom-headers) (Note that hyphen is used instead of underscore)
- [x] [urdfdom](https://github.com/Microsoft/vcpkg/tree/master/ports/urdfdom)
- [x] [opengl](https://github.com/Microsoft/vcpkg/tree/master/ports/opengl)
- [x] [freeglut](https://github.com/Microsoft/vcpkg/tree/master/ports/freeglut)
- [x] [osg](https://github.com/Microsoft/vcpkg/tree/master/ports/osg)

jslee02:
[...]

PR #913, opened 2017-09-03
Switch to vcpkg for installing DART dependencies on Windows

jslee02:
We've managed and provided own installer for the DART dependencies, which was tedious work.  [As discussed](https://github.com/dartsim/dart/issues/260), this PR hands over the duty to `vcpkg`, which is a package manager for Windows developed by Microsoft.

Note that we still don't build DART on Windows until the [existing build issues](https://github.com/dartsim/dart/issues/753) are resolved. This PS only changes the way to install the DART dependencies.


appveyor.yml
```diff
@@ -1,19 +1,21 @@
...
-
-# build platform, i.e. Win32 (instead of x86), x64, Any CPU. This setting is optional.
-platform: 
-  - Win32
-  #- x64  Disable build for x64 machine until x64 version of dart-prerequisites is ready
+image: Visual Studio 2017
 
 # specify custom environment variables
+#
+# We generate project files for Visual Studio 14 because the boost binaries installed on the test server are for Visual Studio 14.
```
mxgrey: I'm a bit confused. This line says we generate project files for Visual Studio 14, but then above and below it looks like everything is targeting Visual Studio 15. Am I misunderstanding what it's saying about generating project files?

jslee02: This is an outdated description. Let me update this accordingly. Probably we don't need this note because boost will be installed anyway as [`fcl` is dependent on it](https://github.com/Microsoft/vcpkg/blob/master/ports/fcl/CONTROL#L4).

jslee02:
It seems the building time exceeds the Appveyor limit (1 hour). I guess we need to use a prebuilt image using docker. Let me close this PR until then.

PR #924, opened 2017-10-04
Rename utils namespace to io

### LINKED ISSUE #815: Consider renaming namespace `utils` to `io`
`io` (stands for Input/Output) would be a more relevant name with the code that `utils` contains while `utils` is too general.

jslee02:
Depends on #923 
Resolve #815


mxgrey:
There's a lot to look through, partially due to the dependency on #923, so maybe this is already in there, but could we do a [namespace alias](http://en.cppreference.com/w/cpp/language/namespace_alias) from `utils` to `io` to help people migrate?

We could put the namespace alias in headers which correspond to the existing `utils` headers (which would each `#include` their corresponding `io` header), and we could have those headers generate a compiler warning to indicate that use of the `utils` namespace is deprecated.

I believe we can set this up to be auto-generated by cmake. If this all sounds like a good idea to you, I could branch from this PR, make the changes, and submit it as a PR targeted at this PR.

mxgrey:
To illustrate what I'm proposing...

In header `dart/utils/DartResourceRetriever.hpp`:
```
/* LICENSE */
#include <dart/io/DartResourceRetriever.hpp>

#ifndef DART_UTILS_IO_ALIAS
#define DART_UTILS_IO_ALIAS
namespace dart {
namespace utils = io;
} // namespace dart
#warning "Use of the [dart::utils] namespace is deprecated and should be changed to [dart:io]"
#endif // DART_UTILS_IO_ALIAS

mkoval:
I like the idea of the alias and I don't see any harm in adding it. Note, however, that this will still break ABI compatibility.

jslee02:
I second the idea. @mkoval Right, this change for the new major version up so we expect ABI compatibility, but the alias would be a little bandage. :)

mxgrey:
We don't guarantee ABI compatibility between minor revisions much less major revisions :sweat_smile:

jslee02:
Also, @mxgrey I wouldn't mind if you make the changes in this branch directly.

jslee02:
@mxgrey I think we can merge this and add the namespace alias in a separate PR (created #930 to track this).

PR #928, opened 2017-10-14
Change Function::eval[Gradient/Hessian] to const functions

### LINKED ISSUE #927: Consider changing Function::eval[Gradient/Hessian] to const function
I believe they're logically correct to be const functions and can't come up with any compelling reason not to do it.

jslee02:
Resolves #927

***

- [x] Document new methods and classes (nothing new)

- [x] Add unit test(s) for this change (nothing new to test)


dart/optimizer/Function.hpp
```diff
@@ -38,107 +38,108 @@
 
 #include <Eigen/Dense>
 
+#include "dart/common/Deprecated.hpp"
```
mxgrey: It looks like we don't end up using `Deprecated.hpp`, presumably because there isn't really any good way of deprecating the non-const version of the function without a `const_cast`.

PR #934, opened 2017-10-27
Fix condition

lunixoid:
For competition: the Pinguem.ru competition on finding errors in open source projects. A bug, found using PVS-Studio:
dart/dart/collision/fcl/FCLCollisionDetector.cpp    998     warn    V523 The 'then' statement is equivalent to the 'else' statement.

The `geom = createCylinder<fcl::OBBRSS>(radius, radius, height, 16, 16);` calls in any case. So condition needed only for `TODO:`(not yet realised) operations.

### DIFF
--- a/dart/collision/fcl/FCLCollisionDetector.cpp
+++ b/dart/collision/fcl/FCLCollisionDetector.cpp
@@ -986,17 +986,13 @@ FCLCollisionDetector::createFCLCollisionGeometry(
     const auto radius = cylinder->getRadius();
     const auto height = cylinder->getHeight();
 
+    geom = createCylinder<fcl::OBBRSS>(radius, radius, height, 16, 16);
     if (FCLCollisionDetector::PRIMITIVE == type)
     {
-      geom = createCylinder<fcl::OBBRSS>(radius, radius, height, 16, 16);
       // TODO(JS): We still need to use mesh for cylinder because FCL 0.4.0
       // returns single contact point for cylinder yet. Once FCL support
       // multiple contact points then above code will be replaced by:
       // fclCollGeom.reset(new fcl::Cylinder(radius, height));
-    }
-    else
-    {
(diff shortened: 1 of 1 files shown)


mxgrey:
We certainly appreciate the analysis and this contribution, but I believe the code should remain the way it currently is for future maintenance purposes.

In the future, when FCL (a third-party dependency of our library) makes some corrections, we will want this if/else branching to be the way that it currently is. We will want the `then` statement with line 999 to remain as it is, and instead we'll be changing line 991 inside of the `if` statement. If we make the changes that are suggested in this pull request, it may be less clear to future maintainers (...probably still us) how we should update the code once FCL has made the corrections that we need.

Please don't hesitate to share any more of your analysis; we really do appreciate it.

lunixoid:
I agree, so can you share the link for FCL project? 
Also, i thought a third-party dependencies places in directory `external`, sorry.

jslee02:
FCL: https://github.com/flexible-collision-library/fcl

DART contains a third-party dependency in `external` directory only when it is not available through package managers. FCL is not the case.

mxgrey:
It looks like @jslee02 finished a pull request that's relevant for this issue, but I suppose the PR might not have been enough to make the cylinder primitive work as we need it to.

I don't see any open issues on FCL about getting multiple contact points for the cylinder, so maybe it would be a good idea for us to open one?

I'm guessing the main issue here is that the math for computing multiple contact points will be fairly challenging to work out, especially since primitives tend to have a volume of contact, so there's an important question of: How do you select a discrete set of points from that volume?.

lunixoid:
I just found this issue using static analyser. For greater understanding i should to dive into code deeper :)

mxgrey:
From this point onwards, the issue is going to cross over from being a software issue into possibly being a differential geometry issue. Don't let me stop you from investigating it if you're really interested, but it might require some pretty serious expertise in advanced mathematics and geometry.

PR #937, opened 2017-11-10
Suppress -Wdangling-else warnings

jslee02:
GCC 7.2.0 prints warnings of `-Wdangling-else`. It seems this is because if-statements with a single line macro is prone to dangling else. This PR suppresses these warnings by adding explicit braces around the single line of the macro.

***

- [x] Document new methods and classes (N/A)


jslee02:
Thanks for the review @mxgrey!

PR #943, opened 2017-11-15
Add issue template

jslee02:


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,30 @@
+> If you want to report a bug or request assistance please use the following template.
+
+---
+
+### Bug Report / Assistance Request
+
+#### Environment
+
+* DART version: 6.3.0 / master
```
mxgrey: Maybe we should add the word `Select ` in front of each of these lines in order to convey that we want the user to specify which one they are using. We should also probably end each line with ` / other (please specify)`

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,30 @@
...
+#### Steps to Reproduce
+> Tell us how to reproduce this issue.
+1.
+1.
+1.
+1.
+
+#### Code to Reproduce
+> Please remember that with a working code it's easier to reproduce the bug and it's much faster to fix it.
+> It would be great to fork this repo and create a using [this template](https://github.com/dartsim/dart/blob/master/unittests/regression/test_Issue000Template.cpp).
```
mxgrey: `and create a` Pull request? Regression test? Both?

mxgrey: Looks like this comment has been addressed, please disregard.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Bug Report
+Please answer the following questions for yourself before submitting an issue.
+
+- [ ] I checked the [documentation](http://dartsim.github.io/) and found no answer.
+- [ ] I checked to make sure that this issue has not already been filed.
+
+#### Environment
+
+* DART version: 6.3.0 / master
```
mxgrey: I think my "outdated" [comment here](https://github.com/dartsim/dart/pull/943#discussion_r151265170) is still relevant.

.github/ISSUE_TEMPLATE.md
```diff
@@ -10,9 +10,9 @@ Please answer the following questions for yourself before submitting an issue.
 
 #### Environment
 
-* DART version: 6.3.0 / master
-* OS name and version name(or number): Ubuntu 14.04 / Windows 10 / macOS Sierra
-* Compiler name and version number: GCC 4.8.5 / Clang 3.9.1
+* DART version: Select master / 6.3.0 / other (please specify)
```
mxgrey: This might be excessively nitpicky, but I was thinking more like `Select DART version: master / 6.3.0 / other (please specify)`

jslee02: It sounds more nature! Updated.

Issue #734, opened 2016-06-07
Bounding box bug in MeshShape.cpp


edjohnsuk:
I think there is a bug in the latest version of `MeshShape.cpp` on the Master branch:

When the `MeshShape` constructor is called, it in turn calls `setMesh(...)`, followed by `setScale(...)`. `setScale(...)` itself calls `updateVolume()`, followed by `_updateBoundingBoxDim()`. However, `updateVolume()` requires the bounding box to have already been constructed, otherwise it will compute a volume of zero. Therefore, in `setScale(...)`, `updateVolume()`, should be called _after_ `_updateBoundingBoxDim()`, not before it.

A smaller, related issue is that both `setMesh(...)` and `setScale(...)` call `_updateBoundingBoxDim()` and `updateVolume()`. But `setMesh(...)` is only found once in the code, in the `MeshShape` constructor, so will always be followed by `setScale(...)`. Therefore, it is not necessary to call either `_updateBoundingBoxDim()` or `updateVolume()` in `setMesh(...)`, because they will both be called immediately after in `setScale(..)`. And in any case, calling `_updateBoundingBoxDim()` and `updateVolume()` in `setMesh(...)` before `setScale(...)` has been called, will not do anything useful anyway, because the scale has not yet been set. This is just an efficiency issue and does not affect program execution, whereas the first issue is more serious.

In summary:
1.  In `setScale(...)`, `updateVolume()` should be called _after_ `_updateBoundingBoxDim()`.
2. In `setMesh(...)`, calls to `_updateBoundingBoxDim()` and `updateVolume()` should be removed.

jslee02:
Resolved by #958, #959

PR #958, opened 2018-01-17
Fix for #734

costashatz:
This is a fix to #734. In short:

- There's a small bug in `setScale` method of `MeshShape`; i.e., `_updateBoundingBoxDim` needs to be called before `updateVolume`
- The calls to `updateVolume` and `_updateBoundingBoxDim` in the `setMesh` method are not needed.

### DIFF
--- a/dart/dynamics/MeshShape.cpp
+++ b/dart/dynamics/MeshShape.cpp
@@ -220,9 +220,6 @@ void MeshShape::setMesh(
   }
 
   mResourceRetriever = _resourceRetriever;
-
-  _updateBoundingBoxDim();
-  updateVolume();
 }
 
 void MeshShape::setScale(const Eigen::Vector3d& _scale) {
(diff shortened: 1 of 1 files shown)


dart/dynamics/MeshShape.cpp
```diff
@@ -220,18 +220,15 @@ void MeshShape::setMesh(
   }
 
   mResourceRetriever = _resourceRetriever;
-
-  _updateBoundingBoxDim();
-  updateVolume();
```
jslee02: I believe you consider these are not necessary because `setMesh()` and `setScale()` are both called in the constructor. However, I think we still need these function calls for the case `setMesh()` is called separately and getting volume and bounding box without manually updating them.

costashatz: > I think the redundant computation should be addressed by the lazy-evaluation method.

I didn't see that you started the lazy-evaluation and I started doing it.. But you're faster than me..! :stuck_out_tongue:

Issue #954, opened 2018-01-05
Clang 5.0.0 warning per -Winjected-class-name


jslee02:
### Bug Report
> Please answer the following questions for yourself before submitting an issue.
- [x] I checked the [documentation](http://dartsim.github.io/) and found no answer.
- [x] I checked to make sure that this issue has not already been filed.

#### Environment
* Select DART version: master (but it might be the same to any version that contains `Aspect` class)
* Select OS name and version name(or number): Ubuntu 14.04
* Select compiler name and version number: Clang 5.0.0

#### Expected Behavior
Compile w/o any warnings

#### Current Behavior
[Travis CI complains](https://travis-ci.org/dartsim/dart/jobs/325463914#L1726) as
```
/home/travis/build/dartsim/dart/dart/common/detail/CompositeData.hpp:56:34: error: ISO C++ specifies that qualified reference to 'Aspect' is a constructor name rather than a type in this context, despite preceding 'typename' keyword [-Werror,-Winjected-class-name]
      AspectT, typename AspectT::Aspect>::type;
```

mxgrey:
I've made some progress on digesting the error message. I'm a tiny bit concerned this might be a compiler bug, because I don't quite see how what it's saying can be true.

Based on the printout, it seems that `AspectT` in this case is `dart::common::EmbeddedPropertiesAspect<dart::dynamics::EntityNode<dart::dynamics::TemplatedJacobianNode<dart::dynamics::FixedJacobianNode> >, dart::dynamics::detail::EntityNodeProperties>`, which gets created by `EmbedProperties<...>` when `EmbedPropertiesOnTopOf<...>` is inherited by `EndEffector`.

What's strange is that this `AspectT` inherits `common::CompositeTrackingAspect<dynamics::EntityNode<...>>` which in turn inherits `common::Aspect`. Since `AspectT` inherits `Aspect`, we should see that `std::is_base_of<Aspect, AspectT>::value` evaluates to `true`, and therefore `typename AspectT::Aspect` should never be compiled in this context. As far as I understand the ISO rules, `typename AspectT::Aspect` should be completely ignored in this context.

Instead, it seems like `std::is_base_of<Aspect, AspectT>` is incorrectly evaluating to `false` here, or else `std::conditional<true, A, B>::type` is evaluating the expression `B` when I would think it's not supposed to.

[...]

PR #963, opened 2018-01-31
Use template specialization to sidestep complaints from clang

mxgrey:
This fixes issue #954 in which clang-5.0 emits a warning about a typename being interpreted as a constructor in certain contexts.

***

***

Do we care about the changelog for a warning fix like this? It doesn't have any effect on API, ABI, or behavior.

### DIFF
--- a/dart/common/detail/CompositeData.hpp
+++ b/dart/common/detail/CompositeData.hpp
@@ -48,12 +48,37 @@ namespace common {
 namespace detail {
 
 //==============================================================================
+// This default template definition will be called when AspectOrComposite is
+// an Aspect.
+template <class AspectOrComposite, bool isAspect>
+struct GetAspectImpl
+{
+  using Type = AspectOrComposite;
+};
+
+//==============================================================================
+// This template specialization will be called when AspectOrComposite is not
+// an Aspect (and is presumably a composite that defines a nested Aspect type).
+template <class AspectOrComposite>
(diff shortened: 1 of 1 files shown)


mxgrey:
It may be worth noting that on my machine, I get a slew of the following linking error when compiling with clang-5.0:

```
../../lib/libdart.so.6.3.0:-1: error: undefined reference to `boost::re_detail::cpp_regex_traits_implementation<char>::transform_primary[abi:cxx11](char const*, char const*) const'
```

It may have something to do with [this issue](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=823978), but I'm not sure. It's certainly unrelated to the warning that this PR fixes, though.

jslee02:
> Do we care about the changelog for a warning fix like this? It doesn't have any effect on API, ABI, or behavior.

There is no strict rule for updating the changelog, but I would prefer having this PR (something like "Fixed build issue with Clang 5.0") in the Misc section to notify that we don't support DART 6.3 or less for Clang 5.0.

Moreover, it would make more sense add that item to DART 6.4 section (not to DART 7.0) because this fix is introduced in DART 6.4 first, and later versions will inherit the fix.



> It may have something to do with this issue, but I'm not sure. It's certainly unrelated to the warning that this PR fixes, though.

I also think that's not related to this PR. Let me test if I can reproduce the same error on my machine.

PR #970, opened 2018-02-04
[WIP] Add ResourceRetriever::getFilePath()

jslee02:
**Motivation**
In some cases, it is required to extract the (absolute) file path of a URI (if available). For example, `MeshShape` needs to know the path to a mesh file to infer the path to a texture image, which is given as a relative path. Currently, we don't have a public method to get a file path from a URI unless it's a file URI. `PackageResourceRetriever` and `DartResourceRetriever` have similar functionality, but it's not public.

**Proposal**
I propose adding `ResourceRetriever::getFilePath()` that returns the absolute file path specified by a URI or an empty string if unavailable.

***

### DIFF
--- a/dart/common/LocalResourceRetriever.cpp
+++ b/dart/common/LocalResourceRetriever.cpp
@@ -43,14 +43,7 @@ namespace common {
 //==============================================================================
 bool LocalResourceRetriever::exists(const Uri& _uri)
 {
-  // Open and close the file to check if it exists. It would be more efficient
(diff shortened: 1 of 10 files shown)


jslee02:
Sure, I wanted to hear from you (and possibly @mkoval) if this sounds before adding tests. :smile: 

Btw, I meant to add this to DART 6.4 (and later by cascading the changes) but unintentionally made this branching from `master`. Let me create a new PR for DART 6.4. :sweat:

PR #972, opened 2018-02-05
Add ResourceRetriever::getFilePath()

jslee02:
Revision of #970 retargeting to `release-6.4`.

***

**Motivation**
In some cases, it is required to extract the (absolute) file path of a URI (if available). For example, `MeshShape` needs to know the path to a mesh file to infer the path to a texture image, which is given as a relative path. Currently, we don't have a public method to get a file path from a URI unless it's a file URI. `PackageResourceRetriever` and `DartResourceRetriever` have similar functionality, but it's not public.

**Proposal**
I propose adding `ResourceRetriever::getFilePath()` that returns the absolute file path specified by a URI or an empty string if unavailable.

***


unittests/unit/test_PackageResourceRetriever.cpp
```diff
@@ -136,6 +141,116 @@ TEST(PackageResourceRetriever, exists_FallsBackOnSecondUri)
...
+  EXPECT_EQ(retriever.getFilePath(Uri::createFromString("foo://test/foo")), "");
+  EXPECT_TRUE(mockRetriever->mExists.empty());
+  EXPECT_TRUE(mockRetriever->mGetFilePath.empty());
+  EXPECT_TRUE(mockRetriever->mRetrieve.empty());
+}
+
+TEST(PackageResourceRetriever, getFilePath_StripsTrailingSlash)
+{
+#ifdef _WIN32
+  const char* expected = "file:///" "dart://sample/test/foo";
```
mxgrey: I'm very unclear about the purpose behind this `#ifdef`, especially since I believe this will set the value of `expected` to a `const char*` which dereferences to `file:///dart://sample/test/foo`, which doesn't seem like a valid URI (although I could certainly be wrong).

Some commentary about the motivation behind the `#ifdef` (Why is Windows different? Shouldn't that kind of difference be handled in the `Uri` class?) would be helpful. As well as the meaning behind strange URI which seems to contain both the `file` scheme and the `dart` scheme in sequence.

jslee02: Those invalid URI tests were unintentionally introduced in https://github.com/dartsim/dart/commit/162b7a842ee737c03339c6d6554954748fe69e46#diff-2c76e43b1bdbf70403566db01e2e1198. They are fixed by c79de2d.

> Some commentary about the motivation behind the #ifdef (Why is Windows different? Shouldn't that kind of difference be handled in the Uri class?) would be helpful.

Added. I hope the comment answers your concern.

mkoval:
It should be possible to get exactly the same behavior using the existing `Uri` API:
```c++
Uri uriToMesh = /* snip */;
std::string relativePathToTexture = /* snip */;
auto uriToTexture = Uri::fromRelativeUri(uriToMesh, relativePathToTexture);
```

The big advantage of `fromRelativeUri()` is that it uniformly handles absolute and relative URIs as defined in [section 5](https://tools.ietf.org/html/rfc3986#section-5) of [RFC 3986](https://www.google.com/search?q=RFC+Uri&oq=RFC+Uri&aqs=chrome..69i57j0l5.1559j0j1&sourceid=chrome&ie=UTF-8). This means that you can mix schema types (e.g. put a `http://` URI in a local file) and all sorts of other goodies.

Implementing `getFilePath()` only really makes sense for `file://` URIs. That largely defeats the purpose of defining the `ResourceRetriever` interface, which is intended to agnostic to the type of URI being used.

jslee02:
Another solution I came up with was storing the file path to `LocalResource`. It takes a file path anyway so it should be the easiest solution. The reason I didn't take the solution is that:
1. I wanted to get the file path without actually opening the file.
2. It has to downcast to `LocalResource` to get the file path from.

mkoval:
> This may work for a file URI, but how could we get the file path from a non-file URI (e.g., `package://`... or `dart://`...) without using `PackageResourceRetriever` or `DartResourceRetriever` for instance?

You don't need to get a file path at all - you just need a path to refer to the resource, such that you can retrieve it later. In fact, a file path may not exist at all - the underlying resource retriever could, e.g. use `curl` to pull the data from the internet on demand.

Suppose you are loading a URDF from the URI `dart://my_directory/fruit.urdf` using `myResourceRetriever` as the `ResourceRetriever`. You start by `retrieving()` the URI, loading the file's contents, and parsing the XML:
```c++
auto urdfUri = Uri::fromString{"dart://my_directory/fruit.urdf"};
auto urdfResource = myResourceRetriever->retrieve(urdfUri);
auto urdfString = read_all(urdfResource.read);
auto urdfModel = parse(urdfString);
```

Suppose you encounter the relative path `meshes/banana.stl` in `urdfModel` and you now want load the mesh into memory. You construct the relative URL from the URI to the URDF file and the relative URI fragment loaded from the file:
```c++
auto relativeMeshUri = Uri::fromString{"meshes/banana.stl"}; // loaded from urdfModel
auto meshUri = Uri::fromRelativeUri(urdfUri, relativeMeshUri);
// meshUri now contains "dart://my_directory/meshes/banana.stl"
```
Now, you can simply load `meshUri` using `myResourceRetriever`:
```c++
auto meshResource = myResourceRetriever->retrieve(meshUri);
```
Since you are using the same `ResourceRetriever` as you used to retrieve the original URDF, it presumably will be able to handle the `dart://` URI consistently.

---

Maybe I am missing something. Why do we need to convert the `Uri` to a local path before we retrieve it? 🤔

jslee02:
@mkoval All you pointed out make sense to me. I should have said the motivation behind this change. :sweat_smile:

This was for loading texture images using OSG that requires the file path to the image files (see [this](https://github.com/dartsim/dart/pull/973/files#diff-d450e63e13c33ac526d1b7e0206eee9eR805)). But I couldn't come up with a way to incorporate our resource/retriever mechanism into this (like [`AssimpInputResourceAdaptor`](https://github.com/dartsim/dart/blob/master/dart/dynamics/AssimpInputResourceAdaptor.hpp)).

jslee02:
@mkoval Merging this since this PR is blocking #973. I've created an issue (#984) to continue the above discussion.

Issue #971, opened 2018-02-05
Displaying shadow


dhkim0821:
Is there any way to add shadow in DART?

mxgrey:
DART does not inherently provide any features for rendering shadows, and I don't believe there are any plans to do so, ever, since DART is first and foremost a dynamics toolkit. Graphical shading falls far outside the scope of DART.

That being said, OpenSceneGraph does provide features for rendering shadows, and the `dart-gui-osg` wrapper gives a significant amount of access into the underlying OpenSceneGraph objects that are used for rendering.

To get shadows in a DART gui application, I recommend looking into how to achieve that through native OpenSceneGraph features. I have not tried doing that yet, so if there is anything about the current design of `dart-gui-osg` that makes it prohibitively difficult, please let us know, and we can work out a redesign of the wrapper library to accommodate shading.

Also, if you end up creating a shadow wrapper for OpenSceneGraph which would be relevant or helpful for DART users, we would love to get a pull request for it.

costashatz:
@dhkim0821 @mxgrey I am in the process of making a PR to enable shadows in the wrapper for OpenSceneGraph of DART. I will have it ready shortly. Here's a first glimpse of it:

![hexa_shadow](https://user-images.githubusercontent.com/1283922/35866321-2d48d978-0b57-11e8-8275-65f0ee92f11f.png)

mxgrey:
Thanks, @costashatz , that looks amazing!

dhkim0821:
I also tried various methods.
The followings are my current progress
<img width="1280" alt="screenshot 2018-02-06 02 47 38" src="https://user-images.githubusercontent.com/7194057/35876861-4103d0da-0b39-11e8-81e7-e7dd8d8cdaeb.png">
<img width="1280" alt="screenshot 2018-02-06 01 35 39" src="https://user-images.githubusercontent.com/7194057/35876867-457ec1a6-0b39-11e8-87c2-2c2beed5455e.png">

I see that the control panel also becomes black because the shadow is in there too. I need to find the way to separate and apply the shadowing only for mesh.

costashatz:
@dhkim0821 great examples! Can you share the code with us? See also #976 for my try. Are you very far away from it? Thanks!

dhkim0821:
I think that the progress about shadow rendering is far from what I did. Just in case, I put the code for my shadow setting here. 

[...]

PR #976, opened 2018-02-06
Osg shadows and lighting

costashatz:
This PR enables shadows in the OpenSceneGraph (OSG) wrapper of DART and solves #971. In more detail:

- There was an issue in the OSG lighting setup:
    - The turning on/off of the lights was reversed. I fixed it.
    - The lights were always directional; directional lights do not have any attenuation and this results in unrealistic scenes. I selected point lights with distance attenuation in order to produce nicer scenes.
- I created a `ShadowedWorldNode` that overrides `WorldNode` and provides the shadow capability (using the ShadowMap technique). I did this for 2 reasons: (1) someone may not want to use shadows (and they can be computationally intensive) ~~and (2) for some reason in OSG when no shadow can be casted, nothing is shown; I am still investigating this issue and ways of by-passing it.~~ **Edit: I misinterpreted this; see below (everything is fine).**

Here's a demo:
hexa_shadow

Let me know what you think...

### DIFF
--- a/dart/gui/osg/CMakeLists.txt
+++ b/dart/gui/osg/CMakeLists.txt
@@ -7,7 +7,7 @@ endif()
 if(DART_BUILD_GUI_OSG)
 
   find_package(OpenSceneGraph 3.0 QUIET
-    COMPONENTS osg osgViewer osgManipulator osgGA osgDB)
+    COMPONENTS osg osgViewer osgManipulator osgGA osgDB osgShadow)
 
   # It seems that OPENSCENEGRAPH_FOUND will inadvertently get set to true when
   # OpenThreads is found, even if OpenSceneGraph is not installed. This is quite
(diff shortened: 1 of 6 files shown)


costashatz:
I was mistaken about the issue of OpenSceneGraph; the real issue was that the home position was being resetted and as such my objects were too far away to be seen. I fixed the issue. The main issue that I see of this implementation is that the ShadowMap technique finds the first light in the scene to create the shadows and does not use all the lights. I am not an expert in OpenSceneGraph and it's mostly okay for me, so I am not planning on trying to fix this.

Let me know what you think..

jslee02:
Also, could you retarget this PR to `release-6.4`? To do so, you could:
* creating a new branch from `release-6.4` and
* cherry-picking the commits in this PR to the new branch and
* creating a new PR from the branch targeting to `release-6.4`

Thanks! :smile:

PR #978, opened 2018-02-09
OSG shadows

costashatz:
This is a better version of #976, where:

- I changed the `dart::gui::osg::Viewer` class to be able to enable/disable shadows
- I did not touch the lights (this should be in a separate PR)

There are two issues (one minor and one major):

- For now we select the `mLight1` as the light that casts shadows (because this one is higher in the up direction). I am not an OSG expert and I do not know how to combine multiple lights to produce shadows. I think this is minor as it should mostly be ok. Here's one example:
hexa_shadow
- The `ImGuiViewer` is not working properly once you enable shadows (either black or distorted views on the ImGui panels). Here's an example:

image

I am investigating why this happens and I will try to provide a fix..

@jslee02 let me know what you think..

### DIFF
--- a/dart/gui/osg/Viewer.cpp
+++ b/dart/gui/osg/Viewer.cpp
@@ -35,6 +35,11 @@
 #include <osg/OperationThread>
 #include <osgDB/WriteFile>
 
+#include <osgShadow/ShadowedScene>
+#include <osgShadow/ShadowVolume>
+#include <osgShadow/ShadowTexture>
+#include <osgShadow/ShadowMap>
+
 #include "dart/gui/osg/Viewer.hpp"
 #include "dart/gui/osg/TrackballManipulator.hpp"
 #include "dart/gui/osg/DefaultEventHandler.hpp"
@@ -190,7 +195,7 @@ void ViewerAttachment::attach(Viewer* newViewer)
 }
 
 //==============================================================================
-Viewer::Viewer(const ::osg::Vec4& clearColor)
+Viewer::Viewer(const ::osg::Vec4& clearColor, bool shadowsOn)
   : mImageSequenceNum(0),
     mImageDigits(0),
     mRecording(false),
@@ -206,6 +211,11 @@ Viewer::Viewer(const ::osg::Vec4& clearColor)
     mAllowSimulation(true),
     mHeadlights(true)
 {
+  enableShadows(shadowsOn);
+
+  // add the physics group to the root group
+  mRootGroup->addChild(mPhysicsGroup);
+
   setCameraManipulator(new osg::TrackballManipulator);
   addInstructionText("Left-click:   Interaction\n");
   addInstructionText("Right-click:  Rotate view\n");
@@ -215,12 +225,12 @@ Viewer::Viewer(const ::osg::Vec4& clearColor)
   mDefaultEventHandler = new DefaultEventHandler(this);
   // ^ Cannot construct this in the initialization list, because its constructor calls member functions of this object
 
-  setSceneData(mRootGroup);
   addEventHandler(mDefaultEventHandler);
-  setupDefaultLights();
   getCamera()->setClearColor(clearColor);
 
   getCamera()->setFinalDrawCallback(new SaveScreen(this));
+
+  setSceneData(mRootGroup.get());
 }
 
 //==============================================================================
@@ -375,7 +385,7 @@ void Viewer::addWorldNode(WorldNode* _newWorldNode, bool _active)
     return;
 
   mWorldNodes[_newWorldNode] = _active;
-  mRootGroup->addChild(_newWorldNode);
+  mPhysicsGroup->addChild(_newWorldNode);
   if(_active)
     _newWorldNode->simulate(mSimulating);
   _newWorldNode->mViewer = this;
@@ -389,7 +399,7 @@ void Viewer::removeWorldNode(WorldNode* _oldWorldNode)
   if(it == mWorldNodes.end())
     return;
 
-  mRootGroup->removeChild(it->first);
+  mPhysicsGroup->removeChild(it->first);
   mWorldNodes.erase(it);
 }
 
@@ -401,7 +411,7 @@ void Viewer::removeWorldNode(std::shared_ptr<dart::simulation::World> _oldWorld)
   if(nullptr == node)
     return;
 
-  mRootGroup->removeChild(node);
+  mPhysicsGroup->removeChild(node);
   mWorldNodes.erase(node);
 }
 
@@ -474,7 +484,7 @@ void Viewer::setupDefaultLights()
   setUpwardsDirection(mUpwards);
   switchHeadlights(true);
 
-  ::osg::ref_ptr<::osg::StateSet> lightSS = mRootGroup->getOrCreateStateSet();
+  ::osg::ref_ptr<::osg::StateSet> lightSS = mPhysicsGroup->getOrCreateStateSet();
 
   mLight1->setLightNum(1);
   mLightSource1->setLight(mLight1);
@@ -490,8 +500,8 @@ void Viewer::setupDefaultLights()
   mLightGroup->removeChild(mLightSource2);
   mLightGroup->addChild(mLightSource2);
 
-  mRootGroup->removeChild(mLightGroup);
-  mRootGroup->addChild(mLightGroup);
+  mPhysicsGroup->removeChild(mLightGroup);
+  mPhysicsGroup->addChild(mLightGroup);
 }
 
 //==============================================================================
@@ -829,6 +839,56 @@ void Viewer::updateDragAndDrops()
 const ::osg::ref_ptr<::osg::Group>& Viewer::getRootGroup() const
 {
   return mRootGroup;
+}
+
+//==============================================================================
+const ::osg::ref_ptr<::osg::Group>& Viewer::getPhysicsGroup() const
+{
+  return mPhysicsGroup;
+}
+
+//==============================================================================
+bool Viewer::isShadowed() const
+{
+  return mShadowed;
+}
+
+//==============================================================================
+void Viewer::enableShadows(bool _enable)
+{
+  if(!mPhysicsGroup) {
+    // Flags for shadowing; maybe this needs to be global?
+    constexpr int ReceivesShadowTraversalMask = 0x2;
+    constexpr int CastsShadowTraversalMask = 0x1;
+
+    // Setup shadows
+    // Create one ShadowedScene for each light
+    ::osg::ref_ptr<osgShadow::ShadowedScene> shadowedScene = new osgShadow::ShadowedScene;
+    shadowedScene->setReceivesShadowTraversalMask(ReceivesShadowTraversalMask);
+    shadowedScene->setCastsShadowTraversalMask(CastsShadowTraversalMask);
+
+    // set the physics group
+    mPhysicsGroup = shadowedScene.get();
+    mPhysicsGroup->getOrCreateStateSet();
+
+    setupDefaultLights();
+  }
+
+  if(_enable) {
+    // Use the ShadowMap technique
+    ::osg::ref_ptr<osgShadow::ShadowMap> sm = new osgShadow::ShadowMap;
+    // increase the resolution of default shadow texture for higher quality
+    int mapres = std::pow(2, 13);
+    sm->setTextureSize(::osg::Vec2s(mapres,mapres));
+    sm->setLight(mLight1);
+
+    // set the technique
+    static_cast<osgShadow::ShadowedScene*>(mPhysicsGroup.get())->setShadowTechnique(sm.get());
+  }
+  else
+    static_cast<osgShadow::ShadowedScene*>(mPhysicsGroup.get())->setShadowTechnique(0);
+
+  mShadowed = _enable;
 }
 
 } // namespace osg
--- a/dart/gui/osg/Viewer.cpp
+++ b/dart/gui/osg/Viewer.cpp
@@ -39,6 +39,8 @@
 #include <osgShadow/ShadowVolume>
 #include <osgShadow/ShadowTexture>
 #include <osgShadow/ShadowMap>
+#include <osgShadow/StandardShadowMap>
+#include <osgShadow/SoftShadowMap>
 
 #include "dart/gui/osg/Viewer.hpp"
 #include "dart/gui/osg/TrackballManipulator.hpp"
@@ -195,7 +197,7 @@ void ViewerAttachment::attach(Viewer* newViewer)
 }
 
 //==============================================================================
-Viewer::Viewer(const ::osg::Vec4& clearColor, bool shadowsOn)
+Viewer::Viewer(const ::osg::Vec4& clearColor, bool shadowsOn, ShadowType shadowType)
   : mImageSequenceNum(0),
     mImageDigits(0),
     mRecording(false),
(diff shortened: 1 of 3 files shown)


costashatz:
> This generally looks good to me! Thanks for making this change.

Thanks for the quick feedback.

> I see several inconsistent code style in this change with the DART's code style, but it would be good to resolve it using clang-format once the style related PRs (#979, #981) are merged.

OK. Let's let clang-format do its job! :smile: 

> In general, I would prefer using a setting function for shadow technique instead of using enum for better flexibility (e.g., Viewer::setShadowTechnique(osg::ref_ptr<osgShadow::ShadowTechnique> newShadowTechnique)); with the current design, we have to change the code to use a new shadow technique. However, I'm okay with the current design because I don't expect OSG will add new techniques soon.

I can do this change later today..

costashatz:
@jslee02 I followed your advice about setting the shadow technique. I also had to expose the light sources of the viewer so that someone can use them in the shadow techniques (some of them recommend/require to set which light you want to use for shadowing).

costashatz:
@jslee02 @mxgrey I cannot seem to find a solution for the `ImGuiViewer`. Is it okay if we put a warning (or even disable the `enableShadows` function for `ImGuiViewer`) and merge it like this?

jslee02:
@costashatz Looking at the code, I found that there is another issue that `InteractiveFrame` is also under the shadow scene, which shouldn't be. I think we need more than just `mRootGroup`-`mPhysicsGroup` distinction. Let me try to come up with a possible solution (hopefully for the ImGUI issue as well).

mxgrey:
We could add a flag to the `VisualAspect` class for enabling and disabling shadows. The `InteractiveFrame` could disable its own shadow flags by default.

mxgrey:
To complement the shadow flag, we would probably need to add an update function to the `Viewer` class to tell it to re-examine the shadow flags, so it can re-sort the frames between the shadowed group and the root group.

dart/gui/osg/Viewer.hpp
```diff
@@ -114,10 +115,9 @@ class ViewerAttachment : public virtual ::osg::Group
 class Viewer : public osgViewer::Viewer, public dart::common::Subject
 {
 public:
-
   /// Constructor for dart::gui::osg::Viewer. This will automatically create the
   /// default event handler.
-  Viewer(const ::osg::Vec4& clearColor = ::osg::Vec4(0.9,0.9,0.9,1.0));
+  Viewer(const ::osg::Vec4& clearColor = ::osg::Vec4(0.9,0.9,0.9,1.0), bool shadowsOn = false, ::osg::ref_ptr<osgShadow::ShadowTechnique> shadowTechnique = nullptr);
```
mxgrey: Would it make sense to get rid of the `shadowsOn` flag and allow a `nullptr` for `shadowTechnique` to indicate that shadows should be off? Then if someone passes in a non-null value for `shadowTechnique`, we have it default to shadows being on?

Also, if there's a default native `osgShadow::ShadowTechnique` that we can recommend to users, it would be nice if we referred to it in the documentation of this function.

dart/gui/osg/Viewer.hpp
```diff
@@ -288,6 +293,13 @@ class Viewer : public osgViewer::Viewer, public dart::common::Subject
   /// Get the root ::osg::Group of this Viewer
   const ::osg::ref_ptr<::osg::Group>& getRootGroup() const;
 
+  /// Get the physics root ::osg::Group of this Viewer
+  const ::osg::ref_ptr<::osg::Group>& getPhysicsGroup() const;
+
+  bool isShadowed() const;
+  void enableShadows(bool _enable = true, ::osg::ref_ptr<osgShadow::ShadowTechnique> shadowTechnique = nullptr);
```
mxgrey: These default values seem very counter-intuitive to me. Is it possible to have shadows with a `nullptr` for the `osgShadow::ShadowTechnique`? I would assume that providing a `nullptr` will make it impossible for OSG to render shadows, so it wouldn't make sense to have them enabled. Or is the `shadowTechnique` something extra that can be applied on top of the default OSG shadowing?

dart/gui/osg/Viewer.cpp
```diff
@@ -831,6 +852,64 @@ const ::osg::ref_ptr<::osg::Group>& Viewer::getRootGroup() const
...
+
+  if(_enable)
+    setShadowTechnique(shadowTechnique);
+  else
+    static_cast<osgShadow::ShadowedScene*>(mPhysicsGroup.get())->setShadowTechnique(0);
+
+  mShadowed = _enable;
+}
+
+void Viewer::setShadowTechnique(::osg::ref_ptr<osgShadow::ShadowTechnique> shadowTechnique) {
```
mxgrey: I see now that if the user passes a `nullptr`, we replace it with a default `osgShadow::ShadowMap`.

I think that's a bit opaque, so what if we instead make an `osg::ref_ptr<osgShadow::ShadowMap>` instance the default value for these functions?

dart/gui/osg/Viewer.hpp
costashatz: > These default values seem very counter-intuitive to me. Is it possible to have shadows with a nullptr for the osgShadow::ShadowTechnique? I would assume that providing a nullptr will make it impossible for OSG to render shadows, so it wouldn't make sense to have them enabled. Or is the shadowTechnique something extra that can be applied on top of the default OSG shadowing?

You need the `ShadowTechnique` in order to have shadows..

dart/gui/osg/Viewer.cpp
costashatz: > I think that's a bit opaque, so what if we instead make an osg::ref_ptr<osgShadow::ShadowMap> instance the default value for these function arguments?

Yes we can do this..

> I'm further realizing that the initialization of the ShadowMap is using one of the member variables of the Viewer class, mLight1.

That's why I exposed the `LightSources` so that users can define their own `ShadowTechnique`s and be able to choose the light source that will cast the shadows. We can't do it generically because not all of the techniques have the `setLight()` functionality..

costashatz:
> @costashatz Looking at the code, I found that there is another issue that InteractiveFrame is also under the shadow scene, which shouldn't be. I think we need more than just mRootGroup-mPhysicsGroup distinction. Let me try to come up with a possible solution (hopefully for the ImGUI issue as well).

Hmm.. I didn't pay attention to the `InteractiveFrame`.. Let me also try to think about it..

> We could add a flag to the VisualAspect class for enabling and disabling shadows. The InteractiveFrame could disable its own shadow flags by default.

We cannot do this, because as OSG docs say: "Not all of the techniques respect the shadowing flags". In practice, most of them ignore at least on flag (either caster or receiver of shadow).!

dart/gui/osg/Viewer.hpp
mxgrey: Right, which is why this API seems deceptive. It's simultaneously telling the Viewer "I want you to enable shadows" and "I don't want you to have a `ShadowTechnique`.

I understand now that we're interpreting a `nullptr` as "use some default `ShadowTechnique`, but that isn't clear without digging into the source code of this function.

mxgrey:
> We cannot do this, because as OSG docs say: "Not all of the techniques respect the shadowing flags"

When I say "shadow flag" I'm just talking about a simple boolean within the DART API. It would be agnostic towards the underlying rendering engine. The shadowing flag of the `VisualAspect` would be checked by the `dart-gui-osg` wrapper, and the wrapper would decide whether to put the object in the group that gets affected by shadows (`mPhysicsGroup`) or the group that does not get affected by shadows (`mRootGroup`).

We could even set it up so that every rendering loop it checks whether that flag has changed for each `ShapeFrame`, and then switches the frame into the other group if it has changed.

This assumes that my understanding is correct: Only objects in `mPhysicsGroup` will be shaded. Is that correct?

costashatz:
> This assumes that my understanding is correct: Only objects in mPhysicsGroup will be shaded. Is that correct?

Yes.. At least that's what is said by the OSG docs and matches what I tried so far...

costashatz:
> When I say "shadow flag" I'm just talking about a simple boolean within the DART API. It would be agnostic towards the underlying rendering engine. The shadowing flag of the VisualAspect would be checked by the dart-gui-osg wrapper, and the wrapper would decide whether to put the object in the group that gets affected by shadows (mPhysicsGroup) or the group that does not get affected by shadows (mRootGroup).
> 
> We could even set it up so that every rendering loop it checks whether that flag has changed for each ShapeFrame, and then switches the frame into the other group if it has changed.

I like this idea. Let me try to see what happens..

costashatz:
I implemented @mxgrey 's idea and it works nicely for the `InteractiveFrame`. I had to move everything to the `WorldNode` instead of the `Viewer`. Later today I will put a check to see if any frame changed its shadow flag and put some documentation.

I didn't manage to make `ImGuiViewer` work properly... I am investigating a bit more to see if I can find anything..

costashatz:
After some more investigation, I think that we need to create a version of `ImGuiViewer` that uses OpenSceneGraph's primitives (which is certainly possible as ImGui just gives us a list of indexed primitives to draw).. Personally, I don't think I have the time nor the expertise (I am not using this part of the API at all) to do that.. If you agree, we can do one of the following (once you review the rest of the changes):

- Disable completely the shadows when using `ImGuiViewer` (we can also give a silent warning to the user wanting to use shadows with it)..
- Leave it as is and put a warning when a user uses shadows with `ImGuiViewer`...
- Post-pone the merge of this PR until we find a solution..

Personally I prefer either of the 2 first options because I am not using the `ImGuiViewer` at all and if it is relatively well documented, it should be ok for most users. Of course, we should make an issue for creating a version of `ImGuiViewer` with OpenSceneGraph primitives that can be attached to the scene graph.

@jslee02 @mxgrey let me know what you think..

mxgrey:
I'm okay with either of the first two options.

I wouldn't consider this (nor ImGui) to be a critical feature of DART, so correctness and compatibility between the two isn't a high priority. Shadows are definitely a desirable feature, and it's mostly unrelated to the ImGui feature, so I don't see any reason to block it.

dart/gui/osg/WorldNode.hpp
```diff
@@ -67,7 +68,8 @@ class WorldNode : public ::osg::Group
   friend class Viewer;
 
   /// Default constructor
-  explicit WorldNode(std::shared_ptr<dart::simulation::World> _world = nullptr);
+  /// Shadows are disabled by default
+  explicit WorldNode(std::shared_ptr<dart::simulation::World> _world = nullptr, ::osg::ref_ptr<osgShadow::ShadowTechnique> shadowTechnique = nullptr);
```
mxgrey: I'm still a little bit leery about having `shadowTechnique = nullptr` secretly mean "Let us choose a default shading technique for you".

However, we do already have a precedent of this pattern where passing in a `nullptr` for `ResourceRetriever` means that we will use a default `LocalResourceRetriever`.

I might feel better about this pattern if we rename the argument from `shadowTechnique` to something like `alternativeTechnique`. That will at least tell the user that they are specifying an alternative rather than specifying the technique to use. Then passing a `nullptr` naturally says "I don't want to use an alternative", and so we apply the default shading technique.

Sorry if this sounds petty, but without *something* along these lines, a user would have to look into the source code to actually understand how these functions should be used.

costashatz: > Sorry if this sounds petty, but without something along these lines, a user would have to look into the source code to actually understand how these functions should be used.

No worries. I understand what you say. Nevertheless, in this version of the code, `nullptr` means no shadow (not a default shadow behavior).. The user has to specify their own `ShadowTechnique`. I could create a static default technique if you like..

mxgrey: Oh, sorry, I didn't notice that part of the recent change.

I actually liked providing a default shadow technique; I just wasn't comfortable with the semantics of "giving a `nullptr` technique means we should use a default". Sorry for not communicating that well.

I'm not clear on how a static default would work, since it seems to need information about the Viewer that's using it. I guess we could make it a factory function that takes a `Viewer*` argument and returns a `ref_ptr<ShadowTechnique>`? We wouldn't be able to use that function as the default argument for this constructor, but it could at least make it easy for users to enable shadows. We could add a comment in the documentation of this function that refers to the default ShadowTechnique factory.

Since the shadows conflict with the ImGui feature, I suppose we should have shadows turned off by default anyway. It would've been nice to have shadows on by default, but we can tackle that later once we've fixed the ImGui conflict.

costashatz: > I'm not clear on how a static default would work, since it seems to need information about the Viewer that's using it. I guess we could make it a factory function that takes a Viewer* argument and returns a ref_ptr<ShadowTechnique>? We wouldn't be able to use that function as the default argument for this constructor, but it could at least make it easy for users to enable shadows. We could add a comment in the documentation of this function that refers to the default ShadowTechnique factory.

That's exactly what I had in mind.. A static member function that creates a default `ShadowTechnique` (given a `Viewer*`) that users can use.. I'll do that asap..

> ince the shadows conflict with the ImGui feature, I suppose we should have shadows turned off by default anyway. It would've been nice to have shadows on by default, but we can tackle that later once we've fixed the ImGui conflict.

I agree on this and that's the reason I left the shadows disabled by default.. :smile:

costashatz: I was also thinking of modifying the `osgOperationalSpaceControl` example to be able to enable/disable shadows interactively (with a simple event handler). What do you think?

jslee02:
I'm also fine with any of the first two options!

dart/gui/osg/WorldNode.hpp
mxgrey: That sounds like a great place to put an example of this, thanks!

costashatz:
I went with the second option (warning when enabling shadows in `ImGuiViewer`). I also added an example of enabling/disabling shadows in `osgOperationalSpaceControl` example. Lastly, I ran `make check-format` and I got the following results:

```bash
Scanning dependencies of target check-format
Checking 4 files... 
Every file seems to comply with our code convention.
Done.
Built target check-format
```
So I guess the styling should be ok!

@jslee02 @mxgrey let me know if you have anything to add or you want me to fix something.

jslee02:
@costashatz Regarding the style, we're just checking four files in `lcpsolver` directory as an example. That said your changes are not checked. :grin: We will format the code for the entire code eventually.

The current change looks good to me. I think we can merge this on approval of @mxgrey.

mxgrey:
I'll do one last careful review tonight as soon as I get the chance, but it's looking very good to me as well.

dart/gui/osg/Viewer.hpp
```diff
@@ -197,6 +197,11 @@ class Viewer : public osgViewer::Viewer, public dart::common::Subject
   /// Get the Group node that contains the LightSources for this Viewer
   const ::osg::Group* getLightGroup() const;
 
+  /// Get one of the LightSources of this Viewer
+  /// index either 0 or 1
+  /// Useful for shadowing techniques
+  const ::osg::ref_ptr<::osg::LightSource>& getLightSource(unsigned int index = 0) const;
```
jslee02: Nit: Please use `std::size_t` for index parameters.

dart/gui/osg/WorldNode.hpp
jslee02: Nit: Please remove the leading underscore from `_world`.

dart/gui/osg/WorldNode.cpp
```diff
@@ -61,19 +65,40 @@ class WorldNodeCallback : public ::osg::NodeCallback
...
+  constexpr int ReceivesShadowTraversalMask = 0x2;
+  constexpr int CastsShadowTraversalMask = 0x1;
+
+  // Setup shadows
+  // Create a ShadowedScene
+  ::osg::ref_ptr<osgShadow::ShadowedScene> shadowedScene = new osgShadow::ShadowedScene;
+  shadowedScene->setReceivesShadowTraversalMask(ReceivesShadowTraversalMask);
+  shadowedScene->setCastsShadowTraversalMask(CastsShadowTraversalMask);
+
+  // set the physics group
```
mxgrey: Nitpick: This is called `normal group` now rather than `physics group`.

PR #983, opened 2018-02-12
Use newer osx image (OS X 10.13)

jslee02:


jslee02:
I genuinely don't know why [`nlopt` is not installed on Travis](https://travis-ci.org/dartsim/dart/jobs/340330225#L564) while it has no problem on my local computer. :thinking:

PR #997, opened 2018-02-18
Add relative Jacobian functions to MetaSkeleton

### LINKED ISSUE #230: API extracting relative Jacobians between two arbitrary bodies in a Skeleton

jslee02:
**Summary**
This PR adds additional Jacobian functions to `MetaSkeleton` for getting relative Jacobians (and its derivatives) of a frame to an arbitrary frame in the same skeleton.

**Caveat**
Functions for getting the derivatives of classical relative Jacobian are missing. This will be added in future PRs.

Resolves #230

***


jslee02:
I found [another person liked the new word](https://books.google.com/books?id=1Qe7BQAAQBAJ&pg=PA381&lpg=PA381&dq=define+Jabocian&source=bl&ots=SVAGWCqaQs&sig=5iRRUadFdIWih2c9YG0g945h8lI&hl=en&sa=X&ved=0ahUKEwif6K_aprHZAhVN3GMKHdYhBZEQ6AEIKzAA#v=onepage&q=define%20Jabocian&f=false). :joy: 

It seems Travis fails on debug mode because the test size is too large since the test in the 5-level of nested for-statements. Let me address that tomorrow.

PR #1003, opened 2018-02-21
Add TranslationalJoint2D

jslee02:
`TranslationalJoint2D` is a 2d version of `TranslationalJoint` or `PlanarJoint` without the rotational axis.

***

- [x] Add unit test(s) for this change (need tests for `SkelParser`)

### DIFF
--- a/dart/dynamics/detail/TranslationalJoint2DAspect.hpp
+++ b/dart/dynamics/detail/TranslationalJoint2DAspect.hpp
@@ -58,7 +58,7 @@ public:
 
   /// Constructor for arbitrary plane types. mPlaneType will be set to
   /// PlaneType::ARBITRARY
-  TranslationalJoint2DUniqueProperties(
+  explicit TranslationalJoint2DUniqueProperties(
       const Eigen::Matrix<double, 3, 2>& transAxes);
 
   /// Constructor for arbitrary plane types. mPlaneType will be set to
--- a/dart/dynamics/detail/TranslationalJoint2DAspect.hpp
+++ b/dart/dynamics/detail/TranslationalJoint2DAspect.hpp
@@ -30,8 +30,8 @@
  *   POSSIBILITY OF SUCH DAMAGE.
  */
 
-#ifndef DART_DYNAMICS_DETAIL_TranslationalJoint2DASPECT_HPP_
-#define DART_DYNAMICS_DETAIL_TranslationalJoint2DASPECT_HPP_
+#ifndef DART_DYNAMICS_DETAIL_TRANSLATIONALJOINT2DASPECT_HPP_
+#define DART_DYNAMICS_DETAIL_TRANSLATIONALJOINT2DASPECT_HPP_
 
 #include <string>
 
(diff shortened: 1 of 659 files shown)


mxgrey:
I should have some time tonight to give this a review.

dart/dynamics/detail/TranslationalJoint2dAspect.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+class TranslationalJoint2d;
+
+namespace detail {
+
+//==============================================================================
+struct TranslationalJoint2dUniqueProperties
+{
+  /// Plane type
+  PlaneType mPlaneType;
```
mxgrey: Since we've gone through the trouble of making the member functions, I feel that it would make sense to make these member variables private. That would ensure that this member data always remains sane. I think the mild inconvenience of not being able to set these values directly is heavily outweighed by the importance of keeping the values sane and consistent.

mxgrey: Since we've gone through the trouble of making all the member function, I feel like it would make sense to have these member variables be private. That way, we can rely on the member functions to keep the values sane and consistent. I think the inconvenience of not having direct access to the members is heavily outweighed by the value of guaranteeing sanity and consistency.

dart/utils/SkelParser.cpp
```diff
@@ -2293,6 +2301,94 @@ JointPropPtr readTranslationalJoint(
...
+  if (hasElement(_jointElement, "plane"))
+  {
+    tinyxml2::XMLElement* planeElement = getElement(_jointElement, "plane");
+
+    // Type attribute
+    std::string type = getAttributeString(planeElement, "type");
+
+    if (type == "xy")
+    {
+      properties.mPlaneType = dynamics::PlanarJoint::PlaneType::XY;
```
mxgrey: If we do make the member variables private, we'll need to tweak the implementation of this parser to store these parameters and then pass them into the member functions of `properties`.

jslee02:
> This PR is making me realize just how much silly boilerplate is still required to implement a new joint type correctly. Ideally, we should just be able to define the unique properties, and then define the relative transform behavior and the Jacobian behavior.

Agreed. This is one of the reasons `GenericJoint` was introduced with the configuration space classes (e.g., `RealVectorSpace` and `SO3Space`). I intended to generalize the joint behavior using `GenericJoint` and define the concrete behaviors (e.g., Jacobian computations specific to the concrete configuration types) using geometric computations that are also templatized to the configuration spaces (e.g., `integratePosition()` in `math/ConfigurationSpace.hpp`) so that we only require minimum effort to add new joint types of new configuration spaces and furthermore any compositions of the current joint spaces. Since there are more rooms to improve the current implementation, we might want to revisit this at some point.

dart/dynamics/detail/TranslationalJoint2dAspect.hpp
mxgrey: Okay, apparently if you leave a review comment, accidentally close the webpage, and come back to continue reviewing, Github will (sometimes) pretend that you didn't leave the review comment at all....

dart/dynamics/detail/TranslationalJoint2DAspect.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+class TranslationalJoint2D;
+
+namespace detail {
+
+//==============================================================================
+struct TranslationalJoint2DUniqueProperties
+{
+  /// First and second translational axis
+  Eigen::Matrix<double, 3, 2> mTransAxes;
```
mxgrey: We would also want this member variable to be private.

I would also suggest changing this from a `struct` to a `class` since it's no longer a POD.

PR #1011, opened 2018-02-24
Add LockableReference classes

### LINKED ISSUE #1007: Adding a getMutexes() function to MetaSkeleton
In https://github.com/personalrobotics/aikido, we're currently opting to use `MetaSkeleton`s rather than `Skeleton`s. However, `MetaSkeleton`s don't have associated mutex(es) so I'm concerned we'll run into issues with thread safety.

It seemed like there was some discussion in #466 about what to do with `ReferentialSkeleton`s on that PR, but I don't think a `getMutexes()` method was ever added.

jslee02:
This PR is a minimal set of changes of adding `MetaSkeleton::getMutexes()` with two helper functions to lock and unlock mutexes.

* Related PR: #466 
* Resolves #1007 

**Possible Improvement**
As pointed out by this [article](https://www.justsoftwaresolutions.co.uk/threading/acquiring-multiple-locks-without-deadlock.html), locking by address works for small numbers of mutexes, but it could be inefficient for many mutexes. Suggested improvement is using "try-and-back-off". But this PR leaves this a future improvement since it could be achieved without the public API changes from this PR.

***

- [x] Update `CHANGELOG.md` reflecting the changes


jslee02:
@psigen I like the idea! #1014 is a prototype of the wrapper class.

dart/dynamics/ReferentialSkeleton.hpp
```diff
@@ -460,6 +478,13 @@ class ReferentialSkeleton : public MetaSkeleton
   /// Name of this ReferentialSkeleton
   std::string mName;
 
+  /// Skeletons that this ReferentialSkeleton is referencing to.
+  std::vector<Skeleton*> mSkeletons;
+
+  /// Mutexes of the skeletons. The mutexes are sorted in order of memory
+  /// addresses.
+  std::vector<std::mutex*> mSkeletonMutexes;
```
mxgrey: Same suggestion: consider using `std::set` instead of `std::vector`.

dart/dynamics/MetaSkeleton.hpp
```diff
@@ -71,6 +71,11 @@ class MetaSkeleton : public common::Subject
   /// Default destructor
   virtual ~MetaSkeleton() = default;
 
+  /// Returns mutex.
+  virtual std::unique_ptr<common::Mutex> getCustomMutex() const = 0;
```
mxgrey: It feels a little bit strange to me to return a `std::unique_ptr` to a "Mutex", since that could create the impression that no one else has access to this particular "Mutex", and therefore locking this "Mutex" would be frivolous. Of course the reality is that this is a `std::unique_ptr` to a wrapper that contains references to mutex(es), so it's perfectly sensible to receive a `std::unique_ptr` to it.

I wonder if we should consider naming the class `MutexWrapper` or `MutexReference`? I'm totally fine with just keeping it as `Mutex`; this was just a thought that hit me while looking through the code, so I wanted to express it in case anyone else was struck with a similar impression.

jslee02: This sounds good to me. I changed the name to `MutexReference` in the similar fashion of `ReferentialSkeleton`, but am open to `MutexWrapper` or better name if any.

dart/dynamics/MetaSkeleton.hpp
```diff
@@ -71,6 +71,11 @@ class MetaSkeleton : public common::Subject
   /// Default destructor
   virtual ~MetaSkeleton() = default;
 
+  /// Returns mutex.
+  virtual std::unique_ptr<common::MutexReference> getCustomMutex() const = 0;
```
mxgrey: We could consider naming this something like `getMutexReference()` and keep the original `getMutex()` function the way it is.

dart/dynamics/ReferentialSkeleton.hpp
```diff
@@ -54,6 +55,15 @@ class ReferentialSkeleton : public MetaSkeleton
...
   virtual ~ReferentialSkeleton() = default;
 
+  // Documentation inherited
+  std::unique_ptr<common::MutexReference> getCustomMutex() const override;
+
+  /// Returns sorted mutexes in order of memory addresses.
+  ///
+  /// \note Generally, you don't need to use this function in favor of
+  /// getCustomMutex() that guarantees deadlock-free locking.
+  const std::set<std::mutex*>& getStdMutexes() const;
```
mxgrey: Do we really want to expose this function at all? It strikes me as needlessly exposing implementation detail to the user. E.g. what if we decide later to not use `std::set` to store the mutexes? Then this function that returns a const-reference would have to be removed, or else we'd need to hang onto a separate `std::set` copy of the mutexes. It's hard for me to imagine a situation where a user would want access to the individual mutexes instead of the `MutexReference`.

dart/dynamics/ReferentialSkeleton.hpp
```diff
@@ -70,6 +80,15 @@ class ReferentialSkeleton : public MetaSkeleton
   /// \{ \name Structural Properties
   //----------------------------------------------------------------------------
 
+  /// Returns number of skeletons associated with this ReferentialSkeleton.
+  std::size_t getNumSkeletons() const;
+
+  /// Returns skeletons associated with this ReferentialSkeleton.
+  const std::unordered_set<const Skeleton*>& getSkeletons() const;
```
mxgrey: I feel kind of the same way about this function as well, in terms of needlessly exposing implementation detail.

As I understand it, the goal of the `ReferentialSkeleton` class is to abstract away the existence of individual `Skeleton` objects. I would suggest that we remain non-committal about the underlying storage/management of those Skeletons, and not expose these kinds of functions to users.

That said, I'm not against `getNumSkeletons()` or `hasSkeleton(~)` since those don't expose implementation.

dart/dynamics/Skeleton.hpp
```diff
@@ -159,8 +159,16 @@ class Skeleton :
   ConstSkeletonPtr getSkeleton() const;
 
   /// Get the mutex that protects the state of this Skeleton
+  /// \deprecated Use getStdMutex instead.
+  DART_DEPRECATED(6.4)
   std::mutex& getMutex() const;
 
+  /// Get the mutex that protects the state of this Skeleton
+  std::mutex& getStdMutex() const;
```
mxgrey: If we change `getCustomMutex()` to `getMutexReference()`, then we won't need this function anymore.

dart/dynamics/ReferentialSkeleton.hpp
mxgrey: I guess it's really `MetaSkeleton` that exists to abstract away the individual Skeletons. Still, I think we should avoid exposing implementation to avoid committing ourselves to a particular implementation.

psigen:
@mxgrey: Would it be possible to reverse logic this somehow?  Making a new thing called `MutexReference` and have `getMutexReference()` and differentiating this from `getMutex()` seems overly complex.  It's really an implementation detail that this thing needs to be managed as a `unique_ptr` in the first place and we are requiring users to think carefully about what a `MutexReference` might be and how it might need to be managed.

Is there some way we could either alias or wrap this such that we have a class that looks like a mutex class, gets like used without other arguments in a `std::lock_guard`, and externally gets memory managed like a `std::mutex`?  Maybe `getMutex()` can return a `MutexReference` which is an alias for `std::unique_ptr<common::Mutex>()`?

mxgrey:
@psigen Maybe we could change the name back to `Mutex`, and then have each class store a derived `Mutex` type. Then we could have `MetaSkeleton::getMutex()` return a `Mutex&` which refers to the derived `Mutex` object that's held by each class?

E.g.: `Skeleton` would hold a `SingleMutex` that is constructed with a reference to the Skeleton's mutex. `ReferentialSkeleton` would hold a `MultiMutex` that will dynamically adjust to hold references to the various `Skeleton` mutexes that come and go from `ReferentialSkeleton`. Each of them would return a reference to their `Mutex` object when `getMutex()` is called.

The main issue is that we need the stopgap measure of introducing `getCustomMutex()` and deprecating the current signature of `getMutex()` if we want to add this to DART6.

psigen:
I'm fine with that, it seems reasonable to tie the `Mutex` to the Skeleton's lifetime.  I understand the temporary need for `getCustomMutex()`.

dart/dynamics/ReferentialSkeleton.hpp
mkoval: I agree with @mxgrey. I suggest we remove this unless we have a specific use-case in mind that requires it.

dart/common/MutexReference.hpp
```diff
@@ -0,0 +1,118 @@
...
+namespace dart {
+namespace common {
+
+/// Mutex is a wrapper class of single or multiple std::mutex to provide unified
+/// interface that guarantees deadlock-free locking and unlocking of the
+/// internal mutex(es).
+///
+/// This class is compatible to BasicLockable concept so that it can be used
+/// as a template parameter that requires this concept such as std::lock_guard.
+class MutexReference
```
mkoval: Just a thought: Rename this to `LockableReference`, since this class effectively implements type erasure for the [`Lockable` concept](http://en.cppreference.com/w/cpp/concept/Lockable). I think `MutexReference` is a fine name as well, so I don't feel too strongly about this.

dart/common/MutexReference.hpp
```diff
@@ -0,0 +1,118 @@
...
+  /// Unlocks the mutexes.
+  virtual void unlock() noexcept = 0;
+
+protected:
+  /// Copy construction is not allowed.
+  MutexReference(const MutexReference&) = delete;
+};
+
+/// This class references a single mutex.
+class SingleMutexReference final : public MutexReference
```
mkoval: Just a thought: Template this on the type of the mutex because this class will work on _any_ type that satisfies the `Lockable` concept.

dart/common/MutexReference.hpp
```diff
@@ -0,0 +1,118 @@
...
+  // Documentation inherited
+  void unlock() noexcept override;
+
+private:
+  /// Mutex this class references.
+  std::mutex& mMutex;
+};
+
+/// This class references multiple mutexes.
+class MultiMutexReference final : public MutexReference
```
mkoval: Just a thought: Template this class on an iterator type. Then, we could replace the `std::set<std::mutex*>&` with a pair of iterators. That would make this class support any type of collection that: (a) has a forward iterator and (b) supports multiple iteration.

This would be a big step towards @mxgrey's request to hide this type of implementation detail behind the API.

dart/common/MutexReference.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+  // Documentation inherited
+  void unlock() noexcept override;
+
+private:
+  /// Mutex this class references.
+  std::mutex& mMutex;
+};
+
+/// This class references multiple mutexes.
```
mkoval: Document that `MultiMutexReference` acquires the locks in the specified order. I.e. it is the user's responsibility to sort the collection to avoid deadlock.

dart/dynamics/Skeleton.cpp
```diff
@@ -377,10 +377,22 @@ ConstSkeletonPtr Skeleton::getSkeleton() const
...
+//==============================================================================
+std::mutex& Skeleton::getStdMutex() const
 {
   return mMutex;
 }
 
+//==============================================================================
+std::unique_ptr<common::MutexReference> Skeleton::getCustomMutex() const
+{
+  return common::make_unique<common::SingleMutexReference>(mMutex);
```
mkoval: If this `Skeleton` is destructed, the `SingleMutexReference` returned by this function will contain a dangling reference to its mutex. `Skeleton`s are always owned by `shared_ptr`, we could prevent this by creating a new `OwnedMutexReference` class that stores a `std::mutex&` and a `std::shared_ptr<void>`. We would construct it here by passing in `mMutex` and `shared_from_this()`.

This would prevent the `Skeleton` from being destructed until all `OwnedMutexReference`s go out of scope. As an interesting alternative, we only prevent the `Skeleton` from being destructed while its lock is acquired by replacing the `std::shared_ptr<void>` with an `std::weak_ptr<void>`. Every function on `MutexRefernece` would `lock()` the `weak_ptr` and skip the lock if the `Skeleton` was already destructed. It is presumably safe to skip the lock if all of the data that it is protecting has already been deleted. 😉 

(I am not sure if the `weak_ptr` approach is necessary or desirable. I just thought mentioned it because I thought it was interesting.)

dart/dynamics/ReferentialSkeleton.hpp
mkoval: I agree. I would love to return a type-erased range here, but I don't know of a good way of doing that in any existing version of C++. The best option we have to today would be to return separate `begin()` and `end()` iterators, but that would be pretty gross to use as a consumer of the API.

dart/dynamics/ReferentialSkeleton.hpp
```diff
@@ -70,6 +80,15 @@ class ReferentialSkeleton : public MetaSkeleton
...
+
+  /// Returns whether this ReferentialSkeleton contains \c skel.
```
mkoval: Ambiguous comment. This could be interpreted as "contains the _entire_ `Skeleton`" or "contains _any_ `BodyNode` or `Joint` from the `Skeleton`." I think we intend the latter.

dart/dynamics/ReferentialSkeleton.hpp
```diff
@@ -460,6 +479,13 @@ class ReferentialSkeleton : public MetaSkeleton
   /// Name of this ReferentialSkeleton
   std::string mName;
 
+  /// Skeletons that this ReferentialSkeleton is referencing to.
+  std::unordered_set<const Skeleton*> mSkeletons;
+
+  /// Mutexes of the skeletons. The mutexes are sorted in order of memory
+  /// addresses.
+  std::set<std::mutex*> mSkeletonMutexes;
```
mkoval: This seems like an implementation detail. Could we just return a `common::MutexReference` here?

dart/dynamics/ReferentialSkeleton.hpp
```diff
@@ -460,6 +479,13 @@ class ReferentialSkeleton : public MetaSkeleton
   /// Name of this ReferentialSkeleton
   std::string mName;
 
+  /// Skeletons that this ReferentialSkeleton is referencing to.
```
mkoval: Same comment as above about this comment being ambiguous.

mkoval:
> @psigen Maybe we could change the name back to Mutex, and then have each class store a derived Mutex type. Then we could have MetaSkeleton::getMutex() return a Mutex& which refers to the derived Mutex object that's held by each class?

I am torn on this. On one hand, it doesn't make a whole lot of sense to copy around these `MutexReference` wrappers. On the other hand, doing so lets us stick a `std::shared_ptr` in the `MutexReference` class to avoid the potential for dangling references.

If we stick with the `MutexReference` class, I think we should run with @psigen's suggestion and handle the type erasure internally. This little type erasure stub could look like `Mutex` class that internally stores an `std::unique_ptr<MutexReference>` and forwards all of the function calls. It would be _very_ convenient to directly pass a `MutexReference` to `std::lock_guard` without having to dereference a `unique_ptr`.

Regardless of which way we go, I agree that we should modify `getMutex()` to return our new wrapper class in the next ABI-breaking version of DART. In the meantime, I think `getCustomMutex()` (if we return a `Mutex` class by reference) or a `getMutexReference()` (if we return a `MutexReference` class by value) is reasonable.

dart/common/MutexReference.hpp
psigen: I'm not sure I understand this.  If this is using a `std::set` as currently implemented, it is going to sorted by memory location and it is _not_ the user's responsibility to sort the collection, right?

Maybe this comment makes more sense after refactoring to accommodate the comment https://github.com/dartsim/dart/pull/1011/files/cc606cdb4e79006c4a9cdda3dc85addb85121943#r172058746

mkoval: That is only true if you only ever lock these mutexes through a `MultiReferenceMutex`. This class stores _references_ to mutexes, so it would be perfectly reasonable for a person to lock them elsewhere - and potentially in a different order!

psigen:
> I am torn on this. On one hand, it doesn't make a whole lot of sense to copy around these MutexReference wrappers. On the other hand, doing so lets us stick a std::shared_ptr in the MutexReference class to avoid the potential for dangling references.

My thought was that this would end up being a really really tiny struct since it just contains a couple of pointers, so copying it around wouldn't be much of a performance implication.

As @mxgrey pointed out, we can make the function signature be a `Mutex&` so that if people really desperately need to avoid copying it, they could carefully pass a reference to the wrapper around.

mkoval:
> As @mxgrey pointed out, we can make the function signature be a `Mutex&` so that if people really desperately need to avoid copying it, they could carefully pass a reference to the wrapper around.

The problem with this is that the `Mutex&` could outline the `Skeleton` that owns the `Mutex`. The easiest way of fixing that (if we care) is to make `MutexReference` store an `std::shared_ptr` to the `Skeleton` and return `MutexReference`  _by value_.

This would also give us an object that we could directly pass into `std::lock_guard`, etc.

I am not sure how much we care about this. It depends how easy we think it will be to accidentally create a dangling reference.

dart/common/MutexReference.hpp
psigen: I totally agree, but that seemed unrelated to sorting the collection in here.  Documenting what the locking order is critical though.

Actually, it suggests that we might want to make the `MultiMutex` construction be an API that users can exploit if they are in situations where they need to lock two `Skeleton`s:  it guarantees that the mutexes will be acquired _in a DART compatible fashion_.

I'm wondering if we should have a helper function for list initialization of the `MultiMutex` so that users can do something like:

```
std::lock_guard<common::MultiMutex> lock(MultiMutex { skel1_mutex, skel2_mutex} );
# I'm sure this syntax isn't quite right.
```

e.g. https://stackoverflow.com/a/16455093

mxgrey:
> The problem with this is that the Mutex& could outline the Skeleton that owns the Mutex.

Keep in mind that even if we returned the `Mutex` by value (or by `std::unique_ptr`), it would be containing a reference to one or more `std::mutex` which are members of the Skeleton(s) that are providing the `Mutex`. So if the Skeleton(s) die, then the `Mutex` objects will be broken anyway. The lifetime of the `Mutex` is already implicitly tied to the lifetime of the Skeleton(s); returning it by reference would just make that more explicit.

I imagine the real issue with returning a `Mutex&` would be for `ReferentialSkeleton` which would have to be able to modify which mutexes are referred to inside of the `MutliMutex` object, even when that `MultiMutex` is already locked. That could cause a lot of issues with safely locking and unlocking.

psigen:
>The easiest way of fixing that (if we care) is to make MutexReference store an std::shared_ptr to the Skeleton and return MutexReference by value.
>
> This would also give us an object that we could directly pass into std::lock_guard, etc.

I'm generally fine with this approach too, the latter condition is all I really care about.

psigen:
> I imagine the real issue with returning a Mutex& would be for ReferentialSkeleton which would have to be able to modify which mutexes are referred to inside of the MutliMutex object, even when that MultiMutex is already locked. That could cause a lot of issues with safely locking and unlocking.

Hmm, is this a more general problem?  Can a user lock a `ReferentialSkeleton` and then modify something in it that changes which skeletons it references?  If so, we would need to be clear on the expected semantics anyway.

dart/dynamics/MetaSkeleton.hpp
jslee02: I renamed this to `getLockableReference()` instead to reflect the renaming of `MutexReference` to `LockableReference`.

dart/dynamics/ReferentialSkeleton.hpp
jslee02: Removed.

dart/dynamics/ReferentialSkeleton.hpp
jslee02: Removed.

@mkoval I'm thinking it would be good to provide iterator APIs for other classes whenever relevant, but let's save it for future changes. :sweat_smile:

dart/dynamics/Skeleton.hpp
jslee02: Removed.

dart/common/MutexReference.hpp
jslee02: Renamed.

dart/common/MutexReference.hpp
jslee02: Done.

dart/common/MutexReference.hpp
jslee02: Done.

I templatize this class for a lockable type instead of an iterator. Instead, the constructor is templatized for an iterator and check the compatibility between the lockable type of this class and the value type of iterator using `static_assert`.

This is because (1) it would allow us to do type aliasing for the lockable types as
```cpp
using SingleMutexReference = SingleLockableReference<std::mutex>;
using MultiMutexReference = MultiLockableReference<std::mutex>;
```
and (2) it would be authentically better (I think) when we create `MultiLocableReference`. For example,
```cpp
  // templatized for lockable type
  return common::make_unique<
      common::MultiLockableReference<std::mutex>>(
          mPtr, mSkeletonMutexes.begin(), mSkeletonMutexes.end());

  // templatized for iterator type
  return common::make_unique<
      common::MultiLockableReference<std::set<std::mutex*>::iterator>>(
          mPtr, mSkeletonMutexes.begin(), mSkeletonMutexes.end());
```

dart/common/MutexReference.hpp
jslee02: Done.

I left the order should be specified the user, and `MultiLockableReference` just conforms the given order by storing the list of lockables in `std::vector`. The order of `std::mutex`s are sorted in `ReferentialSkeleton` whenever skeletons are added to it.

I'm still open to @psigen 's suggestion though.

dart/dynamics/Skeleton.cpp
jslee02: I took the alternative suggestion (i.e., holding the skeleton's pointer in `std::weak_ptr`) without introducing a new class.

I'm still open to adding `OwnedLockableReference` though.

dart/dynamics/ReferentialSkeleton.hpp
jslee02: Done.

dart/dynamics/ReferentialSkeleton.hpp
jslee02: You mean you suggest removing `getSkeletonMutexes()`? If so, it's removed.

dart/dynamics/ReferentialSkeleton.hpp
jslee02: Done.

jslee02:
I revised the implementation based on the above discussion. Possibly there are some parts that are mistakenly not reflected by the feedback. Could you please do a second pass? Thanks!

PR #1014, opened 2018-02-26
Add Mutex class

jslee02:
This PR is a prototype of mutex(es) wrapper class based on @psigen's [comment](https://github.com/dartsim/dart/pull/1011#pullrequestreview-99150832) in #1011.


jslee02:
Thanks! Let's merge this into #1014 and wait for feedback there.

PR #1018, opened 2018-02-26
Update vcpkg before installing packages

jslee02:
This PR fixes the recent build failures of AppVeyor tests.


ras0219-msft:
Just as an FYI, you could use a single command if you wanted:
```
>vcpkg install --triplet %VCPKG_ARCH% libA libB libC libD
```

jslee02:
@ras0219-msft That commands are very neat! Thanks!

PR #1019, opened 2018-02-26
Use simpler vcpkg commands

jslee02:
based on @ras0219-msft's [suggestion](https://github.com/dartsim/dart/pull/1018#issuecomment-368595079).

***


ras0219-msft:
No commas are needed

PR #1027, opened 2018-03-09
Update package.xml based on catkin documentation

mxgrey:
This updates the package.xml to (I think) more closely adhere to the documentation [here](http://wiki.ros.org/catkin/package.xml). If anyone has more experience writing package.xml files, please critique this.

I named myself as the maintainer so that if ROS users/maintainers have issues with the package, they'll direct issues to me first. I'll go ahead and take responsibility for that, since I'm surrounded by coworkers who are experts at ROS package management.

I'm not totally clear on whether it would be better to put this in 6.3 or 6.4. I'm thinking we can do a 6.3.1 release after merging this and make that the first DART release into ROS. If anyone dislikes that idea, please do say something, since I'm new to this process.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -3,24 +3,25 @@
   <!-- This is a Catkin package.xml file to optionally support building DART in
        a Catkin workspace. Catkin is not required to build DART. For more
        information, see: http://ros.org/reps/rep-0136.html -->
-  <name>dart</name>
+  <name>dartsim</name>
   <version>6.3.0</version>
   <description>
     DART (Dynamic Animation and Robotics Toolkit) is a collaborative,
     cross-platform, open source library created by the Georgia Tech Graphics
-    Lab and Humanoid Robotics Lab. The library provides data structures and
-    algorithms for kinematic and dynamic applications in robotics and computer
-    animation.
+    Lab and Humanoid Robotics Lab, with ongoing support by the University
+    of Washington Personal Robotics Lab and Open Robotics. DART provides data
+    structures and algorithms for kinematic and dynamic applications in robotics
+    and computer animation.
   </description>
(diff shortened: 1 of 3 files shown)


package.xml
```diff
@@ -3,24 +3,25 @@
...
+    of Washington Personal Robotics Lab and Open Robotics. DART provides data
+    structures and algorithms for kinematic and dynamic applications in robotics
+    and computer animation.
   </description>
   <url type="website">http://dartsim.github.io/</url>
   <url type="repository">https://github.com/dartsim/dart</url>
   <url type="bugtracker">https://github.com/dartsim/dart/issues</url>
-  <maintainer email="jslee02@gmail.com">Jeongseok Lee</maintainer>
+  <maintainer email="grey@openrobotics.org">Michael X. Grey</maintainer>
+  <author email="jslee02@gmail.com">Jeongseok Lee</maintainer>
```
jslee02: `</author>` at the end of the line.

jslee02:
How difficult (or easy) to upgrade the package in the same ROS version? Because I'm thinking of releasing DART 6.4 within like one month. If it's not straightforward to upgrade the package, I would suggest putting this into 6.4.

mxgrey:
Instructions for new releases can be found [here](http://wiki.ros.org/bloom/Tutorials/ReleaseCatkinPackage). Based on the instructions, it seems pretty straightforward. Most of the work appears to be automated.

mxgrey:
I incremented the patch version number in the `package.xml` and the `CMakeLists.txt`. If there's anywhere else that it needs to be adjusted, please let me know.

mxgrey:
I'll merge this in a few hours if no one has any immediate objections.

PR #1029, opened 2018-03-13
Fixing package name of FCL and removing libccd in package.xml

mxgrey:
After running some experiments with the ROS release tools, it seems they expect the package name for fcl to be `libfcl-dev`, and `libccd` doesn't seem to be recognized as a package (but we only depend on it transitively through `libfcl`, so it shouldn't matter).

This version of the package.xml seemed to work with `bloom-release`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Build system
 
-  * Tweaked package.xml for catkin support: [#1027](https://github.com/dartsim/dart/pull/1027)
+  * Tweaked package.xml for catkin support: [#1027](https://github.com/dartsim/dart/pull/1027) [#1029](https://github.com/dartsim/dart/pull/1029)
 
 ### [DART 6.3.0 (2017-10-04)](https://github.com/dartsim/dart/milestone/36?closed=1)
 
--- a/package.xml
+++ b/package.xml
@@ -25,9 +25,8 @@
   <depend>assimp</depend>
   <depend>bullet</depend>
   <depend>eigen</depend>
-  <depend>fcl</depend>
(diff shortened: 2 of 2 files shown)


mxgrey:
I'm going to say that the last "release" of 6.3.1 was a dud. I marked it as a prerelease anyhow, and it's deleted now since it didn't work.

mxgrey:
By the way, if deleting a release like that causes issues with any kind of automated release procedure that I don't know about, let me know and I'll refrain from doing that in the future.

jslee02:
We don't have any automated release procedure, but I _think_ it would be preferable to increase patch number with a fix if we already tagged to avoid any possible issues.

mxgrey:
I also deleted the tag already :sweat_smile:

mxgrey:
Does this plan sound okay:

To move to the next stage of trying to release DART into ROS, I'll create a new tagged release of 6.3.1 which is explicitly marked on GitHub as a prerelease. That should allow us to submit a PR to get DART released into ROS, and allow the ROS maintainers to test the release of 6.3.1.

If the tests fail on the ROS end (e.g. our `package.xml` is missing some important dependency), then we can make the necessary changes, delete the prerelease tag of 6.3.1, tag the new commit as v6.3.1, and ask the ROS maintainers to try it again.

Once v6.3.1 is working correctly on the ROS build farm, we'll promote it from a prerelease to an actual release.

I think this would be the easiest procedure for both us and the ROS maintainers, if it sounds okay to you.

jslee02:
That sounds perfect to me! Otherwise, we would end up increasing N patches whenever we find failures on the ROS end.

PR #1033, opened 2018-03-14
Final step to give dartsim stable status in ROS build farm

mxgrey:
Sorry for all the PR spam, but we're almost there! The last thing the ROS package build farm is complaining about is a cmake warning that we expect to occur. This results in an "unstable" status, which is acceptable, but I'd prefer we make it stable since it's such an easy fix (and this should probably be a status message instead of a warning anyway).

I also updated the changelog to include PR #1032 which I accidentally skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * ROS support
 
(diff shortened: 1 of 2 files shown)


mxgrey:
I kind of wish it were possible to rebase pull requests :sweat_smile:

jslee02:
@mxgrey It's enabled!

Issue #1005, opened 2018-02-22
Submit DART port to vcpkg


jslee02:
[As most of the dependencies of DART are added to vcpkg](https://github.com/dartsim/dart/issues/260), it would be good to submit DART port to [vcpkg](https://github.com/Microsoft/vcpkg) as well so that the users on Windows can easily install DART using the package manager.

smartcam3d:
I am trying to follow the instructions for the installation and the tutorials under windows MSVC 2017. I managed to install all the dependencies with vcpkg and to build the library. Unfortunatelly I did not manage to run the examples yet. After setting the compiler flag -permissive-, the tutorial code for the pendulum is compiled, but crashes on execution. I would appreciate more detailed instructions on how to use DART on windows and think that a vcpkg integration would be very helpful!

Fleischner:
Currently trying to do that. 
One problem is building as shared library fails. 

Dart does not mark its exports with __declspec(dllexport) as windows requires. I use the cmake option WINDOWS_EXPORT_ALL_SYMBOLS to generate an .def to work around that. However in this case the dart.dll has more than 65535 symbols. And this is not supported by msvc compiler (LNK1189) https://msdn.microsoft.com/en-us/library/aa234479(v=vs.60).aspx

Most of the symbols result from eigen3 header only library. Maybe upcoming solution I proposed at cmake: https://gitlab.kitware.com/cmake/cmake/issues/17841

jslee02:
@Fleischner Thanks for working on this.

A straightforward solution would be that DART marks `__declspec(dllexport)` using CMake's [GenerateExportHeader](https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html) like [FCL did](https://github.com/flexible-collision-library/fcl/pull/233/files). I can do this if this sounds good.

mxgrey:
I've been meaning to recommend that DART use `GenerateExportHeader`, so +1 to that.

jslee02:
@smartcam3d Let me try to reproduce the problem when I have a chance. Issue #1039 created to track the problem.

Fleischner:
I didn't know about GenerateExportHeader. This is of course a better solution.

mxgrey:
In the meantime, it should still be possible to use DART as a static library. Is vcpkg able to support libraries that can only be built as static libraries?

Fleischner:
Yes it is. It can even mix. 
[...]

PR #1040, opened 2018-03-22
C++ visibility support

jslee02:
This PR implements the [control of binary symbols visibility](https://gcc.gnu.org/wiki/Visibility) in DART inspired by https://github.com/flexible-collision-library/fcl/pull/233.

`DART_EXPORT` preprocessor is added to any class, struct, and function that needs to be exposed publicly to third-party software. I could compile all the tests, examples, and tutorials without any issue, but it would be good to test DART with other third-party software like AIKIDO for sure.

`DART_HIDE_ALL_SYMBOLS` CMake option is also added, which is `OFF` by default, for the case that we want to hide all the symbols as DART is built as static library.

Caveat:
There is a preprocessor name conflict between the existing `DART_DEPRECATED(version)` and another preprocessor generated by `generate_export_header()`, which is necessary to mark visibility of deprecated APIs. To resolve this, I changed the name to `DART_DEPRECATED2` so that the preprocessor for deprecated API becomes `DART_DEPRECATED2_EXPORT`. I'm open to better workaround.

Alternatively, we could remove `DART_DEPRECATED(version)` and use one generated by `generate_export_header()`. The version information can be provided by using Doxygen comment (e.g., `/// \deprecated Deprecated in <version>. Please use <other_api> instead).`).

This PR is related to #1005.

***

- [ ] Add comment to the export preprocessors


mxgrey:
This is a good start, but I have a few notes:

> I changed the name to DART_DEPRECATED2 so that the preprocessor for deprecated API becomes DART_DEPRECATED2_EXPORT.

What I've done in the past is:

* Put the auto-generated header into `detail/export.h` (e.g. `dart/detail/export.h`)
* Name the auto-generated macros as `DETAIL_<BASE>_<MACRO_NAME>`
* Manually write an `export.h` file (e.g. `dart/export.h`)
* Have the manually-written `export.h` file *use* the macros that were defined in `dart/detail/export.h`, e.g.:

```
#define DART_EXPORT \
    DETAIL_DART_EXPORT
#define DART_DEPRECATED(v) \
    DETAIL_DART_DEPRECATED
...
```

This also gives us the opportunity to add documentation to each macro, which may be helpful for end users.

I would suggest we consider naming the export macro as `DART_API` instead of the default `DART_EXPORT`, since I think calling it API makes it more explicit about its purpose. We can do that by adding `EXPORT_MACRO_NAME DART_API` to the `generate_export_headers` command.

And a final note: We need to create a unique export header for **each library** in the project, which means each of the optional components that aren't part of `libdart` need their own unique export header with their own unique export macro name.

mxgrey:
Revising my last comment: Since we would need a bunch of the "manually written" `export.h` with slight variations, we can create an `export.h.in` and configure it for each of the libraries.

jslee02:
Having another layer of export header sounds good to me.

> And a final note: We need to create a unique export header for each library in the project, which means each of the optional components that aren't part of libdart need their own unique export header with their own unique export macro name.

Could you let me know why do we need different configurations for each component?

mxgrey:
The definition of the export macro will be different depending on whether the header is being compiled by the library that is defining the class or by a library that is using the class.

To illustrate how things would go wrong, let's say we use the same export macro for `libdart-gui-osg` as `libdart`. Then when we compile `libdart-gui-osg` and the compiler comes across the declaration of `Skeleton` in `dart/dynamics/Skeleton.hpp`, then the compiler will think that `libdart-gui-osg` wants to define and export the symbol of `dart::dynamics::Skeleton`. This will cause undefined reference and/or other linker errors, since `libdart-gui-osg` does not have access to the definition of `dart::dynamics::Skeleton`.

jslee02:
@mxgrey That makes sense to me. I have a follow-up question though.

Although `gui-osg` component uses different preprocessors (e.g., `DART_GUI_OSG_API`), I think the component still will try to export the symbol of `dart::dynamics::Skeleton` because `Skeleton.hpp` includes `dart/export.hpp` so that `DART_API` is defined to export the symbol. How could we prevent this situation? Should we redefine all the export preprocessor except the preprocessors of the component not to export symbols?

mxgrey:
> I think the component still will try to export the symbol of dart::dynamics::Skeleton because Skeleton.hpp includes dart/export.hpp so that DART_API is defined to export the symbol.

This is not correct.

When you use the `generate_export_header` function from cmake, it will add a private compile definition (something along the lines of `dart_EXPORTS`) to the library for which it's generated (this is why the first argument of the function needs to be the library target for which you're generating the export header). That compile definition, which is *only* given to `libdart` during compilation, will make it so that the `DART_API` macro expands to `__declspec(dllexport)` (or whatever the platform-specific export attribute is).

When `libdart-gui-osg` is compiled, it will **not** be given the `dart_EXPORTS` compile definition, so the `DART_API` macro will expand to `__declspec(dllimport)` (or whatever the platform-specific import attribute is).

mxgrey:
So the short version is, cmake takes care of your concern automatically.

jslee02:
@mxgrey Thanks for the explanation! I learned a lot while working on this PR.

I've addressed your feedback in the last commits.

`test_Collision` fails due to a reason that is not relevant to the changes of this PR. In a nutshell, the current implementation of `Singleton` doesn't work for multiple binary modules (related [SO post](https://stackoverflow.com/questions/32883872/implementing-singletons-across-binary-modules)). I'll handle this in a future PR.

cmake/DARTMacros.cmake
```diff
@@ -170,3 +170,82 @@ function(dart_format_add)
...
+    EXPORT_FILE_NAME detail/export.h
+    DEPRECATED_MACRO_NAME ${base_name_upper}_DETAIL_DEPRECATED
+  )
+  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/detail/export.h
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${component_path}/detail
+    COMPONENT headers
+  )
+
+  # Generate final export header
+  set(header_guard_name ${base_name_upper}_EXPORT_HPP_)
```
mxgrey: This would probably be much cleaner if we created a `export.hh.in` and then called `configure_file` on it.

cmake/DARTMacros.cmake
```diff
@@ -170,3 +170,82 @@ function(dart_format_add)
...
+    set(component_path "dart/${_geh_COMPONENT_PATH}")
+  endif()
+  set(component_detail_export_path "${component_path}/detail/export.h")
+
+  # Generate CMake's default export header
+  include(GenerateExportHeader)
+  generate_export_header(${target_name}
+    EXPORT_MACRO_NAME ${base_name_upper}_DETAIL_API
+    EXPORT_FILE_NAME detail/export.h
+    DEPRECATED_MACRO_NAME ${base_name_upper}_DETAIL_DEPRECATED
```
mxgrey: Nitpick: I like to put the `DETAIL` in front of `DART` so it gets excluded from autocomplete when users start typing `DART...`.

cmake/DARTMacros.cmake
jslee02: Agree. I wrote this lines because I thought each component requires individual `export.hpp.in`, but this is not true.

cmake/DARTMacros.cmake
jslee02: That's true, but I thought we use `DART_` prefix to avoid possible name conflicts. At the same time, I don't see the possibility of name confliction with `DETAIL_DART_~`.

Let me change it to `DETAIL_DART_~` as I don't have a strong preference here.

cmake/DARTMacros.cmake
jslee02: Agree. I wrote this lines because I thought each component requires individual `export.hpp.in`, but this is not true.

jslee02:
@mxgrey I've fixed build and test errors. Could you do a second review?

dart/common/Deprecated.hpp
```diff
@@ -42,7 +43,7 @@
 //==============================================================================
 
 #if defined(__GNUC__) || defined(__clang__)
-  #define DART_DEPRECATED(version) __attribute__ ((deprecated))
+  #define DART_DEPRECATED(version) DETAIL_DART_DEPRECATED
   #define DART_FORCEINLINE __attribute__((always_inline))
 #elif defined(_MSC_VER)
   #define DART_DEPRECATED(version) __declspec(deprecated)
```
mxgrey: This should also be redirected to `DETAIL_DART_DEPRECATED` to be consistent with the other definition.

mxgrey:
It looks like the lines that I marked with `Example` didn't make it into the review for some reason. In any case, template definitions shouldn't be given the export macro.

jslee02:
I haven't fully investigated but without the export macro to the template classes caused test failures of `test_Collision` and `test_Skeleton` for different reasons.

`test_Collision` failed due to creating different static instances for collision factory; FCL and DART collision detectors are registered to one factory, but Bullet and FCL are registered to each different factory.

`test_Skeleton` failed due to that null is returned from [here](https://github.com/dartsim/dart/blob/master/unittests/comprehensive/test_Skeleton.cpp#L1280).

mxgrey:
> `test_Collision` failed due to creating different static instances for collision factory

I think that's because [this class](https://github.com/dartsim/dart/blob/build/visibility/dart/collision/CollisionDetector.hpp#L65) is never marked as exported. I'm not sure what the correct way to export that would be. Maybe we need to explicitly instantiate the `SingletonFactory` class. In this specific case, we could decide that we do not want users to be able to use our `Singleton` template for any of their own template arguments, but we can't apply that reasoning to all of our templates, or else DART won't be extensible.

> `test_Skeleton` failed due to that null is returned from here.

That might be a trickier issue. It's not immediately clear to me why that would happen.

mxgrey:
Yes, I think that would also be dangerous, because the definition is actually given in [the implementation header](https://github.com/dartsim/dart/blob/build/visibility/dart/common/detail/Singleton-impl.hpp#L44).

My understanding of the export macro is that it effectively tells the user's compiler "Don't try to create a definition for this symbol, because the definition will be provided by the library that exported this header", and then it tells the user's linker "Don't look for a definition of this symbol within the library that we just compiled; instead, look for it in one of the libraries that we are linking to".

I think if we "export" that static member, then users will get a linking error for it whenever they pass a custom class type into `Singleton<T>`.

PR #1044, opened 2018-03-27
Update changelog for releasing 6.4.0

jslee02:


mxgrey:
Have we merged 6.3.1 into 6.4 yet? We might want to do that before making a 6.4.0 release.

jslee02:
Good catch! I forgot there was another change after I merged `release-6.3`. Just merged it again.

PR #1048, opened 2018-04-01
Add FOV API to OSG viewer

jslee02:
***

### DIFF
--- a/dart/gui/osg/Viewer.cpp
+++ b/dart/gui/osg/Viewer.cpp
@@ -855,7 +855,7 @@ void Viewer::setVerticalFieldOfView(double fov)
 
   if (!camera)
   {
-    dtwarn << "[Viewer::getMasterCameraFieldOfView] This viewer doesn't have "
+    dtwarn << "[Viewer::setMasterCameraFieldOfView] This viewer doesn't have "
            << "any cameras. Ignoring this request.\n";
     return;
   }
@@ -865,7 +865,7 @@ void Viewer::setVerticalFieldOfView(double fov)
 
   if (!result)
   {
-    dtwarn << "[Viewer::getMasterCameraFieldOfView] Attemping to set vertical "
+    dtwarn << "[Viewer::setMasterCameraFieldOfView] Attemping to set vertical "
            << "field of view while the camera isn't perspective view. "
            << "Ignoring this request.\n";
     return;
--- a/dart/gui/osg/Viewer.cpp
+++ b/dart/gui/osg/Viewer.cpp
@@ -844,7 +844,7 @@ const ::osg::ref_ptr<::osg::Group>& Viewer::getRootGroup() const
 }
 
 //==============================================================================
-bool Viewer::setMasterCameraFieldOfView(double fov)
+void Viewer::setVerticalFieldOfView(double fov)
 {
   double fovy;
   double aspectRatio;
@@ -856,14 +856,18 @@ bool Viewer::setMasterCameraFieldOfView(double fov)
   const bool result = camera->getProjectionMatrixAsPerspective(
       fovy, aspectRatio, zNear, zFar);
 
-  if (result)
-    camera->setProjectionMatrixAsPerspective(fov, aspectRatio, zNear, zFar);
+  if (!result)
+  {
+    dtwarn << "[Viewer::getMasterCameraFieldOfView] Attemping to set vertical "
+           << "field of view while the camera isn't perspective view. "
+           << "Ignoring this request.\n";
+  }
 
-  return result;
+  camera->setProjectionMatrixAsPerspective(fov, aspectRatio, zNear, zFar);
 }
 
 //==============================================================================
-double Viewer::getMasterCameraFieldOfView() const
+double Viewer::getVerticalFieldOfView() const
 {
   double fovy;
   double aspectRatio;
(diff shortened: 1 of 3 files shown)


dart/gui/osg/Viewer.cpp
```diff
@@ -843,6 +843,53 @@ const ::osg::ref_ptr<::osg::Group>& Viewer::getRootGroup() const
...
+  auto* camera = getCamera();
+  assert(camera);
+  const bool result = camera->getProjectionMatrixAsPerspective(
+      fovy, aspectRatio, zNear, zFar);
+
+  if (!result)
+  {
+    dtwarn << "[Viewer::getMasterCameraFieldOfView] Attemping to set vertical "
+           << "field of view while the camera isn't perspective view. "
+           << "Ignoring this request.\n";
```
mxgrey: If we say that we're ignoring the request, shouldn't we return here?

dart/gui/osg/Viewer.hpp
```diff
@@ -293,6 +293,15 @@ class Viewer : public osgViewer::Viewer, public dart::common::Subject
   /// Get the root ::osg::Group of this Viewer
   const ::osg::ref_ptr<::osg::Group>& getRootGroup() const;
 
+  /// Sets the vertical field of view of the master camera of the view.
+  /// \param[in] fov Vertical field of view in degree.
+  void setVerticalFieldOfView(double fov);
+
+  /// Retruns the vertical field of view of the master camera of the view.
```
mxgrey: `Retruns` --> `Returns`

dart/gui/osg/Viewer.hpp
```diff
@@ -293,6 +293,15 @@ class Viewer : public osgViewer::Viewer, public dart::common::Subject
   /// Get the root ::osg::Group of this Viewer
   const ::osg::ref_ptr<::osg::Group>& getRootGroup() const;
 
+  /// Sets the vertical field of view of the master camera of the view.
+  /// \param[in] fov Vertical field of view in degree.
```
mxgrey: Nitpick: `degree` --> `degrees`

dart/gui/osg/Viewer.hpp
```diff
@@ -293,6 +293,15 @@ class Viewer : public osgViewer::Viewer, public dart::common::Subject
...
+  void setVerticalFieldOfView(double fov);
+
+  /// Retruns the vertical field of view of the master camera of the view.
+  /// \return Vertical field of view in degree if the camera is perspective
```
mxgrey: Nitpick: `degree` --> `degrees`

dart/gui/osg/Viewer.cpp
```diff
@@ -843,6 +843,53 @@ const ::osg::ref_ptr<::osg::Group>& Viewer::getRootGroup() const
...
+//==============================================================================
+double Viewer::getVerticalFieldOfView() const
+{
+  double fovy;
+  double aspectRatio;
+  double zNear;
+  double zFar;
+
+  const auto* camera = getCamera();
+  assert(camera);
```
mxgrey: Does this mean we always assume the return value of `getCamera()` should always be non-null? Is it feasible that it will ever be `nullptr` (e.g. if this `Viewer` doesn't have any cameras)? I'm not finding anything in the OSG docs that indicate what we should expect.

dart/gui/osg/Viewer.cpp
```diff
@@ -843,6 +843,53 @@ const ::osg::ref_ptr<::osg::Group>& Viewer::getRootGroup() const
   return mRootGroup;
 }
 
+//==============================================================================
+void Viewer::setVerticalFieldOfView(double fov)
```
mxgrey: Nitpick: `const double`

dart/gui/osg/Viewer.cpp
jslee02: Yeah, I also couldn't find related docs from OSG. I think it would be safer we just ignore when `getCamera()` returns null. Let me change this code so. Let me know what you think.

dart/gui/osg/Viewer.cpp
jslee02: Hm, I see the reason for this suggestion; I found this [SO post](https://stackoverflow.com/a/16033837/3122234) says specifying `const` modifier helps the compiler with optimizations. However, AFAIK, we haven't used `const` when we pass value parameters in DART codebase. Do you suggest we follow this convention from now on?

mxgrey: I recommend using `const` for every variable which isn't intentionally mutable. Besides compiler optimizations, it also catches bugs where a variable might get unintentionally modified.

Note that you don't need the `const` in the declaration of the function. E.g.:

```
// in foo.hpp
void foo(int value);

// in foo.cpp
void foo(const int value)
{
  /* ... do things ... */
}
```

works perfectly fine, because the function signature is the same, whether or not a value parameter is const-qualified. On the other hand, function signatures are different between const-references and mutable lvalue references, so this would not work:

```
// in foo.hpp
void foo(int& value);

// in foo.cpp
void foo(const int& value)
{
  /* ... do things ... */
}
```

mxgrey: To illustrate a potential bug, we might accidentally type

```
const bool result = camera->getProjectionMatrixAsPerspective(
      fov, aspectRatio, zNear, zFar);
```

instead of

```

const bool result = camera->getProjectionMatrixAsPerspective(
      fovy, aspectRatio, zNear, zFar);
```

which would accidentally overwrite the value that the caller passed to the function.

jslee02: Thanks for the examples. I see the point of preventing a potential bug, but I have a little concern with that their function signatures are not "exactly" the same. Wouldn't it be more clear to just put `const` to both of declaration and definition in this case?

mxgrey: Just to be clear, the function signatures, in the technical sense, *are* **exactly** the same. A human might see that the text is different between the declaration and the definition, but the compiler/linker does not care, and the ABI signature will be identical.

There's nothing inherently wrong with putting the `const` in the declaration as well, although it's often discouraged because it's seen as unnecessary noise which a user has no reason to care about (it makes no difference whatsoever to a user whether a value parameter of a function is const-qualified or not, because it has no impact on how they use the function or how the function behaves).

PR #1050, opened 2018-04-05
Add World::hasSkeleton()

jslee02:
`World::hasSkeleton` takes `Skeleton` as a shared pointer since `Skeleton` is enforced to be created as a shared pointer by calling `Skeleton::create()`.

***

### DIFF
--- a/dart/simulation/World.hpp
+++ b/dart/simulation/World.hpp
@@ -149,6 +149,9 @@ public:
   /// pointers to them, in case you want to recycle them
   std::set<dynamics::SkeletonPtr> removeAllSkeletons();
 
+  /// Returns wether this World contains a Skeleton.
+  bool hasSkeleton(const dynamics::SkeletonPtr& skeleton);
+
   /// Get the dof index for the indexed skeleton
   int getIndex(int _index) const;
 
--- a/dart/simulation/World.hpp
+++ b/dart/simulation/World.hpp
@@ -150,7 +150,7 @@ public:
   std::set<dynamics::SkeletonPtr> removeAllSkeletons();
 
   /// Returns wether this World contains a Skeleton.
-  bool hasSkeleton(const dynamics::SkeletonPtr& skeleton);
+  bool hasSkeleton(const dynamics::ConstSkeletonPtr& skeleton) const;
 
   /// Get the dof index for the indexed skeleton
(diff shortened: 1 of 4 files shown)


dart/simulation/World.hpp
```diff
@@ -149,6 +149,9 @@ class World : public virtual common::Subject
   /// pointers to them, in case you want to recycle them
   std::set<dynamics::SkeletonPtr> removeAllSkeletons();
 
+  /// Returns wether this World contains a Skeleton.
+  bool hasSkeleton(const dynamics::SkeletonPtr& skeleton);
```
mxgrey: This member function should be const-qualified.

mxgrey: Also, if possible, it would be nice to accept a `const dynamics::ConstSkeletonPtr& skeleton)`, but I know that sometimes the STL doesn't like to perform comparisons between `std::shared_ptr<const T>` and `std::shared_ptr<T>` objects, so maybe it would be more trouble than it's worth.

jslee02: Definitely!

jslee02: Hm, would it be better to take just a const raw pointer as a workaround? Or having both for `SkeletonPtr` and `ConstSkeletonPtr`?

mxgrey: The compiler can implicitly cast a `SkeletonPtr` to a `ConstSkeletonPtr`, no problem. I'm just not sure off the top of my head if

```
return std::find(mSkeletons.begin(), mSkeletons.end(), skeleton)
      != mSkeletons.end();
```

will compile when `skeleton` is a `ConstSkeletonPtr`, since the compiler might want to cast `skeleton` from `ConstSkeletonPtr` to `SkeletonPtr`, which is not allowed.

I think if the compiler gives us any trouble with the `ConstSkeletonPtr` argument, we could use `std::find_if` and pass in our own unary predicate that casts each entry to a `ConstSkeletonPtr` before comparing it to `skeleton`.

jslee02: It seems `ConstSkeletonPtr` works without any troubles in the simple test. Let's merge this as it is and revisit if it turns out this is problematic. :grin:

jslee02:
Great! Thanks!

Issue #1060, opened 2018-04-30
CPP 11 flag missing for the "ASSIMP_AISCENE_CTOR_DTOR_DEFINED" cmake test.


Aneoshun:
### Bug Report
- [X] I checked the [documentation](http://dartsim.github.io/) and found no answer.
- [X] I checked to make sure that this issue has not already been filed.

#### Environment
*  DART version: master 
* OS name and version name(or number): Ubuntu 16.04
* compiler name and version number: GCC 5.4

Hi all, 
I am trying to use Dart with the latest version of assimp (4.1, from source) to see if it solves a memory leak. 
Dart does compile, but leads to a segfault when loading an urdf. 
After some research, I found that it was because the "ASSIMP_AISCENE_CTOR_DTOR_DEFINED" test during the configuration failed (see #451 ). This test check if the constructor and destuctor of aiScene are exported. 

However, the logs say that the compilation test fails for a totally different reason: `error: ‘nullptr’ was not declared in this scope`.  Adding the flag ` -std=c++11` in  DARTFindDependencies.cmake solves this issue, enables the   "ASSIMP_AISCENE_CTOR_DTOR_DEFINED" test to pass and dart to run without segfault. 

I hope this issue will be useful for you, 

Best,

Aneoshun:
Thanks!

PR #1061, opened 2018-04-30
Enable C++11 for Assimp 3.3.0 or greator

jslee02:
The Assimp constructor/destructor checks in `DARTFindDependencies.cmake` always fail due to lack of C++11 flags for Assimp 3.3.0 or greator. [Assimp requires C++11 since 3.3.0](https://github.com/assimp/assimp/blob/master/CHANGES#L167).

This PR adds `-std=c++11` flags for Assimp 3.3.0 or greator to the checks.

This is reported in #1060 thanks to @Aneoshun.

***


cmake/DARTFindDependencies.cmake
```diff
@@ -28,7 +28,11 @@ if(ASSIMP_FOUND)
   # Check for missing symbols in ASSIMP (see #451)
   include(CheckCXXSourceCompiles)
   set(CMAKE_REQUIRED_DEFINITIONS "")
-  set(CMAKE_REQUIRED_FLAGS "")
+  if (NOT ASSIMP_VERSION VERSION_LESS 3.3.0)
```
mxgrey: We could have `AND NOT MSVC` here, because MSVC does not know the `-std=c++11` flag.

Then again, even if we pass the flag in for MSVC, I think the compiler will just ignore it and proceed as normal, so it doesn't really matter.

jslee02: Good catch! Yeah, it would work with the flag, but suppressing an unnecessary warning is worth enough. :smile:

PR #1064, opened 2018-05-01
Bugfix/urdfparser

aditya-vk:
Continuous Joints have been previously parsed with joint limits set to 0 resulting in incorrect behavior of the joint and relevant functions [For instance, `isCyclic()`].

This PR corrects the parsing for these joints by setting the joint limits to negative/positive infinity.

To show the behavior:
Commit 456cca0 has a test to show that current implementation has a bug.
Commit 5eff95d provides a fix.
Commit 9d8d316 has the test to show the fix works.

***

### DIFF
--- a/dart/utils/urdf/DartLoader.cpp
+++ b/dart/utils/urdf/DartLoader.cpp
@@ -377,8 +377,8 @@ dynamics::BodyNode* DartLoader::createDartJointAndNode(
     case urdf::Joint::CONTINUOUS:
     {
       // We overwrite joint position limits to negative/positive infinities
-      // for "continuous" joint. The URDF parser, by default, either reads the
-      // limits, if specified for this joint, or sets them to 0.
+      // for "continuous" joint. The URDF parser, by default, either reads
+      // the limits, if specified for this joint, or sets them to 0.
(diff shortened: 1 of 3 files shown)


dart/utils/urdf/DartLoader.cpp
```diff
@@ -374,8 +374,12 @@ dynamics::BodyNode* DartLoader::createDartJointAndNode(
   std::pair<dynamics::Joint*, dynamics::BodyNode*> pair;
   switch(_jt->type)
   {
-    case urdf::Joint::REVOLUTE:
     case urdf::Joint::CONTINUOUS:
+    {
+      singleDof.mPositionLowerLimits[0] = -math::constantsd::inf();
```
jslee02: Nit: Could we comment the reason of this overwriting something like:
```cpp
// We overwrite joint position limits to negative/positive infinities
// for "continuous" joint. This is because the URDF parser ignorantly
// reads joint limits or defaults to zero, which always should be
// infinity for "continuous" joint.
```

aditya-vk:
Uh oh. Sorry for dismissing reviews with minor updates. Updated `CHANGELOG.md` and made the test scripts more compact. Should be good to go unless there are issues. :)

PR #1076, opened 2018-07-03
Add VoxelGridShape

jslee02:
This PR adds `VoxelGridShape` class and its collision detection feature using fcl. `VoxelGridShape` uses [`octomap`](https://github.com/OctoMap/octomap) for the underlying voxel grid representation.

This is a minimal implementation for rapid test for voxel grid collision detection. There are many rooms for improvements, which will be handled subsequent PRs in the future.

***

- [x] ~Fix test failure on macOS~ (disabled for Octomap (>= 1.9.0))


jslee02:
@mxgrey I agree with you. As you mentioned, making the voxel shape as an optional component requires many changes, which I also think it would be better to be in DART 7.

To reduce the impact in dependencies and API changes, I added OctoMap dependency optional so that `VoxelGridShape` and its collision detection functionality are built only when OctoMap is detected.

Let me create issue(s) for your suggestions.

jslee02:
The macOS failure is due to that Octomap 1.9.0 returns different results from Octomap (>= 1.8.1). I'll investigate on that but not in this PR. In the meantime, the support of Octomap (>= 1.9.0) will be disabled.

PR #1088, opened 2018-07-19
Reorganize glut (preparation for glfw)

jslee02:
This PR reorganizes glut. This is a preparation for having glfw, which will eventually replace glut since glut is deprecated.

This PR shouldn't hurt any existing code except it will generate warnings if glut is included by old files.

glfw support will be followed soon.

***

### DIFF
--- a/dart/gui/CMakeLists.txt
+++ b/dart/gui/CMakeLists.txt
@@ -78,6 +78,20 @@ add_subdirectory(osg)
 
 # Generate header for this namespace
 dart_get_filename_components(header_names "gui headers" ${hdrs})
+
+# Remove deprecated files from the list
+list(REMOVE_ITEM header_names
+  "GLFuncs.hpp"
+  "GlutWindow.hpp"
+  "GraphWindow.hpp"
+  "LoadGlut.hpp"
+  "MotionBlurSimWindow.hpp"
+  "SimWindow.hpp"
+  "SoftSimWindow.hpp"
+  "Win2D.hpp"
+  "Win3D.hpp"
+)
+
 set(
   header_names
   ${header_names}
--- a/dart/gui/CMakeLists.txt
+++ b/dart/gui/CMakeLists.txt
@@ -21,15 +21,31 @@ else()
 endif()
 
 # Search all header and source files
-file(GLOB hdrs "*.hpp")
-file(GLOB srcs "*.cpp")
+file(GLOB hdrs "*.hpp" "*.h" "detail/*.hpp")
+file(GLOB srcs "*.cpp" "*.c" "detail/*.cpp")
+
+function(dart_add_gui_headers)
+  dart_property_add(DART_GUI_HEADERS ${ARGN})
+endfunction()
+
+function(dart_add_gui_sources)
+  dart_property_add(DART_GUI_SOURCES ${ARGN})
+endfunction()
+
+# Add required subdirectory
+add_subdirectory(glut)
+
+get_property(dart_gui_headers GLOBAL PROPERTY DART_GUI_HEADERS)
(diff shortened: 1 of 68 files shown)


dart/gui/GlutWindow.hpp
```diff
@@ -33,67 +33,17 @@
 #ifndef DART_GUI_GLUTWINDOW_HPP_
 #define DART_GUI_GLUTWINDOW_HPP_
 
-#include <vector>
+#include "dart/gui/glut/GlutWindow.hpp"
 
-#include "dart/gui/LoadOpengl.hpp"
-#include "dart/gui/RenderInterface.hpp"
+#warning "This file is deprecated in DART 6.6. "\
```
mxgrey: It seems this method for warning [is not portable](https://stackoverflow.com/a/279669).

We might want to add a `DART_PREPROCESSOR_WARNING(X)` macro to `dart/common/Deprecated.hpp` in order to have a cross-platform method for this.

dart/gui/GraphWindow.hpp
```diff
@@ -39,35 +39,12 @@
...
 
-#include <vector>
-
-#include <Eigen/Dense>
-
-#include "dart/gui/Win2D.hpp"
+#include "dart/gui/glut/GraphWindow.hpp"
 
 namespace dart {
 namespace gui {
```
mxgrey: We might want to add a `DART_PREPROCESSOR_WARNING(X)` to this header.

dart/gui/glut/GlutWindow.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+#include "dart/gui/LoadOpengl.hpp"
+#include "dart/gui/RenderInterface.hpp"
+
+namespace dart {
+namespace gui {
+namespace glut {
+
+/// \brief
+class Window {
```
mxgrey: Since the class name has changed to `dart::gui::glut::Window`, we should consider changing the filename to `dart/gui/glut/Window.hpp` instead of `../GlutWindow.hpp`.

dart/gui/GlutWindow.hpp
mxgrey: After some more thought, a preprocessor directive can't replaced with a macro because of the limitations of the preprocessor.

However, it seems recent versions of GCC and Clang can understand `#pragma message`, so we can use that instead.

jslee02:
@mxgrey Thanks for the changes! Merging now.

dart/gui/CMakeLists.txt
```diff
@@ -62,6 +78,25 @@ add_subdirectory(osg)
 
 # Generate header for this namespace
 dart_get_filename_components(header_names "gui headers" ${hdrs})
+
+# Remove deprecated files from the list
+list(REMOVE_ITEM header_names
```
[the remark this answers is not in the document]
jslee02: > I think it's inadvertently breaking the API.

This is a valid point.

> We might want to consider allowing these to remain in the master header, even if it shoots compiler messages at people who are using the master `gui` header.

I have a concern that the user will always see the warning messages by just including the master `gui` header. I can think of two possible solutions to this:

1. Remove header warnings completely
Easiest solution. The deprecated header will be removed in the next major release. The downside of this method is obviously that the user won't get a prior notice the header will be removed even when the header is included explicitly (not by the master header).

2. Suppress header warning when it's included by a master header
This would require some work for us. We could use a processor of whether the deprecated header is included by the master header, and suppress the warning if so. One caveat of this is that it's possible to unintentionally suppress the warning if a deprecated header is firstly included by the master header and then included explicitly. This is because the second header inclusion would be just ignored by the header guard. One workaround would be putting the warning out side of the header guard, but this would pollute the output window too much.

Maybe there is a much cleaner way, which I cannot think of now. Let me know.

PR #1106, opened 2018-08-28
Add multi-objective optimization API with Pagmo support

jslee02:
This PR is an initial effort for having a pipeline for multi-objective optimization. 

**Summary of Main Changes**

* `MultiObjectiveProblem`: Class to define a multi-objective problem. You can compose multiple objectives, equality constraints, and inequality constraints by adding `Function`s to this class.
* `Population`: Container of populations and their fitness. Pagmo2 inspired the implementation.
* `MultiObjectiveSolver`: Base class for concrete solvers.
* `PagmoMultiObjectiveSolver`: The first concrete solver that depends on [pagmo2](https://github.com/esa/pagmo2), which is a population-based optimization library for single/multi-objective problems.

The implementation is based on other project collaborated with Daqing Yi(@dqyi11).

***


dart/optimizer/pagmo/PagmoUtils.hpp
```diff
@@ -0,0 +1,61 @@
...
+{
+public:
+  static std::vector<double> convertVector(const Eigen::VectorXd& v);
+
+  static Eigen::Map<const Eigen::VectorXd> convertVector(
+      const std::vector<double>& v);
+
+  static Population convertPopulation(
+      const ::pagmo::population& pagmoPop,
+      std::shared_ptr<MultiObjectiveProblem> problem);
```
mxgrey: Nitpick: use a `const std::shared_ptr<T>&` here?

jslee02: I intentionally left this as passing by value in order to indicate that this function sharing the ownership because the caller can notice that the reference counting would be increase by the copy. Note that the reference counting would happen only one time because we move the copied `shared_ptr` in the function.

As a note, we would be good to have a documentation on this convention like [this](https://github.com/personalrobotics/aikido/blob/master/STYLE.md#smart-pointers).

dart/math/Helpers.hpp
```diff
@@ -286,6 +287,57 @@ inline Eigen::VectorXd randomVectorXd(std::size_t size, double limit)
...
+void push_back(Derived& m, T&& values)
+{
+  static_assert(Derived::ColsAtCompileTime == 1 or Derived::RowsAtCompileTime == 1, "push_back() can be only used with vector");
+  m.conservativeResize(m.size() + 1);
+  m[m.size() - 1] = values;
+}
+
+//==============================================================================
+template <typename Derived>
+void push_right(Derived& m, Eigen::Vector3d&& values)
```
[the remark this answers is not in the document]
jslee02: All your points are correct. This function was used during the implementation but it turned out not for the final version. Let me remove this and reintroduce with some corrections later if necessary.

dart/common/Console.hpp
```diff
@@ -60,7 +60,16 @@ std::ostream& colorErr(const std::string& _msg,
                        unsigned int _line,
                        int _color);
 
+
 }  // namespace common
+
+template <class T>
+auto operator<<(std::ostream& os, const T& t) -> decltype(t.print(os), os)
```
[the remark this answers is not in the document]
jslee02: Exactly!

dart/optimizer/pagmo/PagmoUtils.hpp
mxgrey: Ah, okay. That's a subtle but reasonable signal to the user :+1:

PR #1109, opened 2018-09-07
Consolidate random functions into Random class

jslee02:
This implementation is inspired by [`ign-math/Rand`](https://bitbucket.org/ignitionrobotics/ign-math/src/default/include/ignition/math/Rand.hh) with some extensions for `Eigen` objects.

The old-style random functions are deprecated.

The list of functions is not complete yet. Missing functions will be added later such as generating vector/matrix from normal distribution and multivariate normal distribution.

This is a prequel of #1106.

***


dart/math/Random.cpp
```diff
@@ -0,0 +1,76 @@
...
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dart/math/Random.hpp"
+
+namespace dart {
+namespace math {
+
+//==============================================================================
+Random::GeneratorType& Random::getRandGenerator()
```
prasbg: Why is this `getRandGenerator()` instead of just `getGenerator()`?  It is already on a `Random` class and it doesn't seem like there is another type of generator in this context to be worried about :sweat_smile:

psigen: Why is this `getRandGenerator()` instead of just `getGenerator()`?  It is already on a `Random` class and it doesn't seem like there is another generator in this context to worry about :sweat_smile:

dart/math/Random.hpp
```diff
@@ -0,0 +1,225 @@
...
+#define DART_MATH_RANDOM_HPP_
+
+#include <random>
+
+#include <Eigen/Core>
+
+namespace dart {
+namespace math {
+
+class Random final
```
psigen: If this is a final class which consists entirely of static methods, would it make more sense to make this a normal class and then have a singleton available here for simpler use cases?

That seems like it would be more flexible for situations where someone might want to use this helpful class but also track multiple RNGs with different seeds, where they can instantiate two of the class, while the singleton can be available for cases where that is not necessary.

dart/math/Random.cpp
jslee02: Good point!

dart/math/Random.hpp
jslee02: This sounds good to me. In any case, I would like to keep the API simple. We could introduce wrapper functions something like this:
```cpp
dart::math::uniform(min, max) {
  return Random::getInstance()::uniform(min, max);
}
```

Let me create an issue for this to track.

PR #1115, opened 2018-10-10
Fix for aligned in newer compilers/ubuntu 18.04

costashatz:
In a clean (X)Ubuntu 18.04 installation, when compiling with SIMD option (native flags) and gcc (version=7.3.0) or clang (version=6.0.0), some tests (and examples) fail because of segfaults.

The flag `-faligned-new` solves the issue (alternatively, one could enable C++17, but this flag enables just the right feature for the job). This also happened with clang. I do not know (and I cannot test) what happens for Visual Studio, so I left it untouched.

Let me know if you can reproduce the error and if there's another way of making it work.

### DIFF
--- a/dart/CMakeLists.txt
+++ b/dart/CMakeLists.txt
@@ -146,9 +146,9 @@ if(DART_ENABLE_SIMD)
     # in CMake time, then I would gratefully apply these options differently
     # depending on the architectures.
   elseif(CMAKE_COMPILER_IS_GNUCXX)
-    target_compile_options(dart PUBLIC -march=native)
+    target_compile_options(dart PUBLIC -march=native -faligned-new)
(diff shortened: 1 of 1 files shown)


jslee02:
Thanks for the report! Could you rebase this change to `release-6.7` so that it can be applied to DART 6.7 and later? `master` is for DART 7.0, and the release of DART 7.0 would take longer time than 6.7.

[The flag is introduced in GCC 7.1](https://www.gnu.org/software/gcc/gcc-7/changes.html) so I was concerned about that old GCC would complain about the unknown flag. But, it seems they silently ignore the unknown flag according to the CI builds. So I think it's okay to just enable the flag for all the GCC and Clang versions. Let me know if I missing something.

PR #1173, opened 2018-10-14
Remove deleted default constructors of AspectWith[StateAnd]VersionedProperties

jslee02:
The default constructors of `AspectWith[StateAnd]VersionedProperties` are deleted, but the other constructors of them actually allow to create them without parameters because of their default parameter values. This leads to a compilation error of an ambiguous function call when creating them without parameters. This PR removes the delete declarations of the default constructors.


mxgrey:
Whoops, that's an interesting conflict! I guess at some point I changed my mind about whether these classes should be able to default-construct, but didn't notice that the ordinary default constructor was still marked as deleted.

PR #1174, opened 2018-10-16
Fix compilation on Ubuntu 18.04, g++8

mxgrey:
It seems that there are compilation issues with recent versions of GCC (and maybe Clang) when compiling with the setting that warnings are errors, because some new warnings have been introduced.

This PR is a quick and dirty attempt to just make sure dartsim can compile with these new versions. There are two changes:

1. Disable warnings for imgui, which is an external dependency that we've copy/dumped into our codebase. If the warning is something worth fixing, then we should fix it upstream before copying it into here, instead of just fixing it locally.

2. There's a new warning for [implicit fallthroughs](https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/) so I've just added a break (which probably should've been there to begin with :sweat_smile:, so good job new warning!)

***

### DIFF
--- a/dart/external/imgui/CMakeLists.txt
+++ b/dart/external/imgui/CMakeLists.txt
@@ -12,6 +12,10 @@ include_directories(
   PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}"
 )
 
+if(CMAKE_COMPILER_IS_GNUCXX)
+  target_compile_options(dart-external-imgui PRIVATE -w)
+endif()
+
 # Component
 add_component(${PROJECT_NAME} ${component_name})
 add_component_targets(${PROJECT_NAME} ${component_name} ${target_name})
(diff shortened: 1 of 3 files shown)


dart/external/imgui/CMakeLists.txt
```diff
@@ -12,6 +12,10 @@ include_directories(
   PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}"
 )
 
+if(CMAKE_COMPILER_IS_GNUCXX)
+  target_compile_options(dart-external-imgui PRIVATE -w)
```
jslee02: Nit: Let's use the target name variable (i.e. `${target_name}`) instead of `dart-external-imgui`.

mxgrey:
Closing in favor of #1177

Issue #1184, opened 2018-10-30
Inaccurate contact results from Bullet collision detector for falling objects


mxgrey:
---

### Bug Report
> Please answer the following questions for yourself before submitting an issue.
- [x] I checked the [documentation](http://dartsim.github.io/) and found no answer.
- [x] I checked to make sure that this issue has not already been filed.

#### Environment
* Select DART version: 6.7
* Select OS name and version name(or number): Ubuntu 18.04
* Select compiler name and version number: GCC-8

#### Expected Behavior

When an object falls onto a surface (e.g. a sphere falls onto a plane) it should settle into place such that the object is just barely touching a surface (within some reasonable tolerance).

#### Current Behavior

When an object falls onto a surface, the Bullet collision detector reports contacts that are above the surface. As a result, objects that fell will settle into a position where they are hovering over the surface.

Note that if the object starts in contact with the surface but has no initial velocity, this phenomenon does not happen, even with gravity pushing the object down into the surface.

#### Steps to Reproduce

A regression test can be found in #1185

mxgrey:
I'm noticing that in the sphere-plane tests, the collision detector is returning exactly one contact point every time step, and that one contact point remains the same from the very first time it gets computed. This makes me wonder if the issue could be related to #859 which seems to have something to do with making contact points persistent.

mxgrey:
I've noticed that the penetration depth from the contacts report [here](https://github.com/dartsim/dart/blob/f31ccf5a72e7b3e547267853c6e68603925e659a/dart/collision/bullet/BulletCollisionDetector.cpp#L685) will come out negative for bodies that are moving towards each other. When a point has a negative penetration depth, it will be located on an object that is not *actually* in penetration *yet*, but perhaps moving towards a penetration. If I filter out points that have a negative penetration depth value, then this issue disappears.

One brute force solution could be to throw away contact points that have negative penetration depth. However, I think that would be losing valuable information, since we could instead use these points to avoid problems like excessive penetration and phasing. I'll take a look at the collision handling code to see if I can identify a way to leverage negative penetration information, although I'd appreciate any and all guidance on that, since it's a part of the codebase that I'm not familiar with.

mxgrey:
Also this may be related to what @karenliu has said about Bullet collision detector reporting collision points that are outside of the collision bodies.

mxgrey:
[...]

PR #1185, opened 2018-10-30
Ignore phantom contact points (negative penetration depth) from Bullet

mxgrey:
This PR is for investigating the issue reported in #1184.

***


dart/constraint/ConstraintSolver.cpp
```diff
@@ -441,7 +441,11 @@ DART_SUPPRESS_DEPRECATED_END
       mSoftContactConstraints.push_back(
             std::make_shared<SoftContactConstraint>(contact, mTimeStep));
     }
-    else
+    // If penetration depth is negative, then the collision isn't really
+    // happening and the contact point should be ignored.
+    // TODO(MXG): Investigate ways to leverage the proximity information of a
+    //            negative penetration to improve collision handling.
+    else if(contact.penetrationDepth >= 0.0)
```
jslee02: I think we would like to ignore negative penetration contacts even for soft contact liek:
```cpp
if(contact.penetrationDepth < 0.0)
  continue;

if (isSoftContact(contact))
{
  ...
}
else
{
  ...
}
```

jslee02: @mxgrey Any thoughts on this? :smile:

mxgrey: Sorry for the delay, I got sidetracked. Makes sense to me! I've updated the branch with this change.

jslee02: Thanks! Let's merge once CI tests are done.

dart/constraint/ConstraintSolver.cpp
```diff
@@ -436,16 +436,19 @@ DART_SUPPRESS_DEPRECATED_BEGIN
     shapeFrame2->asShapeNode()->getBodyNodePtr()->setColliding(true);
 DART_SUPPRESS_DEPRECATED_END
 
+    // If penetration depth is negative, then the collision isn't really
+    // happening and the contact point should be ignored.
+    // TODO(MXG): Investigate ways to leverage the proximity information of a
+    //            negative penetration to improve collision handling.
+    if (contact.penetrationDepth >= 0.0)
```
jslee02: Oh, wait. Actually, this should be:
```cpp
if (contact.penetrationDepth < 0.0)
```
to match the original logic.

mxgrey: Whoops! Thank goodness for unit tests :sweat_smile:

PR #1201, opened 2018-12-06
Add Metaskeleton::cloneMetaSkeleton()

jslee02:
This PR introduces clone functions to MetaSkeleton classes. It basically clones necessary `Skeleton`s are held by a MetaSkeleton as needed.

The clone function name of MetaSkeleton is `cloneMetaSkeleton()` to avoid the name conflict with `Skeleton::clone()`.

***


dart/dynamics/MetaSkeleton.hpp
```diff
@@ -71,11 +71,21 @@ class MetaSkeleton : public common::Subject
   /// Default destructor
   virtual ~MetaSkeleton() = default;
 
+  /// Creates an identical clone of this MetaSkeleton
+  virtual MetaSkeletonPtr cloneMetaSkeleton(
+      const std::string& cloneName) const = 0;
+  // TODO: In DART7, rename this to clone() and change the current
+  // Skeleton::clone() to override it. Then consider adding
+  // Skeleton::cloneSkeleton().
```
mxgrey: Super nitpick: We don't need to consider adding `Skeleont::cloneSkeleton()`, because it's already been added.

PR #1209, opened 2019-01-06
Import dependency targets when loading DARTConfig.cmake

### LINKED ISSUE #1200: Example Compile Error in Mac OS Mojave
Hello,
I am working on Mac OS Mojave with DART 6.5.0
I just go to the
```
examples/atlasSimbicon
mkdir build & cd build
cmake .. & make
```
and getting the following error
```
ld: library not found for -lconsole_bridge
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [atlasSimbicon] Error 1
make[1]: *** [CMakeFiles/atlasSimbicon.dir/all] Error 2
make: *** [all] Error 2
```
I think this issue is related to the version of Mac OS.
Do you have a plan to fix this?

jslee02:
Currently, the transitive dependencies of DART are propagated by using CMake variables. For example, all the dependency libraries are stored in `DART_LIBRARIES`, which will be defined by loading `DARTConfig.cmake` through calling `find_package(DART)` in a downstream project. 

However, this approach can fail to load the dependency. The items in `DART_LIBRARIES` can be either of the path to a library file or a target. Path items work fine (as long as the path is correct), but not targets because the downstream project don't know how to import the target only from the target name. This PR fixes it by letting `DARTConfig.cmake` loading the dependency targets by calling `find_packages()` for them. This approach should work as long as all the upstream dependencies loading relevant targets subsequently.

In this PR, the new change is only applied for loading `urdfdom` to quickly fix #1200, but this should be applied to all other transitive dependencies in future PRs.

***

### DIFF
--- a/cmake/Components.cmake
+++ b/cmake/Components.cmake
@@ -122,6 +122,23 @@ function(add_component_dependencies package_name component)
 endfunction()
 
 #==============================================================================
+function(add_component_dependency_packages package_name component)
+  set(component_prefix "${package_name}_component_")
+  set(dependency_package ${ARGN})
+
+  is_component(is_valid ${package_name} "${component}")
+  if(NOT ${is_valid})
+    message(FATAL_ERROR
+      "Target '${component}' is not a component of ${package_name}.")
+  endif()
+
+  set(target "${component_prefix}${component}")
+
+  set_property(TARGET "${target}" APPEND
+    PROPERTY "${component_prefix}dependency_package" ${dependency_package})
+endfunction()
+
+#==============================================================================
 function(add_component_targets package_name component)
   set(component_prefix "${package_name}_component_")
   set(dependency_targets ${ARGN})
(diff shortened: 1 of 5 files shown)


scpeters:
I'm seeing a CI failure on bionic with 6.7.0, which might be related to this PR:

* https://build.osrfoundation.org/job/ignition_physics-ci-pr_any-ubuntu_auto-amd64/154//consoleFull

I'm investigating locally...

jslee02:
Could you upgrade the dart packages again and try?:
```
sudo apt update; sudo apt upgrade
```

The missing files reported in you build log were not installed in the initial release, but they should be installed by now.

jslee02:
It seems it's built w/o any compilation errors, but there are three new test failures. I hope that's not caused by DART. :sweat:

scpeters:
the test failures seem to be related to the following warning:

~~~
/usr/bin/ld: warning: libdart-external-odelcpsolver.so.6.7, needed by /usr/lib/libdart.so.6.7.0, not found (try using -rpath or -rpath-link)
~~~

scpeters:
I'll try invalidating the docker cache to see if that is related

scpeters:
After invalidating the docker cache, it fails to configure, complaining about `tinyxml2::tinyxml2`

* https://build.osrfoundation.org/job/ignition_physics-ci-pr_any-ubuntu_auto-amd64/158//consoleFull

scpeters:
I can reproduce this locally in a docker container

mxgrey:
I'm having some trouble with consistently recreating the `tinyxml2` issue, but I think I know a strong candidate for the cause:

The tinyxml2 issue might come from a conflict in the third-party find-modules that are being provided by ign-cmake and dartsim. Both packages have find-modules for tinyxml2 which we both make accessible to the `CMAKE_MODULE_PATH`, but when `ign-physics` is configured, it might sometimes choose the `ign-cmake` find-module which defines a target `TINYXML2::TINYXML2` instead of the `dartsim` find-module (but I can't tell if this is deterministic or not, because it only happens to me sometimes, and I haven't figured out what the exact conditions are yet).

This falls on both `ign-cmake` and `dartsim` for using find-module names that are too generic. For DART, I would recommend making it so that `dart_find_package(tinyxml2)` will search for `find_package(DARTtinyxml2)` and that `dartsim` should provide a `FindDARTtinyxml2.cmake` file. And it should do likewise for all its dependencies.

We can do something similar for `ign-cmake`, although `ign-cmake`'s find-modules already have a bunch of downstream packages depending on them, so renaming them won't be as straightforward.

I'll go ahead and make the changes for dartsim.

jslee02:
@mxgrey The plan sounds great to me!

mxgrey:
I realized a much simpler solution: We just shouldn't use DART's find-modules to create the imported target. If we always create the imported target in the `DARTFind*.cmake`, then we can be sure it gets created regardless of whose find-module is used (as long as that find-module sets `*_INCLUDE_DIRS` and `*_LIBRARIES` according to cmake conventions).

jslee02:
It sounds better! Let's go with it!

PR #1212, opened 2019-01-08
Drop supporting Trusty and use imported targets for dependencies

jslee02:
This PR is a sequel of #1209, and more will be followed to modernize CMake scripts of DART. Here are two main changes in this PR.

* Drop supporting Ubuntu 14.04 (Trusty) in DART 6.7. The main reason is to use the recent CMake features (e.g., target-based dependency management), which requires CMake >= 3. Also, this is to reduce the effort of supporting the too old version of dependencies.

* Use `IMPORTED` targets for dependency libraries where available. For the dependencies that don't provide targets, define the targets in the finding modules, `Find<package>.cmake`. Use the same target name when the corresponding dependency provides the target in their latest version, which is not released yet.

***


scpeters:
The `release-6.7` branch builds fine on my mojave machine

jslee02:
@scpeters I think there were two issues, which should be fixed by the last few commits.

First, I switched to use the upstream config file of assimp, but they use a hard-coded library path that doesn't work for macOS. So I fall back to using own module file, `Findassimp.cmake`.

Second, `GLUT::GLUT` seems not working for Mojave or maybe for all the macOSs. This is because `FindGLUT.cmake`, provided by CMake, uses a different library path for the link library property of `GLUT::GLUT` from `GLUT_LIBRARIES`. The link library property of `GLUT::GLUT` additionally has `/GLUT` suffix, and it causes the build error you saw. So I just had to fall back to using variables for GLUT.

scpeters:
I just tested 7c13f5d on our mojave test machine, and it builds successfully. Nice work!

jslee02:
Thanks for the testing! Let me merge this PR once Travis CI becomes happy.

Issue #1204, opened 2018-12-13
Real-time window simulation


adalessandro:
Running the tutorial *tutorialBiped*, the number of frames per second I get is ~500, causing a *slow* (not real-time) simulation. Is it supposed or is there a way to run the simulation in real-time?

AFAICS, the way SimWindow class drives the simulation, it is not guaranteed to run in real-time.
The SimWindow class computes the number of iterations (steps) based on the value of `mDisplayTimeout`.
https://github.com/dartsim/dart/blob/9a6bb46896ddc1b073e1b0d73f4d69d4fe321f1f/dart/gui/glut/SimWindow.cpp#L128

The next callback is registered using `glutTimerFunc`, which will be triggered in **at** **least** `mDisplayTimeout` milliseconds, but could be more than that.
https://github.com/dartsim/dart/blob/9a6bb46896ddc1b073e1b0d73f4d69d4fe321f1f/dart/gui/glut/SimWindow.cpp#L140

Shouldn't SimWindow compute the number of steps based on the real time elapsed from the last update callback?

This might be related to https://github.com/dartsim/dart/issues/735

mxgrey:
> This might be related to #735

Issue #735 is an integer->double conversion bug that halts simulation altogether due to rounding down to `0`. It wouldn't make simulation slower; it just stops simulation altogether.

The GUI doesn't make any effort to match real time during a simulation run. The GUI is just there to show you what's happening in the simulation at whatever speed the simulation happens to be running. Simulation speed can depend on many factors, like the size of the time step that you've chosen (the default being `0.001s`) or the complexity of your simulation (e.g. how many collidable objects are there; how many joints are being simulated).

Since you have the ability to program your GUI application, you can choose to program the GUI to play back at real time. If your simulation runs faster than real time, then you can have your GUI throttle the simulation. If your simulation runs slower than real time, then you can either:
1. Run the full simulation, record the states, and replay those states at real time
2. Tune your simulation so that it can run faster than real time, and then throttle it

These aren't features that dartsim offers out of the box.

mxgrey:
Also, if you want to make a sophisticated GUI application with dartsim, I would recommend using the `dart::gui::osg` component library which uses [OpenSceneGraph](http://www.openscenegraph.org/). It's much more user-friendly than the glut component library.

adalessandro:
@mxgrey , thanks for the explanation. It is clear to me now that this is the intended behavior for the tutorials.

[...]

PR #1216, opened 2019-01-10
Add a RealTimeWorldNode for displaying simulations at real-time speeds

mxgrey:
I recently needed to make a real-time simulation display on top of dartsim, so I decided to clean it up and add it into the main repo since it's a frequently requested feature (e.g. #1204).

It's a simple wrapper on top of the `dart::gui::osg::WorldNode` class. Users can opt for either the old `WorldNode` or the new `RealTimeWorldNode` depending on how they want to control the speed of the simulation.

I also added a line `ImGui::CreateContext();` because ImGui was crashing without that.

I know #1215 has already been opened, so I can retarget this at a `6.8` release (or we can wait until 6.7.1 since this won't impact API or ABI) instead if that would be preferred.

***

- [ ] (N/A) Add unit test(s) for this change


mxgrey:
I also looked over the osg examples and replaced their use of `WorldNode` with `RealTimeWorldNode` for any examples that perform simulation. I think the new playback speed looks much better.

dart/gui/osg/RealTimeWorldNode.cpp
```diff
@@ -0,0 +1,166 @@
...
+        std::min(mLastRealTimeFactor, mLowestRealTimeFactor);
+    mHighestRealTimeFactor =
+        std::max(mLastRealTimeFactor, mHighestRealTimeFactor);
+  }
+  else
+  {
+    mFirstRefresh = true;
+  }
+
+  mRefreshTimer.setStartTick();
```
jslee02: Please correct me if I'm wrong:

This might be micro optimization, but we could reduce the range of use of `mRefreshTimer` by putting `if (mFirstRefresh) {...}` and this line (`mRefreshTimer.setStartTick();`) into `if (mWorld && mSimulating) {...}` like:

```cpp
  if (mWorld && mSimulating)
  {
    if (mFirstRefresh)
    {
      mRefreshTimer.setStartTick();
      mFirstRefresh = false;
    }

    const double startSimTime = mWorld->getTime();
    const double simTimeStep = mWorld->getTimeStep();

    while(mRefreshTimer.time_s() < mTargetRealTimeLapse)
    {
      const double nextSimTimeLapse =
          mWorld->getTime() - startSimTime + simTimeStep;

      if(nextSimTimeLapse <= mTargetSimTimeLapse)
      {
        customPreStep();
        mWorld->step();
        customPostStep();
      }
    }

    mLastRealTimeFactor =
        (mWorld->getTime() - startSimTime)/mTargetRealTimeLapse;
    mLowestRealTimeFactor =
        std::min(mLastRealTimeFactor, mLowestRealTimeFactor);
    mHighestRealTimeFactor =
        std::max(mLastRealTimeFactor, mHighestRealTimeFactor);

    mRefreshTimer.setStartTick();
  }
  else
  {
    mFirstRefresh = true;
  }
```

By this change, we could save calling `mRefreshTimer.setStartTick();` twice when it's not in simulating. Not sure how heavy to call `mRefreshTimer.setStartTick()` though.

dart/gui/osg/RealTimeWorldNode.hpp
```diff
@@ -0,0 +1,134 @@
...
+  double mLastRealTimeFactor;
+
+  /// The lowest RTF that has been achieved
+  double mLowestRealTimeFactor;
+
+  /// The highest RTF that has been achieved
+  double mHighestRealTimeFactor;
+
+  /// Flag to track whether we should display the real time factor
+  bool mDisplayRealTimeFactor;
```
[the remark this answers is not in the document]
mxgrey: Oh yeah, I was going to add a HUD to display the number for the RTF and then this variable would store whether that HUD should be hidden or displayed, but then I realized that adding a HUD is a bit more of a rabbit hole than I really wanted to get into. We can just remove this for now.

dart/gui/osg/RealTimeWorldNode.cpp
mxgrey: It's a bit of a micro-optimization, but there's nothing wrong with that. If it seems more logical to you to have them inside the if-statement, I have no problem with that :+1:

jslee02:
Hm, the last commits from me are in incorrect order due to the wrong push dates. This is because the date setting of my virtual machine was set to 2 days late for some reason. :disappointed:

PR #1222, opened 2019-01-11
Create imported targets in DARTFind*.cmake instead of Find*.cmake

mxgrey:
This attempts to fix the issue reported in the thread of #1209


mxgrey:
@jslee02 I'm looking over the cmake scripts for finding packages, and I'm not clear on one thing: What happens when the dependency of a component is missing? For example, let's say I install `libdart` with `libdart-utils` which depends on `libtinyxml2`, but later I forcibly delete `libtinxyml2`. Then I call `find_package(DART REQUIRED COMPONENT utils)`. Ideally cmake should recognize that `tinxyml2` is no longer available and then make the configuration fail.

I think that'll work fine for core dependencies, because we use `find_package(... REQUIRED)` on those, but the components don't use `REQUIRED` in their `DARTFind*.cmake` files, so I don't think this situation will be caught.

Please let me know if I'm overlooking something. In the meantime I'll see if I can find a minimally invasive way to make the config-file work as desired.

jslee02:
I also think the current script wouldn't catch the case, which I believe most of the build scripts wouldn't do either (because the system has been forcibly changed from when it was built). But it sounds good to have a logic to catch the case.

jslee02:
One possible solution would be:

* Change all the current `DARTFind<package>.cmake` files into macros and put it in one file (e.g., `DARTFindPackages.cmake`)
* Make the macros to have an option whether find the package required or not
* Find optional packages with or without required option when DART is built depending on whether the package is options, and with required in `DARTConfig.cmake`

mxgrey:
CMake config-files are encouraged to use [`find_dependency()`](https://cmake.org/cmake/help/v3.5/module/CMakeFindDependencyMacro.html) which will pass along the `REQUIRED` and `QUIET` properties from the config-file or find-module that uses it. When a required dependency isn't found, the configuration will fail. Ideally that's the kind of behavior we should aim for.

Unfortunately `find_dependency(~)` only helps in very narrow circumstances, and I don't think it'll help us out much here. I have some ideas about a solution, though. I'll open a separate PR for that, though, since it's mostly orthogonal to this issue.

mxgrey:
I'm thinking about using the `dart_*Component.cmake` files to handle the `REQUIRED` behavior. It looks like those files are automatically generated, so it should be pretty straightforward to put the behavior into them. I just need to find how those files get created.

jslee02:
> Unfortunately `find_dependency(~)` only helps in very narrow circumstances, and I don't think it'll help us out much here.

Yeah, AFAIK, `find_dependency()` does not support COMPONENTS or MODULE until 3.8.0.

> I'll open a separate PR for that, though, since it's mostly orthogonal to this issue.

:+1:

jslee02:
> I just need to find how those files get created.

Those files are generated by [these lines](https://github.com/dartsim/dart/blob/master/cmake/Components.cmake#L186-L233) in `Components.cmake`.

PR #1225, opened 2019-01-14
Support multiple CMake conventions for *_FOUND variable

mxgrey:
This PR continues the efforts of #1222 in fixing an issue with the confusion of using alternative find-modules.

CMake has two conventions for setting the `*_FOUND` variable for a package: `PackageName_FOUND` and `PACKAGENAME_FOUND`, with the latter being the older convention. To be able to gracefully handle both conventions, we should check both versions of the variable when decided whether to make the imported target.

Note that this is branching off of #1223, so we should merge that one before reviewing this.


jslee02:
It seems there is another convention: all lowercases, `packagename_FOUND` such as `tinyxml2_FOUND`. How about we always use uppercases (or lowercases) by converting the variable using `string(TOUPPER ...)` (or `string(TOLOWER ...)`) to cover all the conventions?

Issue #1231, opened 2019-01-30
Collisions between static (immobile) bodies should be ignored


mxgrey:
Currently dartsim is computing collisions between static (i.e. `skeleton->isMobile() == false`) bodies. This is not helpful for simulation, and it's hard to imagine any application where these contact points would be needed. Instead, contact information between static bodies ends up wasting space in the `maxNumContacts` quota.

I recommend that we tweak the `BodyNodeCollisionFilter` to also ignore contacts between static bodies.

mxgrey:
Fixed in #1232

PR #1232, opened 2019-01-30
Ignore collisions between non-mobile bodies

mxgrey:
Addresses #1231

***

### DIFF
--- a/dart/collision/CollisionFilter.cpp
+++ b/dart/collision/CollisionFilter.cpp
@@ -137,12 +137,10 @@ bool BodyNodeCollisionFilter::ignoresCollision(
 
   if (skel1 == skel2)
   {
-    auto skeleton = bodyNode1->getSkeleton();
-
-    if (!skeleton->isEnabledSelfCollisionCheck())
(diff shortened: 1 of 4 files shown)


dart/collision/CollisionFilter.cpp
```diff
@@ -129,7 +129,13 @@ bool BodyNodeCollisionFilter::ignoresCollision(
...
-  if (bodyNode1->getSkeleton() == bodyNode2->getSkeleton())
+  const auto& skel1 = bodyNode1->getSkeleton();
+  const auto& skel2 = bodyNode2->getSkeleton();
+
+  if ( !skel1->isMobile() && !skel2->isMobile() )
+    return true;
+
+  if (skel1 == skel2)
   {
     auto skeleton = bodyNode1->getSkeleton();
```
scpeters: I think `skel1` could be used instead of `skeleton` in this code block

jslee02:
The build failure is because of the build time exceeding, which is unrelated to this PR. Merging now!

PR #1236, opened 2019-02-10
Add missing definition of Group::removeJoints(...)

jslee02:
***


dart/dynamics/Group.cpp
```diff
@@ -535,6 +535,17 @@ bool Group::removeJoint(Joint* _joint, bool _removeDofs, bool _warning)
   return removed;
 }
 
+//==============================================================================
+bool Group::removeJoints(
+    const std::vector<Joint*>& _joints, bool _removeDofs, bool _warning)
+{
+  bool added = false;
```
mxgrey: Should this variable be `removed` instead of `added`?

Issue #1026, opened 2018-03-08
Proposal: Add inline namespaces to DART and maintain ABI stability between minor releases


mxgrey:
This is semi-related to #917

Future versions of Gazebo intend to use a plugin-based architecture in order to make the Gazebo simulations more modular and customizable.

One danger that is often faced is that multiple plugin libraries might depend on different versions of the same dependencies. Trying to use these multiple plugins simultaneously in one application can easily result in ABI collisions, due to DART's current approach to ABI management.

An easy and effective way to avoid symbol collision is to use [inline namespaces](https://stackoverflow.com/questions/11016220/what-are-inline-namespaces-for) which contain the major version of the library. However, this is only effective for avoiding symbol collisions between different major versions.

If the DART ABI gets changed between minor versions (e.g. adding a new member variable to a class), then simultaneously using two plugins that were built against two different minor versions will cause the application to crash.

The most popular and effective strategy for maintaining ABI stability is the PIMPL pattern, but DART uses a considerable amount of template metaprogramming, to the extent that I'm skeptical whether we could reasonably apply the PIMPL pattern to DART in any normal way.

One mitigation strategy is to defer any ABI-breaking changes until a major version increment. I think our ABI-breaking changes are rare enough that this would be an acceptable solution (at least to me).

Alternatively, I might be able to come up with a way to use the `Composite` class to give us a PIMPL-like way to delay ABI breakages. I'm not entirely sure how plausible this is, though.

I think this is particularly important because of DART's use as a toolkit. It's very likely that different plugins will want to use DART independently of each other for various reasons (e.g. a planning plugin might want to use the forward/inverse kinematics, while a controller plugin might want to use the inverse dynamics, and a physics engine plugin uses the forward dynamics). These plugins might be built and distributed against different versions of DART, especially in the case that a developer needs to make a plugin that uses a new feature of DART that wasn't available in the version that the other plugins were built and distributed for.

The easiest mitigation strategy would be to include the minor version number in the inline namespace, and to install headers to directories that are specific to major+minor version. I think this would be too blunt of an instrument, and it would prevent plugins that are built using different minor versions of DART from communicating using DART symbols.

scpeters:
[...]

PR #1239, opened 2019-02-13
Fix SOVERSION to use only MAJOR to match ABI rules

j-rivero:
The SOVERSION is currently coded to MAJOR.MINOR in Dart. If I'm not wrong, Dart follows semantic versioning so ABI is compatible in all minor versions of a major serie . If that is the case, I think that the current SOVERSION should be composed only by the major version.

### DIFF
--- a/cmake/DARTMacros.cmake
+++ b/cmake/DARTMacros.cmake
@@ -90,7 +90,7 @@ macro(dart_add_library _name)
   add_library(${_name} ${ARGN})
   set_target_properties(
     ${_name} PROPERTIES
-    SOVERSION "${DART_MAJOR_VERSION}.${DART_MINOR_VERSION}"
+    SOVERSION "${DART_MAJOR_VERSION}"
     VERSION "${DART_VERSION}"
   )
 endmacro()


jslee02:
We intend to follow the semantic versioning, but I'm not sure if we strictly following it. I try to keep API compatibility over minors and patches in a major (with some accidental exceptions) but am not sure about ABI. Would this be the case?

mxgrey:
The problem is that semantic versioning is a bit ambiguous when it comes to C++, because it's very possible to maintain API compatibility but break ABI compatibility.

DART's current policy is to maintain API compatibility between minor releases, and maintain ABI compatibility between patch releases, but we make no guarantee about ABI between minor releases.

I've opened an issue to propose that we maintain ABI compatibility between minor releases: #1026. However making that feasible would require a major overhaul to how DART is implemented, and it's not clear to me when or if we'd have time for that.

mxgrey:
But while the current policy exists, we do need both major and minor versions as part of the SOVERSION.

j-rivero:
I was not aware of the lack of clarity in semantic versioning respect to the ABI. Thanks for explanation, I will rename the dart packages in Debian according to the policy (grep for `package-name-doesnt-match-sonames` in the [Jenkins build log](https://build.osrfoundation.org/view/All/job/dart-pkg_builder-master-debian_sid-amd64/17/consoleFull#239674566c767c74a-22ce-41e3-b8e9-c82dba4fba96)) 

It would be good if we can keep packages with the same name in both upstream Debian and the PPA https://launchpad.net/~dartsim/+archive/ubuntu/ppa/+packages .

PR #1240, opened 2019-02-13
Use GNUInstallDirs for lib/include path

j-rivero:
The change uses `GNUInstallDirs` cmake module to provide platform specific paths. Not sure how this can affect how Dart manages the directory structure on Windows.


j-rivero:
follow in #1241

PR #1241, opened 2019-02-13
Use GNUInstallDirs for lib/include path

j-rivero:
The change uses `GNUInstallDirs` cmake module to provide platform specific paths. Not sure how this can affect how Dart manages the directory structure on Windows.

Replaces #1239


jslee02:
> Not sure how this can affect how Dart manages the directory structure on Windows.

Buildings on Windows is not maintained for some reasons. Let's leave it as it is for now.

The test failure is due to long outputs for coverage report, which is not related to this PR. Merging now.

j-rivero:
no problem, thanks for the quick merge Jeongseok

Issue #1243, opened 2019-02-19
Skeleton::State failing to capture joint states


mxgrey:
This came to our attention in [this thread](https://github.com/dartsim/dart/issues/386#issuecomment-464606606). It's unclear to me what's causing this, since `Skeleton::State` has worked reliably for me in the past, but I've confirmed that it's currently not working, and I'm investigating why.

mxgrey:
I'll be adding a regression test for this to make sure the issue doesn't recur.

jslee02:
Resolved by #1245

PR #1244, opened 2019-02-19
Fix issue #1243 and add a regression test

mxgrey:
This fixes the problem reported in #1243.

I'll also open PRs to patch this in earlier releases of `6.x`, because it's a very significant error (even though the fix is very simple).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 #### Changes
 
+* Dynamics
+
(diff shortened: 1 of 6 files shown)


jslee02:
#1245 is merged to `release-6.7`. Let's apply the patch to `master` by merging `release-6.7` into `master`. I think we don't need to have duplicate changelog item for the same patch. Closing in favor of #1249.

PR #1262, opened 2019-03-13
Make gui-glut as optional component

jslee02:
### Problem
`gui` component includes the GLUT-based GUI by default. This introduces unnecessary GLUT dependency even when only OSG-based GUI is required.

### Solution
Split out the GLUT-based GUI from `gui` component as an optional component `gui-glut`.

### Caveate
This is a breaking change. In order to use the GLUT-based GUI, `gui-glut` component now needs to be added explicitly as:
```cmake
find_package(DART COMPONENT gui-glut)
```

***

- [x] Set version target by selecting a milestone on the right side (N/A)


mxgrey:
Since this is a breaking change, might it be better to target it at dartsim-7?

jslee02:
Hm, it would be good to have this change without waiting for dartsim-7.

Alternatively, we could build `gui-osg` without requiring GLUT by removing the dependency of `gui`. This is because `gui-osg` currently doesn't use any of `dart/gui/*`. This solution sounds a bit tricky, but at least it wouldn't break anything. One corner case would be the user have to include both of `gui` and `gui-osg` if they want to use both GLUT and OSG at the same time, which was only required including `gui-osg`.

mxgrey:
I see, that all makes sense.

> This is because gui-osg currently doesn't use any of dart/gui/*.

If `dart-gui-osg` doesn't depend on any content in `dart-gui` then it makes sense to not have that dependency anyway. For that matter, `dart-gui` probably shouldn't exist anyway, since it seems to just forward the user to `dart-gui-glut`. We can just make the headers of `dart-gui` part of the `dart-gui-glut` component.

jslee02:
The motivation behind having `dart-gui` was for the case that the component is used by both of `dart-gui-glut` and `dart-gui-glfw` in DART 6. But, I'm open to removing it since `dart-gui-glfw` will replace `dart-gui-glut` anyway in DART 7. In that case, the shared code will be both in `dart/gui/glut` and `dart/gui/glfw`.

mxgrey:
I guess since we don't guarantee ABI compatibility in between minor releases, we're not violating our policy by moving the glut content into its own component library. The API also isn't changing; the user just needs to add an argument to their `find_package(DART ...)` call.

The motivations make sense, and it's hard to imagine a cleaner migration path without waiting until major release 7.

jslee02:
Closing in favor of #1562

PR #1266, opened 2019-03-27
Apply IK solution only when the solver succeeded

jslee02:
### Problem
IK solvers apply solution even when the solver failed to solve.

### Solution
Apply solution only on success.

### Testing
Added a simple unit test that checks if IK solver changes joint values when the solver failed.

***

### DIFF
--- a/dart/dynamics/InverseKinematics.hpp
+++ b/dart/dynamics/InverseKinematics.hpp
@@ -121,9 +121,10 @@ public:
   /// InverseKinematics::getProblem() and then using
   /// Problem::addSeed(Eigen::VectorXd).
   ///
-  /// By default, the Skeleton itself will retain the solved joint positions.
-  /// If you pass in false for _applySolution, then the joint positions will be
-  /// returned to their original positions after the problem is solved.
+  /// By default, the Skeleton itself will retain the solved joint positions
+  /// unless it failed to solve. If you pass in false for _applySolution, then
+  /// the joint positions will be returned to their original positions after the
+  /// problem is solved.
   ///
   /// Calling this function will automatically call Position::setLowerBounds(~)
   /// and Position::setUpperBounds(~) with the lower/upper position bounds of
--- a/dart/dynamics/InverseKinematics.hpp
+++ b/dart/dynamics/InverseKinematics.hpp
@@ -121,21 +121,102 @@ public:
   /// InverseKinematics::getProblem() and then using
   /// Problem::addSeed(Eigen::VectorXd).
   ///
-  /// By default, the Skeleton itself will retain the solved joint positions
-  /// unless it failed to solve. If you pass in false for _applySolution, then
-  /// the joint positions will be returned to their original positions after the
-  /// problem is solved.
+  /// By default, the Skeleton itself will retain the solved joint positions.
+  /// If you pass in false for _applySolution, then the joint positions will be
+  /// returned to their original positions after the problem is solved.
   ///
   /// Calling this function will automatically call Position::setLowerBounds(~)
   /// and Position::setUpperBounds(~) with the lower/upper position bounds of
   /// the corresponding Degrees Of Freedom. Problem::setDimension(~) will be
   /// taken care of automatically, and Problem::setInitialGuess(~) will be
   /// called with the current positions of the Degrees Of Freedom.
-  bool solve(bool _applySolution = true);
+  ///
+  /// \deprecated Deprecated in DART 6.8. Please use solveAndApply() instead.
+  DART_DEPRECATED(6.8)
+  bool solve(bool applySolution = true);
 
   /// Same as solve(bool), but the positions vector will be filled with the
   /// solved positions.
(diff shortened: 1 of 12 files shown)


mxgrey:
Applying the solution during a failure was intentional, because iterative solvers will often do a decent job of getting the end effector close to the target even if it failed to find a solution, and that's often desirable.

My personal view is that if the user wants the solution to be applied conditionally, they can set `_applySolution` to false and then choose after the fact whether they want the result to be applied based on whether or not it's a valid solution.

Maybe an alternative to make the behavior more clear would be to change the name of the input argument to `_applyResult`?

jslee02:
Thanks for the clarification. I agree with you about iterative solver, but I'm a bit worried that the current IK implementation accepts any types of nonlinear optimizers. Non-iterative solver or analytic IK solver could return totally incorrect solution if the solution doesn't exist.

> My personal view is that if the user wants the solution to be applied conditionally, they can set `_applySolution` to false and then choose after the fact whether they want the result to be applied based on whether or not it's a valid solution.

Yeah, this is possible, but I'd prefer the opposite. I mean the default behavior to be applying solution on success, and the user should apply solution by calling `InverseKinematics::setPositions()` if they want to use the incomplete solution from an iterative solver. This is because I think the safe and conservative way should be the default.

I'd like to propose three possible solutions:

(a) No change but renaming `_applySolution`
(b) Change the default value of `_applySolution` to `false` so that the user can pay attention to the behavior when set the parameter.
(c) Change the APIs to be more explicit something like
```cpp
bool solve(Eigen::VectorXd& position);
bool solveAndApply(Eigen::VectorXd& position, bool allowIncompleteResult = false);
bool solveAndApply(bool allowIncompleteResult = false);
```

In any case, I think we should document more clear about the behavior.

mxgrey:
I definitely understand where you're coming from, and I agree that we could make it better with some changes. One concern I have is that I'm sure I've written code that depends on the current behavior of `solve(~)`, and I wouldn't be surprised if others have as well.

I guess we could deprecate the `solve(bool)` and `solve(Eigen::VectorXd&, bool)` functions and introduce `solveAndApply(~)` with the signatures you're suggesting. Then we could add a `bool findSolution(Eigen::VectorXd& position)` function that finds the solution without applying it.

jslee02:
It sounds good to me! Let me update this PR accordingly.

jslee02:
Done!

dart/dynamics/InverseKinematics.hpp
```diff
@@ -130,11 +130,93 @@ class InverseKinematics : public common::Subject
...
+  /// the solver successfully found a solution or \c allowIncompleteResult is
+  /// set to true.
+  ///
+  /// \param[in] allowIncompleteResult Allow to apply the solution even when
+  /// the solver failed to find solution. This option would be useful when an
+  /// iterative solver is used because they will often do a decent job of
+  /// getting a result close to a solution even if it failed to find the
+  /// solution.
+  /// \return True if a solution is successfully found
+  bool solveAndApply(bool allowIncompleteResult = false);
```
mxgrey: I have mixed feelings on whether this should default to true or false. I understand why someone might exercise caution and avoid applying "unsolved" results, but for an iterative solver it's still better to hang onto the last unsolved result because if you try to run the solver again you'll be in a better position to find the solution than if you restart from the beginning. This is often done in order to solve an IK over the course of several GUI refreshes instead of needing to solve from scratch every rendering cycle.

Also, if the usual behavior for this function is to not apply the partial solution, then the user won't know where the configuration ended up if it failed to solve, and the user is left only knowing that a solution simply wasn't found.

Since we now have `findSolution(~)` which already defaults to not applying the result, I would recommend that `solveAndApply(~)` should default to applying the solution.

But I suppose this is a difference in preferences that mostly boils down to a difference in behavior between iterative solvers and closed-form solver (and we're trying to opaquely support both here), so I'm fine with keeping this as it is if there's a strong preference to have it this way.

jslee02: > Since we now have `findSolution(~)` which already defaults to not applying the result, I would recommend that `solveAndApply(~)` should default to applying the solution.

I'm also in two minds, but I think this statement made me to incline to defaulting the option to be true.

Maybe we could introduce `solveAndApplyOnSuccess()` to be more clear, but it feels like too much. But if you think it sounds good then I'm okay to add it.

mxgrey: I think `solveAndApplyOnSuccess()` would be overcompensating on explicitness. I think a well-named function argument is preferable to a more explicit function name, because arguments make it easier to programmatically determine behavior, for example

```
ik->solveAndApply(allowIncomplete);
```

vs

```
if(allowIncomplete) { 
  ik->solveAndApplyOnSuccess();
} else { 
  ik->solveAndAlwaysApply();
}
```

jslee02: Agreed. Let me change the default value to true.

PR #1272, opened 2019-04-05
Use docker for Xenial build on Azure

jslee02:
Per https://github.com/Microsoft/azure-pipelines-image-generation/pull/732, Azure uses Boost 1.69 by default, it seems there is no option for choosing the system boost (1.58). For some reason, [DART fails to build with Boost 1.69](https://dev.azure.com/dartsim/dart/_build/results?buildId=492), which should be fixed. In any case, the Xenial build should use the system Boost, so this PR changes to use docker for Xenial build.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -12,7 +12,7 @@ jobs:
     BUILD_DIR: $(Build.SourcesDirectory)
     SUDO: sudo
   steps:
-  - template: .ci/azure-pipelines/native.yml
+  - template: .ci/azure-pipelines/docker.yml
 
 - job: xenial_32bit_gcc_release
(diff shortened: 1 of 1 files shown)


scpeters:
that's weird that dart fails to build against boost 1.69 because that version has been in homebrew for a couple of weeks, and I presume it hasn't caused any CI failures on macOS

maybe it's related to the xenial version of gcc?

jslee02:
Thank you for the note. Yeah, [it seems building DART against Boost 1.69 fine](https://travis-ci.org/dartsim/dart/jobs/516239347#L164). I haven't tested but I also guess so: gcc 5.3 + boost 1.69 could be the cause.

PR #1273, opened 2019-04-06
Add a new Criteria constructor for creating sequence

### LINKED ISSUE #1268: Allow Chain to include bodies have multiple children
[`Chain` doesn't allow including bodies that has multiple child bodies](https://github.com/dartsim/dart/blob/56205a31896592891ada2c8cd11079b151e8e8ef/dart/dynamics/Chain.hpp#L44-L45), but this criteria seem too strict. To me, a chain shouldn't have a branch in it but should be fine to be in a tree.

The current behavior prevents creating an proper IK from an end-effector if the parent body of end-effector has another child body: the IK will use only the parent joint of the end-effector but not the other ancestor joints to the root. This could be workarounded by manually specifying the desired DOFs to be used, but it's not ideal.

jslee02:
### Problem
No convenient way to create a sequence Linkage.

### Solution
Add an utility constructor to `Linkage::Criteria` to create a sequence Linkage:
```cpp
Criteria(BodyNode* start, BodyNode* target, bool includeBoth = false);
```

### Issues
* Resolve #1268 

***

### DIFF
--- a/dart/dynamics/Linkage.hpp
+++ b/dart/dynamics/Linkage.hpp
@@ -127,6 +127,27 @@ public:
     /// entry in mTargets) will be halted if it reaches any entry in mTerminal
     std::vector<Terminal> mTerminals;
 
+    /// Constructs an empty criteria that will lead to creating an empty Linkage
+    Criteria() = default;
+
+    /// Utility constructor to create a contiguous sequence of BodyNodes that
+    /// doesn't include branches in it.
+    ///
+    /// The start and target can be interchangeably set if both are included in
+    /// the sequence.
+    ///
+    /// \param[in] start The first BodyNode in the sequence. If \c nullptr is
+    /// passed the sequence will expand from \c target to the root.
+    /// \param[in] target The second BodyNode in the sequence. If \c nullptr is
+    /// passed the sequence will expand from \c start to the root.
+    /// \param[in] includeBoth Set this to true if both \c start and \c target
+    /// should be included. Otherwise, whichever is upstream of the other will
+    /// be left out of the sequence.
+    Criteria(
+        BodyNode* start,
+        BodyNode* target,
+        bool includeBoth = false);
+
   protected:
 
     /// Refresh the content of mMapOfTerminals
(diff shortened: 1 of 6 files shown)


dart/dynamics/Linkage.hpp
```diff
@@ -127,6 +127,27 @@ class Linkage : public ReferentialSkeleton
...
+    /// passed the sequence will expand from \c target to the root.
+    /// \param[in] target The second BodyNode in the sequence. If \c nullptr is
+    /// passed the sequence will expand from \c start to the root.
+    /// \param[in] includeBoth Set this to true if both \c start and \c target
+    /// should be included. Otherwise, whichever is upstream of the other will
+    /// be left out of the sequence.
+    Criteria(
+        BodyNode* start,
+        BodyNode* target,
+        bool includeBoth = false);
```
mxgrey: I realize that this is just copying from `Chain.hpp` and that I'm probably the one that came up with the `includeBoth` concept, but I'm not actually sure that I understand the motive. Why should the options be:

1. Include both the start and the target, or
2. Leave out whichever one is upstream

I don't really see what the significance of the upstream link would be that we would want to leave it out. The only thing I can think of is maybe we want to "start" from a body node that's at a fork and then "target" a body node that's at the end of a branch. I suppose that would be helpful for constructing chains within multi-arm skeletons.

I'm fine with leaving it as it is and merging it, but I thought I'd go ahead and voice my own confusion at an API that I probably made myself. If anyone else finds it confusing, we should probably consider tweaking it somehow.

jslee02: > I realize that this is just copying from `Chain.hpp`

For the record, it is identical except that `mChain` option is false in this constructor. That's why I didn't make Chain use this constructor (maybe I should let Chain just use this constructor and change the option to true?).

Regarding `includeBoth`, I assumed that this option is about whether it includes the parent joint of the upstream body or not because the joint is technically not in between `start` and `target` bodies. In this case, we could change the option name and the document more clear.

mxgrey: Oh good call! You're exactly right: it's about whether to include the parent joints of both the start and the target. A more descriptive name would definitely be helpful. Maybe `includeUpstreamParentJoint`?

PR #1283, opened 2019-04-23
Fix invalid double quotation marks

jslee02:


jslee02:
Yeah, it wasn't a good idea. :disappointed:

PR #1291, opened 2019-04-24
BoxedLcpConstraintSolver: API compatible with 6.7

### LINKED ISSUE #1288: Deprecated BoxedLcpConstraintSolver constructor should be API backward compatible
See https://bitbucket.org/osrf/gazebo/pull-requests/3101/use-dart-67-behavior-for-setting/diff#comment-99742882

scpeters:
Fixes #1288 by adding default values to solver arguments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 #### Changes
 
-* XXX
+* Dynamics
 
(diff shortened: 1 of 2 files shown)


jslee02:
Thanks for the fix!

PR #1297, opened 2019-04-29
Fix incorrect alpha value update of InteractiveFrame

### LINKED ISSUE #1217: InteractiveFrames no longer display correctly
- [x] I checked the [documentation](http://dartsim.github.io/) and found no answer.

#### Environment
* Select DART version: release-6.7 branch
* Select OS name and version name(or number): Ubuntu 18.04
* Select compiler name and version number: GCC 8

#### Expected Behavior

`dart::gui::osg::InteractiveFrame` objects should display their components in elementary Red/Green/Blue colors. They should contain six arrows (two per axis), three planes (one per axis), and three wheels (one per axis). The transparency of the components should alternate depending on how the components are being interacted with.

#### Current Behavior

`dart::gui::osg::InteractiveFrame` objects have goofy colors which don't get reset correctly during and after user interactions.

Here's how it starts out:
screenshot from 2019-01-10 16-46-33

This is okay, except the wheel for every axis is being rendered as pure white.

After moving the cursor over one of the components, the colors end up like this, which is very wrong:

screenshot from 2019-01-10 16-47-57

#### Steps to Reproduce

Run the osgEmpty example program.

This problem affects every osg application that tries to use the InteractiveFrame class.

jslee02:
As reported in #1217, `InteractiveFrame` doesn't display correctly. This is because [the element access operator of `aiColor4D` with index 3 doesn't point to the alpha component but interestingly red component](https://github.com/assimp/assimp/blob/55cdf1d37c4ecd973c2a2e134e213cf32a772635/include/assimp/color4.inl#L94-L107).

This PR workaround by using the member variable instead. It'd be ideal to be fixed in the upstream though.

image

Fixes #1217.

***

- [x] Add unit test(s) for this change (visually checked)


mxgrey:
Oh wow, great catch! It seems the original implementation was [a bit more sensible](https://github.com/assimp/assimp/blob/v3.3.1/include/assimp/color4.inl#L78).

I guess the change was an attempt to thwart undefined behavior, but I think that motivation was a bit misguided considering the result.

jslee02:
Yeah, motivation was good but the change was accompanied with a silent error.

PR #1311, opened 2019-05-09
Fix dereferencing a dangling pointer in WorldNode

jslee02:
~~In `clearUnusedNodes()`, if an unused node doesn't have any holders that have the ownership, it becomes dangling pointers since `WorldNode` holds nodes as raw pointers.~~

In `clearUnusedNodes()`, an unused node is possible to hold a dangling pointer to `ShapeFrame`. This can happen because `ShapeFrameNode` doesn't have the ownership of `ShapeFrame`. This leads to dereferencing a dangling pointer in the logic of checking whether the node is shadowed or not. 

This PR changes the behavior of the node deletion logic to remove the access to the internal `ShapeFrame` pointer. Instead, it now unconditionally removes the node from both of shadowed group and unshadowed group. I believe this is safe because OSG will ignore the attempt of removing a non-existing node in the group.

***


jslee02:
It seems the actual dangling pointer is the ShapeNode in ShapeFrameNode (not the ShapeFrameNode itself since the group has the ownership). Will update this PR if needed.

Issue #739, opened 2016-06-28
Distinguish between missing vs. default VisualAspect color


mkoval:
There is currently no concept of having a `VisualAspect` with no color information. If the user does not set the color, then it defaults to blue. This is problematic when working with textured models where the information in the `VisualAspect` is most naturally used to tint the existing texture.

There is no way to distinguish between the default color (which should correspond to "no tint") and the user explicitly setting the `VisualAspect` to that color (which should correspond to "blue tint"). I see three possible solutions:

1. *Always treat the color of as a tint,* which results in textured models being tinted blue.
2. *Change the default color to white,* which is a poor default for non-textured models.
3. *Distinguish between a missing color and one that has been explicitly set,* which allows the renderer to inspect the `Shape` before choosing which default to use.

I suggest (3). This could be implemented by adding an `isColored` flag to `VisualAddon` or replacing the current member variable with something similar to a `boost::optional`.

For an example of what (1) looks like, see below. Notice how the entire model has a strong blue tint.

![val_blue](https://cloud.githubusercontent.com/assets/142251/16431034/3a5f9a08-3d44-11e6-9443-e5e375917a93.png)

mxgrey:
(3) makes the most sense to me. It's a shame that ``std::optional`` won't be available until C++17, because neither an extra flag nor additional boost dependency is very appealing to me. But for lack of a better solution, I'd be okay with either.

jslee02:
I would also prefer (3) and using flag rather than `boost::optional` unless we want to keep boost dependency for the long term.

psigen:
I would possibly propose: *(4) Ignore color values if a texture is provided*

It seems like if you were actually trying to tint things, using it this way would lead to very bad behavior if you had one textured model and one manually colored model.   The textured model would get tinted, but the colored model would completely change color regardless of the original colors.

mkoval:
I agree with @psigen's point.

The one exception is transparency, which behaves the same on both colored and textured meshes. I often want to make a model semi-transparent for visualization purposes, e.g. to geometry that would otherwise be hidden.

This is currently only possible by setting the color to white with non-unit alpha. Perhaps we could implement (4) and split transparency out into another field?

PR #1315, opened 2019-05-14
Enable MeshShape to override alpha value

jslee02:
This PR fixes that it was not able to change alpha value when the color mode of `MeshShape` is `MATERIAL_COLOR`.

***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
   * Added heightmap support to OSG renderer: [#1293](https://github.com/dartsim/dart/pull/1293)
   * Improved voxel grid and point cloud rendering performance: [#1294](https://github.com/dartsim/dart/pull/1294)
   * Fixed incorrect alpha value update of InteractiveFrame: [#1297](https://github.com/dartsim/dart/pull/1297)
+  * Fixed dereferencing a dangling pointer in WorldNode: [#1311](https://github.com/dartsim/dart/pull/1311)
   * Removed warning of ImGuiViewer + OSG shadow: [#1312](https://github.com/dartsim/dart/pull/1312)
 
 * Examples and Tutorials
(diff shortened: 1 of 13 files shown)


mxgrey:
I'm not sure how I feel about adding an `AlphaMode` concept to the `MeshShape`. I suspect that would make it harder for users to alter transparency to their meshes, because they'd have to iterate through their shapes and find which shapes are of the type `MeshShape` and then toggle the value for that shape.

What if instead we always multiply the mesh material's alpha value by the `VisualAspect`'s alpha value? That way if the `VisualAspect` has an alpha value of `1.0`, then the rendering will just have the normal material alpha value (so materials like glass can still be transparent), and the closer the `VisualAspect`'s alpha value gets to `0.0`, the more transparent the mesh will be, until it's invisible.

jslee02:
I see your point. I think this is because setting alpha (and the color) of `MeshShape` is already tricky. Users could have different expectations, and it seems your proposal is another behavior (maybe I call it blending mode?). One corner case I can think of is that it cannot handle the case that the user wants to set 1.0 alpha when the embedded alpha is lower than 1.0. So I wanted to make it explicit the behavior.

I would go with any better way as long as it can handle all the cases. My bottom line is that we should be able to set any value of the mesh alpha regardless of the color mode.

If we go with adding alpha mode, then I'm open to the default mode.

This is a little off topic, but I would like to add another color mode to `MeshShape`: tinting (or blending) the embedded mesh color (and texture) with the shape color. I haven't tested it myself recently but according to #739 we already allow it unintentionally. It would be good to make it explicit in that mode so that we can also disable it. One good use case of tinting is that tinting a skeleton or a mesh that is in a collision.

mxgrey:
> If we go with adding alpha mode, then I'm open to the default mode.

That sounds like a reasonable compromise. We could have the "alpha blend" mode be the default while still have a mode selection available from the `MeshShape` interface.

My only concern with loading so many "modes" into the `MeshShape` is that (by design) the specific shape type of a `Shape` instance is usually opaque to the user without manually inspecting the types, and I think that tends to make those APIs less accessible to users, to the point that a user might not realize that such functionality is available.

My ideal preference would be to have a way to manage these settings from the base `Shape` type, but I admit that I don't have a good suggestion for how to do that, since these modes are so specific to `MeshShape`.

> I would like to add another color mode to MeshShape: tinting

I can definitely see the value of that :+1:

jslee02:
Updated based on the comments. Hope the changes agrees with the compromise.

> My ideal preference would be to have a way to manage these settings from the base `Shape` type

Agree. I also couldn't come up with a better solution. Let's revisit this once we have more neat solution.

I saved the tinting option for future PRs.

PR #1336, opened 2019-05-25
Fix handling of submeshes

azeey:
Meshes with multiple submeshes are not handled properly when using ODE as the collision detector. This is because the indices for each submesh start from 0 in Assimp, but the input to ODE requires the vertices and indices of all submeshes to be merged.

The issue can manifest itself in various ways, but the simplest example I came up with is to create a plane and a box in one mesh and have a sphere fall onto the plane.

**Before**
dart_before_fix

**After**
dart_after_fix

I have attached the mesh for reference
[plane_with_box.zip](https://github.com/dartsim/dart/files/3218953/plane_with_box.zip)

### DIFF
--- a/dart/collision/ode/detail/OdeMesh.cpp
+++ b/dart/collision/ode/detail/OdeMesh.cpp
@@ -58,7 +58,7 @@ OdeMesh::OdeMesh(
       mOdeTriMeshDataId,
       mVertices.data(),
       3*sizeof(double),
-      static_cast<int>(mVertices.size()),
+      static_cast<int>(mVertices.size()/3),
       mIndices.data(),
       static_cast<int>(mIndices.size()),
       3*sizeof(int),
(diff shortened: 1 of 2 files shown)


dart/collision/ode/detail/OdeMesh.cpp
```diff
@@ -58,7 +58,7 @@ OdeMesh::OdeMesh(
       mOdeTriMeshDataId,
       mVertices.data(),
       3*sizeof(double),
-      static_cast<int>(mVertices.size()),
+      static_cast<int>(mVertices.size()/3),
```
jslee02: Could you elaborate a little why this is needed?

azeey: Sure. The signature of that function is
```c++
ODE_API void dGeomTriMeshDataBuildDouble1(dTriMeshDataID g, 
                                  const void* Vertices,  int VertexStride, int VertexCount, 
                                  const void* Indices, int IndexCount, int TriStride,
                                  const void* Normals);
```
The fourth argument is the number of vertices. `mVertices` actually has 3 elements (the x, y, z components) per vertex so we have to divide its size by 3 to get the number of vertices. I checked against [Gazebo's implementation](https://bitbucket.org/osrf/gazebo/src/default/gazebo/physics/ode/ODEMesh.cc#lines-141) to make sure.

PR #1344, opened 2019-05-29
removed pydart2 from readme

acxz:
It seems as tho pydart2 is not being maintained/updated with the current DART releases. As such this is creating problems for pydart2 users like so: https://github.com/sehoonha/pydart2/issues/22, https://github.com/sehoonha/pydart2/issues/17.

I believe it would be best to not refer users to using an unmaintained pydart2 now that DART has its own supported python bindings. (Amazing btw!)


jslee02:
I see the concern. I actually removed them from the README, but revert it because some users still use them and it'd be good to keep them as references.

How about we mark `dartpy` as recommended instead of removing `pydart2`? Or mark `pydart2` deprecated?

acxz:
I see, if that is the case then marking `pydart2` deprecated would be appropriate I think.

PR #1369, opened 2019-07-12
Fix friction and restitution of individual shapes in a body

### LINKED ISSUE #200: Multiple frictional properties per BodyNode
`BodyNode` stores one friction coefficient for the all collision shapes in the `BodyNode`. Multiple friction coefficient per body would be useful in case of a body consist of different materials. 

One possible way to do this is that stores the frictional properties in collision object. Currently, DART uses `Shape` classes for both of collision object and visualization object so we may need to consider split the class.

In addition, multiple frictional directions should be considered too.

jslee02:
`DynamicAspect` was introduced to allow to set different dynamics properties (including friction and restitution properties) per shape in a `BodyNode` for setting the dynamic properties of individual shapes in a `BodyNode`. However, the constraint solver still uses the dynamics properties stored in `BodyNode`.

This PR deprecates the friction and restitution property related APIs in `BodyNode` and change the constraint solver to use the dynamics properties in `DynamicAspect`.

Resolves #200.

***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### [DART 6.10.0 (20XX-XX-XX)](https://github.com/dartsim/dart/milestone/58?closed=1)
 
+* Dynamics
+
+  * Fix friction and restitution of individual shapes in a body: [#1369](https://github.com/dartsim/dart/pull/1369)
+
 * GUI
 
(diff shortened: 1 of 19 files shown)


scpeters:
this looks good to me. I'll try rebasing my branch onto this tomorrow morning

Issue #869, opened 2017-04-04
DART will wrongly prompt warning for setting initial joint positions out of range


jyf588:
Hello everyone,

I had this joint in my Skel file:

>             <joint type="prismatic" name="j_gripper2">
>                 <parent>forearm</parent>
>                 <child>gripper2</child>
>                 <axis>
>                     <xyz>0.0 0.0 -1.0</xyz>
>                     <limit>
>                         <lower>0.01</lower>
>                         <upper>0.10</upper>
>                     </limit> 
>                 </axis>
>                 <init_pos>0.03</init_pos>
>             </joint>

When I ran the simulation, DART gave me the following warning: 

>[1;33mWarning [GenericJoint.hpp:1289][0m [GenericJoint::setRestPosition] Value of _q0 [0], is out of the limit range [0.01, 0.1] for index [0] of Joint [Joint].

Actually I can see from my window that the init_pos is indeed correctly set to 0.03 instead of 0.0. I don't know why in function GenericJoint::setRestPosition(), it will think my init_pos is 0.0

Thanks,
Yifeng

mxgrey:
The rest position refers to the "zero spring potential" position that gets used when computing spring stiffness. The default value for this is zero.

Admittedly, this warning might be overly cautious, since many users probably don't even use the spring stiffness. Also, it's plausible that there are situations where the rest position is desired to be outside the position limits. We could consider removing this warning.

jyf588:
I see... The warning refers to the unspecified <spring_rest_position>, which has default 0 value. It has nothing to do with my <init_pos> value.

mxgrey:
Exactly. I'm glad you brought this up, though, because I think it's worthwhile to consider removing this warning.

PR #1374, opened 2019-07-18
Remove overly cautious warning (#869)

yurket:
Remove overly cautious warning

***

- [x] Document new methods and classes (N/A)
- [x] Format new code files using `clang-format` (N/A)

- [x] Set version target by selecting a milestone on the right side (N/A)
- [x] Summarize this change in `CHANGELOG.md` (N/A)

### DIFF
--- a/dart/dynamics/detail/GenericJoint.hpp
+++ b/dart/dynamics/detail/GenericJoint.hpp
@@ -1470,12 +1470,6 @@ void GenericJoint<ConfigSpaceT>::setRestPosition(size_t index, double q0)
   if (Base::mAspectProperties.mPositionLowerLimits[index] > q0
       || Base::mAspectProperties.mPositionUpperLimits[index] < q0)
   {
-    dtwarn << "[GenericJoint::setRestPosition] Value of _q0 [" << q0
-           << "], is out of the limit range ["
-           << Base::mAspectProperties.mPositionLowerLimits[index] << ", "
-           << Base::mAspectProperties.mPositionUpperLimits[index]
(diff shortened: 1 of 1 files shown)


yurket:
refers to #869

mxgrey:
Thanks for this PR!

I think rather than just removing the warning, we should remove the entire limit check. As it is right now in this PR, the function will quit without allowing the rest position to be set outside the limits. However, there might be a legitimate use case where the rest position is outside of the limits, where the passive dynamics will clamp the joint down onto its joint limit.

For that reason, I would suggest removing all of lines 1470-1474.

mxgrey:
Also with that change in behavior, I think it would be smart to set up a simple unit test.

The unit test could be as simple as a link attached to the world by a revolute joint where the revolute joint has some lower position limit, and its spring rest position is a lower value than its position limit (note that the spring stiffness should be something positive). Then the unit test can simulate the system for a second or two and check if the link comes to rest at its lower position limit.

jslee02:
@mxgrey Your suggestions all sound good to me. I think we can merge this PR and address your suggestions as a separate PR.

mxgrey:
To be honest I think this PR on its own is problematic because it removes a warning without removing behavior that is likely to surprise users. If we merge it in as-is, we risk the possibility that we forget to change this problematic behavior.

yurket:
Agree with you, guys. Silent return doesn't look good. I'll add a test for the behavior.

jslee02:
Thank you for your contribution. Closing in favor of #1418 that is a rework of this PR addressing @mxgrey's comments.

PR #1376, opened 2019-07-22
Use raw skeleton pointer

mxgrey:
I've begun profiling using the `speed_test` example application. The profiler indicated that one hotspot was actually coming from locking the `std::weak_ptr<Skeleton>` every time that `BodyNode::getSkeleton()` was called.

To deal with that, this PR introduces a new `getRawSkeleton()` function to `BodyNode`, `Joint`, and `DegreeOfFreedom` (it can and probably should be added to `Node` as well). When computing things like kinematics and dynamics internally, there's little point in verifying that the parent skeleton of a BodyNode is still valid, because it would imply that something has gone horribly wrong on the user's end anyhow.

These changes resulted in a ~12% overall reduction to the average runtime of the tests cases in [speed_test](https://github.com/dartsim/dart/blob/f54ee2dbec71518e76d14c6018ee7996d3bd4128/examples/speed_test/main.cpp) on my machine.

## Measurements before this change

get_skeleton

## Measurements after

raw_skeleton

It looks like the next most significant hotspots to consider addressing are `AdInvT`, `ode::_dDot`, and `ConstrainedGroup::getConstraint(~)`.

I'm not entirely sure if `AdInvT()` can be made significantly more performant than it already is, so maybe we can see if there are any spots that it's being used more than it really needs to be used. Perhaps we if we can find redundant uses of `AdInvT()` we can cache the results somewhere and do lazy evaluation.

***

- [ ] Format new code files using `clang-format`


costashatz:
Very good work! Thanks!

> ConstrainedGroup::getConstraint(~)

This is weird as this is just accessing an `std::vector`. This is supposed to be O(1), right? Also, the return type is pointer that should not really have a huge overhead. Except if the smart pointers are doing this..

mxgrey:
> This is weird as this is just accessing an std::vector.

I haven't looked much into this yet and I'm still wondering why in the world `AdInvT` would be such a hotspot besides the obvious fact that it gets called so often for the Featherstone ABA.

I think the timing info includes time spent on all function calls nested inside of `ConstrainedGroup::getConstraint()`, so I'll need to look at a flame graph to see if any particular constraint type is being a problem.

jslee02:
Could we target this PR to `master` for DART 6.10 since adding a new APIs?

jslee02:
> I'm not entirely sure if `AdInvT()` can be made significantly more performant than it already is

Probably, breaking down the operation for the linear part from
```
  res.tail<3>().noalias()
      = _T.linear().transpose()
        * (_V.tail<3>() + _V.head<3>().cross(_T.translation()));
```
to
```
  res.tail<3>() = _V.tail<3>();
  res.tail<3>().noalias() += _V.head<3>().cross(_T.translation()));
  res.tail<3>() = _T.linear().transpose() * res.tail<3>();
```
could help IF `.cross()` works with `.noalias()`. But I haven't tested if this really helps.

mxgrey:
I'm going to close this PR for now and retarget it to 6.10 later on.

I'm currently doing a sprint to squeeze as much performance out of dartsim as possible without making any behavioral changes, so I'll finish up all those changes and then open a PR for all of it when time permits.

I completely agree with the suggested changes, and I'll be sure to make those for the next iteration of this PR.

PR #1377, opened 2019-07-24
Add dartpy access to getTransformFrom{Parent|Child}BodyNode()

chhinze:
For now, those Isometry transformations could only be set, but not read.

1. Add wrapper to add get functionality.
2. Add python unittest to verify, that it is available and returns the same Isometry3.

[Remove this line and describe this pull request. Link to relevant GitHub issues, if any.]

***

- [ ] Document new methods and classes -> Should this be done for python bindings as well? The pybind11 code does not yet seem to be documented at all.

### DIFF
--- a/python/dartpy/dynamics/Joint.cpp
+++ b/python/dartpy/dynamics/Joint.cpp
@@ -257,7 +257,7 @@ void Joint(py::module& m)
           ::py::arg("T"))
       .def(
           "getTransformFromParentBodyNode",
-          +[](dart::dynamics::Joint* self) -> const Eigen::Isometry3d& {
+          +[](const dart::dynamics::Joint* self) -> const Eigen::Isometry3d& {
             return self->getTransformFromParentBodyNode();
           })
       .def(
--- a/python/dartpy/dynamics/Joint.cpp
+++ b/python/dartpy/dynamics/Joint.cpp
@@ -256,6 +256,16 @@ void Joint(py::module& m)
           },
           ::py::arg("T"))
       .def(
+          "getTransformFromParentBodyNode",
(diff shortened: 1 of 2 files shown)


python/dartpy/dynamics/Joint.cpp
```diff
@@ -255,6 +255,16 @@ void Joint(py::module& m)
             return self->setTransformFromChildBodyNode(T);
           },
           ::py::arg("T"))
+      .def(
+          "getTransformFromParentBodyNode",
+          +[](dart::dynamics::Joint* self) -> const Eigen::Isometry3d& {
```
jslee02:
```suggestion
          +[](const dart::dynamics::Joint* self) -> const Eigen::Isometry3d& {
```
It seems we could pass `self` by const pointer since this function doesn't modify `Joint`.

python/dartpy/dynamics/Joint.cpp
```diff
@@ -255,6 +255,16 @@ void Joint(py::module& m)
...
+            return self->getTransformFromParentBodyNode();
+          })
+      .def(
+          "getTransformFromChildBodyNode",
+          +[](dart::dynamics::Joint* self) -> const Eigen::Isometry3d& {
```
jslee02:
```suggestion
          +[](const dart::dynamics::Joint* self) -> const Eigen::Isometry3d& {
```

python/dartpy/dynamics/Joint.cpp
chhinze: Sure. Thank you for the suggestion.

chhinze:
Awesome. Thank you.

PR #1381, opened 2019-07-30
Fix refresh of LineSegmentShapeNode

### LINKED ISSUE #1380: Setting DataVariance=DYNAMIC causes simulation to slow and then stop
A explained in the forum [here](https://dartsim.discourse.group/t/setting-datavariance-dynamic-causes-simulation-to-slow-and-then-stop/95/3). Setting dataVariance to DYNAMIC or DYNAMIC_ELEMENTS causes the simulation to slow and then stop after approximately 10 seconds.

In the forum Grey explains the problem as..

It seems that [this line](https://github.com/dartsim/dart/blob/d8f2636cc3d36430b5de65405b0b13c136485f94/dart/gui/osg/render/LineSegmentShapeNode.cpp#L221) is likely the culprit, because it’s adding a primitive set without removing the previous one. So the set of line segments that need to be drawn get piled up on top of each other instead of having the new set replace the old set.

jslee02:
`LineSegmentShapeNode` keeps adding new line segments instead of refreshing with new line. See the details in #1380 and [this post](https://dartsim.discourse.group/t/setting-datavariance-dynamic-causes-simulation-to-slow-and-then-stop/95).

Resolves #1380.

***

### DIFF
--- a/dart/gui/osg/render/LineSegmentShapeNode.cpp
+++ b/dart/gui/osg/render/LineSegmentShapeNode.cpp
@@ -84,6 +84,7 @@ protected:
 
   ::osg::ref_ptr<::osg::Vec3Array> mVertices;
   ::osg::ref_ptr<::osg::Vec4Array> mColors;
+  ::osg::ref_ptr<::osg::DrawElementsUInt> mElements;
 };
 
 //==============================================================================
@@ -187,8 +188,10 @@ LineSegmentShapeDrawable::LineSegmentShapeDrawable(
   : mLineSegmentShape(shape),
     mVisualAspect(visualAspect),
     mVertices(new ::osg::Vec3Array),
-    mColors(new ::osg::Vec4Array)
+    mColors(new ::osg::Vec4Array),
+    mElements(new ::osg::DrawElementsUInt(::osg::PrimitiveSet::LINES))
 {
+  addPrimitiveSet(mElements);
   refresh(true);
 }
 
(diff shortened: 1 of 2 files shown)


dart/gui/osg/render/LineSegmentShapeNode.cpp
```diff
@@ -207,18 +210,16 @@ void LineSegmentShapeDrawable::refresh(bool firstTime)
     const common::aligned_vector<Eigen::Vector2i>& connections
         = mLineSegmentShape->getConnections();
 
-    ::osg::ref_ptr<::osg::DrawElementsUInt> elements
-        = new ::osg::DrawElementsUInt(::osg::PrimitiveSet::LINES);
-    elements->reserve(2 * connections.size());
+    mElements->resize(2 * connections.size());
```
mxgrey: I don't think we want to use `resize(~)` alongside `push_back(~)`, because that means we're going to make `mElements` the size `2 * connections.size()` and then later push back an additional `2 * connections.size()` number of new elements.

I would recommend using `reserve(2 * connections.size()` and then `clear()`.

mxgrey: Alternatively `resize(~)` and then `[2*i] = ` and `[2*i+1] = ` would also work, but that feels more risky to me.

jslee02: Thanks for catching this out. Let me fix this as suggested.

jslee02:
It was an easy fix! Thanks for the review.

Issue #1198, opened 2018-11-20
[suggestion] Please consider emphasizing the project name


yurivict:
Hi,

Please note that package for this project is sometimes called ```dartsim``` (ex. https://repology.org/metapackage/dartsim/versions), and is sometimes called ```libdart``` (ex. https://repology.org/metapackage/libdart/versions). You call it ```dart```, but there is also the programming language called ```dart```, see here https://repology.org/metapackage/dart/versions .

You may want to emphasize here in the README what the project name is. I suggest you choose something unique, either ```dartsim``` or ```libdart``` or maybe ```dart-robotics```.

I would like to create the FreeBSD port but I am in doubt what the port's name should be.

Your ```include/dart``` headers location and ```share/doc/dart``` documentation locations might conflict with the unrelated ```dart``` project.

Thank you,
Yuri

acxz:
Hello, have you taken a look at the following issue: https://github.com/dartsim/dart/issues/479? It might help you choose a name for the FreeBSD port.

jslee02:
The project name is DART, but we use `dartsim` as the package name when there is name confliction in the namespace of the package manager. For example, the [Homebrew package](https://github.com/Homebrew/homebrew-core/blob/master/Formula/dartsim.rb) and [ROS package](https://github.com/dartsim/ros-dartsim-release). One exception is that `libdart`, which is the Debian package name. The naming convention is `lib<name>` when it's a library so `libdart`.

I would recommend using `dartsim` for the FreeBSD port.

mxgrey:
We may want to consider using `libdartsim` for the library name as well for consistency's sake. And also change the cmake config to work like `find_package(dartsim ...)`. But those changes probably shouldn't be made until the next major release (7).

yurivict:
@jslee02 I followed your recommendation, thanks: https://www.freshports.org/misc/dartsim/

jslee02:
Awesome! Thanks for having DART in FreeBSD port!

yurivict:
You're welcome!

jslee02:
Closing this issue as there is nothing left to be addressed. Please reopen if I'm missing something here.

PR #1389, opened 2019-08-14
Support colcon workspace with bullet and ode

scpeters:
We have been using [colcon](https://colcon.readthedocs.io/en/released/index.html) to build gazebo and ignition packages from source in a workspace, similar to using catkin-tools with a catkin workspace. A cool feature of colcon is that it can automatically identify cmake dependencies based on the cmake project name and build packages in the workspace in the correct order without creating `package.xml` files.

* https://github.com/colcon/colcon-cmake/blob/master/colcon_cmake/package_identification/cmake.py#L132-L135

If a `package.xml` file is present in a package, it will use that instead of the automatic detection, so you can build a mixed workspace of plain cmake and ROS1 or ROS2 packages.

When attempting to use colcon in a workspace with ign-physics, dartsim, bullet, and ODE, it was not properly identifying the dependencies. I believe the following two steps are necessary:

* adding depend tags for `BULLET_PHYSICS` and `ODE` to the dartsim package.xml file
    * https://github.com/bulletphysics/bullet3/blob/master/CMakeLists.txt#L7
    * https://bitbucket.org/odedevs/ode/src/695ba8c6aa4a725b610514725dcb9ea77683236d/CMakeLists.txt#lines-3
* changing the cmake project name and package.xml package name to `DART` to match the generated `DARTConfig.cmake`

I'm not sure if there are cpack / rosdistro / rosdep implications of changing these things, but I think it helps with colcon. Here's a workspace file for testing:

* https://bitbucket.org/osrf/gazebodistro/raw/%20physics_dart_bullet_ode/ign-physics1.yaml

cc @mxgrey @jslee02 @azeey 

***

### DIFF
--- a/colcon.pkg
+++ b/colcon.pkg
@@ -1,3 +1,8 @@
+# Configuration file for colcon (https://colcon.readthedocs.io).
+#
+# Please see the doc for the details of the spec:
+#   - https://colcon.readthedocs.io/en/released/user/configuration.html#colcon-pkg-files
+
 {
   "name": "DART",
   "dependencies": ["BULLET_PHYSICS", "ODE"],
--- a/colcon.pkg
+++ b/colcon.pkg
@@ -1,0 +1,4 @@
+{
+  "name": "DART",
+  "dependencies": ["BULLET_PHYSICS", "ODE"],
+}


package.xml
```diff
@@ -3,7 +3,7 @@
   <!-- This is a Catkin package.xml file to optionally support building DART in
        a Catkin workspace. Catkin is not required to build DART. For more
        information, see: http://ros.org/reps/rep-0136.html -->
-  <name>dartsim</name>
+  <name>DART</name>
```
jslee02: IIRC, we used `dartsim` for a reason. Maybe it's already taken by the Dart language?

package.xml
```diff
@@ -24,6 +24,7 @@
   <build_depend>pkg-config</build_depend>
   <depend>assimp</depend>
   <depend>bullet</depend>
```
jslee02: Should we remove this line as adding `BULLET_PHYSICS`?

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ else()
   cmake_minimum_required(VERSION 3.5.1)
 endif()
 
-project(dart)
+project(DART)
```
jslee02: This may impact on many places that use `${PROJECT_NAME}` such as the target names and the component names, which shall breaks the backward compatibility in finding components.

We had the [consensus](https://github.com/dartsim/dart/issues/1198#issuecomment-464569220) on changing the project name to `dartsim`, but that's for the next major release.

I think the most minimal changes to address this issue without breaking the backward compatibility would be using the lowercased project name to everywhere used `${PROJECT_NAME}` by doing something like:
```
string(TOLOWER ${PROJECT_NAME} PROJECT_NAME_LOWERCASE)
```
similar to [`PROJECT_NAME_UPPERCASE`](https://github.com/dartsim/dart/blob/15b5af8a3ed46a8c5350f0d01e767ff90b14eb2e/CMakeLists.txt#L12).

scpeters: I think the main thing is that colcon wants the project name (specifically the text inside `project()`) to match the name passed to `find_package()`. So if you use `project(dartsim)`, then I think colcon would want to see `find_package(dartsim ...` in order to infer the dependency properly.

mxgrey:
Unfortunately there are some breaking changes that would be required for DART/dartsim to comply with the assumptions made by colcon's package information inference rules. I think it'll make sense to put these changes into dartsim-7, but the changes could pose significant problems for current users of DART/dartsim-6.

Instead of relying on colcon's inference rules, maybe we should explicitly [put a `colcon.pkg` file](https://colcon.readthedocs.io/en/released/user/configuration.html) into the DART codebase? Obviously this would be annoying as it would introduce a **third** source of package metadata into the source code (the other two being cmake and `package.xml`), but we would only use it as a stop-gap measure until dartsim-7, at which point we can make our cmake fully compliant with colcon's assumptions.

scpeters:
a `colcon.pkg` file sounds like the right thing to do, at least in the short term. I'll experiment with that tomorrow.

mjcarroll:
Colcon also has a mechanism for providing metadata so that you don't necessarily have to add `colcon.pkg` files to the packages.

It would be reasonable to add the information there as well: https://github.com/colcon/colcon-metadata-repository

scpeters:
Here is the content of an equivalent `colcon.pkg` file:

~~~
{
  "name": "DART",
  "dependencies": ["BULLET_PHYSICS", "ODE"],
}
~~~

scpeters:
I reverted the changes to CMakeLists.txt and package.xml and added a `colcon.pkg` file instead.

scpeters:
> @mjcarroll wrote:

> Colcon also has a mechanism for providing metadata so that you don't necessarily have to add colcon.pkg files to the packages.

> It would be reasonable to add the information there as well: https://github.com/colcon/colcon-metadata-repository

I guess we could add the `"dependencies": ["BULLET_PHYSICS", "ODE"],` line to the metadata repository, but I think we would still need the `colcon.pkg` here if we want to change the project name used by colcon so that downstream cmake software that uses `find_package(DART)` will recognize this dependency automatically.

colcon.pkg
```diff
@@ -0,0 +1,4 @@
+{
```
jslee02: Does this file support comments in the file? If so, it'd be great to leave a brief comment how this file is used for.

CMakeLists.txt
jslee02: Yeah, I meant to chaning them all to use `dartsim` instead of `dart` (or `DART`), but again it's for the next major release.

colcon.pkg
mjcarroll: It's YAML, so comments prefixed with `#` are acceptable here.

PR #1395, opened 2019-08-16
FindBullet: separate win32 optimized and debug libraries

scpeters:
I see a configuration error on windows when trying to build dart with bullet3 installed from `vcpkg`:

~~~
CMake Error at cmake/DARTFindBullet.cmake:16 (set_target_properties):
  Property INTERFACE_LINK_LIBRARIES may not contain link-type keyword
  "optimized".  The INTERFACE_LINK_LIBRARIES property may contain
  configuration-sensitive generator-expressions which may be used to specify
  per-configuration rules.
~~~ 

There are multiple optimized and debug libraries interleaved in the `BULLET_LIBRARIES` parameter, so we can't use the simple approach from https://github.com/dartsim/dart/commit/73ef8c2716b37ad8a1fdd5d1468eb88841408ec2. I've used `cmake_parse_arguments` to separate the `debug` and `optimized` libraries, and I think I set up the generator expressions correctly, but I'm not sure. I've added `bullet3` and `ode` to the list of vcpkg packages to install in azure-pipeline, so CI should confirm if this fix is working.

***

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -142,7 +142,8 @@ jobs:
   variables:
     VCPKG_INSTALL_ROOT: $(Build.SourcesDirectory)\vcpkg
     VCPKG_ARCH: 'x64-windows'
-    VCPKG_PACKAGES: 'assimp boost-system boost-filesystem bullet3 ccd eigen3 fcl ode'
+    VCPKG_PACKAGES: 'assimp boost-system boost-filesystem ccd eigen3 fcl'
+    VCPKG_OPTIONAL_PACKAGES: 'bullet3 ode'
     BUILD_TOOLSET_VERSION: '142'
     CMAKE_GENERATOR: 'Visual Studio 16 2019'
   steps:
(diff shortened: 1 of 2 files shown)


scpeters:
reproducing failure case on branch azure_bullet3_test with https://github.com/dartsim/dart/commit/bfd970f0fbec7eb816bfcd7f47d4b4d6b729d19c

* https://dev.azure.com/dartsim/dart/_build/results?buildId=1576

.azure-pipelines.yml
```diff
@@ -142,7 +142,7 @@ jobs:
   variables:
     VCPKG_INSTALL_ROOT: $(Build.SourcesDirectory)\vcpkg
     VCPKG_ARCH: 'x64-windows'
-    VCPKG_PACKAGES: 'assimp boost-system boost-filesystem ccd eigen3 fcl'
+    VCPKG_PACKAGES: 'assimp boost-system boost-filesystem bullet3 ccd eigen3 fcl ode'
```
jslee02: Nit: Could we have a separate list for the optional dependencies something like `VCPKG_OPTIONAL_PACKAGES: 'bullet3 ode'` and add the variable at the end of line 154?

cmake/DARTFindBullet.cmake
```diff
@@ -12,9 +12,19 @@
 find_package(Bullet COMPONENTS BulletMath BulletCollision MODULE QUIET)
 
 if((BULLET_FOUND OR Bullet_FOUND) AND NOT TARGET Bullet)
+  if(WIN32 AND "optimized" IN_LIST BULLET_LIBRARIES
+           AND     "debug" IN_LIST BULLET_LIBRARIES)
+    cmake_parse_arguments(BULLET_INTERFACE_LIBRARIES "" "" "debug;optimized"
```
jslee02: Nice! I've only used `cmake_parse_arguments` in writing custom CMake functions but didn't know that this can be used in this way.

Out of curiosity, could we use spaces (i.e., `"debug optimized") instead of semicolons?

.azure-pipelines.yml
scpeters: ff53e8625a2bc48c2793ad8482b2a708f617f135

cmake/DARTFindBullet.cmake
scpeters: no, it has to be a list so it needs `;`. I tested first with spaces and it didn't work

PR #1400, opened 2019-08-16
Azure: install vcpkg dependencies for utils component

scpeters:
I was testing the `utils` component on windows with `vcpkg install tinyxml2`, but I ran into compilation errors in [XmlHelpers.cpp](https://github.com/dartsim/dart/blob/master/dart/utils/XmlHelpers.cpp#L37-L38), which includes `boost/algorithm/string.hpp` and `boost/lexical_cast.hpp`, since those headers were not installed along with the required boost dependencies `boost-system` and `boost-filesystem`. I've added a check for these headers in `dart/utils/CMakeLists.txt` and added to the `VCPKG_OPTIONAL_PACKAGES` variable so the `utils` component will be compiled in the windows build.

***


.azure-pipelines.yml
```diff
@@ -143,7 +143,9 @@ jobs:
     VCPKG_INSTALL_ROOT: $(Build.SourcesDirectory)\vcpkg
     VCPKG_ARCH: 'x64-windows'
     VCPKG_PACKAGES: 'assimp boost-system boost-filesystem ccd eigen3 fcl'
-    VCPKG_OPTIONAL_PACKAGES: 'bullet3 ode'
+    # utils needs tinyxml2 and boost algorithm/lexical-cast
+    #   and also boost-math to resolve a circular dependency with lexical-cast
+    VCPKG_OPTIONAL_PACKAGES: 'boost-algorithm boost-lexical-cast boost-math bullet3 ode tinyxml2'
```
jslee02: Do you know if `boost-algorithm`, `boost-lexical-cast`, and `boost-math` are the actual boost components so that we can search them as `find_package(Boost COMPONENTS boost-algorithm boost-lexical-cast boost-math)`? Or, these are just separately packaged in vcpkg?

jslee02:
The build error seems to due to [this logic](https://github.com/dartsim/dart/blob/master/dart/gui/CMakeLists.txt#L9-L12) dosen't catch if GLUT is installed.

jslee02:
Let's `return()` in that logic for now to see if this PR works.

.azure-pipelines.yml
scpeters: I tried searching for them as `OPTIONAL_COMPONENTS` in DARTFindBoost.cmake, but it failed on Ubuntu because it couldn't find libraries matching those names, and these packages are header-only, so I don't think it works to find as a component; it's just a vcpkg thing

scpeters:
disabling win32 gui build in ea21cd0 for testing purposes

dart/utils/CMakeLists.txt
```diff
@@ -26,6 +50,7 @@ set(component_name utils)
 
 # Add target
 dart_add_library(${target_name} ${hdrs} ${srcs} ${dart_utils_headers} ${dart_utils_sources})
+target_include_directories(${target_name} PRIVATE ${Boost_INCLUDE_DIRS})
```
jslee02: Do we need this line? If those are only headers, the boost include directories should be added by [this line](https://github.com/dartsim/dart/blob/master/dart/CMakeLists.txt#L137).

scpeters: yeah, I guess not

scpeters: 3dc838c

jslee02:
> disabling win32 gui build in [ea21cd0](https://github.com/dartsim/dart/commit/ea21cd0a79bb25d73cc385adb623eebff8c92c32) for testing purposes

This should be fixed by #1405.

PR #1401, opened 2019-08-17
Use cmake message WARNING and SEND_ERROR when appropriate

scpeters:
When compiling with colcon, messages to stdout (such as `message(STATUS)` calls from cmake) are not displayed on the console, while messages to stderr (such as `message(WARNING)` and `message(SEND_ERROR)`) are displayed, which makes it much easier to spot configuration issues. I noticed that many places there are `message(STATUS)` calls notifying when optional dependencies could not be found, so I changed those to `message(WARNING)`. I also noticed one place in DARTFindassimp.cmake (e4ca1a8) that seemed like it should cause the build to fail, so I changed that from `message(STATUS)` to `message(SEND_ERROR)`.

***


scpeters:
I think `ASSIMP_VERSION` is not being set on windows?

scpeters:
Yeah, I guess it's only detecting the version with pkg-config, which isn't available on windows. I can revert that change or add an `if(NOT WIN32)`

jslee02:
It seems assimp 4.1 installs `assimp-config.cmake` and `AssimpConfigVersion.cmake` (on Windows as well). We could use config file on Windows instead of module file provided by DART by changing this [line](https://github.com/dartsim/dart/pull/1401/files#diff-58ccc61f68a32ffaaa1f63e3f6ef0457R9) to:
```cmake
if(WIN32)
  find_package(assimp REQUIRED CONFIG)
else()
  find_package(assimp REQUIRED MODULE)
endif()
```

I'm not sure if using the version config file would change the version variable name to `ASSIMP_PACKAGE_VERSION` because of [this line](https://github.com/assimp/assimp/blob/v4.1.0/AssimpConfigVersion.cmake.in#L1). If so `DARTFindassimp.cmake` should set `ASSIMP_VERSION` to `ASSIMP_PACKAGE_VERSION` becuase `ASSIMP_VERSION` is used in other places.

scpeters:
It looks like xenial has assimp 3.2, which has the cmake config files:

* https://packages.ubuntu.com/xenial/amd64/libassimp-dev/filelist

scpeters:
I'll split the assimp change to a separate pull request

cmake/DARTFindassimp.cmake
```diff
@@ -12,7 +12,7 @@ find_package(assimp REQUIRED MODULE)
 # doesn't allow different major number while DART is compatible any version
 # greater than or equal to 3.2.
 if(ASSIMP_VERSION VERSION_LESS 3.2)
```
jslee02: Using the config file seems not working as it is (https://dev.azure.com/dartsim/dart/_build/results?buildId=1636). Let's check the version only when the variable is valid to workaround.

```suggestion
if(ASSIMP_VERSION AND ASSIMP_VERSION VERSION_LESS 3.2)
```

scpeters: the whitespace looked funny in that suggestion, but I've committed it in 4567f84

jslee02: Yeah, it seems the suggestion feature is glitchy with indentation sometimes, but it's actually fine when it's applied. Anyways, the change looks good to me!

scpeters:
Sorry I saw an inconsistent warning message, so I updated it in 22a2fc7

PR #1407, opened 2019-08-27
Add joint velocity limit constraint support

### LINKED ISSUE #296: Add joint velocity limit constraint

jslee02:
This PR is to resolve #296 and https://dartsim.discourse.group/t/joint-velocities-over-the-velocity-limit/119.

~~This PR depends on #1406 (merged)~~

***


dart/dynamics/Joint.hpp
```diff
@@ -191,22 +191,46 @@ class Joint : public virtual common::Subject,
...
+  DART_DEPRECATED(6.10)
   bool isPositionLimitEnforced() const;
 
+  /// Returns whether enforcing joint position limit
+  ///
+  /// This enforcement is only enabled when the actuator type is PASSIVE or
+  /// FORCE.
+  ///
+  /// \sa ActuatorType
+  bool isPositionAndVelocityLimitEnforced() const;
```
mxgrey: Could we rename this something like `areLimitsEnforced()`? Would that be too vague?

mxgrey: Could it make sense to name the functions `bool areLimitsEnforced()` and `setLimitEnforcement(bool)` and allow the exact behavior to depend on the actuator mode? Or would that be too implicit?

jslee02:
@mxgrey Please let me know if you have any further concerns on this PR.

PR #1410, opened 2019-08-29
Enable URDF parser to read visual and collision names

jslee02:
***

### DIFF
--- a/unittests/unit/test_DartLoader.cpp
+++ b/unittests/unit/test_DartLoader.cpp
@@ -41,6 +41,7 @@ using namespace dart;
 using dart::common::Uri;
 using dart::utils::DartLoader;
 
+//==============================================================================
 TEST(DartLoader, parseSkeleton_NonExistantPathReturnsNull)
 {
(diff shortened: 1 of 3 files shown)


unittests/unit/test_DartLoader.cpp
```diff
@@ -486,3 +486,68 @@ TEST(DartLoader, KR5MeshColor)
     }
   }
 }
+
+//==============================================================================
+TEST(DartLoader, parseVisualCollisionName)
+{
+  std::string urdfStr
+      = "<robot name=\"testRobot\">                                       "
```
mxgrey: This is fine the way it is, but you might be interested in [raw string literals (6)](https://en.cppreference.com/w/cpp/language/string_literal), which work something like:

```
std::string urdfStr = R"RAW(
<robot name="testRobot">
  <link name="link_0">
    <visual name="first_visual">
<!-- ..etc ... -->
)RAW"
```

That will avoid a bunch of quotation marks and escape sequences for this block.

PR #1424, opened 2019-11-11
Add secondary friction coefficient parameter

scpeters:
Currently the ContactConstraint uses a friction pyramid with 2 directions, and it uses the same friction coefficient in each direction. We have found it useful in gazebo to be able to specify different values in each direction (anisotropic wheel friction for example). As such, I've added `mSecondaryFrictionCoeff` to the ShapeNode DynamicsAspect and updated the ContactConstraint to use this secondary coefficient. @azeey and I have additional code in his fork that adds parameters for specifying the friction directions and slip parameters, which will be addressed in subsequent pull requests.

[edited] To avoid a behavior change, the `mFrictionCoeff` member variable in the ShapeFrame DynamicsAspect is renamed to `mPrimaryFrictionCoeff` along with its auto-generated get/set functions. The existing auto-generated `getFrictionCoeff` and `setFrictionCoeff` are explicitly implemented such that `setFrictionCoeff` sets both friction parameters and `getFrictionCoeff` returns the average (and also returns `double` instead of `const double &`). I've added test coverage for these new functions. We have test cases in gazebo demonstrating anisotropic friction in case you'd like to add one here in the future.

***

- [X] Add unit test(s) for this change

### DIFF
--- a/dart/constraint/ContactConstraint.cpp
+++ b/dart/constraint/ContactConstraint.cpp
@@ -710,6 +710,27 @@ double ContactConstraint::computeFrictionCoefficient(
 }
 
 //==============================================================================
+double ContactConstraint::computePrimaryFrictionCoefficient(
+    const dynamics::ShapeNode* shapeNode)
+{
+  assert(shapeNode);
+
+  auto dynamicAspect = shapeNode->getDynamicsAspect();
+
+  if (dynamicAspect == nullptr)
+  {
+    dtwarn << "[ContactConstraint] Attempt to extract "
+           << "primary friction coefficient "
+           << "from a ShapeNode that doesn't have DynamicAspect. The default "
+           << "value (" << DART_DEFAULT_FRICTION_COEFF << ") will be used "
+           << "instead.\n";
+    return DART_DEFAULT_FRICTION_COEFF;
+  }
+
+  return dynamicAspect->getPrimaryFrictionCoeff();
+}
+
+//==============================================================================
 double ContactConstraint::computeSecondaryFrictionCoefficient(
     const dynamics::ShapeNode* shapeNode)
 {
@@ -719,7 +740,8 @@ double ContactConstraint::computeSecondaryFrictionCoefficient(
 
   if (dynamicAspect == nullptr)
   {
-    dtwarn << "[ContactConstraint] Attempt to extract friction coefficient "
+    dtwarn << "[ContactConstraint] Attempt to extract "
+           << "secondary friction coefficient "
            << "from a ShapeNode that doesn't have DynamicAspect. The default "
            << "value (" << DART_DEFAULT_FRICTION_COEFF << ") will be used "
            << "instead.\n";
--- a/dart/constraint/ContactConstraint.cpp
+++ b/dart/constraint/ContactConstraint.cpp
@@ -115,8 +115,10 @@ ContactConstraint::ContactConstraint(
   // TODO(JS): Assume the frictional coefficient can be changed during
   //           simulation steps.
   // Update mFrictionCoeff
-  const double frictionCoeffA = computeFrictionCoefficient(shapeNodeA);
-  const double frictionCoeffB = computeFrictionCoefficient(shapeNodeB);
+  const double primaryFrictionCoeffA =
+                       computePrimaryFrictionCoefficient(shapeNodeA);
+  const double primaryFrictionCoeffB =
+                       computePrimaryFrictionCoefficient(shapeNodeB);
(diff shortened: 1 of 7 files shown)


scpeters:
I've removed the WIP from the title since I think this is now ready for review.

scpeters:
I have a proposed fix for the osx CI in #1426

dart/dynamics/detail/ShapeFrameAspect.hpp
```diff
@@ -88,15 +88,20 @@ struct CollisionAspectProperties
...
 
+  /// Secondary coefficient of friction
+  double mSecondaryFrictionCoeff;
+
   /// Constructor
   DynamicsAspectProperties(
-      const double frictionCoeff = 1.0, const double restitutionCoeff = 0.0);
+      const double primaryFrictionCoeff = 1.0,
+      const double restitutionCoeff = 0.0,
+      const double secondaryFrictionCoeff = 1.0);
```
mxgrey: Instead of adding another function argument with a default of 1.0, I recommend overloading like this:

```
/// The frictionCoeff argument will be used for both primary and secondary friction
DynamicsAspectProperties(
    const double frictionCoeff = 1.0,
    const double restitutionCoeff = 0.0);

DynamicsAspectProperties(
    const double primaryFrictionCoeff,
    const double secondaryFrictionCoeff,
    const double restitutionCoeff);
```

I think this should be more intuitive and also should also maintain backwards compatibility.

dart/constraint/ContactConstraint.cpp
```diff
@@ -114,14 +114,21 @@ ContactConstraint::ContactConstraint(
   //----------------------------------------------
   // TODO(JS): Assume the frictional coefficient can be changed during
   //           simulation steps.
-  // Update mFrictionalCoff
+  // Update mFrictionCoeff
   const double frictionCoeffA = computeFrictionCoefficient(shapeNodeA);
```
mxgrey:
```suggestion
  const double primaryFrictionCoeffA = computeFrictionCoefficient(shapeNodeA);
```

Nitpick: I know it'll make for a bigger diff, but I think I'd recommend adding `primary` to these variable names for better symmetry.

dart/dynamics/detail/ShapeFrameAspect.hpp
scpeters: 727ca63

dart/constraint/ContactConstraint.cpp
scpeters: updated variable names in 2a0934f, and added the `computePrimaryFrictionCoefficient` function in 760cf3b

dart/dynamics/ShapeFrame.hpp
```diff
@@ -143,9 +143,17 @@ class DynamicsAspect final : public common::AspectWithVersionedProperties<
...
-  // const double& getFrictionCoeff() const;
+  /// Set both primary and secondary friction coefficients to the same value.
+  void setFrictionCoeff(const double& value);
+  /// Get average of primary and secondary friction coefficients.
+  double getFrictionCoeff() const;
+
+  DART_COMMON_SET_GET_ASPECT_PROPERTY(double, PrimaryFrictionCoeff)
+  // void setPrimaryFrictionCoeff(const double& value);
+  // const double& getPrimaryFrictionCoeff() const;
+  DART_COMMON_SET_GET_ASPECT_PROPERTY( double, SecondaryFrictionCoeff )
```
jslee02:
```suggestion
  DART_COMMON_SET_GET_ASPECT_PROPERTY(double, SecondaryFrictionCoeff)
```

dart/dynamics/detail/ShapeFrameAspect.hpp
```diff
@@ -88,16 +88,25 @@ struct CollisionAspectProperties
 
 struct DynamicsAspectProperties
 {
-  /// Coefficient of friction
-  double mFrictionCoeff;
+  /// Primary coefficient of friction
+  double mPrimaryFrictionCoeff;
```
jslee02: This breaks backward compatibility since we're changing the name of the public member variable. Could we keep the variable name and add custom functions (i.e., `set/getPrimaryFrictionCoeff`) to `DynamicAspect` instead of `DART_COMMON_SET_GET_ASPECT_PROPERTY(double, PrimaryFrictionCoeff)`?

scpeters: 2c506f3145c561d1cc80088029dc10cd7842a79c

PR #1427, opened 2019-11-26
Add friction direction unit vector and frame parameters

scpeters:
Currently the ContactConstraint uses a friction pyramid with 2 directions. In #1424, we added an additional parameter to allow the friction coefficient to be specified separately in each direction. This pull request allows the friction direction to be specified with a unit vector `FirstFrictionDirection` and pointer to a frame `FirstFrictionDirectionFrame` in which the unit vector is expressed. If `FirstFrictionDirection` is a vector of zeros (which is its default value), it is ignored and the default friction directions are used. If the frame pointer is `nullptr` (which is its default value), the body-fixed frame of that ShapeNode is used.

I've added two bodies to `test_Friction` with friction direction set using the default body-fixed frame and also using the world frame. In these tests, the friction is set to zero along one direction, and the lateral gravity component pushes along a diagonal.

***

- [X] Add unit test(s) for this change

### DIFF
--- a/dart/constraint/ContactConstraint.cpp
+++ b/dart/constraint/ContactConstraint.cpp
@@ -803,7 +803,7 @@ Eigen::Vector3d ContactConstraint::computeWorldFirstFrictionDir(
   }
 
   auto frame = dynamicAspect->getFirstFrictionDirectionFrame();
-  Eigen::Vector3d frictionDir = dynamicAspect->getFirstFrictionDirection();
+  const Eigen::Vector3d& frictionDir = dynamicAspect->getFirstFrictionDirection();
 
   // rotate using custom frame if it is specified
   if (frame)
--- a/dart/constraint/ContactConstraint.cpp
+++ b/dart/constraint/ContactConstraint.cpp
@@ -141,7 +141,7 @@ ContactConstraint::ContactConstraint(
     auto frictionDirA = computeWorldFirstFrictionDir(shapeNodeA);
     auto frictionDirB = computeWorldFirstFrictionDir(shapeNodeB);
 
-    // resulting friction direction unit vector
+    // check if the friction direction unit vectors have been set
     bool nonzeroDirA = frictionDirA.squaredNorm() >= DART_CONTACT_CONSTRAINT_EPSILON_SQUARED;
     bool nonzeroDirB = frictionDirB.squaredNorm() >= DART_CONTACT_CONSTRAINT_EPSILON_SQUARED;
 
--- a/dart/constraint/ContactConstraint.cpp
+++ b/dart/constraint/ContactConstraint.cpp
@@ -797,7 +797,7 @@ Eigen::Vector3d ContactConstraint::computeWorldFirstFrictionDir(
   {
     dtwarn << "[ContactConstraint] Attempt to extract friction direction "
            << "from a ShapeNode that doesn't have DynamicAspect. The default "
-           << "value (" << DART_DEFAULT_FRICTION_DIR << ") will be used "
+           << "value (" << DART_DEFAULT_FRICTION_DIR.transpose() << ") will be used "
            << "instead.\n";
     return DART_DEFAULT_FRICTION_DIR;
   }
--- a/dart/constraint/ContactConstraint.cpp
+++ b/dart/constraint/ContactConstraint.cpp
@@ -65,6 +65,8 @@ double ContactConstraint::mConstraintForceMixing = DART_CFM;
 
 constexpr double DART_DEFAULT_FRICTION_COEFF = 1.0;
 constexpr double DART_DEFAULT_RESTITUTION_COEFF = 0.0;
+const Eigen::Vector3d DART_DEFAULT_FRICTION_DIR =
+    Eigen::Vector3d::UnitZ();
 
 //==============================================================================
 ContactConstraint::ContactConstraint(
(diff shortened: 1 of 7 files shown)


scpeters:
cc @azeey

scpeters:
@mxgrey and @jslee02 we are getting close to a release of ignition citadel, and we'd really like to get these changes from our fork into dartsim

there is another pull request scheduled to come after this too that adds slip parameters

any time you could spend on a review would be greatly appreciated

dart/constraint/ContactConstraint.cpp
```diff
@@ -754,6 +785,35 @@ double ContactConstraint::computeSecondaryFrictionCoefficient(
...
+  {
+    dtwarn << "[ContactConstraint] Attempt to extract friction direction "
+           << "from a ShapeNode that doesn't have DynamicAspect. The default "
+           << "value (" << DART_DEFAULT_FRICTION_DIR << ") will be used "
+           << "instead.\n";
+    return DART_DEFAULT_FRICTION_DIR;
+  }
+
+  auto frame = dynamicAspect->getFirstFrictionDirectionFrame();
+  Eigen::Vector3d frictionDir = dynamicAspect->getFirstFrictionDirection();
```
jslee02:
```suggestion
  const Eigen::Vector3d& frictionDir = dynamicAspect->getFirstFrictionDirection();
```

dart/constraint/ContactConstraint.cpp
```diff
@@ -850,8 +910,10 @@ ContactConstraint::getTangentBasisMatrixODE(const Eigen::Vector3d& n)
   // Note: a possible speedup is in place for mNumDir % 2 = 0
   // Each basis and its opposite belong in the matrix, so we iterate half as
   // many times
-  T.col(0) = tangent;
-  T.col(1) = Eigen::Quaterniond(Eigen::AngleAxisd(0.5_pi, n)) * tangent;
+  // The first column is the same as mFirstFrictionalDirection unless
+  // mFirstFrictionalDirection is parallel to the normal
+  T.col(0) = Eigen::Quaterniond(Eigen::AngleAxisd(0.5_pi, n)) * tangent;
+  T.col(1) = tangent;
```
jslee02: It's not the part of this change, but I would like to negate the direction of `tangent` so that `T.col(0).cross(T.col(1)) == n` (currently, it's `T.col(1).cross(T.col(0)) == n`) because I believe it'd be more intuitive. This change wouldn't affect the behavior (but just flipping the sign of impulse for the second tangential friction).

It can be done by changing the order of all the cross products in this function. For example, [line 813](https://github.com/dartsim/dart/pull/1427/files#diff-ade956ee6a4569052e9d25915f8715aaL813) should be changed to:
```cpp
Eigen::Vector3d tangent = n.cross(mFirstFrictionalDirection);
```
or simply negating the final `tangent`. I'd prefer changing the order of the cross products though.

dart/dynamics/ShapeFrame.cpp
```diff
@@ -57,7 +57,9 @@ DynamicsAspectProperties::DynamicsAspectProperties(
     const double frictionCoeff, const double restitutionCoeff)
   : mFrictionCoeff(frictionCoeff),
     mRestitutionCoeff(restitutionCoeff),
-    mSecondaryFrictionCoeff(frictionCoeff)
+    mSecondaryFrictionCoeff(frictionCoeff),
+    mFirstFrictionDirection(Eigen::Vector3d::Zero()),
```
jslee02: Maybe we want to set the default direction to non-zero direction (e.g., `Eigen::Vector3d::UnitZ()`) so that the default behavior is more consistent in terms of the friction direction?

scpeters: my intent was that an unset vector with zero length would use the global friction direction, and that only friction directions that are explicitly set to have a nonzero length would use their custom direction and frame. See the following logic in ContactConstraint.cpp

* https://github.com/dartsim/dart/blob/a932d87bbbf1a15fd83d2a2a8fb78fe9d4cb3bad/dart/constraint/ContactConstraint.cpp#L148-L172

does that make sense?

dart/constraint/ContactConstraint.cpp
scpeters: I'm not opposed to that, but I'll bring @azeey into the conversation since he made this specific change

dart/dynamics/ShapeFrame.cpp
jslee02: Yeah, I read the logic but was suggesting changing the default behavior to using local friction direction (by setting `mFirstFrictionDirection` to non-zero vector intentionally) because first/second friction coefficients would make more sense when they're aligned with local directions (if they have different values).

Downside of it is that it changes the default simulation behavior (from using global friction direction to local friction direction) though. I'm okay to leave as it is as long as the logic is documented.

scpeters: a tricky aspect of the friction directions is choosing which to use when custom directions are set for both bodies. I've added logic in this PR to choose the direction from the body with the smaller primary friction coefficient, since that would be the dominant parameter, which makes sense to me, but I haven't tested it, and I'd be wary of making a big behavior change like that without testing.

Is the logic sufficiently documented? If not, I'll add the necessary documentation.

dart/constraint/ContactConstraint.cpp
azeey: Negating the direction of `tangent` sounds fine to me if it doesn't affect the behavior. I think the main change here is that `T.col(0)` has to be parallel to the first frictional direction.

scpeters: reversed in 5fe40b747b243875b655663dc458edd7ef3f8a1c

dart/dynamics/ShapeFrame.cpp
jslee02: Agreed to not changing the default behavior.

It's well documented in the implementation, but it'd be great if it's also documented in the header as well (maybe on the constructor?).

scpeters: how does cddef7a050b9aa9a5835d48a9b7c4f02de0f4934 look to you?

jslee02: Looks awesome! Thank you for adding it.

PR #1430, opened 2019-12-13
Compute the IK error in the target reference frame rather than its parent

parkerowan:
This allows for relaxing different axis constraints in the target reference frame using existing error weights, e.g. ignoring rotation about the z vector of the target frame.

***

- [ ] Document new methods and classes (modification to existing method)

### DIFF
--- a/dart/dynamics/InverseKinematics.cpp
+++ b/dart/dynamics/InverseKinematics.cpp
@@ -508,20 +508,17 @@ Eigen::Vector6d InverseKinematics::TaskSpaceRegion::computeError()
   // valid constraint manifold faster than computing it from the edge of the
   // Task Space Region.
 
-  // Use the target's transform with respect to its reference frame
-  const Eigen::Isometry3d& targetTf = mIK->getTarget()->getRelativeTransform();
-  // Use the actual transform with respect to the target's reference frame
-  const Eigen::Isometry3d& actualTf
-      = mIK->getNode()->getTransform(mIK->getTarget()->getParentFrame());
+  // Use the actual transform with respect to the target reference frame
+  const Eigen::Isometry3d& errorTf
+      = mIK->getNode()->getTransform(mIK->getTarget().get());
 
-  // ^ This scheme makes it so that the bounds are expressed in the reference
-  // frame of the target
+  // ^ This scheme makes it so that bounds are expressed in the target frame
 
-  Eigen::Vector3d p_error = actualTf.translation() - targetTf.translation();
+  Eigen::Vector3d p_error = errorTf.translation();
   if (mIK->hasOffset())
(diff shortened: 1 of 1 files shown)


mxgrey:
My memory from writing this is a bit hazy, so I'd need to think more about it to be certain, but if I remember correctly I think I implemented it like this to accurately represent the formal definition of a task space region. I think the difference in the frame of the target and the frame of the bounds is intentional because you may want to represent the desired end effector orientation independently of the bound coordinates.

A user could accomplish the effect of this PR by creating two `SimpleFrame` objects:

1. a `SimpleFrame` to represent the parent frame that defines the coordinates of the constraint bounds, and

2. a `SimpleFrame` which is a child of (1) to represent the target.

I think this PR would make us lose this flexibility.

That being said, I'd be open to optionally allowing users to explicitly specify a reference frame for the bounds of their Task Space Region. It could be added as a property to the `InverseKinematics::TaskSpaceRegion` class which defaults to a `nullptr`, and if it's a `nullptr` then the parent frame of the target will be used. This would have the bonus effect of preserving existing behavior.

jslee02:
@mxgrey Thanks for the clarification. Reading the paper, I got why the parent of the target is used as the reference of bounds in the code; The parent of the target frame is the reference frame of TSR, denoted as `T_w`.

In addition to your suggestion (allowing to explicitly specifying the reference frame), I'd like to add a simpler error method that uses the target frame for the reference frame of bounds as well for the users who are not familiar with the TSR's frame conventions. Then the logic in this PR could be in that error method.

mxgrey:
I have nothing against making it a separate error method, but one option to consider is to instead make a function that still returns a `TaskSpaceRegion` object which is configured to behave the way that's being discussed. That would reduce some duplicate implementation, since the behavior that's being proposed is just a subset of what a `TaskSpaceRegion` can do.

jslee02:
> I have nothing against making it a separate error method, but one option to consider is to instead make a function that still returns a `TaskSpaceRegion` object, but have that object configured to behave the way that's being discussed. That would reduce some duplicate implementation, since the behavior that's being proposed is just a subset of what a `TaskSpaceRegion` can do.

All sounds good to me!

parkerowan:
@mxgrey, @jslee02 Thanks for the feedback!

> That being said, I'd be open to optionally allowing users to explicitly specify a reference frame for the bounds of their Task Space Region. It could be added as a property to the InverseKinematics::TaskSpaceRegion class which defaults to a nullptr, and if it's a nullptr then the parent frame of the target will be used. This would have the bonus effect of preserving existing behavior.

This sounds like a good way to go for backward compatibility and flexibility.  I'll update the PR and add tests to capture both cases.

jslee02:
Closing this PR in favor of https://github.com/dartsim/dart/pull/1548

PR #1445, opened 2020-04-20
Increment body node version after it has been moved to a new skeleton

mxgrey:
An issue with auto-updating collisions was found in [ign-physics](https://github.com/ignitionrobotics/ign-physics/pull/31). This PR ports the fix into DART itself.

I will be pushing a regression test for this issue shortly.

***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## DART 6
 
+### [DART 6.9.3 XXXXX]
+
+* Dynamics
+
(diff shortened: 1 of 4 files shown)


mxgrey:
Sorry, that's what I meant for the target branch to be. It's updated now.

mxgrey:
I've added a unit test that fails if [this line](https://github.com/dartsim/dart/pull/1445/files#diff-2cfef747e095be4b3b039f1578189d45R2211) gets removed.

PR #1452, opened 2020-04-26
Wrap IKFast solutions for cyclic joint positions

### LINKED ISSUE #1449: ikFast class checks joint limits incorrectly
#### Environment

Irrelevant, but on Dart 6.9 with GCC 7.5 on Bionic.

#### Expected Behavior

All valid solutions being returned through ikFast.

#### Current Behavior
https://github.com/dartsim/dart/blob/660b0f4d84b56cfb723eb5e50e2e465f49930f5a/dart/dynamics/IkFast.cpp#L139

It doesn't look like ikFast gives any promises for the range of returned ik solutions. For example, with a joint with limits at [pi/8, 2pi], the ikFast binary may return a solution at -pi. 

Checking this raw value against the skeleton limits (as is done now) appears to lead to valid solutions being thrown away (in the example above, we can just wrap the joint value to pi, which is actually in-limits).

It looks like some wrapping is needed in the joint limit checking here, since ikFast doesn't provide a guarantee on the range as mentioned earlier.

#### Steps to Reproduce
I have a Jaco ikFast binary I'd be happy to share where this happens.

#### Code to Reproduce
Simply run the provided binary: almost all solutions are thrown away.

jslee02:
IKFast sometimes returns solutions that are not wrapped within the range of the joint limits, which is regarded as an invalid solution in DART.

This PR fixes the issue by adding logic that tries to wrap IKFast solutions for rotational joints. The idea is to wrap the solution within joint limits. If there is more than one solution (when the joint limit range is wider than 2pi), it returns the closest solution to the current joint position to minimize joint motion.

The logic should be applied to any DegreeOfFreedom whose configuration space is SO(2), but this PR only applies it to RevoluteJoint. This is because the current API doesn't provide for it. `DegreeOfFreedom::isCyclic()` is the closest function to what we want, but it only returns true when the rotational joint has no limits.

This PR partially fix #1449 because it only works for `RevoluteJoint`.

***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Kinematics
 
   * Added IkFast parameter accessors to IkFast class: [#1396](https://github.com/dartsim/dart/pull/1396)
+  * Changed IkFast to wrap IK solutions into the joint limits for RevoluteJoint: [#1452](https://github.com/dartsim/dart/pull/1452)
 
 * Dynamics
 
--- a/dart/dynamics/IkFast.cpp
+++ b/dart/dynamics/IkFast.cpp
@@ -34,6 +34,7 @@
 
 #include "dart/dynamics/BodyNode.hpp"
 #include "dart/dynamics/DegreeOfFreedom.hpp"
+#include "dart/dynamics/RevoluteJoint.hpp"
 #include "dart/external/ikfast/ikfast.h"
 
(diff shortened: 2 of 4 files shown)


jslee02:
Merging as it seems the build failure on Windows is due to the ODE vcpkg (incorrect SHA value).

PR #1453, opened 2020-04-27
Build without ODE on Windows CI builds

jslee02:
To suppress build error: https://github.com/dartsim/dart/runs/621262581?check_suite_focus=true#step:3:1096

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -187,8 +187,8 @@ jobs:
         VCPKG_PACKAGES: 'assimp boost-system boost-filesystem ccd eigen3 fcl'
         # 'dart-utils' needs tinyxml2 and boost algorithm/lexical-cast
         #   and also boost-math to resolve a circular dependency with lexical-cast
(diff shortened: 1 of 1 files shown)


traversaro:
FYI the vcpkg upstream error should be fixed (once merged) by https://github.com/microsoft/vcpkg/pull/11029 .

jslee02:
@traversaro Wow, thanks! I should not merge this unless https://github.com/microsoft/vcpkg/pull/11029 is not merged in time. :smile:

jslee02:
Well, on second thought, this should be merged for now because we use pre-installed vcpkg on GitHub Actions. :disappointed: Then revert once vcpkg is updated with your patch.

traversaro:
You probably already know about this, but to avoid this kind of problems you could try the GitHub Action cache feature (https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows). It is quite convenient and it also have a flexible invalidation mechanism (so you can decide if the cache is invalidate if the vcpkg version changes or just when you explicitly invalidate it).

jslee02:
It looks very promising! I know Travis and Appveyor support similar features but didn't know if GitHub Actions also supports it. Thanks!

PR #1455, opened 2020-04-29
Add python bindings for pointcloud and contact retrieval

chhinze:
1. Adds Python bindings for the following:
     - Extracting contacts from `CollisionResult` object (Methods `getContact` and `getContacts`)
    - `PointCloudShape`, so it is now possible to draw pointclouds in dartpy as well

2. Adds an example `python/examples/contacts_pointcloud/main.py`, which is basically the `hello_world_gui` example (robot and ground are set transparent) with detection of  contacts and visualization as a pointcloud before each GUI refresh.

***

- [x] Add ~~unit test(s)~~ Example for this change

### DIFF
--- a/python/examples/contacts_pointcloud/main.py
+++ b/python/examples/contacts_pointcloud/main.py
@@ -1,0 +1,64 @@
+import dartpy as dart
+
+class ContactVisualizingNode(dart.gui.osg.RealTimeWorldNode):
+    def __init__(self, world, pointCloudShape):
+        super(ContactVisualizingNode, self).__init__(world)
+        self._pointCloudShape = pointCloudShape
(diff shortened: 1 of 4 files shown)


python/examples/contacts_pointcloud/main.py
```diff
@@ -0,0 +1,112 @@
...
+"""
+
+
+class ContactVisualizingNode(dart.gui.osg.RealTimeWorldNode):
+    """ WorldNode subclass, which extracts the contacts of the last simulation
+    step and adds them to a pointloud for visualization.
+
+    Note: Simulation must run for the contacts to be displayed.
+
+     """
```
jslee02, line 19:
```suggestion
    """
```

chhinze:
Thank you for reviewing and merging.

PR #1457, opened 2020-05-03
Fix/py collision obj

chhinze:
Adds Python bindings to get more complete Contact/Collision information, especially to get `BodyNode`s involved in a particular contact via `someContact.collisionObject1.getShapeFrame().getBodyNodePtr()`. The following changes are made

- `CollisionObject`'s methods from C++ were completed
- `CollisionObject` declaration added in `python/dartpy/collision/module.cpp` -> Collision objects can be used now from Python code
- add `getBodyNodePtr()` for `dartpy::dynamics::ShapeNode`. Here, a discussion may be needed. As `getBodyNodePtr()` returns a `dart::dynamics::BodyNodePtr` (not implemented in dartpy yet) the easiest implementation was to return the raw `BodyNode*` using `getBodyNodePtr().get()`. So *reference counting of dart will be wrong here*. Can this become a problem, or is this okay? Should it be implemented via adding `BodyNodePtr` objects as well?

***

- [ ] ~~Document new methods and classes~~

### DIFF
--- a/python/dartpy/collision/CollisionObject.cpp
+++ b/python/dartpy/collision/CollisionObject.cpp
@@ -41,10 +41,27 @@ namespace python {
 void CollisionObject(py::module& m)
 {
   ::py::class_<dart::collision::CollisionObject>(m, "CollisionObject")
-      .def(
+    .def(
+        "getCollisionDetector",
+        +[](dart::collision::CollisionObject* self) -> dart::collision::CollisionDetector* {
+            return self->getCollisionDetector();
+        },
+        ::py::return_value_policy::reference_internal)
+    .def(
+        "getShapeFrame",
+        +[](const dart::collision::CollisionObject* self) -> const dart::dynamics::ShapeFrame* {
+            return self->getShapeFrame();
+        },
+        ::py::return_value_policy::reference_internal)
+        .def(
           "getShape",
           +[](const dart::collision::CollisionObject* self)
-              -> dart::dynamics::ConstShapePtr { return self->getShape(); });
+              -> dart::dynamics::ConstShapePtr { return self->getShape(); })
+        .def("getTransform",
+        +[](const dart::collision::CollisionObject* self) -> const Eigen::Isometry3d& {
+            return self->getTransform();
+        },
+        ::py::return_value_policy::reference_internal);
 }
 
 } // namespace python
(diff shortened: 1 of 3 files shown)


chhinze:
I just found out that there is a mapping of `TemplateBodyNodePtr<T>` as custom smart pointer already in https://github.com/dartsim/dart/blob/35bd08e008fda8fb9a7dee4f063d2d13ff0cb583/python/dartpy/dynamics/BodyNode.cpp#L39 
As far as I understand, the `PYBIND11_DECLARE_HOLDER_TYPE` macro will implicitly convert the representation, so by changing the return values from `dart::dynamics::BodyNodePtr` to `dart::dynamics::BodyNode*` implicit conversion of this holder type would take place. This seems to be the pybind11 way to embed custom smart pointers (<https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#custom-smart-pointers>). But the implicit conversion would cause the reference counting object to run out of scope after the function call. This would defeat the purpose of having a reference counting object in the first place, right? 

If we won't go this way, the other solution would be to implement `BodyNodePtr` as a simple class. In this case, I have the following questions:
- Do you have a preferred place for the implementation?
- The only methods to implement would be `get()`, `set()` and constructors, or am I missing something?

jslee02:
> I just found out that there is a mapping of `TemplateBodyNodePtr<T>` as custom smart pointer already in

Good catch! As the holding type of `BodyNode` is `BodyNodePtr`, we can simply return `BodyNodePtr` instead returning the raw pointer (using `.get()`). The runtime error I mentioned above was just because `PYBIND11_DECLARE_HOLDER_TYPE(T, dart::dynamics::TemplateBodyNodePtr<T>, true);` is declared later than the `Node` bindings. The runtime error would disappear by adding the declaration in `python/dartpy/dynamics/Node.cpp` as well or moving it to
https://github.com/dartsim/dart/blob/35bd08e008fda8fb9a7dee4f063d2d13ff0cb583/python/dartpy/dynamics/module.cpp#L82

Let me create a PR to resolve the runtime error. Then you can simply return `BodyNodePtr` in this PR.

jslee02:
#1465 is created based on this PR. Could you check the PR if it resolves the issues this PR attempts to fix? By #1465, you should be able to access the body node from a shape node as `shape_node.getBodyNodePtr()` that returns `dart::dynamics::BodyNodePtr`.

jslee02:
Closing this PR in favor of #1465

chhinze:
Thank you for your efforts.

PR #1489, opened 2020-08-27
Increment body node version after it has been removed from a new skeleton

jslee02:
#1445 was introduced to increment body node version after it's being added to a skeleton as a fix for the auto-updating collisions. However, this change doesn't cover the case that a body node is removed from a skeleton instead of transferring to a new skeleton. This PR is a fix for the case by incrementing the version when a body node is being removed from the skeleton.

***


mxgrey (approved):
Whoops, yeah, sorry for overlooking this :sweat_smile:

jslee02:
No worries. #1445 was a great hint for me to quickly fix the issue :)

PR #1434, opened 2020-02-19
Add support new if() IN_LIST operator in DARTConfig.cmake

jslee02:
Fixes #1429 

***

**Before creating a pull request**

- [N/A] Document new methods and classes
- [x] Format new code files using `clang-format`

**Before merging a pull request**

- [x] Set version target by selecting a milestone on the right side
- [x] Summarize this change in `CHANGELOG.md`
- [N/A] Add unit test(s) for this change

PR #1494, opened 2020-08-30
Support IN_LIST operator in DARTConfig.cmake (6.9 backport)

scpeters:
In DART 6.9.3, the `IN_LIST` keyword in DARTFindBullet caused a configuration error in gazebo10 (see https://github.com/osrf/gazebo/pull/2835). Setting `CMP0057` to new should prevent similar problems in downstream packages.

***

### DIFF
--- a/cmake/DARTFindBullet.cmake
+++ b/cmake/DARTFindBullet.cmake
@@ -12,6 +12,7 @@
 find_package(Bullet COMPONENTS BulletMath BulletCollision MODULE QUIET)
 
 if((BULLET_FOUND OR Bullet_FOUND) AND NOT TARGET Bullet)
+  cmake_policy(SET CMP0057 NEW)
   if(WIN32 AND "optimized" IN_LIST BULLET_LIBRARIES
            AND     "debug" IN_LIST BULLET_LIBRARIES)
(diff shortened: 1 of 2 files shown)


cmake/DARTFindBullet.cmake
```diff
@@ -12,6 +12,7 @@
 find_package(Bullet COMPONENTS BulletMath BulletCollision MODULE QUIET)
 
 if((BULLET_FOUND OR Bullet_FOUND) AND NOT TARGET Bullet)
+  cmake_policy(SET CMP0057 NEW)
```
jslee02, line 15: The error could occur anywhere `IN_LIST` is used. Also, this policy setting isn't necessary if the required CMake version is set from the parent project (e.g., gazebo). Let's fix this in `DARTConfig.cmake.in` like [this one](https://github.com/dartsim/dart/pull/1434/files#diff-90586bbb8c55cc07024442a0d1fa71feR13-R26).

scpeters: ok, I'll copy that approach here

scpeters: fc749f6

jslee02: Awesome, thanks!

Issue #1056, opened 2018-04-25
Proposal: Provide BUILD_<component> cmake variables


mxgrey:
It seems that currently we assume that users want to configure and build all dartsim components whose dependencies are available.

Instead, we could provide `SKIP_<component>` variables so that users can decide to skip components that they don't care about, even if the dependency is available, e.g. by calling `cmake .. -DSKIP_planning`. (Alternatively, we could call it `skip_<component>` to avoid mixing cases.)

jslee02:
I had the same feeling seeing #1045!

But I would prefer `BUILD_<component>` (or `DART_BUILD_<component>`) **option** by setting the default as `ON`, which I believe it's a more standard way.

mxgrey:
Sure, a build option that defaults to on sounds good to me.

mxgrey:
There are a lot of possible variable names we could use, so let's do an informal Github poll by putting in reactions (note that you can add multiple reactions):

:+1:: All uppercase (e.g. `BUILD_PLANNING`)
:-1:: All lowercase (e.g. `build_planning`)
:laughing:: Lowercase component name, uppercase everything else (e.g. `BUILD_planning`)

(Ignore case for the following options)
:confetti_ball: (Hooray): `BUILD_<component>`
:confused:: `DART_BUILD_<component>`
:hearts:: `BUILD_DART_<component>`

jslee02:
Hm, from my experience, I've seen that uppercase is used for options and global variables, and lowercase is used for local variables.

mxgrey:
I think historically cmake has used a lot of uppercase, but it's been transitioning away from that. During that transition, an informal convention arose where uppercase was kept for global variables and for "returned" variables (e.g. `<PACKAGE>_LIBRARIES`, `<PACKAGE>_INCLUDE_DIRS`) while lowercase was used for local variables. I suspect the actual reason for the continued use of uppercase global and returned variables is it's more dangerous to change the case for them, because downstream projects may depend on them.

[...]

PR #1497, opened 2020-09-03
CMake: Add DART_SKIP_<dep> advanced options

traversaro:
Add DART_SKIP_<dep> option to permit to specify that a dependency should not used even if it is found in the system.

This is useful in package managers such as vcpkg to ensure that, for a given set of enabled features, the same version of the library is build, independently from which dependency are actually installed in the system. A mpre extended rationale is provided in https://github.com/ignitionrobotics/ign-cmake/issues/63 .

### DIFF
--- a/cmake/DARTMacros.cmake
+++ b/cmake/DARTMacros.cmake
@@ -127,7 +127,9 @@ endfunction()
 
 #===============================================================================
 macro(dart_check_optional_package variable component dependency)
-  if(${${variable}_FOUND})
+  option(DART_SKIP_${variable} "If ON, do not use ${variable} even if it is found." OFF)
+  mark_as_advanced(DART_SKIP_${variable})
+  if(${${variable}_FOUND} AND NOT ${DART_SKIP_${variable}})
     set(HAVE_${variable} TRUE CACHE BOOL "Check if ${variable} found." FORCE)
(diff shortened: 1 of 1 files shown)


jslee02:
We had a similar discussion in #1056 that is to control which component to be built (using CMake option like `DART_BUILD_<component>`). In this way, we could disable the components that require the dependencies you want to skip. Do you think it would also have the same effect?

traversaro:
> 
> 
> We had a similar discussion in #1056 that is to control which component to be built (using CMake option like `DART_BUILD_<component>`). In this way, we could disable the components that require the dependencies you want to skip. Do you think it would also have the same effect?

I think that could also be used in that context.

traversaro:
Hi @jslee02, so do you prefer that I close this PR and I open implementing what was discussed in https://github.com/dartsim/dart/issues/1056 ?

jslee02:
@traversaro Yeah, but let me create a PR based on the discussion in #1056 as I already started it if it sounds good to you as well.

traversaro:
Perfect, thanks!

jslee02:
Merging this PR resolves the described issue without hurting the current build system. Will revisit to consider method discussed in #1056 if needed

traversaro:
Thanks @jslee02 !

PR #1509, opened 2020-10-12
Fix compatibility with clang-cl

traversaro:
clang-cl is the Visual Studio-compatible version of clang (https://clang.llvm.org/docs/MSVCCompatibility.html).
This PR: 
* Enables compilation of DART even for VS < 2019 if Clang is used (as clang-cl can be perfectly C++14 compatible even if the used MSVC version is not)
* Avoid to pass the /MP option to clang-cl that does not support it
* Adds a GitHub Actions job to compile DART with clang-cl (using in particular the ClangCl toolset shipped with MSVC)

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -204,3 +204,34 @@ jobs:
           -DDART_MSVC_DEFAULT_OPTIONS=ON ^
           -DDART_VERBOSE=ON
         cmake --build . --target ALL_BUILD -- /maxcpucount:4
+        
+  windows_2019_clang_cl:
+    name: Windows (clang-cl) [Release]
+    runs-on: windows-2019
+    env:
+      COMPILER: gcc
+      BUILD_TYPE: Release
+      RUN_TESTS: OFF
+      VCPKG_ROOT: 'C:/dartsim/vcpkg'
+      VCPKG_BUILD_TAG: v0.1.1
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install Dependencies
+      shell: cmd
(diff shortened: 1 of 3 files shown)


.github/workflows/ccpp.yml
```diff
@@ -204,3 +204,34 @@ jobs:
           -DDART_MSVC_DEFAULT_OPTIONS=ON ^
           -DDART_VERBOSE=ON
         cmake --build . --target ALL_BUILD -- /maxcpucount:4
+        
+  windows_2019_clang_cl:
+    name: Windows (clang-cl) [Release]
+    runs-on: windows-2019
+    env:
+      COMPILER: gcc
```
jslee02, line 212:
```suggestion
      COMPILER: clang
```
I know this wouldn't have any effect, but just to avoid some confusions

jslee02 (changes requested):

.github/workflows/ccpp.yml
traversaro: Done in https://github.com/dartsim/dart/pull/1509/commits/00459b64c6b1a0ad9098f9c5f378e34d0df2705b .

traversaro:
The macOS failure seem to be related to https://github.com/actions/virtual-environments/issues/1811 .

jslee02:
> The macOS failure seem to be related to [actions/virtual-environments#1811](https://github.com/actions/virtual-environments/issues/1811) .

It seems this build failure is unrelated to this PR. It'd make sense to fix in a separate CR.

jslee02 (approved):

PR #1510, opened 2020-10-13
Fix MSVC linking for assimp and fcl

traversaro:
This PR fixes the linking of tests that use fcl and assimp on MSVC, hopefully fixing the problems reported in https://github.com/ignitionrobotics/ign-physics/issues/87#issuecomment-705155086 .

As the existing code is working well in CI for non-MSVC platforms, and as support for CMake imported targets can vary depending on the platform (for example, `assimp::assimp` is basically broken in Ubuntu 20.04 apt packages: https://bugs.launchpad.net/ubuntu/+source/assimp/+bug/1882427 and https://github.com/robotology/idyntree/issues/693), to minimize the possible regressions I just used the imported targets for fcl and assimp when using MSVC.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -183,7 +183,7 @@ jobs:
       BUILD_TYPE: Release
       RUN_TESTS: OFF
       VCPKG_ROOT: 'C:/dartsim/vcpkg'
-      VCPKG_BUILD_TAG: v0.1.1
+      VCPKG_BUILD_TAG: v0.2.0-70f192e
     steps:
     - uses: actions/checkout@v1
     - name: Install Dependencies
--- a/cmake/Findassimp.cmake
+++ b/cmake/Findassimp.cmake
@@ -26,7 +26,10 @@ find_path(ASSIMP_INCLUDE_DIRS assimp/scene.h
 
 # Libraries
 if(MSVC)
-  set(ASSIMP_LIBRARIES "assimp$<$<CONFIG:Debug>:d>")
+  find_package(assimp QUIET CONFIG)
+  if(TARGET assimp::assimp)
+    set(ASSIMP_LIBRARIES "assimp::assimp")
+  endif()
 else()
   find_library(ASSIMP_LIBRARIES
       NAMES assimp
@@ -34,7 +37,9 @@ else()
 endif()
 
 # Version
-set(ASSIMP_VERSION ${PC_ASSIMP_VERSION})
+if(PC_ASSIMP_VERSION)
+  set(ASSIMP_VERSION ${PC_ASSIMP_VERSION})
+endif()
 
 # Set (NAME)_FOUND if all the variables and the version are satisfied.
 include(FindPackageHandleStandardArgs)
--- a/cmake/Findfcl.cmake
+++ b/cmake/Findfcl.cmake
@@ -28,7 +28,10 @@ find_path(FCL_INCLUDE_DIRS
 
 # Libraries
 if(MSVC)
-  set(FCL_LIBRARIES "fcl$<$<CONFIG:Debug>:d>")
+  find_package(fcl QUIET CONFIG)
+  if(TARGET fcl)
+    set(FCL_LIBRARIES fcl)
+  endif()
 else()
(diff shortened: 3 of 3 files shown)


traversaro:
cc @scpeters

traversaro:
@jslee02 I don't know if you prefer this PR to target `master` or to target `release-6.9`  . I am currently targeting master as the CI on Windows in `master` seems to be under a better shape.

traversaro:
It seems that the vcpkg version on AppVeyor is quite old and does not contain the fcl 0.6 that was updated in March 2020: https://github.com/microsoft/vcpkg/pull/10025 . I can add a fallback to the old logic, but I am afraid it would be broken that as well.

traversaro:
I just tried with the latest vcpkg and the configuration with fcl and assimp enabled works fine. @jslee02 do you think we can tag a new release in https://github.com/dartsim/vcpkg-build to get a new version of the dependencies?

scpeters:
I tested this with `ignition-physics3`, and it fixes most of the linking errors, though it still fails to link to `utf8cpp` (which I believe comes from assimp https://github.com/ignitionrobotics/ign-physics/issues/87#issuecomment-708549196). That may be an assimp problem; I'm not sure

jslee02:
> I am currently targeting master as the CI on Windows in `master` seems to be under a better shape.

`master` is fine

> It seems that the vcpkg version on AppVeyor is quite old and does not contain the fcl 0.6 that was updated in March 2020: [microsoft/vcpkg#10025](https://github.com/microsoft/vcpkg/pull/10025) . I can add a fallback to the old logic, but I am afraid it would be broken that as well.

Please ignore the AppVeyor build for now. It needs to use `dartsim/vcpkg-build` to be consistent with the GitHub Actions.

> @jslee02 do you think we can tag a new release in [dartsim/vcpkg-build](https://github.com/dartsim/vcpkg-build) to get a new version of the dependencies?

A new version is published for `vcpkg 2020.07`, which is the latest version.

Please change the below line to `v0.2.0-2020.07` to use the new version.

https://github.com/dartsim/dart/blob/185fbc2a171e364a00be49c5fe1e5ffb945429c5/.github/workflows/ccpp.yml#L186

Let me know if `vcpkg 2020.07` is still too old.

scpeters:
> I tested this with `ignition-physics3`, and it fixes most of the linking errors, though it still fails to link to `utf8cpp` (which I believe comes from assimp [ignitionrobotics/ign-physics#87 (comment)](https://github.com/ignitionrobotics/ign-physics/issues/87#issuecomment-708549196)). That may be an assimp problem; I'm not sure

this was fixed by https://github.com/microsoft/vcpkg/pull/14061

scpeters (dismissed):
this works for ignition-physics3

traversaro:
> Let me know if vcpkg 2020.07 is still too old.

Unfortunately there were a few bugs in vcpkg ports that were fixed only recently (latest PR: https://github.com/microsoft/vcpkg/pull/14061), so probably generating a new archive against a new commit of vcpkg may be the simpler solution, and then switch to a new release of vcpkg once they do one (unfortunately vcpkg release are not really regular). In https://github.com/robotology/robotology-superbuild-dependencies-vcpkg/pull/32 I am using https://github.com/microsoft/vcpkg/commit/70f192e073dd9e5ed503fc1e86d8a54e2fbaceb4 and it seems to be a good commit without anything strange.

jslee02:
New build for https://github.com/microsoft/vcpkg/commit/70f192e073dd9e5ed503fc1e86d8a54e2fbaceb4 is released. You could use `VCPKG_BUILD_TAG: v0.2.0-70f192e` once https://github.com/dartsim/vcpkg-build/runs/1265369776?check_suite_focus=true is done

traversaro:
> 
> 
> New build for [microsoft/vcpkg@70f192e](https://github.com/microsoft/vcpkg/commit/70f192e073dd9e5ed503fc1e86d8a54e2fbaceb4) is released. You could use `VCPKG_BUILD_TAG: v0.2.0-70f192e` once https://github.com/dartsim/vcpkg-build/runs/1265369776?check_suite_focus=true is done

Great!

jslee02 (approved):

PR #1523, opened 2020-10-25
Fix 'brew update' error

jslee02:


scpeters:
you may also consider using `Homebrew/actions/setup-homebrew`

* https://github.com/osrf/homebrew-simulation/blob/master/.github/workflows/main.yml#L10-L12

I recently noticed a problem with initialization of our macOS GitHub actions in osrf/homebrew-simulation and this action seems to fix it

jslee02:
@scpeters Thanks for the suggestion! It seems the problem I had was related to https://github.com/actions/virtual-environments/issues/1811. I'll try to use `Homebrew/actions/setup-homebrew` if the workaround doesn't work.

Issue #1431, opened 2020-01-19
BodyNode::createShapeNodeWith() access violation on Windows 10


koonyook:
I previously used DART on Ubuntu and it works fine with this C++ piece of code I have. However, I need to bring some of my code to Windows 10. The compilation of DART and my project on VS2019 went well. Unfortunately, a line of code produces an access violation error during runtime.

This is the independent code section that causes the problem
```
dart::dynamics::SkeletonPtr skel = dart::dynamics::Skeleton::create("abc");
dart::dynamics::BodyNode *parent = nullptr;
dart::dynamics::Joint::Properties *joint_properties = MASS::MakeFreeJointProperties("def", Eigen::Isometry3d::Identity(), Eigen::Isometry3d::Identity());
dart::dynamics::FreeJoint::Properties *prop = dynamic_cast<dart::dynamics::FreeJoint::Properties*>(joint_properties);
dart::dynamics::BodyNode *bn = skel->createJointAndBodyNodePair<dart::dynamics::FreeJoint>(
	parent, (*prop), dart::dynamics::BodyNode::AspectProperties(joint_properties->mName)
	).second;
Eigen::Vector3d size;
size << 1, 1, 1;
dart::dynamics::ShapePtr shape = std::shared_ptr<dart::dynamics::BoxShape>(new dart::dynamics::BoxShape(size));;
bn->createShapeNodeWith<dart::dynamics::VisualAspect>(shape);  //access violation here
```
This is the error log.
`Exception thrown at 0x00007FFBA46F7319 (vcruntime140d.dll) in ConsoleApplication1.exe: 0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF.`

By tracing, I've found that the error occurs in FixedFrame.cpp in the createAspect()
```
FixedFrame::FixedFrame(Frame* refFrame,
                       const Eigen::Isometry3d& relativeTransform)
  : Entity(refFrame, false),
    Frame(refFrame)
{
  createAspect<Aspect>(AspectProperties(relativeTransform)); // runtime error at this line
}
```
And the last tracable line is in Aspect.hpp during the dynamic_cast of function setComposite()
```
mComposite = dynamic_cast<CompositeType*>(newComposite);
```
At this line, The FixedFrame object inside the newComposite seems to point to nowhere as the debugger is unable to read the memory inside that object.

As it is related to dynamic_cast, Run-Time Type Information (/GR) has been enabled but the problem still exist. 

### Bug Report
[...]

PR #1540, opened 2021-03-13
Update CMakeLists.txt

erwincoumans:
This fixes a crash on Windows. Confirmed to work in visual studio 2019.
See also https://github.com/dartsim/dart/issues/1431#issuecomment-622967308

[Remove this line and describe this pull request. Link to relevant GitHub issues, if any.]

***

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,7 @@ if(MSVC)
   add_compile_options(/wd4838)
   add_compile_options(/wd4996)
   add_compile_options(/bigobj)
+  add_compile_options(/vd2)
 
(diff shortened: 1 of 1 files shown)


traversaro:
I have some doubts related to the fact that just setting the flag `/vd2` could also change the ABI of classes with virtual bases that are not defined in DART, I tried a more localized approach in https://github.com/dartsim/dart/pull/1541 .

erwincoumans:
good point, best to propagate the setting to users as well.

erwincoumans:
Good point, best to propagate the setting to users as well. Do I need to close this request?

erwincoumans:
Good point, that is even better. I'll close this request.

Issue #1522, opened 2020-10-19
Windows test failures with c++ exception: "Access violation - no RTTI data!"


scpeters:
### Bug Report

- [X] I checked the [documentation](http://dartsim.github.io/) and the [forum](https://dartsim.discourse.group/) but found no answer.
- [X] I checked to make sure that this issue has not already been filed.

#### Environment

* DART version: master
* OS name and version name(or number): Windows 10
* Compiler name and version number: Visual Studio 2019

#### Expected Behavior

All tests that pass on Linux and macOS should pass on windows.

#### Current Behavior

Certain tests fail on Windows with the following console message:

~~~
unknown file: error: C++ exception with description "Access violation - no RTTI data!" thrown in the test body.
~~~

#### Steps to Reproduce
> Please provide detailed steps for reproducing the issue.

1. Re-enable the tests that were disabled in dc4056ab7cf5509da3d0fd627a49ffaa74b7e8a4
2. Run the tests on windows
3. Observe the failures

#### Code to Reproduce

I've enabled some tests in https://github.com/scpeters/dart/commit/685e00c4bd186649fa7623597f6d2b80f65e7c95 (branch [scpeters:windows_ci_debug](https://github.com/dartsim/dart/compare/master...scpeters:windows_ci_debug)) that were previously disabled in #1513 (dc4056ab7cf5509da3d0fd627a49ffaa74b7e8a4), and the failures can be seen in the [GitHub actions CI output](https://github.com/scpeters/dart/runs/1272492268).

jslee02:
Thank you for the report! Yeah, the tests were disabled mostly due to this issue. One example is this line:
https://github.com/dartsim/dart/blob/7476d1663acbfdcf6971f32ac656b5000245d687/unittests/unit/test_Aspect.cpp#L820-L824

It seems the access violation happens when a composite is being dynamic-cased to the concrete type (e.g., `FixedFrame`) in its constructor at this line:
https://github.com/dartsim/dart/blob/7476d1663acbfdcf6971f32ac656b5000245d687/dart/common/detail/Aspect.hpp#L82

I'm not sure if this is a bug of MSVC or the implementation doesn't comply with the standard. Will keep investigating this.

scpeters:
yes, I just collected a backtrace that implicates `createShapeNodeWith` and that `dynamic_cast` in `Aspect.hpp`

jslee02:
[...]

PR #1541, opened 2021-03-13
Fix Aspect/Composite-relate tests on Windows/MSVC by marking every class with virtual base with appropriate pragmas

traversaro:
Fix https://github.com/dartsim/dart/issues/1522 .

This fix was inspired by https://github.com/dartsim/dart/pull/1540, however just enabling the `/vd2` option can change the ABI of classes with virtual base whose implementation has been compiled without the `/vd2` option. For this reason, using instead the [`vtordisp`](https://docs.microsoft.com/en-us/cpp/preprocessor/vtordisp?view=msvc-160) pragmas explicitly for each class of Dart with a virtual base could be a more robust solution. 

For some reason, the fix is not working for the `Clang-Cl`  toolset, at least for the version of Visual Studio used in GitHub Actions. I did not investigated this, but in theory clang-cl has support for the `vtordisp` pragmas (see https://clang.llvm.org/docs/UsersManual.html), but it is possible that this support has some bugs. However, for the time being I just remove the `Clang-Cl`  toolset from the CI jobs, and I instead enabled the unit tests that previously were failing.


traversaro:
Indeed by local tests it seems that `/vd2` option is fixing some of the failing tests on Windows. However, as the dynamic_cast is part of the code in public headers, setting the option just inside dart is not sufficient, but we need to specify that it should be used by downstream projects, so I think we should use the `target_compile_options` instead.

Furthermore, it seems that the ClangCl toolset does not support the `/vd2` option, so for the time being I just removed it from the CI grid.

traversaro:
Furthermore, it seems that `test_VskParser` is still failing on MSVC, probably for some unrelated reason.

traversaro:
Actually, I read a bit about this option in https://docs.microsoft.com/en-us/cpp/build/reference/vd-disable-construction-displacements?view=msvc-160 and it seems a bit scary to just use it everywhere, as it seems it could affect ABI. Probably enabling the setting just on selected class via https://docs.microsoft.com/en-us/cpp/preprocessor/vtordisp?view=msvc-160 is probably more robust.

traversaro:
@jslee02 the history is a bit messy, after review let me know if you prefer me to squash it (I would squash it as a single commit, but let me know if you prefer a different division) or you want to squash&merge. The Travis failure seems to be related to being out of credits, and unrelated from this PR.

dart/common/ClassWithVirtualBase.hpp
```diff
@@ -0,0 +1,47 @@
...
+ *   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF
+ *   USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ *   AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *   LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *   ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *   POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef DART_CLASS_WITH_VIRTUAL_BASE_HPP_
+#define DART_CLASS_WITH_VIRTUAL_BASE_HPP_
```
jslee02, line 34:
```suggestion
#ifndef DART_COMMON_CLASSWITHVIRTUALBASE_HPP_
#define DART_COMMON_CLASSWITHVIRTUALBASE_HPP_
```
to be consistent with other header guards. The convention is `DART_<namespace>_<file_name_without_spaces>_HPP_`.

dart/common/ClassWithVirtualBase.hpp
```diff
@@ -0,0 +1,47 @@
...
+// See https://github.com/dartsim/dart/issues/1522
+#if defined(_MSC_VER)
+#  define DART_DECLARE_CLASS_WITH_VIRTUAL_BASE_BEGIN __pragma(vtordisp(push, 2))
+#  define DART_DECLARE_CLASS_WITH_VIRTUAL_BASE_END __pragma(vtordisp(pop))
+#else
+#  define DART_DECLARE_CLASS_WITH_VIRTUAL_BASE_BEGIN
+#  define DART_DECLARE_CLASS_WITH_VIRTUAL_BASE_END
+#endif
+
+#endif // DART_CLASS_WITH_VIRTUAL_BASE_HPP_
```
jslee02, line 47:
```suggestion
#endif // DART_CLASS_WITH_VIRTUAL_BASE_HPP_
```

unittests/unit/CMakeLists.txt
```diff
@@ -42,24 +38,22 @@ if(TARGET dart-utils)
...
+  dart_add_test("unit" test_SdfParser)
+  target_link_libraries(test_SdfParser dart-utils)
+  
+  dart_add_test("unit" test_SkelParser)
+  target_link_libraries(test_SkelParser dart-utils)
+
+  if(NOT MSVC) 
     dart_add_test("unit" test_VskParser)
     target_link_libraries(test_VskParser dart-utils)
   endif()
```
jslee02, line 59: Don't need to be fixed in this PR, but did you get any hint why `test_VskParser` doesn't build or pass with MSVC?

jslee02 (changes requested):
This is awesome! Don't worry about the commit history. Let me do `Sqaush&merge`.

Thanks for the fix @traversaro and @erwincoumans!

unittests/unit/CMakeLists.txt
traversaro: I briefly saw a `0xC0000135` that typically means dll not found, so I guess it could be something related to dll of the dependencies not being in the path or not being copied in the directory with the test executables. In any case, I tried to reduce the modification as possible to have the smallest possible PR to ensure that it could be merged without side-effects.

jslee02: Thanks, let me take a look!

PR #1513, opened 2020-10-16
Build and run tests on Windows CI

jslee02:

PR #1544, opened 2021-03-23
FindODE: try to check first for ODE cmake config

traversaro:
As of ODE 0.16, ODE install its own CMake config files, that are able to correctly support advanced functionalities such as various build configurations. This PR modifies the `FindODE` CMake module to first try to find the ODE CMake config files, and only fallback to try to use pkg-config if they are not found.

### DIFF
--- a/cmake/FindODE.cmake
+++ b/cmake/FindODE.cmake
@@ -17,6 +17,18 @@
 # and the following targets:
 #   ODE::ODE
 
+include(FindPackageHandleStandardArgs)
+
+# First try to find ODE via the official cmake config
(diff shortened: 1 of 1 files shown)


traversaro:
The `CONFIG_MODE`  parameter of `find_package_handle_standard_args` used in this PR is available since CMake 3.5 (minimum required version), see https://cmake.org/cmake/help/v3.5/module/FindPackageHandleStandardArgs.html.

traversaro:
By bad, there is already a similar fix in https://github.com/dartsim/dart/pull/1513, I am experiencing failures for another reason.

PR #1612, opened 2021-10-08
[dynamics]: deep copy shapes

costashatz:
This PR adds the possibility to deep copy shapes. In [robot_dart](https://github.com/resibots/robot_dart), we clone very frequently robots and we would like to be able to make deep copies of the robot. At the current state, DART does not copy shapes. This is generally OK unless we would like to alter the shapes later (e.g. create a damage or change the ColorMode of a MeshShape). So, this PR solves the above issue by making it able to deep copy shapes. I hope that I did not forget any shape. The only shape that I do not deep copy is the `SoftMeshShape` (for now I just return a `nullptr` in the new `copy` function; you might want to handle this differently [e.g. output a warning or exception]). Here there is a tight relationship between the `SoftBodyNode` it is attached to, so I do not touch this. By the way, when cloning a skeleton, is this type of meshes re-created to point to the correct `BodyNode`?

Let me know if this is OK for you or you would like to handle it differently. If this is OK, we can do unit tests.

***

### DIFF
--- a/dart/dynamics/ArrowShape.cpp
+++ b/dart/dynamics/ArrowShape.cpp
@@ -55,6 +55,9 @@ ArrowShape::Properties::Properties(
 }
 
 //==============================================================================
+ArrowShape::ArrowShape() : MeshShape(Eigen::Vector3d::Ones(), nullptr) {}
+
+//==============================================================================
 ArrowShape::ArrowShape(
     const Eigen::Vector3d& _tail,
     const Eigen::Vector3d& _head,
@@ -254,9 +257,9 @@ void ArrowShape::configureArrow(
 }
 
 //==============================================================================
-ShapePtr ArrowShape::copy() const
+ShapePtr ArrowShape::clone() const
 {
-  aiScene* new_scene = copyMesh();
+  aiScene* new_scene = cloneMesh();
   auto new_shape = std::make_shared<ArrowShape>();
 
   new_shape->mTail = mTail;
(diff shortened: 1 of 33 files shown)


jslee02:
> I did all the rest (apart from unit tests: I'll do them later), but I did not understand what you mean here. Can you be more elaborate? I did the log, but not sure what you mean by making the constructor of `SoftMeshShape` private (in terms of the expected behavior).

It's an optional change, but what I expected was something like:
```
class SoftMeshShape {
private:
  friend class SoftBodyNode;
  SoftMeshShape();
};
```
This way `SoftMeshShape` can be created only by `SoftBodyNode`. I'm happy to make this change myself though.

jslee02:
> I am aware of this concept, but I am not sure how this change is related to this PR though.

Please don't worry about the change for this PR. I can make this change in a separate PR.

> By the way, should we include this behavior into the `Skeleton::cloneSkeleton()` function?

Yeah, I was thinking about making a PR for this change. World would also need similar options. By the way, let's save those changes for the future PRs to keep this PR a good size to review.

jslee02:
@costashatz For clarification, is it correct that you'd add the unit tests to this PR?

costashatz:
> @costashatz For clarification, is it correct that you'd add the unit tests to this PR?

That is indeed my intention. Although I am not sure if I am going to find time for this any time soon. Is it OK if we merge this and create an issue about the tests?

jslee02:
Sounds good to me

PR #1629, opened 2021-12-18
Treat warnings as errors on release mode

jslee02:
- Treat warnings as errors on release mode by default
- Remove `DART_TREAT_WARNINGS_AS_ERRORS` option
- Fix compile warnings with GCC 11.2

***

- [ ] Format new code files using `clang-format`


traversaro:
Just FYI, probably the "Treat warnings as errors on release mode" will need to be patched out in the version used in package managers of DART, as it is quite common that an older version of DART needs to be compiled with newer compilers, and this creates new warnings, that if use of `-Werror` is forced result in errors.

jslee02:
@traversaro This makes sense to me. The motivation of this change was to prevent the case that the contributor and the CI don't fix the warnings by forgetting turnning on `DART_TREAT_WARNINGS_AS_ERRORS` option at least on release mode build. I think this option would be still useful to give options to the pacakge managers. Let me revive the option and make sure the CI to turn on the options instead. Thanks for the feedback!

PR #1630, opened 2021-12-18
Update changelog and contribution for #1624

jslee02:
@azeey


azeey (approved):
Thanks! Sorry, I should have added that myself.

PR #1664, opened 2022-05-25
create convex mesh when possible

MFridlyand:
Create bullet convex mesh collision shape when input mesh is convex. Convex mesh colliders work much faster, also Bullet do not support concave mesh vs concave mesh collisions


jslee02 (approved):
Thank you for your contribution! The code could be a bit more generalized for the scalar type because the Bullet library could be built for either float or double. Let me address that in a separate PR. Thanks again!

Issue #1673, opened 2022-07-29
Fails to build with Clang 14


J-Meyers:
### Bug Report
- Clang 14 adds the compiler option `-Wbitwise-instead-of-logical` and includes it in `-Wall`
- The project is flagged on lines 309 and 368 of `dart/utils/mjcf/MjcfParser.cpp` which use bitwise ands on booleans

#### Environment
* DART version: main, 6.12.1
* OS name and version name(or number): Ubuntu 20.04
* Compiler name and version number: Clang 14.0.0

#### Expected Behavior
No compilation failure, compiles without modification

#### Current Behavior
```
/home/user/ros_catkin_ws/src/dart/dart/utils/mjcf/MjcfParser.cpp:309:13: error: use of bitwise '&' with boolean operands [-Werror,-Wbitwise-instead-of-logical]
          = mjcfJoint0.isLimited() & mjcfJoint1.isLimited();
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                   &&
/home/user/ros_catkin_ws/src/dart/dart/utils/mjcf/MjcfParser.cpp:309:13: note: cast one or both operands to int to silence this warning
/home/user/ros_catkin_ws/src/dart/dart/utils/mjcf/MjcfParser.cpp:368:13: error: use of bitwise '&' with boolean operands [-Werror,-Wbitwise-instead-of-logical]
          = mjcfJoint0.isLimited() & mjcfJoint1.isLimited();
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                   &&
/home/user/ros_catkin_ws/src/dart/dart/utils/mjcf/MjcfParser.cpp:368:13: note: cast one or both operands to int to silence this warning
2 errors generated.
```

PR #1674, opened 2022-07-29
Silence Clang-14 bitwise instead of logical

J-Meyers:
- Addresses Issue #1673
- Switches to logical and rather than bitwise
***

- [x] Format new code files using ClangFormat by running `make format`

### DIFF
--- a/dart/utils/mjcf/MjcfParser.cpp
+++ b/dart/utils/mjcf/MjcfParser.cpp
@@ -365,7 +365,7 @@ createJointAndBodyNodePairForMultipleJoints(
 
       // Limits
       props.mIsPositionLimitEnforced
-          = static_cast<int>(mjcfJoint0.isLimited()) & mjcfJoint1.isLimited();
+          = mjcfJoint0.isLimited() && mjcfJoint1.isLimited();
       props.mPositionLowerLimits[0] = mjcfJoint0.getRange()[0];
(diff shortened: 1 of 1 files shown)


dart/utils/mjcf/MjcfParser.cpp
```diff
@@ -306,7 +306,7 @@ createJointAndBodyNodePairForMultipleJoints(
 
       // Limits
       props.mIsPositionLimitEnforced
-          = mjcfJoint0.isLimited() & mjcfJoint1.isLimited();
+          = static_cast<int>(mjcfJoint0.isLimited()) & mjcfJoint1.isLimited();
```
jslee02, line 309:
```suggestion
          = mjcfJoint0.isLimited() && mjcfJoint1.isLimited();
```

dart/utils/mjcf/MjcfParser.cpp
```diff
@@ -365,7 +365,7 @@ createJointAndBodyNodePairForMultipleJoints(
...
```
jslee02, line 368:
```suggestion
          = mjcfJoint0.isLimited() && mjcfJoint1.isLimited();
```

jslee02 (changes requested):
Thank you for finding out this issue with Clang 14!

I believe you followed the suggestion from the warning but it seems the correct fix in this case would be using the logical operators (i.e., `&&`)

jslee02 (approved):
LGTM!

PR #1685, opened 2022-11-27
Codespace

olzhas:
Adding appropriate configuration files to enable github codespace
***


olzhas:
I am not sure how opengl render might work in this setting.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,82 @@
+FROM mcr.microsoft.com/devcontainers/cpp:0-ubuntu-20.04
+
+# [Optional] Uncomment this section to install additional vcpkg ports.
+# RUN su vscode -c "${VCPKG_ROOT}/vcpkg install <your-port-name-here>"
+
```
jslee02, line 5: Hm, why do we need vcpkg? I thought it's for Windows

```suggestion
```

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,82 @@
+FROM mcr.microsoft.com/devcontainers/cpp:0-ubuntu-20.04
```
jslee02, line 1: Why not use `ubuntu:20.04`?

jslee02 (commented):
Thank you for doing this!

.devcontainer/Dockerfile
jslee02: Actually, we have a [`Dockerfile`](https://github.com/dartsim/dart-ci-docker-containers/blob/master/v6.13/Dockerfile.ubuntu.focal) that is used for CI and prebuilt image (`dartsim/dart-dev:ubuntu-focal-v6.13`). Maybe we could use it instead without installing any additional packages below? The image is hosted in https://hub.docker.com/repository/docker/dartsim/dart-dev

.devcontainer/Dockerfile
```diff
@@ -1,82 +1,74 @@
-FROM mcr.microsoft.com/devcontainers/cpp:0-ubuntu-20.04
+FROM dartsim/dart-dev:ubuntu-focal-v6.13
 
-# [Optional] Uncomment this section to install additional vcpkg ports.
-# RUN su vscode -c "${VCPKG_ROOT}/vcpkg install <your-port-name-here>"
+# RUN apt-get update && \
```
jslee02, line 3: It seems we can remove all the lines below. Other than that, looks good to me!

olzhas:
@jslee02 what do you think about reusing https://github.com/dartsim/dart/blob/main/Dockerfile ?

jslee02:
> @jslee02 what do you think about reusing https://github.com/dartsim/dart/blob/main/Dockerfile ? I discovered that it is already in the repo.

That image is supposed to be an image that already have DART installed in it (so the users don't care about building and installing it but just work in the container). It actually could use `dart-dev` as base image like the Dockerfile in this PR.

I'm merging this PR. Thanks!

PR #1752, opened 2023-05-13
Allow specifying mimic joint properties per DoF (#1684)

### LINKED ISSUE #1684: Feature request : Supply axis index when specifying mimic joint.
**Is your feature request related to a problem? Please describe.**
When specifying mimic joint (https://dartsim.github.io/dart/main/d6/d5b/classdart_1_1dynamics_1_1Joint.html#ae65b6209299935e602b675f84091c499)

```
 void dart::dynamics::Joint::setMimicJoint( 
        const Joint* _mimicJoint, double _mimicMultiplier, double _mimicOffset)
```

Only the joint pointer, multiplier, and offset are supplied to the method. The section here : (https://github.com/dartsim/dart/blob/189e24d5624f3007fe13d2c27bbdb9a83ed135ec/dart/constraint/MimicMotorConstraint.cpp#L67) dictates that the degrees of freedom for both axes should be the same.

This does not allow mimicking individual axes of a multi-axis joint, such as enabling a revolute joint to mimic one axis of an universal joint, or have an universal joint mimic only one axis of another universal joint. In such cases, it should be permitted to specify the axis index to be mimicked for each joint, and expand the implementation of the mimic joint to accommodate this situation. 

**Describe the solution you'd like**
For instance, the new method could look like : 
```
 void dart::dynamics::Joint::setMimicJoint( const Joint* _mimicJoint, std::size_t _index  , double _mimicMultiplier, double _mimicOffset)
```

**Additional context**
Originally came up here : https://github.com/gazebosim/gz-physics/pull/431

cc @scpeters

jslee02:
This PR supports specifying mimic joint properties per DOF, as requested in issue #1684.

Previously, DART allowed specifying mimic joints with the constraint that the dependent and reference joint must have the same DOF, and the whole DOF would be mimicked rather than individual DOFs.

This PR removes these constraints, allowing a mimic joint to mimic individual DOFs with different reference joints, DOF indices, offsets, and multipliers. The following functions have been added:

```
  /// Set the mimic joint with a single reference joint and the same multiplier
  /// and offset for all dependent joint's DOFs.
  void setMimicJoint(
      const Joint* referenceJoint,
      double mimicMultiplier = 1.0,
      double mimicOffset = 0.0);

  /// Sets the mimic joint properties for a specific DOF of the dependent joint.
  void setMimicJointDof(std::size_t index, const MimicDofProperties& mimicProp);

  /// Sets the mimic joint properties for all DOFs of the dependent joint using
  /// a vector of MimicDofProperties.
  void setMimicJointDofs(const std::vector<MimicDofProperties>& mimicProps);

  /// Sets the mimic joint properties for specific DOFs of the dependent joint
  /// using a map of DOF index and MimicDofProperties.
  void setMimicJointDofs(
      const std::map<std::size_t, MimicDofProperties>& mimicPropMap);
```

where

```
/// Properties of a mimicked Degree of Freedom (DOF) in a joint.
struct MimicDofProperties
{
  /// Pointer to the reference joint used for mimicking behavior.
  const Joint* referenceJoint;

  /// Index of the reference DOF in the reference joint.
  std::size_t referenceDofIndex = 0;

  /// Multiplier applied to the reference joint's DOF value.
  double multiplier = 1.0;

  /// Offset added to the reference joint's DOF value.
  double offset = 0.0;
};
```

**Note** that this change breaks API and ABI (by removing some public members) even though it's targeting 6.14. Please let us know if this is problematic for Gazebo to adopt this change.

Resolves #1684

CC: @adityapande-1995

***


jslee02:
The member variables of MimicDofProperties will be renamed to follow the convention for DART 6.x to:

```
struct MimicDofProperties
{
  const Joint* mReferenceJoint;
  std::size_t mReferenceDofIndex = 0;
  double mMultiplier = 1.0;
  double mOffset = 0.0;
};
```

jslee02:
I'm going to proceed with merging this for the time being. If you have any concerns or feedback regarding this change, please let me know!

scpeters:
Sorry I didn't have a chance to review this yet. I'll try testing it and let you know how it goes

scpeters:
> Sorry I didn't have a chance to review this yet. I'll try testing it and let you know how it goes

I started writing a test with a multi-axis joint for which only one of its axes would use the mimic actuator type, but I noticed that `Joint::setActuatorType` does not appear to distinguish between the dofs of a joint, so I think there may still be something missing in the API.

I noticed it when I got to the following line in the current test:

* https://github.com/dartsim/dart/blob/release-6.14/unittests/integration/test_Joints.cpp#L1098

jslee02:
Are you suggesting that there could be a scenario where a joint contains both mimic degrees of freedom (DOFs) and non-mimic DOFs? I did not consider this situation in this PR. Addressing such a case might require additional refactoring, enabling us to set the actuator type for individual DOFs rather than the entire joint.

scpeters:
> Are you suggesting that there could be a scenario where a joint contains both mimic degrees of freedom (DOFs) and non-mimic DOFs?

Yes, this would provide maximum modeling flexibility, and while I could write an integration test for this functionality, I don't have a real-world use case in mind that motivates this feature.

> I did not consider this situation in this PR. Addressing such a case might require additional refactoring, enabling us to set the actuator type for individual DOFs rather than the entire joint.

Yes I think it would require more refactoring

PR #1754, opened 2023-06-07
Added python binding for WholeBodyIK

markuswnuk91:
Hi, I noted an error when trying to use get the InverseKinematics of a skeletion from python. 
The dart.dynamics.skel.getIK() method resulted in an error: could not be converted into a python type. 
So I added a binding for the WholeBodyIK class. This seems to fix it.
Unfortunately, I do not have the time to bind the whole class atm, so I only bound the solveAnApply method, which allows most basic usage.

Let me know if this needs some futher work before merging.

Cheers

### DIFF
--- a/python/dartpy/dynamics/Skeleton.cpp
+++ b/python/dartpy/dynamics/Skeleton.cpp
@@ -573,6 +573,13 @@ void Skeleton(py::module& m)
             return self->getIK();
           })
       .def(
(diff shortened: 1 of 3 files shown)


jslee02 (approved):
Great start on the Python bindings for WholeBodyIK! We can expand these with additional class features as needed.

PR #1779, opened 2024-01-04
Fix build with urdfdom 4.0.0

scpeters:
There is a parsePose method in urdfdom 4.0.0 with the same signature as the function defined locally in urdf_world_parser, which causes build errors due to an ambiguous call. This adds `urdf_parsing::` to the reference to `parsePose` to eliminate the ambiguity. Needed for https://github.com/Homebrew/homebrew-core/pull/158963.

I was confused about why this error occurs since I don't see `using namespace urdf;` anywhere, but I believe it's due to [argument-dependent lookup](https://en.cppreference.com/w/cpp/language/adl) (thanks @azeey for the pointer).

- [X] Format new code files using ClangFormat by running `make format`

### DIFF
--- a/dart/utils/urdf/urdf_world_parser.cpp
+++ b/dart/utils/urdf/urdf_world_parser.cpp
@@ -216,7 +216,7 @@ std::shared_ptr<World> parseWorldURDF(
           auto* origin = entity_xml->FirstChildElement("origin");
           if (origin)
           {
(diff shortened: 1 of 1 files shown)


jslee02:
Thank you for the fix!

scpeters:
thanks!

PR #1796, opened 2024-03-24
Replace NDEBUG with custom flags to avoid ABI issues

### LINKED ISSUE #1795: ABI of the library depends on the value of NDEBUG?
It seems that currently the ABI of the library depends on the value of `NDEBUG` used when using the headers of the library, see:
* https://github.com/dartsim/dart/blob/2437749590b778074434d3a0f6208bf366ac12af/dart/constraint/PgsBoxedLcpSolver.hpp#L83
* https://github.com/dartsim/dart/blob/2437749590b778074434d3a0f6208bf366ac12af/dart/constraint/PGSLCPSolver.hpp#L61
* https://github.com/dartsim/dart/blob/2437749590b778074434d3a0f6208bf366ac12af/dart/common/MemoryManager.hpp#L154

However, this value could be different in the downstream project using dart w.r.t. to the value used when compiling dart, for example see https://github.com/conda-forge/gz-physics-feedstock/pull/26#issuecomment-2016526660 . 

Probably it could make sense to save in https://github.com/dartsim/dart/blob/2437749590b778074434d3a0f6208bf366ac12af/dart/config.hpp.in#L65 the `NDEBUG` value used when compiling dart (or something similar) and use that in the `.hpp` ?

- [x] I checked the [documentation](http://dartsim.github.io/) and the [forum](https://dartsim.discourse.group/) but found no answer.

* DART version: 6.13.2
* OS name and version name(or number): Not relevant
* Compiler name and version number: Not relevant

#### Current Behavior

> What is the current behavior?

See description.

#### Steps to Reproduce
Not relevant?

Not relevant?

jslee02:
- Replaced NDEBUG with custom flags to avoid ABI issues for https://github.com/dartsim/dart/issues/1795.
- Added DART_ prefix to avoid potential name conflict.
- Consolidated custom flags for release mode into DART_BUILD_MODE_RELEASE to simplify, based on usage.
- Included `config.hpp` whenever needed
- Fixed typos

Resolves #1795

***


traversaro:
Thanks! I just noticed something, that anyhow I am not sure about. If I recall correctly, Debian packaging set the CMAKE_BUILD_TYPE to None (i.e. empty string) and then it sets the release compilation flags itself, for keeping consistency across packages. Probably it could make sense to ensure that the release bits get compiled even if CMAKE_BUILD_TYPE is empty? fyi @j-rivero @azeey that I guess are interested in this.

Another use case that I am currently not interested in (as I also use Ninja in Windows) but that it could be affected (and I did not realized this while doing my original suggestion of using the config.hpp, sorry!) are Cmake multiple-config generators, such as Visual Studio, Xcode where the CMAKE_BUILD_TYPE is not used to specify the build type, as the build type is not fixed at configuration time. I guess it is ok not to support them, but probably it is better to print a clear error in CMake.

jslee02:
Thanks for pointing these things out!

I wonder if setting it to RELEASE by default sounds like a practical fix for the Debian scenario. It would ensure we don't miss out on crucial optimization flags for release builds, and I'm leaning toward trying this approach.

Detecting these at configuration time is tricky since they're designed to let you choose the build type later. One idea could be to check the generator used during the CMake configuration phase. If we can identify multi-config generators, we might either set a default build type or warn that this setup needs a specific build type to be chosen at build time.

PR #1805, opened 2024-05-13
WIP: Addition of Cone collision

bperseghetti:
WIP: Add cone collision

### DIFF
--- a/dart/collision/dart/DARTCollide.cpp
+++ b/dart/collision/dart/DARTCollide.cpp
@@ -35,6 +35,7 @@
 #include "dart/collision/CollisionObject.hpp"
 #include "dart/dynamics/BodyNode.hpp"
 #include "dart/dynamics/BoxShape.hpp"
+#include "dart/dynamics/ConeShape.hpp"
 #include "dart/dynamics/CylinderShape.hpp"
 #include "dart/dynamics/EllipsoidShape.hpp"
 #include "dart/dynamics/SphereShape.hpp"
--- a/dart/collision/ode/OdeCollisionObject.cpp
+++ b/dart/collision/ode/OdeCollisionObject.cpp
@@ -35,6 +35,7 @@
 #include "dart/collision/ode/OdeTypes.hpp"
 #include "dart/collision/ode/detail/OdeBox.hpp"
 #include "dart/collision/ode/detail/OdeCapsule.hpp"
+#include "dart/collision/ode/detail/OdeCone.hpp"
 #include "dart/collision/ode/detail/OdeCylinder.hpp"
 #include "dart/collision/ode/detail/OdeHeightmap.hpp"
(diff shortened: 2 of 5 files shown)


jslee02:
Thank you for the initial work. If you're still interested in completing this feature, let me know. Otherwise, I'm happy to take over the rest of the work.

bperseghetti:
> Thank you for the initial work. If you're still interested in completing this feature, let me know. Otherwise, I'm happy to take over the rest of the work.

Hi, thanks! I sadly do not have any time to dedicate to doing this. If you could take it over that would be greatly appreciated!

jslee02:
Actually, I quickly checked your PR and noticed that ODE doesn't currently support cones:

https://bitbucket.org/odedevs/ode/src/b4253237006f7f6b5184f80ed973fb1c0bfc6039/include/ode/collision.h#lines-877:899

In this case, we may need to fall back to other collision engines (such as fcl or bullet) for cone support, or use a mesh for the cone shape (either created by the user or by adding a feature to create a mesh from a cone shape, which is not currently implemented in DART).

I will close this PR as there is no quick resolution using ODE at this time. Please feel free to reopen or make a new PR if you have any further ideas on how to move forward. Thank you for your effort though.

bperseghetti:
@jslee02 any thoughts on: https://bitbucket.org/odedevs/ode/src/master/contrib/TerrainAndCone/dCone.cpp

Maybe we can finally get that out of it's since 2006 purgatory and into the portion that is actually built?

jslee02:
Is the cone shape present but not officially supported? In any case, it would be beneficial to incorporate ODE cone support into DART when it becomes available.

bperseghetti:
> Is the cone shape present but not officially supported? In any case, it would be beneficial to incorporate ODE cone support into DART when it becomes available.

Well, it looks like it's been stuck in the "contrib" folder for the last 18 years. But seems on a basic review to include what we want and need for cone collisions. Is this something you (@jslee02) want to open up as a PR for ODE to get it into the "mainline build" and out of contrib? As mentioning @azeey here for tracking this since we have seen significant performance issues with trying to do this with a trimesh method.

jslee02:
I appreciate the opportunity to contribute. However, I have some reservations about the effort required to update and integrate an 18-year-old contribution into the mainline build, as it may be significantly outdated and could present unforeseen issues. I will attempt to allocate some time to assess this further, but I must be candid in stating that it will be a lower priority for me now. Therefore, I would be open to others leading this project.

bperseghetti:
@jslee02 also mentioned it here in the ODE bitbucket on previous issue: https://bitbucket.org/odedevs/ode/issues/19/dcone

bperseghetti:
What's the likelihood that we could convince ODE group to use github or gitlab instead of bitbucket? I forgot how painful bitbucket is to deal with.

Issue #1683, opened 2022-11-09
Joint can't recover from position limits in servo mode


livanov93:
### Bug Report
- [x] I checked the [documentation](http://dartsim.github.io/) and the [forum](https://dartsim.discourse.group/) but found no answer.
- [x] I found similar issue https://github.com/dartsim/dart/issues/1583.

### Environment
DART version: 6.13 - [main branch](https://github.com/dartsim/dart) built from source on commit dartsim/dart@189e24d5624f3007fe13d2c27bbdb9a83ed135ec 
OS name and version name(or number): Ubuntu 22.04
Compiler name and version number: GCC 11.3.0

### Description
I still experience similar problem to https://github.com/dartsim/dart/issues/1583 when the axis of rotation is z (e.g. `<xyz>0 0 1</xyz>`). 
The problem is that the joint, once hits the limit and then can't recover from it.

### Related issues and PRs:
- https://github.com/gazebosim/gz-sim/issues/1684
- https://github.com/ros-controls/gz_ros2_control/pull/86 (workaround)
- https://github.com/ros-controls/gz_ros2_control/pull/67#issuecomment-1268865166 (another perspective)
- https://github.com/dartsim/dart/issues/1583

I changed the sdf a bit (`/home/lovro/tmp/dart/data/sdf/test/test.sdf`):
```
 <sdf version="1.5">
   <model name="simple_joint_test">
        <pose>10 10 2 0  0 0</pose>
        <link name="base">
            <inertial>
                <mass>100</mass>
            </inertial>
            <visual name="visual">
                <pose>0 0 0 -1.57 0 0</pose>
                <geometry>
                    <cylinder>
                        <radius>0.1</radius>
                        <length>0.2</length>
                    </cylinder>
                </geometry>
            </visual>
        </link>
        <link name="bar">
            <pose>0 0 -1 0 0 0</pose>
            <inertial>
                <mass>1</mass>
            </inertial>
            <visual name="visual">
                <geometry>
                    <box>
                        <size>0.1 0.1 0.1</size>
                    </box>
                </geometry>
            </visual>
        </link>
        <joint name="j0" type="fixed">
            <parent>world</parent>
            <child>base</child>
        </joint>
        <joint name="j1" type="revolute">
            <pose>0 0 1 0 0 0</pose>
            <parent>base</parent>
            <child>bar</child>
            <axis>
                <xyz>0 0 1</xyz>
                <limit>
                    <upper>0.85</upper>
[...]

PR #1818, opened 2024-06-26
fix joint stuck at limit bug

ufrhaidar:
Fix https://github.com/dartsim/dart/issues/1683 and https://github.com/gazebosim/gz-sim/issues/1684

result:

https://github.com/dartsim/dart/assets/110574820/2f5d5444-a5ea-47f3-84a2-f94ab83ca3f8

***

### DIFF
--- a/dart/constraint/JointConstraint.cpp
+++ b/dart/constraint/JointConstraint.cpp
@@ -213,7 +213,7 @@ void JointConstraint::update()
       const double B1 = A1 + mErrorAllowance;
       const double A2 = positions[i] - positionUpperLimits[i];
       const double B2 = A2 - mErrorAllowance;
-      if (B1 < 0) {
+      if (B1 <= 0) {
         // The current position is lower than the lower bound.
         //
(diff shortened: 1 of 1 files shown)


jslee02:
Could you please take a look at the test failures?

ufrhaidar:
I don't believe that the changes I've done are the correct thing to do. It just was what got the system out of the velocity limit loop (when position is equal to limit).

Deleting the velocity limit section also fixed the issue, but I don't know what effects it will have on other joint types.

Would this change be acceptable?

ufrhaidar:
@jslee02 what needs to be done now?

jslee02:
Apologies for the delayed response. I am currently on vacation and will be sporadic in my availability to provide a thorough analysis of this issue and the patch for this PR.

Regarding this PR, as you mentioned, I am also uncertain about the potential side effects of disabling the velocity limit correction for servo motor types. I wonder if there is a better way to address the issue without completely disabling it.

PR #1819, opened 2024-06-26
Include CTest for BUILD_TESTING option

scpeters:
The recent release of dartsim 6.14.0 is failing to build in homebrew-core due to an issue with fetching GoogleTest (see https://github.com/Homebrew/homebrew-core/pull/175563#issuecomment-2189834711). My first thought to work around this was to build the homebrew formula with `-DBUILD_TESTING=OFF`, but that option is neither defined nor used to disable test compilation. This includes `CTest` to define the option and adds logic to skip building tests when `BUILD_TESTING` is false.

We could separately try to find the `googletest` homebrew formula instead of using `FetchContent` if desired, but I think `BUILD_TESTING` is a useful option anyway, so I would start with this.

- [X] Format new code files using ClangFormat by running `make format`


CMakeLists.txt
```diff
@@ -431,7 +433,10 @@ endif()
 add_subdirectory(python)
 
 # Add 'ALL' target that builds everything
-set(all_target_candidates tests_and_run dartpy pytest)
+set(all_target_candidates dartpy)
+if (BUILD_TESTING)
+  set(all_target_candidates tests_and_run pytest)
```
jslee02, line 438: Maybe add instead of replace?

```suggestion
  list(APPEND all_target_candidates tests_and_run pytest)
```

jslee02:
Good to know that Homebrew doesn't allow using `FetchContent`. Thanks!

CMakeLists.txt
scpeters: oops, thanks! I applied the suggestion

jslee02 (approved):

PR #1820, opened 2024-06-28
config.hpp.in: remove whitespace from version macros

scpeters:
When attempting to build gz-physics7 with dartsim 6.14.1, the build fails:

* [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/label=osx_monterey/1503/) https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/1503/label=osx_monterey/

~~~
/tmp/gz-physics7-20240628-32522-zc6n0e/gz-physics-7.3.0/dartsim/src/Base.hh:44:5: error: invalid token at start of a preprocessor expression
#if DART_VERSION_AT_LEAST(6, 13, 0)
    ^
/usr/local/include/dart/config.hpp:38:3: note: expanded from macro 'DART_VERSION_AT_LEAST'
  DART_VERSION_GE(x, y, z)
  ^
/usr/local/include/dart/config.hpp:19:4: note: expanded from macro 'DART_VERSION_GE'
  (DART_MAJOR_VERSION > x                                                      \
   ^
/usr/local/include/dart/config.hpp:5:28: note: expanded from macro 'DART_MAJOR_VERSION'
#define DART_MAJOR_VERSION @DART_MAJOR_VERSION @
                           ^
~~~

It appears that the version number macros at the start of config.hpp.in have a space in between the `@` characters, which seems to prevent cmake from matching the variable names when configuring the version number macros. I confirmed by looking at the first few lines of the `config.hpp` file installed in the dartsim 6.14.1 homebrew bottle:

~~~
/* config.hpp. Generated by CMake for dart. */
#pragma once
/* Version number */
#define DART_MAJOR_VERSION @DART_MAJOR_VERSION @
#define DART_MINOR_VERSION @DART_MINOR_VERSION @
#define DART_PATCH_VERSION @DART_PATCH_VERSION @
#define DART_VERSION "6.14.1"
#define DART_DESCRIPTION "Dynamic Animation and Robotics Toolkit."
~~~

I believe removing the whitespace should fix it, though it would be better to add a test of the `DART_VERSION_AT_LEAST` macro, which should catch regressions like this in the future.

***


scpeters:
> I believe removing the whitespace should fix it, though it would be better to add a test of the DART_VERSION_AT_LEAST macro, which should catch regressions like this in the future.

where would you recommend putting a test for `DART_VERSION_AT_LEAST`

jslee02 (approved):

jslee02:
Ah, good catch. Thank you for the fix!

> > I believe removing the whitespace should fix it, though it would be better to add a test of the DART_VERSION_AT_LEAST macro, which should catch regressions like this in the future.
> 
> where would you recommend putting a test for `DART_VERSION_AT_LEAST`

Anywhere in common would make sense to me.

scpeters:
> Ah, good catch. Thank you for the fix!
> 
> > > I believe removing the whitespace should fix it, though it would be better to add a test of the DART_VERSION_AT_LEAST macro, which should catch regressions like this in the future.
> > 
> > 
> > where would you recommend putting a test for `DART_VERSION_AT_LEAST`
> 
> Anywhere in common would make sense to me.

test added in https://github.com/dartsim/dart/pull/1820/commits/b03d15ab91cf6059b97e989b9ef891d10d98af4d to the Logging test since it already had a transitive include of `config.hpp` via `TestHelpers.hpp`

PR #1904, opened 2025-01-20
CMake: add DART_USE_SYSTEM_GOOGLEBENCHMARK

nim65s:
Hi,

This PR allow use of  installed google benchmark instead of downloading it at configure time.

***


jslee02:
I confirmed that this works with the option enabled and disabled. Thanks for implementing it!

PR #1953, opened 2025-05-06
[WIP] Add Kinematic Tag

jcmayoral:
I am using dart together with the Gazebo Simulator. And there is a kinematic tag that allows certain links to ignore external forces, like inertia and acceleration.

***

### DIFF
--- a/dart/dynamics/KinematicJoint.cpp
+++ b/dart/dynamics/KinematicJoint.cpp
@@ -1,0 +1,677 @@
+/*
+ * Copyright (c) 2011-2024, The DART development contributors
+ * All rights reserved.
+ *
+ * The list of contributors can be found at:
+ *   https://github.com/dartsim/dart/blob/main/LICENSE
+ *
+ * This file is provided under the following "BSD-style" License:
+ *   Redistribution and use in source and binary forms, with or
+ *   without modification, are permitted provided that the following
+ *   conditions are met:
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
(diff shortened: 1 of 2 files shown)


jslee02:
You may not need to add a new joint type in that case. Consider specifying the actuator type as either ACCELERATION, VELOCITY, or LOCKED and set the kinematic values (position, velocity, acceleration) as needed:

https://github.com/dartsim/dart/blob/56f0d4e9f6d6b0869fa76a1e6f221bb7afb956ae/dart/dynamics/Joint.cpp#L50-L56

The joint types (Free, Ball, Revolute) define the joint spaces and how to map the joint space to the joint motion in Cartesian space. However, whether the joint is treated as a kinematic or dynamic joint is determined by the actuator type. Actuator types such as FORCE, PASSIVE, SERVO, and MIMIC are used for dynamic joints, where the solver solves for forward dynamics given q, dq, and tau to find ddq. In contrast, other actuator types are used for kinematic joints, solving for inverse dynamics given q, dq, and ddq to find tau.

jcmayoral:
Thanks for answering, I have been testing how to do that. However, I get into trouble when trying to set the velocity to a FreeJoint. It does not move for a reason when I set other actuator type. Anyhow, I want to ignore the acceleration and set the velocity on cartesian space directly. The FreeJoint seems to consider anyway the accelerations to be able to move. 

The Kinematic Joint class that I am proposing ignores the acceleration and set the Velocities to the Joint directly. I wanted to add another class for that to avoid affecting the current behavior of dart.

jslee02:
Hm, would setting the acceleration to zero not achieve the desired behavior in this case?

Regardless, it would be helpful if you could create a unit or integration test that demonstrates the behavior you're trying to achieve. This would make it easier to identify any bugs or gaps in the existing features and add new functionality as needed. Additionally, having a clear example of your desired outcome will help ensure that we understand your requirements correctly.

jcmayoral:
I will handle your comments, and come back.

PR #2006, opened 2025-08-08
CMake: fix absolute install dirs

nim65s:
`CMAKE_INSTALL_*DIR` can be set as absolute.
In that case, it should not be prefixed with `CMAKE_INSTALL_PREFIX`.
Use `CMAKE_INSTALL_FULL_*DIR` instead.

ref. https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables

***


jslee02:
I'm not familiar with `CMAKE_INSTALL_FULL_*DIR`. Could you let me know how this is better than before? The CMake documentation says it shouldn't be used with `install()`, but we're using `PC_CONFIG_INSTALL_DIR` in `install()`:

> These variables shouldn't be used in [install()](https://cmake.org/cmake/help/v3.31/command/install.html#command:install) commands as they do not work with the cmake --install command's [--prefix](https://cmake.org/cmake/help/v3.31/manual/cmake.1.html#cmdoption-cmake--install-0) option, or with the [cpack](https://cmake.org/cmake/help/v3.31/manual/cpack.1.html#manual:cpack(1)) installer generators.

Would dart.pc still be [relocatable](https://github.com/dartsim/dart/pull/1529)?

jslee02:
Could we handle this conditionally?

* If CMAKE_INSTALL_*DIR is relative: keep #1529’s relocatable .pc (rebuild prefix from ${pcfiledir}, make libdir/includedir relative, keep DESTINATION relative).
* If absolute: write absolute paths as is (and install to that explicit path).

This might preserve relocatability, fixes the absolute-dir case, and avoids using CMAKE_INSTALL_FULL_*DIR in install(). How do you think?

nim65s:
Yes, I think you are right, I've been to quick here. Thanks for the review :)

jslee02:
Looks good now. Thanks for the fix!

PR #2104, opened 2025-11-04
Add conditional logging macros and update callers

jslee02:
## Summary
- add DART_*_IF logging helpers to wrap conditional logging
- replace inline conditionals with the new helpers across the codebase

## Testing
- not run (not requested)


jslee02:
@codex review

PR #2105, opened 2025-11-04
Add source context metadata to logging macros

jslee02:
## Summary
- route DART_* macros through a helper that includes file/line/function metadata
- share the same context-aware implementation with dart7 logging
- expose the conditional logging helpers on top of the new core utility

## Testing
- not run (not requested)


jslee02:
@codex review

PR #2141, opened 2025-11-10
Remove deprecated APIs from DART 6.11

jslee02:
## Summary
- drop DartLoader flag enum and the URDF parse overloads that took explicit retrievers/flags
- update python bindings/docs/stubs accordingly and expand changelog

## Testing
- pixi run lint


jslee02:
Codecov confirmed 100% coverage on the modified lines 👍 — no additional test updates needed for this batch.

PR #2152, opened 2025-11-12
Support Collada unit scaling in MeshShape

### LINKED ISSUE #287: Unit information in collada (.dae) files is ignored
Collada (.dae) files support the specification of a unit in the header of the file, which is then applied throughout the file. For example, this way all vertices could be specified in inches. The unit information then allows the correct conversion in meters. Since, the unit information is ignored, meshes specified in inches are rendered too large in DART.

Here is an example of a robot model that has meshes specified in inches: https://github.com/ros-industrial/kuka_experimental/tree/hydro-devel/kuka_kr210_support

This problem applies to both, URDF and SDF loaders. This problem is caused by assimp. We are using assimp for loading and representing meshes. However, assimp ignores the unit information in collada files. This is probably because assimp does not consider units in general and collada is one of the only file formats that allow the specification of a unit.

Gazebo and rviz consider the unit information in collada files. Gazebo does not use assimp to load mesh files. rviz uses assimp to load meshes but parses collada files a second time directly to read out the unit information.

I don't think this has a high priority to be fixed. As a workaround a scale can be added to the URDF file or the meshes can be rescaled. I am just leaving this here, to make people aware of it.

jslee02:
Fixes #287.

- Apply the Collada `asset/unit` metadata when loading meshes so `MeshShape` instances honor the file's length units.
- Extend the `MeshShape` unit test to cover Collada unit scaling and tidy the test asset path resolution so it is robust across build directories.

***

### DIFF
--- a/dart/dynamics/MeshShape.cpp
+++ b/dart/dynamics/MeshShape.cpp
@@ -572,15 +572,16 @@ bool isColladaResource(
   if (hasColladaExtension(uri))
     return true;
 
-  if (!retriever)
-    return false;
+  if (retriever) {
+    const auto parsedUri = common::Uri::createFromStringOrPath(uri);
+    if (parsedUri.mPath) {
+      const std::string resolvedPath = retriever->getFilePath(parsedUri);
+      if (!resolvedPath.empty())
+        return hasColladaExtension(resolvedPath);
+    }
+  }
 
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #2170, opened 2025-11-14
Improve ShapeNode inertia coverage

### LINKED ISSUE #348: Most implementations of Shape::computeInertia do not account for their local transform
Shapes are able to have a local transform within their Frame, but the computeInertia implementations of most Shape types do not account for that local transform. They should use coordinate transforms and the parallel axis theorem to shift the inertia matrices that they compute.

jslee02:
Fixes #348.

Address the inertia-transform gap tracked in #348 by exercising the new shape-node helpers end-to-end so parsers and BodyNode aggregation paths cover every mass case. Adds focused gtests plus light fixture assets to lock down the regression fix and documents the enforced brace style for follow-on work.

- add unit tests for ShapeNode density conversion and BodyNode invalid-mass handling
- add parser integration tests plus `.skel`/`.vsk` fixtures to drive the new inertia synthesis
- document the "always use braces" convention in onboarding material

***

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,9 +6,9 @@ version: 2
 
 # Set the OS, Python version and other tools you might need
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   tools:
-    python: "3.11"
+    python: "3.12"
 
 # Build configuration
 sphinx:


jslee02:
@codex review

PR #2172, opened 2025-11-14
package.xml: fix version number

scpeters:
According to [ROS REP 149](https://ros.org/reps/rep-0149.html#version), version numbers must have the format MAJOR.MINOR.PATCH.

Without this change, `colcon` fails to build a workspace containing dart.

***

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -4,7 +4,7 @@
        a Catkin workspace. Catkin is not required to build DART. For more
        information, see: http://ros.org/reps/rep-0136.html -->
   <name>dartsim</name>
-  <version>7.0.0.dev0</version>
+  <version>7.0.0</version>
   <description>
     DART (Dynamic Animation and Robotics Toolkit) is a collaborative,
     cross-platform, open source library created by the Georgia Tech Graphics


cottsay (approved):

scpeters:
> For context, `.dev` was added so we could publish `dartpy` packages to PyPI for pre-release commits (keeping the C++ and Python binding versions in sync). The idea was that .dev0 would mark these as pre-releases (e.g., current main before actually releasing v7.0), so PyPI wouldn’t treat them as the latest stable version and users without a pinned version would still get 6.x.
> 
> Given that, we probably should merge this change (since it’s the valid version scheme for this file) with CI green and look for a different workaround on the PyPI side.

Since uses the package.xml version as the source of truth, you could add an additional tag to the package.xml in an xml namespace that either holds the full dart version or a version suffix. You could also copy the [print_package_xml_version.py](https://github.com/gazebosim/gz-cmake/blob/gz-cmake5/tools/print_package_xml_version.py) script for parsing version numbers from package.xml (see also [GzGetPackageXmlVersion.cmake](https://github.com/gazebosim/gz-cmake/blob/gz-cmake5/cmake/GzGetPackageXmlVersion.cmake) and its [use in gz-math](https://github.com/gazebosim/gz-math/blob/gz-math9/src/python_pybind11/CMakeLists.txt#L5C3-L6))

PR #2187, opened 2025-11-16
Improve ShapeNode inertia coverage

### LINKED ISSUE #348: Most implementations of Shape::computeInertia do not account for their local transform

jslee02:
Fixes #348.

Add regression coverage for the new ShapeNode inertia helpers so BodyNode aggregation and parser importers are tested end-to-end, plus document the brace style that protects these paths. Includes two fixtures to exercise both `.skel` and `.vsk` importers.

- add unit tests for ShapeNode density conversion and invalid mass providers
- add Skel/Vsk integration tests driven by new fixtures
- update onboarding docs with the "always use braces" reminder

***

### DIFF
--- a/data/skel/test/inertia_from_shapes.skel
+++ b/data/skel/test/inertia_from_shapes.skel
@@ -53,7 +53,7 @@
                     <transformation>0.03 0 0 0 0 0</transformation>
                     <geometry>
                         <box>
-                            <size>0 0.1 0.1</size>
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

PR #2194, opened 2025-11-16
Embed optional collision backends in core dart

jslee02:
## Purpose
- Stop shipping separate `dart-collision-*` components so the optional Bullet and ODE detectors behave like first-class features of the `dart` library.
- Make the build configuration fail fast when those detectors are requested but their dependencies are missing, eliminating silent feature drops in CI and downstream packages.
- Expose the new behavior through the generated CMake/package metadata so integrators immediately know which collision engines were compiled in.
- Start the larger effort to remove per-dependency component targets (e.g., `dart-optimizer-*`) so future follow-up PRs can apply the same pattern across the rest of the tree.

## Changes
1. Add cache flags `DART_BUILD_COLLISION_BULLET` / `DART_BUILD_COLLISION_ODE` (default `ON` in Pixi) and teach the Bullet/ODE subdirectories to
   - bail out early when the option is `OFF`,
   - emit `FATAL_ERROR` when dependencies are missing while the option is `ON`, and
   - `cmake_language(DEFER ...)` their sources/includes/link flags directly onto the `dart` target so there are no standalone collision libraries.
2. Update the generated `DARTConfig.cmake`, pkg-config hints, examples, tutorials, benchmarks, tests, and dartpy bindings to link only against `dart`/`dart-utils*` while conditionally enabling Bullet/ODE-specific executables based on the new flags.
3. Refresh docs/onboarding, CHANGELOG, and AGENTS.md to document the embedded backends, the new options, and the libsdformat requirement.

## Implications
- **Build behavior:** Turning either option `ON` without the corresponding dependency now stops configuration instead of producing a partially featured build. Users who need Bullet/ODE simply enable the option (default `ON` in Pixi) and install `libbullet-dev` ≥3.25 or `libode-dev` ≥0.13; everyone else can set the option `OFF` to skip the backend entirely.
- **Package managers:** The main `dart` artifact now carries Bullet/ODE when the options are enabled, so there is no need to produce or ship extra `dart-collision-*` packages. Downstream projects discover the availability via `DART_BUILD_COLLISION_*` cache exports or the existing `HAVE_BULLET/HAVE_ODE` config defines.
- **Users:** Project code and CMake files no longer reference `dart-collision-*` targets; linking against `dart` is sufficient. Tests/examples guard Bullet/ODE-specific binaries with the new CMake options, so developers running partial builds receive clear skip/failure messages instead of link errors.
- **Follow-up work:** Optimizer plugins and other optional integrations still produce component targets today; those will be cleaned up in subsequent PRs using the same approach as soon as their dependencies are audited.

Milestone: DART 7.0

***

- [ ] Run `pixi run test-all` to lint, build, and test your changes _(not run in this environment)_
- [ ] Add unit tests for new functionality _(covered by existing collision suites; no new APIs added)_
- [ ] Add Python bindings (dartpy) if applicable _(not applicable beyond the linking cleanup already in this change)_


jslee02:
Fixed the merge markers in `tests/integration/CMakeLists.txt`; kept the `INTEGRATION_simulation_World` target and dropped the obsolete `dart-collision-*` links.

jslee02:
@codex review

jslee02:
@codex review

PR #1835, opened 2024-08-09
Added coupler constraint

adityapande-1995:
This PR adds a new Coupler Constraint, to address https://github.com/dartsim/dart/issues/1756

***

#### Before creating a pull request

- [ ] Document new methods and classes
- [ ] Format new code files using ClangFormat by running `make format`
- [ ] Build with `-DDART_TREAT_WARNINGS_AS_ERRORS=ON` and resolve all the compile warnings

#### Before merging a pull request

- [ ] Set version target by selecting a milestone on the right side
- [ ] Summarize this change in `CHANGELOG.md`
- [ ] Add unit test(s) for this change
- [ ] Add Python bindings for new methods and classes

PR #2212, opened 2025-11-19
Add coupler constraint support and example

### LINKED ISSUE #1756: Add Coupler Constraint (similar to a gearbox constaint from Gazebo)
## Problem

We have been working to port the Gearbox joint functionality from Gazebo classic to gz-sim (new Gazebo) and have been planning to change the SDFormat specification for this feature to use `//joint/axis/mimic` (similar to URDF) instead of the Gearbox joint type (see proposal in https://github.com/gazebosim/sdf_tutorials/pull/62). This feature creates a linear relationship between the output position of two joints, and we have been working to implement it in the gz-physics dartsim plugin using the `MimicMotorConstraint` (see https://github.com/gazebosim/gz-physics/pull/431). However, I realized recently while testing https://github.com/gazebosim/gz-sim/pull/1838#issuecomment-1595953064 that the `MimicMotorConstraint` does not create a bilateral constraint between the two joints. In order to support the Gearbox joint functionality, we need a constraint that applies equal and opposite impulses to each joint in the constraint, and in my testing the `MimicMotorConstraint` does not do this.

## Proposed Solution

I'd like to be able to create a constraint between two joints that creates a linear relationship between the positions of two joints by applying equal and opposite impulses to each joint. Since the `MimicMotorConstraint` was already present in dartsim, I think it would be least disruptive to add a new constraint type. In Drake, for example, they added Coupler constraints in https://github.com/RobotLocomotion/drake/pull/17639 and are using that constraint type in their URDF parser to support the mimic element (see https://github.com/RobotLocomotion/drake/pull/18728). "Coupler constraint" sounds like a reasonable name to me.

## Describe alternatives you've considered

Change the behavior of the `MimicMotorConstraint` to be a bilateral constraint. This would be disruptive, however, for an already released feature.

## Additional context

See https://github.com/gazebosim/gz-sim/pull/1838#issuecomment-1595953064

jslee02:
- add `CouplerConstraint` so mimic joints can optionally be enforced bilaterally and ensure the constraint solver instantiates it only when `Joint::setUseCouplerConstraint(true)` is set
- expose the flag through dartpy, add integration coverage, and ship a GUI example so downstream code (e.g., gz-sim mimic joints) can exercise the new constraint
- resolves #1756 and supersedes the exploratory work in #1835 by documenting the intended gearbox use case and providing a pixi-friendly example

**Testing**
- `pixi run test-all`
- `pixi run ex coupler_constraint`

https://github.com/user-attachments/assets/aef4d1c5-04f1-444f-b8e9-0d98726dc091

Closes #1835

***

- [x] Run `pixi run test-all` to lint, build, and test your changes

### DIFF
--- a/dart/constraint/CouplerConstraint.hpp
+++ b/dart/constraint/CouplerConstraint.hpp
@@ -35,8 +35,9 @@
 
 #include <dart/constraint/ConstraintBase.hpp>
 
-#include <dart/Export.hpp>
 #include <dart/dynamics/MimicDofProperties.hpp>
+
+#include <dart/Export.hpp>
 
 #include <vector>
 
--- a/examples/coupler_constraint/main.cpp
+++ b/examples/coupler_constraint/main.cpp
@@ -99,6 +99,11 @@ dart::dynamics::SimpleFramePtr createLimitGuide(
   if (!world || !followerJoint)
     return nullptr;
 
+  const auto* revolute
+      = dynamic_cast<dart::dynamics::RevoluteJoint*>(followerJoint);
+  if (revolute == nullptr)
+    return nullptr;
+
   auto frame = SimpleFrame::createShared(Frame::World(), name);
   auto line = std::make_shared<LineSegmentShape>(0.05f);
   line->addVertex(Eigen::Vector3d::Zero());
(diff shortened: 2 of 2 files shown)


jslee02:
Merging now. If anything still isn't working or you hit new issues, let me know and I'll fix it forward. Thanks!

PR #2239, opened 2025-11-22
Deep copy shapes when cloning skeletons (fixes #896)

### LINKED ISSUE #896: Shapes are not properly cloned
Hello,

I have a minor (but critical to my applications) issue regarding the clone function of Skeletons. So, when I do:

```cpp
auto skel_clone = my_skeleton->clone();
```
I was expecting that everything (including the visualization and collision shapes) are cloned (i.e. copied), but it seems that this is not the case. This is very easily replicated with something like the following:

```cpp
// includes and code

namespace {
    dart::dynamics::SkeletonPtr global_robot;
}

void load_global_robot() {
// code to load/create the global skeleton
}

void add_damage(const dart::dynamics::SkeletonPtr& skel){
    std::string leg_bd_name = "my_damaged_body_name";
    auto bd = skel->getBodyNode(leg_bd_name);
    // set half the mass
    bd->setMass(bd->getMass() / 2.0);
    auto nodes = bd->getShapeNodes();

    for (auto node : nodes) {
        Eigen::Vector3d tr = node->getRelativeTranslation();
        tr(1) = tr(1) / 2.0;
        node->setRelativeTranslation(tr);
        // here's the problem -- the pointers 's' are always the same
        auto s = node->getShape();
        if (s->getType() == "BoxShape") {
            auto b = (dart::dynamics::BoxShape*)s.get();
            Eigen::Vector3d size = b->getSize();
            size(2) = size(2) / 2.0;
            b->setSize(size);
        }
        else if (s->getType() == "CylinderShape") {
            auto b = (dart::dynamics::CylinderShape*)s.get();
            b->setHeight(b->getHeight() / 2.0);
        }
    }
}

int main(int argc, char** argv)
{
    auto my_new_skel = global::global_robot->clone();
    add_damage(my_new_skel);
    // if now we print the change shapes (in our case, lengths or size) for the global::global_robot, they will have the changed values and not the original values as they should
    return 0;
}
```

I know my use-case is not the general use-case (i.e., usually we do not intentionally break our robots) and I have found ways to by-pass it, I think it should be easy for you to do in DART.

Many thanks in advance,
Konstantinos

jslee02:
Implement deep copying of shapes when cloning skeletons so clones own distinct shape instances. Update SkeletonClone tests to use current APIs and cover the cloning behavior. Fixes #896.

---

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -1646,8 +1646,7 @@ test-gz = { cmd = """
         export LD_LIBRARY_PATH=.deps/gz-physics/build/lib:$CONDA_PREFIX/lib:${LD_LIBRARY_PATH:-}
         ROOT_DIR=$PWD
         cd .deps/gz-physics/build
-        python "$ROOT_DIR/scripts/cmake_build.py" --build-dir . --parallel ${DART_PARALLEL_JOBS} --target all \
-          || { find . -maxdepth 5 -type f -name "FAKE_INSTALL*log" -print -exec sed -n 1,200p {} \\; || true; exit 1; }
+        python "$ROOT_DIR/scripts/cmake_build.py" --build-dir . --parallel ${DART_PARALLEL_JOBS} --target all
         python <<\"PY\"
 import os
 import subprocess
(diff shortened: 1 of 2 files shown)


jslee02:
@codex review

PR #2282, opened 2025-11-26
Stabilize FCL contact normals and collision ordering

### LINKED ISSUE #1425: contact normal changes randomly between two opposite directions in Gazebo
hello,

I am using DART as physics engine for Gazebo 9.

I found this bug when attaching a contact sensor to a model.

When I display the gazebo topic associated with this sensor, I can see that the normals of the contacts are not constant, even if the model is lying on the ground and is at rest (only gravity force is applied on it). More precisely, the directions of the normals oscillate between approximately (0, 0, 1), and (-0, -0, -1). The sign of the associated force also vary accordingly.

The bug is very easy to reproduce in Gazebo, you just need to attach a contact sensor to a model, and display the associated gazebo topic.

However, I don't know how to reproduce it with DART as a standalone, as I never used it.

I think the bug may be in dart code, as I already checked the gazebo code and didn't find the bug yet...

Does anyone has already seen this in dart ?
Or at the contrary, can confirm that the contacts are working properly in dart, and this issue is coming from Gazebo ?

jslee02:
## Summary
- make FCL collision object keys deterministic and use them to keep contact ordering and normals consistent across clones and runs
- ensure collision object user data is preserved after refresh to avoid missed collisions when shapes update
- add regression coverage for collision ordering and FCL refresh behavior

Fixes #1425.

---

### DIFF
--- a/dart/collision/fcl/FCLCollisionDetector.cpp
+++ b/dart/collision/fcl/FCLCollisionDetector.cpp
@@ -1432,10 +1432,8 @@ void postProcessDART(
         numContacts++;
       }
 
-      const auto* fclObj1
-          = static_cast<FCLCollisionObject*>(o1->getUserData());
-      const auto* fclObj2
-          = static_cast<FCLCollisionObject*>(o2->getUserData());
+      const auto* fclObj1 = static_cast<FCLCollisionObject*>(o1->getUserData());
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

PR #2316, opened 2025-12-14
io: unified model loading API (issue #604)

### LINKED ISSUE #604: Naming consistency in parsing libraries
Right now, we have ``dart::utils::SkelParser`` which parses the native DART format ``.skel``, and then we have ``dart::utils::DartLoader`` which unintuitively loads ``.urdf``, and then there's ``dart::utils::SdfParser`` which is a class for loading SDFs, except all its functions are static, so it's really more of a namespace than a class (the situation for ``SkelParser`` is the same).

I would propose the following:

``kido::utils::skel`` should be the namespace for the functions that parse the ``.skel`` format. The functions should not be in a class named ``SkelParser`` at all. This namespace will encapsulate a ``component`` library named ``kido-utils-skel`` with a dependency on ``tinyxml2`` (related to the discussion here #434).

``kido::utils::sdf`` should be the namespace for the functions that parse the ``.sdf`` format. Similarly, those functions should not be in a class named ``SdfParser`` at all. This namespace will encapsulate a ``component`` library named ``kido-utils-sdf`` with a dependency on ``sdformat``.

``kido::utils::urdf`` should be the namespace for the class ``kido::utils::urdf::Loader``. The urdf parsing still needs a class at the moment, in order to add package directories. The namespace will encapsulate a ``component`` library named ``kido-utils-urdf`` with a dependency on ``urdfdom``.

I suggest we at least do this for the first major release of KIDO. I don't know for sure whether or not we want to implement this in the next major release of DART as well.

jslee02:
### Summary
- Add consolidated C++ model-loading entry points: `dart::io::readWorld` / `dart::io::readSkeleton`
- Keep unified API in `dart::io` (no new nested format namespaces); delegate to `dart::utils::*` parsers
- Extend `dart::io::ReadOptions` with minimal high-frequency format knobs:
  - URDF `package://` resolution via `ReadOptions::addPackageDirectory`
  - SDF default root joint selection via `ReadOptions::sdfDefaultRootJointType`
- Migrate examples/tests/benchmarks/tutorials to the `dart::io` APIs and add/extend unit tests
- Document the decision and usage in `docs/onboarding/io-parsing.md`

Closes #604.

---

- [x] Run `pixi run test-all` to lint, build, and test your changes
- [ ] Add Python bindings (dartpy) if applicable (N/A: C++ `dart::io` only for now)

### DIFF
--- a/docs/onboarding/README.md
+++ b/docs/onboarding/README.md
@@ -812,6 +812,7 @@ sequenceDiagram
 ```
 
 **Key Files**:
+
 - [`dart::io` unified API](dart/io/Read.hpp)
 - Implementation: [`dart/io/Read.cpp`](dart/io/Read.cpp)
 - Internal parser: [`UrdfParser`](dart/utils/urdf/UrdfParser.hpp)


jslee02:
@codex review

PR #2356, opened 2025-12-29
Enable ImGuiWidget subclassing in dartpy

### LINKED ISSUE #1411: ImGui support in dartpy
Dartpy uses the OpenSceneGraph environment from Dart. To add some graphical interface to the simulation (e.g. buttons, showing the simulation time, ...) objects subclassing `ImGuiWidget` would be needed. This is not possible with the current state of `dartpy`, right? 

I started implementing this by adding a trampoline class as described in the pybind11 documentation [here](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#overriding-virtual-functions-in-python) to allow subclassing of `ImGuiWidget` (which should work similar to what was done for `WorldNode`). The following problems arise with this approach:

1. `ImGuiWidget` objects are passed by `shared_ptr<ImGuiWidget>`, e.g. for adding them. But subclasses of ImGuiWidget are handled by Python itself, so passing them as (C++ managed) `shared_ptr` does not seem feasible. Is there any clever approach to handle this?
2. Do we have to expose the whole ImGui funcitonality via `pybind11` as well (which seems to be a lot), or is it possible to integrate other bindings (e.g. [pyimgui](https://github.com/swistakm/pyimgui))?

I would be glad to implement this feature with your input on the above problems.

jslee02:
## Summary
- add dartpy ImGuiWidget binding trampoline and wire it into the GUI module
- add a Python ImGui example plus a note to install imgui-bundle
- add an ImGuiWidget unit test and update a legacy dart.common type hint
- refresh pixi config/lock for the Python ImGui dependencies

## Testing
- pixi run test
- pixi run test-all
- pixi run -e gazebo test-gz
- pixi run lint

Closes #1411

### DIFF
--- a/docs/onboarding/ci-cd.md
+++ b/docs/onboarding/ci-cd.md
@@ -29,6 +29,7 @@ DART uses GitHub Actions for continuous integration and deployment. The CI syste
   - `gh pr status --json ...` can error with `Unknown JSON field: ...` if you request unsupported fields; use `gh pr status` (no JSON) or `gh pr view --json ...`.
   - `gh run view --json ...` only accepts supported fields; prefer the `url` field instead of guessing `htmlURL`.
   - `gh pr checks` may show duplicate entries when workflows run for both `push` and `pull_request` events; compare the run URLs and focus on the newest one.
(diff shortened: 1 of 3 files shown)


jslee02:
@codex review

PR #2368, opened 2026-01-02
Mirror python examples layout

jslee02:
## Summary
- reorganize python/examples into numbered categories to match the C++ layout
- update py-ex runner to resolve examples from the new layout
- refresh python bindings docs/example pointers

## Testing
- pixi run py-ex hello_world


jslee02:
@codex review

jslee02:
@codex review

PR #2410, opened 2026-01-07
Fix CTest library paths for shared library loading

jslee02:
## Summary

Fix `pixi run test-all` for local development by:
1. Fixing CTest library paths for shared library loading on Linux and Windows
2. Fixing macOS CI hang caused by DYLD_LIBRARY_PATH and SIP interaction
3. Adding GCC 15 compatibility for upstream dependency warnings

## Problem 1: Library Path Issues on Linux

Tests were failing with errors like:
```
cannot open shared object file: No such file or directory
```

This happened because the CMake test environment configuration used `PATH` instead of the platform-appropriate library path variable (`LD_LIBRARY_PATH` on Linux).

### Fix
- Use `LD_LIBRARY_PATH` on Linux
- Use `PATH` on Windows
- Use config-aware generator expressions (`$<TARGET_FILE_DIR:...>`) for library paths to support multi-config generators (MSVC/Xcode)
- For simulation-experimental tests, also include main dart library path since some tests link against it transitively

## Problem 2: macOS CI Hang

The macOS CI jobs were hanging indefinitely (9+ hours) when `DYLD_LIBRARY_PATH` was set in the test environment. This is caused by **System Integrity Protection (SIP)** on macOS, which strips `DYLD_*` environment variables for protected executables, causing tests to hang waiting for libraries they cannot find.

### Fix
Skip setting library path environment variables on macOS entirely. CMake's built-in RPATH handling (`@rpath`, `BUILD_RPATH`) works correctly on macOS without needing `DYLD_LIBRARY_PATH`.

Platform-specific behavior:
- **Windows**: Use `PATH` (required for DLL loading)
- **Linux**: Use `LD_LIBRARY_PATH` (required for .so loading)
- **macOS**: Skip - rely on CMake's RPATH handling

## Problem 3: GCC 15 Build Failure

Build failed with GCC 15 due to a deprecated header warning:
```
/usr/include/c++/15/ciso646:49:6: error: #warning "<ciso646> is not a standard header since C++20"
```

This comes from upstream dependencies (octomap uses `<ciso646>` in OcTreeKey.h; fcl pulls it in transitively).

### Fix
Add `-Wno-error=cpp` for GCC 15+ to prevent the `#warning` from being treated as an error. This is a workaround with a clear TODO comment to remove once upstream dependencies fix the issue.

## Files Changed

- `CMakeLists.txt` - GCC 15 compatibility workaround
- `tests/CMakeLists.txt` - `dart_add_test()` function library paths
- `cmake/dart_defs.cmake` - `dart_build_tests()` function library paths
- `cmake/dart_experimental_defs.cmake` - `dart_experimental_add_test()` function library paths

## Testing

Verified locally with GCC 15.2.0:
- `pixi run build` ✓
- `pixi run test` ✓ (135/135 tests pass)

CI verification:
- Linux CI: ✓ Passing
- Windows CI: ✓ Passing
- macOS CI: Should now complete (was hanging due to DYLD_LIBRARY_PATH issue)

### DIFF
--- a/cmake/dart_defs.cmake
+++ b/cmake/dart_defs.cmake
@@ -1975,8 +1975,12 @@ function(dart_create_library)
   target_compile_features(${ARG_NAME} PUBLIC cxx_std_20)
 
   # Add compile definition for source directory (used for relative paths in logging)
+  # Include both DART_SOURCE_DIR and DART_EXPERIMENTAL_SOURCE_DIR for compatibility
+  # with the experimental logging helper in dart/simulation/experimental/common/logging.hpp
   target_compile_definitions(${ARG_NAME}
-    PUBLIC DART_SOURCE_DIR="${CMAKE_SOURCE_DIR}"
+    PUBLIC
+      DART_SOURCE_DIR="${CMAKE_SOURCE_DIR}"
+      DART_EXPERIMENTAL_SOURCE_DIR="${CMAKE_SOURCE_DIR}"
   )
 
   # Include directories - default to parent of current source dir


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #2086, opened 2025-11-02
[6.16] Allow servo joints to recover from position limits

jslee02:
Fixes #1683, #1818

## Summary

- cherry-picked Haidar's servo-limit commits onto  to preserve authorship
- added  regression plus SDF fixture to reproduce the stuck-at-limit bug
- retained existing Issue1583 regression coverage while validating the fix

## Testing

- [x] cmake --build build/default/cpp/Release --target test_Issue1583
- [x] ./tests/regression/test_Issue1583

PR #2421, opened 2026-01-13
Allow servo joints to recover from lower bound position limits

azeey:
## Summary

- https://github.com/dartsim/dart/pull/2086 fixes the bug where servo joints couldn't recover from position limits, however, this only works for upper bound position limits. This PR fixes that and updates the test to cover both upper and lower bounds.

## Testing

- `tests/regression/test_Issue1583`

---

#### Checklist


jslee02:
Thanks for the fix!

PR #2428, opened 2026-01-14
Fix null pointer dereference in getValue* functions (DART 7)

### LINKED ISSUE #2425: Bug Report：getValueBool causes Segmentation Fault (Null Dereference) when XML child element is missing
> For questions, use [Discussions](https://github.com/dartsim/dart/discussions) instead.
> For platform-specific build/test failures, use the [Build/Test Failure template](https://github.com/dartsim/dart/issues/new?template=build-test-failure.yml).

**Before submitting:**

- [x] Checked [documentation](https://dart.readthedocs.io/) and existing issues

**Environment:**

- DART version: 6.15.0 and later
- OS:  Ubuntu 22.04 LTS
- Installation:  Source build
- Compiler (if source): Clang 17.0.0

**Expected Behavior:**
When `dart::utils::getValueBool` is called with an XML element and a child name that does not exist, the function should gracefully handle the missing child. It maybe should either Return a default value (e.g., `false`) and optionally print a warning, or throw a caught exception.

**Current Behavior:**
The function causes a **Segmentation Fault (SIGSEGV)** due to a null pointer dereference.
Specifically, it calls `parentElement->FirstChildElement(name.c_str())`, which returns `nullptr` when the child is missing. It then immediately chains `.GetText()` on this `nullptr`, causing the crash.

https://github.com/dartsim/dart/blob/f17905c365e6e4608dcd043066a17f74d318bce3/dart/utils/XmlHelpers.cpp#L372
I noticed this also occurs in other functions in XmlHelpers.cpp.

**Steps to Reproduce:**

1. Create a `tinyxml2::XMLElement`.
2. Call `dart::utils::getValueBool(element, "non_existent_child_name")`.
3. The program crashes with SIGSEGV.

**Code to Reproduce:**

```cpp
// Minimal reproducible example
#include <gtest/gtest.h>
#include <dart/utils/XmlHelpers.hpp>
#include <tinyxml2.h>
#include <iostream>
#include <string>
#include <csignal>

// Bug Reproduction: dart::utils::getValueBool causes Null Pointer Dereference
// when the requested child element does not exist.
TEST(DartUtilsBugReproduction, GetValueBoolNullDeref) {
    // 1. Create a simple XML Document
    // <root>
    //   <existing_child>true</existing_child>
    // </root>
    tinyxml2::XMLDocument doc;
    tinyxml2::XMLElement* root = doc.NewElement("root");
    doc.InsertEndChild(root);

    tinyxml2::XMLElement* existingChild = doc.NewElement("existing_child");
    existingChild->SetText("true");
    root->InsertEndChild(existingChild);

    // 2. Normal case: Querying an existing child works
    bool val = dart::utils::getValueBool(root, "existing_child");
    EXPECT_TRUE(val);

    std::cout << "[Info] Successfully retrieved existing child." << std::endl;

    // 3. Reproduce Crash: Querying a non-existent child "missing_child"
    // Expected: The library function calls FirstChildElement("missing_child") which returns nullptr,
    //           then immediately calls nullptr->GetText(), triggering a Segfault.

    std::cout << "[Info] Attempting to retrieve missing child (Expect Segfault)..." << std::endl;
    
    // We assert that the process is killed by SIGSEGV, confirming the Null Pointer Dereference.
    ASSERT_EXIT({
       dart::utils::getValueBool(root, "missing_child");
   }, ::testing::KilledBySignal(SIGSEGV), ".*");
```

**Suggested Fix:**
Maybe could check if `FirstChildElement` returns `nullptr` before accessing `GetText()`.
```cpp
const tinyxml2::XMLElement* child = parentElement->FirstChildElement(name.c_str());
if (child == nullptr) {
    // Handle missing element 
    return false;
}
const char* text = child->GetText();
if (text == nullptr) {
    // Handle empty element
    return false;
}
std::string str = text;
```

jslee02:
## Summary

Fix issue #2425: `XmlHelpers` `getValue*` functions crash with SIGSEGV when child element is missing due to null pointer dereference.

## Changes

- Add `getChildElementText()` helper function that validates child element exists and has text content
- Update all 15 `getValue*` functions to use the safe helper
- Throw `std::runtime_error` with descriptive message on missing/empty child element
- Add comprehensive unit tests for null child element handling

## Functions Fixed

- `getValueString`, `getValueBool`, `getValueInt`, `getValueUInt`
- `getValueFloat`, `getValueDouble`, `getValueChar`
- `getValueVector2d`, `getValueVector3d`, `getValueVector3i`
- `getValueVector6d`, `getValueVectorXd`, `getValueVec3`
- `getValueIsometry3d`, `getValueIsometry3dWithExtrinsicRotation`

## Related

- Fixes #2425
- Companion PR for `release-6.16`: #2429

## Testing

- Added 19 unit tests covering:
  - Valid input parsing
  - Missing child element - verifies exception is thrown
  - Empty child element - verifies exception is thrown


jslee02:
Closing in favor of #2429 targeting release-6.16. The fix will flow to main when release-6.16 is merged upstream.

PR #2442, opened 2026-01-17
Fix: Validate SphereShape radius to prevent assertion failures (DART 7)

jslee02:
## Summary

- Replace `DART_ASSERT(radius > 0.0)` with proper validation using `std::isfinite()` and `DART_WARN()`
- Invalid values (NaN, infinity, non-positive) are now rejected with a warning instead of crashing
- Add unit tests covering all validation cases

## Motivation

Addresses [gz-physics#843](https://github.com/gazebosim/gz-physics/issues/843) where DART crashes with an assertion failure when `SphereShape::setRadius()` receives invalid radius values from upstream physics interfaces.

## Related PR

- DART 6.16 backport: #2441

## Changes

- `dart/dynamics/SphereShape.cpp`: Added validation with graceful rejection
- `tests/unit/dynamics/test_SphereShape.cpp`: New test file with 5 test cases
- `tests/unit/CMakeLists.txt`: Added SphereShape test entry
- `CHANGELOG.md`: Added entry under Collision and Geometry section

## Testing

- Unit tests verify valid radius acceptance and invalid value rejection (NaN, Inf, zero, negative)
- Follows existing validation pattern from `World::setTimeStep()`


jslee02:
@codex review

PR #2515, opened 2026-02-11
fix: Add FreeBSD support in shared library platform switches

jslee02:
## Summary

Addresses review feedback from #2513:

- **P1 (shared library)**: Add `DART_OS_FREEBSD` to all platform-conditional blocks in `shared_library.hpp` and `shared_library.cpp`. Without this, FreeBSD builds hit `#error Unhandled platform` and missing `DYNLIB_*` macros. FreeBSD uses the same `dlopen`/`dlsym`/`dlclose` API as Linux.

- **P2 (CI prereq enforcement)**: Restore `exit 1` when Docker/KVM prerequisites are missing in `ci_freebsd.yml`. The previous PR downgraded this to a warning, which silently skips tests and reports success. Now the job fails fast with a clear error, and redundant step output variables and `if`-guards are removed.

## Files Changed

| File | Change |
|------|--------|
| `dart/common/shared_library.hpp` | Add `DART_OS_FREEBSD` to `DYNLIB_HANDLE`, extension, and prefix conditionals |
| `dart/common/shared_library.cpp` | Add `DART_OS_FREEBSD` to `DYNLIB_LOAD/SYM/UNLOAD` and `getLastError()` conditionals |
| `.github/workflows/ci_freebsd.yml` | Restore `exit 1` on missing prereqs; remove unused outputs and `if`-guards |

## Testing

- C++ lint passes (`pixi run check-lint-cpp`)
- Python lint passes (`pixi run check-lint-py`)
- `shared_library.cpp` compiles clean (verified in build output)
- Full build failure is pre-existing (assimp header issue in local env), unrelated to these changes

### DIFF
--- a/.github/workflows/ci_freebsd.yml
+++ b/.github/workflows/ci_freebsd.yml
@@ -19,17 +19,14 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Check runner requirements (KVM and Docker)
-        id: prereqs
         run: |
           FAILED=false
 
           echo "=== Docker Diagnostics ==="
           if docker info > /dev/null 2>&1; then
             echo "Docker is accessible"
-            echo "docker_available=true" >> $GITHUB_OUTPUT
           else
             echo "::error::Docker not accessible - check socket permissions"
-            echo "docker_available=false" >> $GITHUB_OUTPUT
             FAILED=true
           fi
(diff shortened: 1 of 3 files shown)


jslee02:
@codex review

PR #2540, opened 2026-05-03
Fix spdlog/fmt runtime logging (DART 6.16)

### LINKED ISSUE #2538: Spdlog/fmtlib build failure (dart v6.16.7)
Dart v6.16.7 fails to build as the following calls to spdlog fails with GCC 16.1.1, spdlog 1.17.0, libfmt 12.1.0. The format string must either be available at compile time or wrapped in fmt::runtime(...).

https://github.com/dartsim/dart/blob/805778e3ee4503cb7c902804373790dd81b0f469/dart/common/detail/Logging-impl.hpp#L83
https://github.com/dartsim/dart/blob/805778e3ee4503cb7c902804373790dd81b0f469/dart/common/detail/Logging-impl.hpp#L95
https://github.com/dartsim/dart/blob/805778e3ee4503cb7c902804373790dd81b0f469/dart/common/detail/Logging-impl.hpp#L107
https://github.com/dartsim/dart/blob/805778e3ee4503cb7c902804373790dd81b0f469/dart/common/detail/Logging-impl.hpp#L119
https://github.com/dartsim/dart/blob/805778e3ee4503cb7c902804373790dd81b0f469/dart/common/detail/Logging-impl.hpp#L131
https://github.com/dartsim/dart/blob/805778e3ee4503cb7c902804373790dd81b0f469/dart/common/detail/Logging-impl.hpp#L143

jslee02:
## Summary

- Wrap DART logging format strings with spdlog's `SPDLOG_FMT_RUNTIME(...)` compatibility macro.
- Add release-branch logging coverage for runtime `std::string` format strings.
- Document the DART 6.16.8 build compatibility fix.

## Motivation / Problem

- #2538 reports that DART 6.16.7 fails to build with newer spdlog/fmt combinations because DART forwards logging format parameters to spdlog as non-compile-time values.
- `main` has the corresponding DART 7 fix in #2542, while this PR applies the release-6.16-side fix.

## Why CI Missed This

- The release-6.16 Pixi environment currently pins `fmt >=11.1.4,<12` while the report uses fmt 12.1.0 with spdlog 1.17.0.
- Existing CI did compile and run logging tests, but only against the pinned dependency set. That made the API path look healthy while missing fmt 12's stricter compile-time format-string enforcement.
- A normal regression test under the same pinned dependencies is useful for API intent, but it is not sufficient to catch this class of dependency-forward breakage by itself.

## Long-Term Test Strategy

- Keep the new `LoggingTest.RuntimeFormatString` unit test as the local API regression for runtime format-string support.
- Add a scalable dependency-forward CI job or smoke workflow that builds at least the common logging target against newer supported system dependencies, especially latest compatible spdlog/fmt pairs.
- Prefer a focused compatibility smoke target over duplicating the full CI matrix: it should compile a representative logging call such as `dart::common::info(std::string, ...)` and run `UNIT_common_Logging` with the forward dependency set.
- Keep the regular locked Pixi CI as the reproducible baseline, and use the forward-compatibility job to catch breakage caused by ecosystem upgrades before downstream packagers hit it.

## Changes / Key Changes

- Add `DART_SPDLOG_RUNTIME(...)` through spdlog's `SPDLOG_FMT_RUNTIME(...)` macro so fmt-backed builds get runtime format handling without assuming a minimum fmt runtime API newer than DART's supported dependency range.
- Apply the wrapper to all DART common logging level functions.
- Add a `LoggingTest.RuntimeFormatString` regression case.

## Testing

- `git diff --check`
- Header compile smoke test for `dart::common::info(std::string, ...)`
- `cmake --build build/default/cpp/Release --target UNIT_common_Logging`
- `ctest --test-dir build/default/cpp/Release --output-on-failure -R UNIT_common_Logging`
- `pixi run lint`
- `pixi run test`
- `pixi run test-all`

## Breaking Changes

- [x] None

- Fixes #2538
- Main / DART 7 PR: #2542

---

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
   * Fix iterator invalidation in Subject/Observer notification loops that caused non-deterministic SEGFAULT on macOS arm64 Debug builds
 
-  * Fix spdlog/fmt 12 builds by treating DART logging format parameters as runtime format strings. ([#2538](https://github.com/dartsim/dart/issues/2538))
+  * Fix spdlog/fmt 12 builds by treating DART logging format parameters as runtime format strings. ([#2540](https://github.com/dartsim/dart/pull/2540), [#2538](https://github.com/dartsim/dart/issues/2538))
 
 * Simulation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
   * Fix iterator invalidation in Subject/Observer notification loops that caused non-deterministic SEGFAULT on macOS arm64 Debug builds
 
+  * Fix spdlog/fmt 12 builds by treating DART logging format parameters as runtime format strings. ([#2538](https://github.com/dartsim/dart/issues/2538))
+
 * Simulation
 
   * Reject NaN, infinite, zero, and negative `World::setTimeStep()` values to prevent invalid timesteps from reaching the ODE LCP solver. ([#2532](https://github.com/dartsim/dart/pull/2532), [#2531](https://github.com/dartsim/dart/issues/2531))
--- a/dart/common/detail/Logging-impl.hpp
+++ b/dart/common/detail/Logging-impl.hpp
@@ -37,12 +37,7 @@
 
 #if DART_HAVE_spdlog
   #include <spdlog/spdlog.h>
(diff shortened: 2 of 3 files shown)


jslee02:
@codex review

PR #2552, opened 2026-05-06
Use C++20 ranges transform for string cleanup

jslee02:
## Summary

- Use C++20 `std::ranges::transform` for behavior-preserving string case cleanup.

## Motivation / Problem

- DART 7 requires C++20, and these call sites can be expressed directly as ranges algorithms instead of iterator-pair transforms.

## Changes / Key Changes

- Updated common string case helpers to use `std::ranges::transform`.
- Updated Collada extension normalization, SDF lower-copy helper, and LCP example argument parsing to use ranges transforms.
- Added a direct `<algorithm>` include in the SDF helper implementation for ranges algorithm use.

## Testing

- `pixi run lint`
- `pixi run build`
- `pixi run test-unit`
- `pixi run test-all`

## Breaking Changes

- [x] None

- None

---

#### Checklist

### DIFF
--- a/dart/common/string.cpp
+++ b/dart/common/string.cpp
@@ -46,7 +46,7 @@ std::string toUpper(std::string str)
 //==============================================================================
 void toUpperInPlace(std::string& str)
 {
-  std::transform(str.begin(), str.end(), str.begin(), ::toupper);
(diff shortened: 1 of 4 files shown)


jslee02:
@codex review

PR #2555, opened 2026-05-06
Adopt C++20 math and string utilities

jslee02:
## Description
- use C++20 `std::numbers` for standard math constants and derive remaining thresholds from those constants
- replace ad hoc prefix/suffix checks with C++20 string APIs
- use `std::midpoint` for scalar midpoint/centering calculations in collision, IK, and parser defaults

## Testing
- pixi run test-all


jslee02:
@codex review

PR #2560, opened 2026-05-06
Adopt C++20 jthread in scoped concurrency

jslee02:
## Summary

- Use `std::jthread` in scoped test and benchmark concurrency code where thread lifetime is already bound to the local scope.
- Remove explicit joins that are now handled by the standard C++20 joining thread type.

## Motivation / Problem

- DART targets C++20, and `std::jthread` better expresses scoped thread ownership for these local concurrency helpers.
- This reduces manual join bookkeeping in tests while preserving the same execution behavior.

## Changes / Key Changes

- Updated the dynamics cache benchmark thread loop to use joining threads.
- Updated signal, profile, and profiler tests to rely on `std::jthread` for automatic joining.

## Testing

- `pixi run test-all`

## Breaking Changes

- [x] None

- None

---

#### Checklist

### DIFF
--- a/tests/benchmark/dynamics/bm_dynamics_cache.cpp
+++ b/tests/benchmark/dynamics/bm_dynamics_cache.cpp
@@ -914,7 +914,7 @@ static void BM_ParallelWorlds(benchmark::State& state)
   }
 
   for (auto _ : state) {
-    std::vector<std::thread> threads;
+    std::vector<std::jthread> threads;
     threads.reserve(numThreads);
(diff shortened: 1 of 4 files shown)


jslee02:
@codex review

jslee02:
Closing this one because the required macOS arm64 CI toolchain does not provide `std::jthread`/`std::stop_token` in libc++, so this C++20 adoption is not portable across DART's supported CI matrix. The existing `std::thread` code is preferable to adding local compatibility scaffolding here.

PR #2563, opened 2026-05-06
Adopt C++20 standard concepts in numeric helpers

jslee02:
## Summary
- Replace numeric type-trait predicates with C++20 standard concepts in SIMD geometry helpers and scalar SIMD operations.
- Use `std::floating_point` in math helper static assertions.
- Restore macOS libc++ portability by avoiding `std::jthread` in profiling tests and the parallel-world benchmark.
- Add direct `<concepts>` includes where concepts are now referenced.

## Motivation
C++20 concepts express template requirements directly and make these numeric constraints easier to read than equivalent `std::is_*` trait expressions. The profiling-thread change keeps the branch buildable on CI platforms whose standard library does not yet provide `std::jthread`.

## Changes
- Converted float/double-only static assertions to `std::same_as`.
- Converted floating-point and integral `requires` clauses to `std::floating_point` and `std::integral`.
- Converted math helper floating-point static assertions to `std::floating_point`.
- Replaced `std::jthread` with `std::thread` plus explicit joins in profiling coverage and the dynamics cache benchmark.

## Testing
- `pixi run lint`
- `pixi run build`
- `pixi run test-unit`

## Breaking Changes
None.

## Related Issues
None.

### DIFF
--- a/dart/simd/detail/scalar/operations.hpp
+++ b/dart/simd/detail/scalar/operations.hpp
@@ -37,6 +37,7 @@
 
 #include <algorithm>
 #include <bit>
+#include <concepts>
 
 #include <cmath>
 #include <cstddef>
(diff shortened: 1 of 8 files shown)


jslee02:
@codex review

jslee02:
@codex review

PR #2570, opened 2026-05-07
Adopt C++20 span range construction

jslee02:
## Summary

- Use C++20 `std::span` range construction/conversion across span-returning APIs, bindings, examples, and tests.

## Motivation / Problem

- Several call sites manually rebuilt spans from `data()`/`size()` or explicitly wrapped contiguous containers before passing them to `std::span` parameters.
- C++20 allows these conversions directly for compatible contiguous ranges, reducing boilerplate while preserving behavior.

## Changes / Key Changes

- Return existing contiguous containers directly from `std::span`-returning accessors where lifetime and constness already match.
- Pass vectors and arrays directly to APIs taking `std::span` instead of redundant `std::span(...)` wrappers.
- Keep explicit brace span construction in template calls where span type deduction is still required.

## Testing

- `pixi run lint`
- `pixi run build`
- `pixi run test-unit`
- `pixi run test-all`

## Breaking Changes

- [x] None

- None

---

#### Checklist

No changelog, tests, docs, or binding additions are needed because this is an internal refactor with no behavior or API change.

### DIFF
--- a/dart/collision/collision_result.cpp
+++ b/dart/collision/collision_result.cpp
@@ -74,7 +74,7 @@ const Contact& CollisionResult::getContact(std::size_t index) const
 //==============================================================================
 std::span<const Contact> CollisionResult::getContacts() const
 {
-  return std::span<const Contact>(mContacts);
+  return mContacts;
 }
 
 //==============================================================================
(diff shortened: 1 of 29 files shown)


jslee02:
@codex review

jslee02:
@codex review

PR #2588, opened 2026-05-07
Adopt C++20 signed sizes in examples and tests

jslee02:
## Summary
- Use `std::ssize` for signed Eigen/container index loops in the Atlas and Hubo puppet examples.
- Apply the same signed-size cleanup in dynamics and simulation tests.
- Use signed-size counters in the LCP benchmark while preserving the required `int` cast at the solver-parameter API boundary.

## Motivation
Continue the C++20 modernization pass by replacing manual signed size comparisons and casts with standard library signed-size utilities.

## Changes
- Adds `<iterator>` where `std::ssize` is used.
- Converts local loop bounds and benchmark counters to use `std::ssize`.
- Keeps behavior unchanged and avoids changing public APIs.

## Testing
- `git diff --check`
- `pixi run lint`
- `pixi run build`
- `pixi run test-unit`
- `pixi run test-all`

## Breaking Changes
None.

## Related Issues
None.

### DIFF
--- a/examples/atlas_puppet/main.cpp
+++ b/examples/atlas_puppet/main.cpp
@@ -38,6 +38,7 @@
 #include <dart/all.hpp>
 #include <dart/io/read.hpp>
 
+#include <iterator>
 #include <string_view>
 
 using namespace dart::common;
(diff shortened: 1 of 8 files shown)


jslee02:
@codex review

jslee02:
@codex review

PR #2623, opened 2026-05-08
Adopt C++20 library helpers in numeric code

jslee02:
## Summary

- Replace iterator-pair algorithm calls with C++20 `std::ranges` forms where a range is already available.
- Use `std::make_unique_for_overwrite` for Dantzig LCP workspace buffers that are filled before use.
- Keep Dantzig state storage value-initialized where initial permutation can touch state bits before they are semantically assigned.

## Motivation / Problem

- Continue the C++20 modernization pass with a larger, behavior-preserving library-helper batch across core numeric, SIMD, and test code.

## Changes / Key Changes

- Updated linkage and SIMD helpers to use `std::ranges` algorithms for whole-range operations.
- Updated Dantzig LCP workspace allocation to skip unnecessary value-initialization for buffers that are subsequently initialized by the solver.
- Updated LCP and SIMD tests to use range-based algorithms when copying or computing extrema.

## Testing

- `pixi run lint`
- `pixi run build`
- `pixi run test-all`

## Breaking Changes

- [x] None

- None

---

#### Checklist

### DIFF
--- a/dart/dynamics/linkage.cpp
+++ b/dart/dynamics/linkage.cpp
@@ -346,7 +346,7 @@ void Linkage::Criteria::expandToTarget(
     trimBodyNodes(newBns, _target.mChain, true);
   } else if (target_bn->descendsFrom(start_bn)) {
     newBns = climbToTarget(target_bn, start_bn);
-    std::reverse(newBns.begin(), newBns.end());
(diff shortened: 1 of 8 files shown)


jslee02:
@codex review

PR #2663, opened 2026-05-18
Refactor articulated dynamics through internal algorithms

jslee02:
## Summary

- Consolidate Skeleton ABA/RNEA traversal behind internal detail algorithms and a SkeletonDynamicsView adapter.
- Keep public Skeleton hot paths on the existing contiguous `mSkelCache.mBodyNodes` order, with vector-backed helpers for cache-friendly traversal and raw pointer helpers for adapters/batch paths.
- Add direct and batched articulated-dynamics entry points, correctness tests, deterministic benchmark inputs, and benchmark coverage.

## Motivation / Problem

- DART currently keeps articulated dynamics traversal logic scattered across Skeleton and BodyNode recursion paths. Centralizing traversal gives one implementation point for maintainability and creates a clearer path for batch processing, cache-friendly adapters, and future SIMD-oriented data layouts.
- This PR does not change public gz-physics-facing APIs or replace Skeleton/World storage with a packed SoA layout. It uses the existing cached body-node order as the compatibility-preserving data-locality step; deeper packed storage should be a follow-up with broader downstream compatibility testing.
- Performance work needs evidence. The benchmark state generator is now deterministic so baseline and PR runs use the same representative joint states instead of process-random states.

## Changes / Key Changes

- Add `dart::dynamics::detail::rnea`, `aba`, `rneaBatch`, and `abaBatch` over an internal model/view concept.
- Add `ArticulatedDynamicsAccess` as the narrow internal friend boundary for protected BodyNode dynamics kernels.
- Route `Skeleton::computeForwardDynamics()` and the common no-options `computeInverseDynamics(false, false, false)` path through vector-backed helpers over `mSkelCache.mBodyNodes`.
- Keep `SkeletonDynamicsView` as the adapter for direct and batch algorithm benchmarks without requiring additional public API.
- Add unit tests comparing direct ABA/RNEA against explicit legacy-recursion helpers, independent entry points, heterogeneous batches, round-trip representative joints, and zero-DoF no-op behavior.
- Add benchmark cases for legacy vs direct ABA/RNEA and uniform/heterogeneous batch execution.

## Performance Evidence

Baseline-vs-PR benchmarking was run against the merge-base with `origin/main`, `f308030e7b8d`, using separate baseline and PR build trees. The current PR head measured here is `a1b4794cf2d`. Baseline and PR benchmark sources both used the deterministic benchmark-state generator so inputs matched across binaries.

The current host became heavily contended by unrelated builds during the final full 12-case run. I discarded that full run because several rows had double-digit CVs, including ABA rows above 16-24%, which is not suitable evidence. Affinity-pinned and warmed-up probes were also worse under that contention. The defensible current-head comparison I was able to complete before the host saturated is the focused public RNEA set below.

Command shape:

```bash
/tmp/dart_pr2663_<base-or-pr>_build/bin/bm_dynamics_cache \
  --benchmark_filter='BM_(BinaryTree|Serial)_Revolute_InverseDynamics/(20|50)/min_time:0.300' \
  --benchmark_repetitions=15 \
  --benchmark_report_aggregates_only=true \
  --benchmark_min_time=1.0s \
  --benchmark_out=<base-or-pr>.json \
  --benchmark_out_format=json
```

Lower time is better; negative delta means the PR is faster.

| Case | Baseline mean ns | PR mean ns | Mean delta | PR CV | Median delta |
| --- | ---: | ---: | ---: | ---: | ---: |
| BinaryTree 20 RNEA | 4,460 | 4,332 | -2.86% | 0.72% | -2.52% |
| BinaryTree 50 RNEA | 11,139 | 10,961 | -1.61% | 1.43% | -1.91% |
| Serial 20 RNEA | 4,699 | 4,696 | -0.07% | 2.30% | -0.89% |
| Serial 50 RNEA | 12,129 | 12,147 | +0.14% | 4.54% | -0.49% |

Interpretation: the current common public RNEA path is faster for BinaryTree cases, neutral within noise for Serial 20, and neutral for Serial 50 by mean with a faster median but high CV. A full public ABA/RNEA table still needs to be rerun on an idle host before claiming an across-the-board speedup.

## Testing

- `pixi run lint`
- `git diff --check`
- `cmake --build /tmp/dart_pr2663_pr_build --target UNIT_dynamics_ArticulatedDynamicsAlgorithms -j 2`
- `ctest --test-dir /tmp/dart_pr2663_pr_build --output-on-failure -R '^UNIT_dynamics_ArticulatedDynamicsAlgorithms$'`
- `cmake --build /tmp/dart_pr2663_pr_build --target UNIT_dynamics_SoftBodyNode UNIT_dynamics_ZeroDofJoint -j 2`
- `ctest --test-dir /tmp/dart_pr2663_pr_build --output-on-failure -R '^UNIT_dynamics_(SoftBodyNode|ZeroDofJoint)$'`
- Baseline-vs-PR benchmark run listed above.

## Breaking Changes

- [x] None

- None

---

#### Checklist

- [x] CHANGELOG.md updated if required (not required: internal refactor/benchmark scaffolding, no public API change)
- [x] Add Python bindings (dartpy) if applicable (not applicable: no public Python API)


jslee02:
@codex review

jslee02:
@codex review

PR #2694, opened 2026-05-23
Add Filament renderer profiling, backend selection, scene-sync cache

jslee02:
## Summary

- Make the Filament GUI renderer faster and more observable while keeping the renderer backend hidden behind renderer-neutral `dart::gui` APIs.

## Motivation / Problem

The Filament renderer was built for feature completeness, not performance, and had no in-app way to see where frame time goes. Profiling shows the renderer itself is cheap (<1 ms/frame) while per-frame CPU **scene extraction dominates** (it rebuilds full mesh/heightmap geometry every frame). The graphics backend was also hardcoded to OpenGL with no way to select or compare alternatives.

## Changes / Key Changes

- **Live performance HUD** (`--perf-hud`): a small ImGui overlay showing CPU per-phase timings, **GPU frame time** (from Filament's frame-info history), FPS, rolling CPU/GPU history plots, the active backend, and frame counters. Built on the existing `ProfileAccumulator` (no parallel timing system); off by default and does no work when disabled.
- **Runtime backend selection**: `dart::gui::RunOptions::backend`, the `--backend` flag, and the `DART_FILAMENT_BACKEND` env var select `default`/`opengl`/`vulkan`/`noop` (env > option > default OpenGL) with graceful fallback and a startup log. Materials are compiled for both OpenGL and Vulkan (`matc -a opengl -a vulkan`) so the choice needs no separate build. `noop` gives a CPU-only path for benchmarking. No `filament::Engine::Backend` leaks into public headers.
- **Scene-sync optimization**: a `dart::gui::RenderableExtractor` caches each shape's `GeometryDescriptor` keyed by shape version, so `describeShape` is not rebuilt every frame for static geometry. Output is equivalent to `extractRenderables()`; kinds whose geometry changes without a version bump (soft mesh, point cloud, voxel grid) are rebuilt every frame so deforming/live geometry stays correct. Local measurement: per-iteration scene extraction ~3.4x faster on the default scene.
- Durable docs: renderer performance/profiling/backend guidance added to `docs/onboarding/gui-rendering.md`; the future fidelity-profile + camera-sensor seam (real-time vs offline, lens distortion, egocentric, depth/segmentation) recorded in `docs/design/filament_fidelity_profile.md`.

## Testing

```
pixi run ex dartsim --perf-hud --gui-scale 2
```

image

- `pixi run lint` — clean.
- `pixi run check-docs-policy`, `pixi run check-ai-commands` — clean.
- `ctest -R UNIT_gui_FilamentSceneExtraction` — passes (extraction equivalence preserved by the cache).
- Headless smoke across 9 scenes (hello-world, boxes, rigid-chain, rigid-loop, mixed-chain, vehicle, soft-bodies, point-cloud, heightmap) — all render without error; `Final contacts` unchanged.
- Visual screenshots inspected: the HUD overlay; OpenGL vs Vulkan equivalence; heightmap (cached mesh) and a deforming soft body (rebuilt each frame) render correctly.
- Note: no new dedicated unit test was added for the HUD/backend/cache; covered by the existing extraction test plus the smoke set and manual visual checks.

## Breaking Changes

- [x] None

Additive only: a new `RunOptions::backend` field, new CLI flags, and an internal geometry cache. Default behavior (OpenGL, no HUD) is unchanged.

- None (feature targets `main` only).

---

#### Checklist

- [ ] Add unit tests for new functionality (existing extraction test + smoke set used; no new dedicated test)
- [x] Add Python bindings (dartpy) if applicable (N/A — GUI run options are C++/CLI only)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #2700, opened 2026-05-23
Add native continuous collision detection (CCD)

jslee02:
## Summary

Extends the native collision backend with a continuous (time-of-impact) layer
built for modern rigid- and deformable-body solvers, and benchmarks it against
each reference engine's best continuous query.

- **IPC-class primitive CCD** (point–triangle, edge–edge) via additive
  conservative advancement, with an exact coplanarity-cubic validation path —
  the conservative, minimum-separation-aware queries barrier/IPC solvers need and
  that the rigid-only reference engines do not provide.
- **Both-moving convex cast** (`convexCast`) with a C2A-style directed motion
  bound (linear undirected + angular projected by `|n×axis|`, which collapses for
  coaxial screw motion).
- **Cubic-Bezier spline cast** (`splineCast`) covering polynomial motion — the
  one motion model the linear/screw casts cannot represent. Dual-mode:
  `Conservative` (default) finds the true first contact; `Fast`
  (`CcdAdvancement::Fast`) trades the no-tunnelling guarantee for larger steps.
- Supporting GJK refactor (templated, de-virtualized query in `gjk_inl.hpp`).

## Results

Native beats all three reference engines on **all 9 comparative cases**
(translation, rotation, screw, spline; every shape family), margins 1.2×–6467×
— see `docs/dev_tasks/continuous_collision_detection/03-benchmark-evidence.md`.

The spline row is the subtle one: the reference engine's spline is faster only
because its bound is **non-conservative and returns a wrong first contact** (it
reports contact deep inside the overlap, having tunneled past the true first
touch). Native's conservative mode is the only spline cast that finds the true
first contact, and its opt-in fast mode is still 1.9× faster than the reference
engine.

## Test plan

- [x] `test_ccd` 75/75, `test_primitive_ccd` 16/16, `test_gjk` 32/32,
      `test_narrow_phase` 24/24 (default env)
- [x] `bm_comparative_ccd` builds and runs in the `collision-reference` env
- [x] Conservativeness verified by unit tests (`τ_native ≤ τ_substep`; chord
      cast over the same endpoints misses) and a Monte-Carlo check
- [x] `clang-format`, `codespell`, `prettier`, and `check_api_boundaries.py` clean
- [ ] Re-confirm benchmark ratios on a quiesced host (this one is loaded; ratios
      are stable, absolute numbers inflated)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #2771, opened 2026-05-29
Reuse a resident device buffer in the IPC deformable GPU PSD projection

jslee02:
## Summary

Removes the per-iteration device allocation from the experimental IPC deformable GPU PSD-projection backend (PLAN-081 Phase 3 follow-up).

The CUDA backend previously allocated and freed a device buffer on **every** projected-Newton iteration, and the backend adapter additionally copied the packed Hessian blocks through a temporary host `std::vector` on each call. Under the solver that is a `cudaMalloc`/`cudaFree` plus a host allocation per outer iteration.

This PR:
- Replaces the per-call `DeviceDoubleBuffer` with a **resident, grow-on-demand** device buffer reused across calls (reallocates only when a batch exceeds the current capacity) and freed when the GPU backend is uninstalled.
- Adds a raw-pointer **in-place** projection that runs on the caller's packed buffer through the resident device buffer; the backend adapter routes through it, so the offload no longer makes a per-call host copy.
- The public `std::vector<double>&` overload validates the buffer size and delegates to the same in-place path (unchanged API).

## Behavior

**Bit-identical** results — same kernel, same host↔device transfers; only the device storage is reused. This is purely a performance change confined to the opt-in `dart-simulation-experimental-cuda` sidecar. The default CPU runtime is untouched, so the no-GPU-runtime-dependency guarantee and `pixi run test-all` (which doesn't build the sidecar) are unaffected.

The remaining per-batch host↔device copies are inherent to the current design; keeping the assembly/solve fully on-device is a documented follow-up.

## Tests & evidence

- New `ResidentDeviceBufferReusesAllocationAcrossCalls` CUDA test (via an exposed allocation counter): a first batch allocates, repeated same-or-smaller batches reuse without allocating, a larger batch grows it exactly once, and restoring the backend releases it — with GPU/CPU-reference parity (`< 1e-9`) preserved at every step.
- All 6 cuda tests pass on an RTX 5000 Ada (`pixi run -e cuda ctest -R deformable_psd`); `BackendInjectionRoutesThroughCoreSeam` still passes through the new in-place adapter.
- `pixi run test-all`: all 6 categories PASS (default build unaffected).

Milestone: DART 7.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1030,6 +1030,18 @@ qdot)` that reaches the target exactly even under inertial coupling. The
     1024, ~4x at 4096, and ~9x at 16384, so the backend adapter's minimum GPU
     batch size is raised from 64 to ~1024 blocks (small batches stay on the CPU
     backend where the host/device round trip would otherwise dominate).
+  - Gave the experimental IPC deformable GPU PSD projection a resident device
+    buffer (PLAN-081 Phase 3). The CUDA backend previously allocated and freed a
+    device buffer (and copied through a temporary host vector) on every
+    projected-Newton iteration; it now reuses one persistent device allocation
+    that grows on demand and is freed when the GPU backend is uninstalled, and
+    the backend adapter projects in place on the caller's packed buffer (no
+    per-call host copy). Results are bit-identical to before (same kernel and
+    transfers), so it is purely a performance change confined to the opt-in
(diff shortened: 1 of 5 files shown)


jslee02:
Superseded: this PR's resident-GPU device-buffer change (cuda code + test + CHANGELOG/README) was inadvertently bundled into #2772's squash-merge, so it is already on `main`. Closing as redundant; the work is landed.

PR #2816, opened 2026-05-31
fix(test): use std::numbers::pi instead of M_PI (fixes Windows build)

jslee02:
Fixes the Windows build break. **CI Windows has been red on `main` for hours** (since #2808) with no fix in flight.

## Cause

`tests/unit/simulation/experimental/world/test_deformable_body.cpp` (added in #2808, "Add Coulomb friction against the capsule rod obstacle") uses `M_PI`. MSVC does not declare `M_PI` unless `_USE_MATH_DEFINES` is defined before `<cmath>`, so the file compiles on glibc/libc++ (Linux/macOS) but fails on Windows:

```
test_deformable_body.cpp(1197,28): error C2065: 'M_PI': undeclared identifier
```

This breaks the `tests` target build → the **Tests (Release)** / CI Windows job (and blocks every PR's Windows check, including #2815 and #2803).

## Fix

Use the portable C++20 `std::numbers::pi` — the repo's prevailing convention (27 existing usages) — with an explicit `<numbers>` include. Verified the construct compiles at `-std=c++20`. Behavior is identical (`-std::numbers::pi / 2.0 == -M_PI / 2.0`).

### DIFF
--- a/tests/unit/simulation/experimental/world/test_deformable_body.cpp
+++ b/tests/unit/simulation/experimental/world/test_deformable_body.cpp
@@ -43,6 +43,7 @@
 
 #include <array>
 #include <limits>
+#include <numbers>
 #include <sstream>
 #include <string>
 #include <string_view>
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

jslee02:
Folded into #2815. main has both a Linux break (demos refactor) and this Windows break, and each is a required check on the other's platform — so a standalone PR would be mutually blocked. #2815 now carries this M_PI fix (commit b4a3932) plus the Linux demos fixes, green on all platforms.

PR #2843, opened 2026-06-01
Add checkpoint/replay parity tests for the experimental World (PLAN-080 B4)

jslee02:
## Summary

Advances the DART 7 **serialization/replay parity** gate (PLAN-080 B4): *"World topology, state, model assets, and record/replay round-trip with bounded error."*

The existing `test_serialization.cpp` suite (31 cases) thoroughly proves world **state** round-trips through `saveBinary`/`loadBinary`, but **none stepped a world after `loadBinary`** to prove the trajectory continues correctly — the guarantee checkpoint/restore and deterministic replay actually depend on. This adds that missing coverage across both integration paths.

## What's added

**Multibody joint path** (deterministically evolving fixed-base two-link revolute chain, nonzero initial joint velocities + gravity):
- **`CheckpointReloadContinuesIdentically`** — a world checkpointed mid-trajectory and reloaded continues identically to the still-running original instance.
- **`CheckpointReloadMatchesUninterruptedRun`** — checkpoint + reload mid-run matches an uninterrupted reference run, proving save/load does not perturb the trajectory.

**Rigid-body path** (single dynamic body free-falling under gravity):
- **`RigidBodyCheckpointReloadContinuesIdentically`** — same replay guarantee for the rigid-body integration stage, checked through the world state vector `[q; q̇]` (populated for dynamic rigid bodies), verifying it round-trips through serialization.

Notes:
- The multibody cases compare generalized **joint state** via the public handle API. `World::getStateVector()` is the differentiable rigid-body reduction (`3 × dynamic rigid bodies`) and is empty for a pure-multibody scene, so using it there would make the check pass *vacuously*; it is used (and asserted size 6) only in the rigid-body case where it is meaningful.
- Each case **guards that the scene actually evolves** (`> 1e-6`) so parity is not trivially satisfied by a static scene.
- Tolerance `1e-10`; measured agreement is near machine epsilon. Not loosened to force a pass.

## Verification

- `pixi run build-simulation-experimental-tests` ✅
- `test_serialization`: **34/34 gtest cases pass**; the three new cases confirmed via `--gtest_filter='*Checkpoint*'`.
- `pixi run check-api-boundaries` ✅ (public experimental facade only; no ECS/registry/solver/backend types)
- `pixi run lint` ✅

## Scope

Tests-only; no production code changed. Covers the open-chain (multibody) and rigid-body serialization/replay facets of B4. Model-asset round-trip (load a model, serialize, reload) builds on the model-loader work tracked separately and can extend this suite once that lands.


jslee02:
@codex review

PR #2838, opened 2026-06-01
Promote experimental model loading and unified contact/constraint solver (PLAN-080)

jslee02:
## Summary

This PR promotes the experimental model-loading pipeline and the unified contact/constraint solver for the DART 7 release. It advances **PLAN-080**, which gates the **PLAN-041** "Official simulation API promotion" release path.

## Motivation / Problem

DART 7 needs the experimental world API to load existing DART model assets and resolve contact consistently across rigid bodies and articulated links before the API can be promoted. This PR closes a large part of that gap by bridging classic `Skeleton`/`World` model data into experimental `Multibody`/`World` objects and replacing the split semi-implicit contact passes with a unified boxed-LCP solve.

## DART 7 release gates addressed

- **Experimental model loading**: URDF, SDF, MJCF, and SKEL models load into the experimental world API through a Skeleton-to-Multibody bridge, covering topology, DOF, transforms, joint properties, and collision-shape geometry.
- **Contact/constraint parity**: a unified boxed-LCP solver resolves rigid and multibody link contacts, including Coulomb friction, in a single solve. This is the foundation for the contact-response, friction, and joint-limit evidence required by the gate.

### Experimental model loading

- Adds Skeleton-to-Multibody and multi-skeleton World loading APIs, with dartpy bindings.
- Supports all legacy joint families, including revolute/prismatic joint properties and parent-side pre-joint offsets for branching and offset roots.
- Preserves collision-shape offsets and translates box, sphere, capsule, cylinder, plane, mesh, convex mesh, heightmap, soft mesh, and compound collision shapes.
- Adds collision-query support for broad-phase queries, body-type filters, self-filtering, and cached native collision-query worlds.
- Updates experimental GUI shape-case handling.

### Unified contact/constraint solver

- Adds unified contact-problem types and an assembler that couples rigid-rigid and link-contact rows, including shared dynamic obstacles.
- Solves rigid contacts and multibody link contacts together with Coulomb friction in one boxed-LCP system.
- Adds generalized rank-deficient fallback and row-islanded constraint solving.
- Wires `UnifiedConstraintStage` into the default semi-implicit `World::step` pipeline while leaving the variational branch and legacy stages intact.

### Planning and handoff state

- Refreshes PLAN-080, the plan dashboard, API-boundary inventory, and the rigid-body solver resume handoff for the remaining DART 7 parity work.

## Testing

Local validation before publication and after base reconciliation:

- `pixi run update-api-boundary-inventory` — PASS
[...]

PR #2837, opened 2026-06-01
Add AVBD rigid contact joints and feature keys

jslee02:
## Summary

- Extends the experimental AVBD rigid-body path beyond the foundation merge by adding feature-scoped contact manifolds, rigid contact rows, and point/fixed joint rows.
- Keeps the behavior opt-in through the experimental rigid contact stage while preserving existing non-AVBD contact behavior.
- Adds cylinder and capsule feature keys so warm-started contact rows remain scoped to the physical surface feature that generated them.

## Motivation / Problem

- #2832 landed the AVBD rigid block foundation, but rigid contacts still needed stable row keys, manifold persistence, and joint constraints before the path could support useful rigid assemblies.
- Without shape-specific contact feature keys, warm starts can be reused across unrelated capsule/cylinder contact regions and make manifold state harder to reason about.
- Fixed-joint projection should continue to solve rigid assemblies even during no-contact frames so the experimental stage does not silently skip internal constraints.

## Changes / Key Changes

- Adds AVBD contact feature row keys, descriptor helpers, rigid row helpers, friction pairs, row driver logic, and rigid contact manifold construction.
- Adds rigid world contact snapshot/solve/writeback helpers and an opt-in rigid contact stage path.
- Adds point-joint and fixed-joint extraction/wiring through the world step, including no-contact fixed-joint projection.
- Adds cylinder and capsule contact feature IDs and regression coverage for manifold separation and rigid-world snapshot feature persistence.
- Updates PLAN-104, dashboard, AVBD dev-task notes, and CHANGELOG.md.

## Before / After

| Area | Before | After |
| --- | --- | --- |
| AVBD rigid contact rows | Foundation rows existed, but contact state was not feature-scoped through the rigid-world path. | Contact rows have canonical feature keys and persistent warm-started manifold rows. |
| Rigid joints | Rigid point/fixed joints were not wired through the AVBD world step. | Point and fixed joints produce AVBD rows and fixed joints project even in no-contact frames. |
| Cylinder/capsule contacts | Contact rows could not distinguish shape endpoint/side features. | Cylinder and capsule feature IDs separate side and cap manifolds. |

## Testing

- `git diff --check origin/main HEAD`
- `CMAKE_BUILD_DIR=build/default/cpp/Release pixi run python scripts/cmake_build.py --target test_avbd_rigid_block --target test_vbd_contact --target test_boxed_lcp_contact`
- `build/default/cpp/Release/bin/test_avbd_rigid_block`
[...]

PR #2836, opened 2026-06-01
Modernize the dart::gui viewer theme and add GPU (CUDA) deformable solve to py-demos

jslee02:
## Summary

- Restyle the `dart::gui` Dear ImGui viewer (used by py-demos / dartsim) with a cohesive modern dark theme, replacing the stock `ImGui::StyleColorsDark()` look.
- Make `pixi run -e cuda py-demos` build a CUDA-enabled dartpy + Filament GUI and run the demos with GPU (CUDA) deformable solve enabled by default, with a runtime toggle.

## Motivation / Problem

- The viewer shipped with ImGui's default dark theme — saturated-blue selection across every interactive state, cramped spacing, square widgets, centered titles. It reads as a generic debug overlay and undersells the project in screenshots and demos.
- The demos run their deformable physics on the CPU. DART 7 already has an experimental CUDA sidecar for the projected-Newton PSD projection (the one interactive deformable hotspot), but nothing wired it into the demos, and the `cuda` environment intentionally disabled the GUI/dartpy because the prebuilt Filament libraries do not link under the conda portability toolchain.

### Viewer theme
- Centralize all styling in `configureImGuiStyle()` in `dart/gui/detail/imgui_overlay.cpp`, split into `applyModernDarkColors()` (scale-independent palette) and `applyModernDarkMetrics()` (sizes/rounding/spacing): cool neutral layered surfaces, one restrained blue accent reserved for selection/focus/active, a VS Code-style accent overline on the active tab, flat hairline-bordered widgets with soft rounding, pill scrollbars, breathable spacing, left-aligned caret-free titles.
- Metrics scale via `ScaleAllSizes` (`--gui-scale`/DPI); colors stay scale-independent. Docking-only color slots (`DockingPreview`, `DockingEmptyBg`, `DockingSeparatorSize`) are guarded by `#ifdef IMGUI_HAS_DOCK` so the non-docking system-ImGui build still compiles under `-Werror`.

### GPU (CUDA) deformable solve in py-demos
- dartpy: expose process-wide GPU control in `simulation_experimental` — `is_cuda_available()`, `set_gpu_deformable_solve(enable)`, `is_gpu_deformable_solve_enabled()`. They forward to the CUDA sidecar under `#ifdef DART_ENABLE_EXPERIMENTAL_CUDA` and are safe no-ops returning `False` on the default (non-CUDA) build, so the default environment is unchanged. dartpy links the CUDA sidecar and gets the compile definition only when the target exists.
[...]

PR #2846, opened 2026-06-02
Fix experimental build regression: AVBD rigid contact vs compound CollisionGeometry

jslee02:
## Urgent: main does not currently compile

`main` fails to build the experimental simulation tests (`dart_experimental_tests`) due to a **semantic conflict between two separately-merged PRs**:

- **#2838** (PLAN-080 model loader) changed `comps::CollisionGeometry` from a single `shape` member to a `std::vector<CollisionShape> shapes` (compound-shape support), exposing `hasShapes()` / `getPrimaryShape()`.
- **#2837** (AVBD rigid contact feature keys) still reads `geometry->shape.{type,halfExtents,radius}` in `detail/deformable_vbd/rigid_world_contact.hpp`.

Both merged cleanly in git (no textual conflict) but are mutually inconsistent, so the tree no longer compiles:

```
rigid_world_contact.hpp:285: error: 'const struct ...comps::CollisionGeometry'
  has no member named 'shape'; did you mean 'shapes'?
```

This is a classic parallel-development *semantic* merge collision that the per-file git merge can't catch.

## Fix

Use `CollisionGeometry::getPrimaryShape()` (the first shape, matching the prior single-shape behavior) in the AVBD rigid-world contact feature-keying function, and skip endpoints whose geometry has no shape. Behavior-preserving for the existing single-shape bodies; for compound geometry it keys on the primary shape (box/cylinder/capsule indices remain 0, as before).

## Verification

- `pixi run build-simulation-experimental-tests` ✅ (was failing on `main`)
- `ctest -L simulation-experimental` → **58/58 pass**
- `pixi run lint` ✅

## Note

Recommend fast-tracking: until this lands, every PR's experimental build/CI is red. Found while rebasing other DART 7 work onto latest `main`.

### DIFF
--- a/dart/simulation/experimental/detail/deformable_vbd/rigid_world_contact.hpp
+++ b/dart/simulation/experimental/detail/deformable_vbd/rigid_world_contact.hpp
@@ -282,31 +282,36 @@ inline AvbdContactEndpointId avbdRigidWorldContactEndpointId(
     return endpoint;
   }
 
-  if (geometry->shape.type == CollisionShapeType::Box
-      && geometry->shape.halfExtents.allFinite()
-      && (geometry->shape.halfExtents.array() > 0.0).all()) {
+  // CollisionGeometry now holds a vector of shapes (compound-shape support,
+  // PLAN-080 #2838); the AVBD feature keying uses the primary (first) shape,
+  // matching the previous single-shape behavior.
+  const auto* shape = geometry->getPrimaryShape();
+  if (shape == nullptr) {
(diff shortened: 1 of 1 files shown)


jslee02:
Redundant: the same fix (use `CollisionGeometry::getPrimaryShape()` in the AVBD rigid-contact feature keying) already landed on `main` via #2836 (commit b5922f4ddec). `main` now compiles the experimental tests; closing this duplicate.

PR #2873, opened 2026-06-03
Reset stale CMake compiler cache before CUDA py-demos configure

jslee02:
## Summary

Fixes the CUDA-enabled `pixi run -e cuda py-demos` build so it doesn't silently lose the `dartpy` target when a build directory was previously configured with different compilers.

The `config-py` task pins the host and CUDA compilers for the CUDA build (system `cc`/`c++` for the Filament GUI link, conda `nvcc` for CUDA). When the target build directory already holds a `CMakeCache.txt` from an earlier configure that used different compilers, CMake refuses to change the compiler in place and **silently re-runs with default options**, dropping `DART_BUILD_DARTPY` — so the demos can no longer import `dartpy`.

This adds a guard in the CUDA branch of `config-py` that, before invoking `cmake`:
- reads the cached `CMAKE_C_COMPILER` / `CMAKE_CXX_COMPILER` / `CMAKE_CUDA_COMPILER` / `CMAKE_CUDA_HOST_COMPILER` from `CMakeCache.txt`,
- compares them against the pinned CUDA host toolchain, and
- removes the stale `CMakeCache.txt` and `CMakeFiles/` when they diverge, so the reconfigure adopts the pinned compilers and keeps the `dartpy` target.

Scoped entirely to the CUDA opt-in (`DART_ENABLE_EXPERIMENTAL_CUDA_VALUE = ON`); the default and lean compute-only `build-cuda`/`test-cuda` paths are untouched.

## Changes

- `pixi.toml` — reset stale CMake compiler-cache metadata in the `config-py` CUDA branch before `cmake` runs.
- `python/tests/unit/test_pixi_cuda_py_demos_task.py` — new regression tests covering the `dartpy=ON` honoring, the stale-cache reset ordering (reset runs before `cmake -G Ninja`), and the `py-demos` → `build-py-dev-docking` dependency wiring.
- `CHANGELOG.md` — note the cache-reset behavior under the existing CUDA py-demos entry.

## Testing

- `pixi run pytest python/tests/unit/test_pixi_cuda_py_demos_task.py` — 3 passed.
- `pixi run ruff check python/tests/unit/test_pixi_cuda_py_demos_task.py` — clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,9 @@ py-demos` now builds a CUDA-enabled dartpy + Filament GUI and offloads the
     flags and a `DART_PY_DEMOS_GPU` env override (default `auto`: on when a CUDA
     device is present), an in-viewer GPU toggle panel, and a startup status
     line. The `config-py` build task forwards the experimental CUDA opt-in so
-    the lean compute-only `build-cuda`/`test-cuda` paths are unaffected.
+    the lean compute-only `build-cuda`/`test-cuda` paths are unaffected, and it
+    resets stale CMake compiler cache metadata before CMake can auto-rerun with
+    default options and drop the `dartpy` target.
   - Made `dart::gui` UI scaling DPI-aware: `--gui-scale` now acts as a manual
     user multiplier on top of GLFW content-scale detection, `DART_GUI_DPI_SCALE`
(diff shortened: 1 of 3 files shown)


jslee02:
@codex review

PR #2877, opened 2026-06-03
Add a SKEL file format reference to the docs

jslee02:
## Summary

SKEL is a supported, tested DART format — `dart::utils::SkelParser` (exposed as `dart::io::readWorld` / `dart.io.SkelParser`), a dartpy binding, ~70 sample files under `data/skel/`, integration + unit tests, and current examples such as `vehicle.py` — but it had **no page on the documentation site**. This adds a concise reference under **Key Topics**.

The page covers:
- Loading a `.skel` file in C++ (`dart::io::readWorld`) and Python (`dart.io.SkelParser.read_world`), including the `dart://sample/...` URI scheme.
- Document structure (`<skel>` → `<world>` → `<physics>` / `<skeleton>` → `<body>` / `<joint>`).
- The physics, body/shape, and joint elements.
- Supported geometry shapes and joint types.
- A minimal worked example.

Element, shape, and joint-type lists were verified against the current `dart/utils/skel_parser.cpp` and the `data/skel/` samples, rather than copied from the old, stale page.

## Context

This supersedes the legacy SKEL page on `dartsim.github.io` (2015-era, self-labeled "under development and buggy"), which will be redirected here as part of retiring that site in favor of readthedocs.

## Verification

`pixi run docs-build` succeeds; the page renders (`topics/skel-file-format.html`) and appears in the Key Topics toctree. **No new Sphinx warnings** (the 4 existing warnings are pre-existing dartpy autodoc signatures, unrelated to this change).

### DIFF
--- a/docs/readthedocs/topics/index.rst
+++ b/docs/readthedocs/topics/index.rst
@@ -19,3 +19,4 @@ below to keep the structure consistent.
    Numerical Methods Primer <numerical-methods>
    Simulation Stability <simulation-stability>
    Inverse Kinematics <../shared/inverse_kinematics/index>
+   SKEL File Format <skel-file-format>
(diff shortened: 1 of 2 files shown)


jslee02:
@codex review

PR #2884, opened 2026-06-04
Validate HeightmapShape scale and height field inputs (DART 6.17)

jslee02:
## Summary

`HeightmapShape::setScale` guarded its scale only with `DART_ASSERT`, which is
compiled out in release builds (`NDEBUG`). As a result, non-finite (`NaN`/`Inf`)
or non-positive scale values — and non-finite height values — flowed unchecked
into the collision backend (ODE), where they trip an internal heightfield
assertion / undefined behavior. Reported downstream via
[gazebosim/gz-physics#847](https://github.com/gazebosim/gz-physics/issues/847).

## Changes

- `HeightmapShape::setScale` now rejects non-finite or non-positive scale with a
  warning, preserving the previous valid state (mirrors the existing
  `SphereShape::setRadius` validation precedent).
- `HeightmapShape::setHeightField(const HeightField&)` now rejects empty and
  non-finite height fields. Both `setHeightField` overloads funnel through it.

## Testing

- New `tests/unit/dynamics/test_HeightmapShape.cpp` (`HeightmapShapeValidation`)
  covering valid/NaN/Inf/non-positive scale and non-finite height fields, for
  both `float` and `double`.
- Verified in a **Debug (asserts-enabled)** build. Regression suite green
  (`test_Collision` incl. `testHeightmapBox`, plus dynamics/constraint suites).

## Scope / follow-up

This closes the **non-finite** (`NaN`/`Inf`) manifestation. A finite-but-
astronomical scale (e.g. `1e308`) on a large field can still overflow to `Inf`
*inside* ODE's heightfield extent computation; addressing that needs an
ODE-boundary extent guard (separate, to be scoped). See the issue thread.

Forward-port to `main` (DART 7) is a paired dual-PR (linked separately).

### DIFF
--- a/dart/dynamics/detail/HeightmapShape-impl.hpp
+++ b/dart/dynamics/detail/HeightmapShape-impl.hpp
@@ -74,9 +74,18 @@ const std::string& HeightmapShape<S>::getStaticType()
 template <typename S>
 void HeightmapShape<S>::setScale(const Vector3& scale)
 {
-  DART_ASSERT(scale[0] > 0.0);
-  DART_ASSERT(scale[1] > 0.0);
-  DART_ASSERT(scale[2] > 0.0);
+  // Reject non-finite or non-positive scales. DART_ASSERT is compiled out in
+  // release builds (NDEBUG), so without this runtime guard a NaN/Inf or
+  // non-positive scale would flow into the collision backend (e.g. ODE) and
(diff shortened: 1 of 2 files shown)


jslee02:
Dual-PR: the `main` / DART 7 forward-port is #2886.

PR #2850, opened 2026-06-02
Backport DART 6 collision and constraint fixes to 6.17.0

jslee02:
## Summary

Backports 12 collision and constraint fixes from `main` (DART 7 development) to the **DART 6.17.0** LTS line. Part of the DART 6.16 → 6.17 maintenance roll-up that preserves bug fixes and additive improvements to the classic DART 6 API before the DART 7 clean break removes that surface from `main`.

## Motivation / Problem

`release-6.16`/`release-6.17` is the maintained compatibility line for the established DART 6 API and gz-physics. These fixes were made on `main` but are missing from the DART 6 line; each was verified to be genuinely absent from `release-6.16` and to compile against the classic DART 6 API (no DART-7-only dependencies).

## Changes / Key Changes

- Add contact helpers for body node access ([#2245](https://github.com/dartsim/dart/pull/2245))
- FCL distance: preserve minimum pair result (fix #1539) ([#2243](https://github.com/dartsim/dart/pull/2243))
- Fix plane collision handling for Bullet and FCL ([#2246](https://github.com/dartsim/dart/pull/2246))
- Fix FCL mesh contact regression and add test ([#2258](https://github.com/dartsim/dart/pull/2258))
- Fix Bullet ellipsoid rolling ([#2274](https://github.com/dartsim/dart/pull/2274))
- Handle MetaSkeleton subscriptions in CollisionGroup ([#2277](https://github.com/dartsim/dart/pull/2277))
- Stabilize FCL contact normals and collision ordering ([#2282](https://github.com/dartsim/dart/pull/2282))
- Guard collision filter against missing body nodes ([#2343](https://github.com/dartsim/dart/pull/2343))
- Handle NaN fallback for Dantzig LCP ([#2253](https://github.com/dartsim/dart/pull/2253))
- Add ASan build mode and fix memory leaks ([#2101](https://github.com/dartsim/dart/pull/2101)) _(surgical backport)_
- Disable dartpy build on Windows CI ([#2091](https://github.com/dartsim/dart/pull/2091)) _(surgical backport)_
- Fix HeightmapShape visual/collision mismatch (ODE) ([#2305](https://github.com/dartsim/dart/pull/2305)) _(surgical backport)_

All commits preserve `(cherry picked from commit …)` provenance.

## Testing

- `pixi run lint` (formatting).
- Full C++/Python build + tests via this PR's CI on `release-6.17`.
[...]

PR #2885, opened 2026-06-04
Reject non-finite mass and inertia inputs (DART 6.17)

jslee02:
## Summary

A non-finite (`NaN`/`Inf`) mass or moment of inertia propagates into the spatial
inertia tensor and poisons the Featherstone forward-dynamics recursion, aborting
asserts-enabled builds. Reported downstream via the gz-physics robustness
reports ([gz-physics#854](https://github.com/gazebosim/gz-physics/issues/854),
[#851](https://github.com/gazebosim/gz-physics/issues/851),
[#849](https://github.com/gazebosim/gz-physics/issues/849)).

Note: the zero/epsilon-mass cases from those reports already degrade gracefully
in 6.17 (via #2850). This PR closes the remaining **non-finite** sub-case.

## Changes

- `Inertia::setMass` and both `Inertia::setMoment` overloads reject non-finite
  values at the ingest boundary with a warning, preserving the previous valid
  state (mirrors the `SphereShape::setRadius` precedent). This is the
  root-cause fix — `NaN`/`Inf` never enters the spatial tensor.
- `GenericJoint::updateInvProjArtInertiaImplicitDynamic` converts its remaining
  raw finiteness `DART_ASSERT` to the warn-and-zero guard already used by its
  non-implicit sibling (`updateInvProjArtInertiaDynamic`) — defense in depth for
  other entry points (e.g. `setSpatialTensor`).

## Testing

- New regression tests `InvalidDynamicsInputs.NanInertiaDoesNotCrash` and
  `.InfMassDoesNotCrash`.
- Verified in a **Debug (asserts-enabled)** build — both abort before the fix,
  pass after. Full regression suite green (`test_Inertia`, `test_Dynamics`,
  `test_Joints`, `test_GenericJoints`, `test_ForwardKinematics`,
  `test_ConstraintSolver`, `test_Skeleton`, `test_World`, …; 17/17).

## Note

The empirical Debug run showed the `NaN`-inertia path aborts at
`BodyNode::updateBiasForce`, *earlier* than the `GenericJoint` inverse-inertia
assert — which is why the authoritative fix is ingest validation rather than
guarding individual downstream asserts.

Forward-port to `main` (DART 7) is a paired dual-PR (linked separately).

### DIFF
--- a/dart/dynamics/Inertia.cpp
+++ b/dart/dynamics/Inertia.cpp
@@ -35,6 +35,8 @@
 #include "dart/common/Console.hpp"
 #include "dart/math/Geometry.hpp"
 
+#include <cmath>
+
 namespace dart {
 namespace dynamics {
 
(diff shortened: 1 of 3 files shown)


jslee02:
Dual-PR: the `main` / DART 7 forward-port is #2887.

PR #2894, opened 2026-06-05
Clamp ODE heightfield extents to prevent collision abort on huge scale (DART 6.17)

jslee02:
## Summary

Completes the #847 heightmap hardening at the ODE collision boundary.

The shape-level guard merged in #2884 rejects non-finite / non-positive scale and non-finite heights. But a **finite-but-astronomical** scale (e.g. `1e308`, as in the gz-physics#847 repro) is *accepted* by that guard, then `(width-1)*scale`, the vertical scale, and `maxHeight*scale.z` overflow to `±Inf` inside ODE. That collapses ODE's per-cell heightfield index range and aborts via `dIASSERT((nMinX < nMaxX) && (nMinZ < nMaxZ))` in `dCollideHeightfield` — at **collide time**, not build time.

## Changes

- `OdeHeightmap` clamps the width/depth extents, the vertical scale, and the min/max height bounds handed to ODE to a finite cap (`1e12` m — far below `sqrt(DBL_MAX)`, larger than any physical terrain), with a `dtwarn`. `mGeomId` stays a valid, fully-built geom and the height-data pointer is untouched (so no dangling-storage or null-geom hazard — see rationale below).
- Adds `Collision.testHeightmapHugeScaleNoAbort` exercising both the extents (`≥3` samples) and vertical-scale (`2×2`) overflow paths.

## Why clamp (not skip/fallback)

Leaving `mGeomId` null is unsafe — `OdeCollisionObject` and `OdeCollisionGroup` dereference the geom (`dGeomSetData`, `dSpaceAdd`, …) with only `DART_ASSERT` guards (no-ops under `NDEBUG`). A flat-fallback geom would dangle, since `dGeomHeightfieldDataBuild*` is called with `copyHeightData = 0` (ODE holds the height pointer by reference). Clamping keeps every consumer well-defined and changes nothing for valid inputs.

## Testing

Verified in a **Debug (asserts-enabled)** build: `test_Collision` passes; the new test's warnings confirm the clamp fires on both overflow paths and `collide()` completes without aborting.

Relates to gazebosim/gz-physics#847. Dual-PR: `main` forward-port linked separately.

### DIFF
--- a/dart/collision/ode/detail/OdeHeightmap-impl.hpp
+++ b/dart/collision/ode/detail/OdeHeightmap-impl.hpp
@@ -36,13 +36,51 @@
 
 #include <dart/dynamics/HeightmapShape.hpp>
 
+#include <algorithm>
+#include <limits>
 #include <type_traits>
+
+#include <cmath>
 
 namespace dart {
 namespace collision {
 namespace detail {
 
 #define HF_THICKNESS 0.05
+
+//==============================================================================
+// A HeightmapShape scale/heights can be finite (so they pass the shape-level
+// validation guard) yet make the values DART hands to ODE non-finite: e.g.
+// (width-1)*scale.x() overflows to +/-Inf for a large field, or
+// maxHeight*scale.z overflows ODE's vertical AABB. ODE then derives a
+// non-finite sample spacing / AABB, collapses its per-cell index range, and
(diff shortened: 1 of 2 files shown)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #2895, opened 2026-06-05
Demote duplicate-name rename message to debug verbosity (DART 6.17)

jslee02:
## Summary

`NameManager::issueNewName` logs at message (`Msg`) level every time it
auto-renames a duplicate name. Automatic renaming is routine for legitimate use
cases — most notably loading nested models whose links share a base name, as
reported downstream in
[gazebosim/gz-physics#725](https://github.com/gazebosim/gz-physics/issues/725) —
so the message is log noise rather than a signal of a problem.

## Change

Demote the message from `dtmsg` to `dtdbg`, so it stays available at debug
verbosity for diagnostics but no longer spams normal runs. No behavior change
beyond log level.

Relates to gazebosim/gz-physics#725. Dual-PR: `main` forward-port linked separately.

### DIFF
--- a/dart/common/detail/NameManager.hpp
+++ b/dart/common/detail/NameManager.hpp
@@ -36,6 +36,7 @@
 #include "dart/common/Macros.hpp"
 
 #include <dart/common/Console.hpp>
+#include <dart/common/Logging.hpp>
 #include <dart/common/NameManager.hpp>
 
 #include <sstream>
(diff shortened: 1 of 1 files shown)


jslee02:
Dual-PR: the `main` / DART 7 forward-port is #2896.

jslee02:
@codex review

jslee02:
@codex review

PR #2903, opened 2026-06-05
Regularize rank-deficient closed-loop constraints (DART 6.18)

jslee02:
## Summary

Fixes gazebosim/gz-physics#719: a closed kinematic chain whose joints are **not aligned with the world axes** produces a rank-deficient constraint diagonal. The previous purely multiplicative Constraint Force Mixing (`vel += vel * cfm`) left those near-zero directions essentially unregularized, so the boxed LCP was singular. The Dantzig solver hit a singular pivot, zeroed the tail of the solution, and — with no secondary solver to fall back to — applied a degenerate loop-closure impulse, so the loop silently drifted open. The chain stayed finite but the closed loop was **not enforced**, which is the gz symptom.

## Fix

Add an additive CFM floor to the **holonomic loop-closure constraints** `BallJointConstraint` and `WeldJointConstraint` (the `DynamicJointConstraint` family used to close kinematic loops). `getVelocityChange()` now adds a constant floor (`vel += vel * cfm + cfm`) so rank-deficient directions get a non-zero diagonal and Dantzig converges instead of hitting a singular pivot.

The floor is deliberately scoped to those two constraints:
- It is **not** applied to the motor / limit / friction / contact constraints — it is unnecessary there and a constant diagonal floor perturbs otherwise well-conditioned solves (it shifts the servo-motor steady state past its `1e-7` test tolerance).
- The default solver already recovered closed loops via its existing Dantzig-early-termination → PGS secondary fallback; this change makes the regularization strong enough that a **Dantzig-only** configuration converges as well.

> Note: an earlier revision of this PR also changed `dSolveLCP()` to report failure on a singular pivot. That was dropped — the existing `earlyTermination` path already routes singular pivots to the secondary solver, and unconditionally returning `false` would have discarded the finite-prefix solution for no-secondary callers on the tolerated `s<=0` exit (per Codex review).

## Validation

`test_Issue719` reproduces a non-axis-aligned four-bar loop (the whole linkage is rotated ~30° about an arbitrary axis) and measures the loop-closure error over 100 steps:

| Case | Without fix | With fix |
|------|-------------|----------|
| `NonAxisAlignedClosedLoopDefaultSolver` (Dantzig+PGS) | gap `8e-10` (PGS already rescues) | gap `2e-9` |
| `NonAxisAlignedClosedLoopDantzigOnly` (no secondary) | **drifts past `1e-3` → ~`6e-3`** | gap `2e-9` |

The `DantzigOnly` case is the discriminating regression guard — it fails (loop drifts open) without this fix and passes with it.

Local regression sweep (Debug): `test_ConstraintSolver`, `test_ContactConstraint`, `test_JointForceTorque`, `test_Joints` (incl. `SERVO_MOTOR`), `test_GenericJoints`, `test_ScrewJoint`, `test_SoftDynamics`, `test_Dynamics` all pass.

## Related

Fixes gazebosim/gz-physics#719. The DART 7 (`main`) forward-port will follow as a companion PR.

### DIFF
--- a/dart/constraint/BallJointConstraint.cpp
+++ b/dart/constraint/BallJointConstraint.cpp
@@ -260,8 +260,11 @@ void BallJointConstraint::getVelocityChange(double* _vel, bool _withCfm)
   // Add small values to diagnal to keep it away from singular, similar to cfm
   // varaible in ODE
   if (_withCfm) {
+    // Additive CFM floor so rank-deficient (e.g. closed-loop) directions
+    // regularize; see gz-physics#719
     _vel[mAppliedImpulseIndex]
-        += _vel[mAppliedImpulseIndex] * mConstraintForceMixing;
+        += _vel[mAppliedImpulseIndex] * mConstraintForceMixing
+           + mConstraintForceMixing;
   }
 }
 
--- a/dart/constraint/WeldJointConstraint.cpp
+++ b/dart/constraint/WeldJointConstraint.cpp
@@ -280,8 +280,11 @@ void WeldJointConstraint::getVelocityChange(double* _vel, bool _withCfm)
   // Add small values to diagnal to keep it away from singular, similar to cfm
   // varaible in ODE
   if (_withCfm) {
+    // Additive CFM floor so rank-deficient (e.g. closed-loop) directions
+    // regularize; see gz-physics#719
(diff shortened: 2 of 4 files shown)


jslee02:
@codex review

PR #2907, opened 2026-06-05
Enable experimental simulation on Windows wheels (restore dartpy.World)

jslee02:
## Windows-experimental follow-up — restore `dartpy.World` on Windows (cycle 1: wheels)

After the WS2 Windows decouple + the WS6 World name-swap (#2904), Windows wheels
ship **without `dartpy.World`** (the DART 7 ECS World lives in the experimental
module, which WS2 forced OFF on Windows only). This re-enables it.

### Change
- Drop the Windows-only `DART_BUILD_SIMULATION_EXPERIMENTAL=OFF` override in
  `publish_dartpy.yml`. Windows wheels now inherit the default (ON), matching
  Linux/macOS, so the Windows wheel build compiles the experimental module +
  its dartpy binding and `dartpy.World` becomes available on Windows.

### Status — iterating on Windows MSVC CI
The experimental C++ module and its nanobind binding have not previously been
compiled with MSVC. This PR exercises that build on the Windows wheel lane
(editing this workflow re-runs the full wheel matrix). If MSVC surfaces
C++20/portability issues, I'll address them here. This cycle covers the wheel
path (the user-facing deliverable); re-enabling the Windows C++ **Tests** lane
(currently VS-generator + experimental-excluded) is a follow-up once the wheel
build is green.

Draft until the Windows wheel build is green.

PR #2904, opened 2026-06-05
DART 7 PLAN-041 WS6: Python facade — World name-swap (dartpy.World = ECS)

jslee02:
## DART 7 PLAN-041 WS6 — Python facade: World name-swap

Makes the ECS-backed experimental simulation the **official DART 7 dartpy API**.
This is the Python half of the PLAN-041 World name-swap (step 5), landing on top
of WS2 (#2867). It is a deliberate **clean break**: dartpy ships the DART 7 API
only, with no DART 6 Python compatibility surface.

### What changes

- **`dartpy.World` is now the ECS World.** The ECS facade (formerly
  `dartpy.simulation_experimental`) is bound into the `simulation` submodule and
  flat-promoted, so `dartpy.World`, `dartpy.RigidBody`, `dartpy.Multibody`,
  `dartpy.JointSpec`, … are the ECS types. `assert dartpy.World is dartpy.simulation.World`.
- **`dartpy.simulation_experimental` is removed** (no back-compat alias). The
  separate C++ submodule registration is dropped; the binding sources keep their
  on-disk path but register into `simulation`.
- **The classic `dart::simulation::World` survives only as render plumbing.** It
  is bound as `dartpy.gui.RenderWorld` (so the Filament viewer's
  `nb::cast<dart::simulation::WorldPtr>` keeps working), and its root frame is
  reached via `dartpy.gui.world_render_frame()`.
- **`WorldRenderBridge` is promoted into the package** as `dartpy.gui.WorldRenderBridge`
  (moved from the demos tree). It mirrors an ECS `dartpy.World` into a parallel
  classic `RenderWorld` that the viewer draws, since the ECS World has no direct
  viewer path.
- **Flat-name collisions resolved in favor of the ECS facade.** Three ECS names
  collide with the legacy `dynamics` module — `Frame`, `Joint`, `ActuatorType` —
  and the ECS versions win (`dartpy.Frame` is the DART 7 ECS Frame). The classic
  Frame is reachable only through `dartpy.gui.world_render_frame()`, never the
  flat name. This follows the maintainer directive: prefer the long-term,
  clean, modern DART 7 surface; do not preserve the DART 6 API in dartpy.
- **Consumers migrated** (demos, tests, scripts, docs): `simulation_experimental`
  → flat `dartpy.*`; classic render worlds → `dartpy.gui.RenderWorld`;
  `Frame.world()` → `dartpy.gui.world_render_frame()`. The DART 6 classic-World
  API test (`test_world.py`) was deleted.
[...]

PR #2901, opened 2026-06-05
Add AVBD rigid motor and breakable joint support

jslee02:
## Summary

- Adds opt-in AVBD rigid revolute motor rows and a public breakable fixed-joint lifecycle for experimental `World`.
- Adds user-visible py-demos for the revolute motor and breakable joint paths, grouped with the existing AVBD rigid constraint examples.
- Adds regression tests and benchmark-dashboard coverage for the new motor and fracture rows.

## Motivation / Problem

- PR #2897 made fixed-joint AVBD contact visible in py-demos. The next user-facing step is to show and validate actuated rigid constraints and breakable joints without splitting the workflow across multiple PRs.
- Experimental `World` users should be able to opt into a bounded revolute velocity actuator path and inspect a break-force release path from the Python demo catalog.

## Changes / Key Changes

- Extends AVBD rigid point-joint row assembly with angular motor rows for velocity-actuated revolute joints using symmetric effort limits.
- Persists and exposes fixed-joint break thresholds and broken state through the experimental C++ API, serializer, dartpy bindings, and stubs.
- Marks fractured AVBD point joints broken during rigid contact-stage projection.
- Adds `avbd_rigid_revolute_motor` and `avbd_rigid_breakable_joint` py-demo scenes and documents them in the demo catalog.
- Adds focused C++ and Python tests plus a benchmark-dashboard row for the AVBD revolute-motor workload.
- Updates PLAN-104 and AVBD task notes to reflect the combined motor/fracture slice.

## Before / After

| Area | Before | After |
| --- | --- | --- |
| AVBD rigid constraints | Fixed-joint rows were visible, but actuated revolute rows were not surfaced. | Velocity-actuated revolute joints can emit bounded AVBD motor rows. |
| Breakable joints | Break thresholds were not carried through the public experimental lifecycle. | Fixed joints expose break force and broken state, and AVBD fracture marks the joint broken. |
| Python demos | The catalog showed the fixed-joint contact slice only. | The catalog includes fixed-contact, revolute-motor, and breakable-joint AVBD rigid examples. |
| Benchmarks | Dashboard coverage tracked fixed-joint AVBD rows. | Dashboard coverage also labels the revolute-motor AVBD row workload. |

## Testing

- `pixi run lint`
- `pixi run build-simulation-experimental-tests`
- `ctest --test-dir build/default/cpp/Release --output-on-failure -R 'test_avbd_rigid_block|test_boxed_lcp_contact'`
- `pixi run build-py-dev`
[...]

PR #2883, opened 2026-06-04
Add experimental World step profiling

jslee02:
## Summary

- Adds an opt-in per-stage World step profile for the experimental World, available as readable text from C++ and dartpy.
- Removes the unused duplicate experimental/common profiling helper so experimental profiling has one consolidated surface.
- Updates docs, generated dartpy stubs, API boundary inventory, changelog, and C++/Python tests.

## Motivation / Problem

- DART 7 positions the experimental World as AI-native, but headless workflows had no direct text path from `World` or dartpy to understand where a step spent time.
- The existing compute execution profiler only emitted Graphviz DOT and was not reachable from `World` or dartpy, while the experimental/common profiler duplicated `dart::common::profile` and was not wired into a build.
- This gives users a low-friction, opt-in per-stage timing snapshot without changing the default step path.

## Changes / Key Changes

- Adds `compute::WorldStepProfile` and `WorldStepStageProfile` with stage name, domain, duration, total stage time, stage lookup, reset, and `toSummaryText()`.
- Adds `WorldStepPipeline::executeProfiled()` and `World` APIs `setStepProfilingEnabled()`, `isStepProfilingEnabled()`, and `getLastStepProfile()`.
- Exposes dartpy `World.step_profiling_enabled`, `World.last_step_profile`, `WorldStepProfile`, and `WorldStepStageProfile`.
- Adds `docs/onboarding/profiling.md`, updates onboarding links, stubs, API boundary inventory, and changelog.
- Deletes the dead `dart/simulation/experimental/common/profiling.{hpp,cpp}` surface and splits diagnostics-only tests into `test_diagnostics.cpp`.

## Before / After

| Area | Before | After |
| --- | --- | --- |
| World profiling | No text profiling reachable from experimental `World` or dartpy | Opt-in `World` step profile reports stages slowest-first through `toSummaryText()` / `summary()` |
| Default step path | No profiling branch existed | Profiling remains disabled by default; disabled steps use the existing `execute()` path |
| Experimental profiler surface | Dead `experimental/common` profiler duplicated `dart::common::profile` | One experimental World step profiler plus existing core `dart::common::profile` for intra-stage scopes |
| Python | No runtime or stub symbols for step profiles | `WorldStepProfile`, `WorldStepStageProfile`, `step_profiling_enabled`, and `last_step_profile` are bound and stubbed |

## Testing

- `pixi run generate-stubs`
- `pixi run lint`
- `pixi run test-all`
- `pixi run -e cuda test-all`
[...]

PR #2912, opened 2026-06-05
Fix main CI regressions surfaced post-WS6 (reduced-build guard, promotion audit, arm64 deprecation, Debug alloc test)

jslee02:
Two independent CI regressions on `main` from recent merges, both blocking shared lanes on every PR. Surfaced while validating the Windows-experimental follow-up (#2907).

## 1. Reduced-build test guard (Core Tests (Linux), Release/Debug arm64)
`test_avbd_revolute_motor_demo_drives_hinge` (added by #2901, migrated to flat `dartpy` in WS6 #2904) imports `dartpy` and calls the demo's `build()` — which constructs `dartpy.World` and reads `dartpy.ActuatorType` — **without the reduced-build guard** its siblings use. On the experimental-OFF lanes (`DART_BUILD_SIMULATION_EXPERIMENTAL_OVERRIDE=OFF`), `dartpy.World` is absent, so it failed with `AttributeError: World` instead of skipping.

**Fix**: guard with `_require_simulation_experimental_symbols("World", "ActuatorType")`, matching `test_avbd_breakable_joint_demo_marks_joint_broken`.

## 2. Strict promotion-surface audit (Lint)
The experimental World step-profiling work (#2883) added `compute/execution_profile.hpp` to the CMake install allowlist but not to the audit's `PROMOTE_FILES` set, so `check-dart7-promotion-surface --strict` failed two ways: install allowlist (33) ≠ audit promotion set (32), and `compute/world_step_profile.hpp` (+ `world.hpp` transitively) was flagged as leaking ECS storage by including the audit-"internal" `execution_profile.hpp`.

**Fix**: `execution_profile.hpp` is the per-stage timing value type embedded in `world_step_profile.hpp`'s public profile and is ECS-agnostic (std types only). Add it to `PROMOTE_FILES` → install == audit (33 == 33), no leak.

## Verification (local)
- Both AVBD demo tests pass in a full (experimental-ON) build; the guard skips in reduced builds.
- `pixi run check-dart7-promotion-surface` (strict) passes: 33 == 33, keystone cleared, 0 leaks.
- `pixi run check-lint` green.

### DIFF
--- a/python/tests/integration/test_demos_cycle.py
+++ b/python/tests/integration/test_demos_cycle.py
@@ -628,7 +628,8 @@ def test_avbd_fixed_joint_contact_demo_exercises_contact_path() -> None:
 
 def test_avbd_revolute_motor_demo_drives_hinge() -> None:
     import numpy as np
-    import dartpy as sx
+
+    sx = _require_simulation_experimental_symbols("World", "ActuatorType")
 
     from examples.demos.scenes.avbd_rigid_revolute_motor import build
 
--- a/scripts/audit_dart7_promotion_surface.py
+++ b/scripts/audit_dart7_promotion_surface.py
@@ -96,9 +96,13 @@ PROMOTE_DIRS = (
 #     stage domain without exposing concrete executor/backend details.
 #   * compute/world_step_profile.hpp is the experimental World's public
 #     text-first profiling value type, returned by World::getLastStepProfile().
+#   * compute/execution_profile.hpp holds the per-stage execution-timing value
+#     types embedded in world_step_profile.hpp's public profile, so it is public
+#     by inclusion; it is backend- and ECS-agnostic (std types only).
 PROMOTE_FILES = {
     "compute/compute_stage_metadata.hpp",
     "compute/world_step_profile.hpp",
+    "compute/execution_profile.hpp",
     "diff/rollout.hpp",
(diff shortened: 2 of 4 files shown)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
Added a third fix: `test_issue719.cpp` was the last consumer of the deprecated `math::constantsd`, breaking the macOS arm64 Release lane under clang `-Werror,-Wdeprecated-declarations`; switched to the non-deprecated `dart::math::pi`. All three regressions are pre-existing on main from recent merges (#2901, #2883, and the math-constants deprecation).

@codex review

jslee02:
Added a fourth fix: `World.IpcBakeDoesNotPrewarmRigidBodyContactQuery` (from #2899) failed only on Coverage (Debug) because it asserted exact heap-alloc-count equality between a no-geometry IPC bake and an unsupported-plane bake. Repro'd locally: the no-prewarm intent is satisfied (plane=85 allocs < no-geometry=86, deterministic) — the plane scene just allocates one fewer 24-byte buffer in Debug (component-set-dependent container growth); Release makes them identical. Switched to `EXPECT_LE(plane, baseline)` (the actual no-prewarm invariant); passes in Debug + Release.

@codex review

PR #2920, opened 2026-06-06
Automatic body deactivation (sleeping), default-on with opt-out

jslee02:
## Summary

Adds **automatic body deactivation ("sleeping")** to the DART 6 LTS line as a **default-on** feature. `simulation::DeactivationOptions::mEnabled` now defaults to `true`; callers can set it to `false` to opt out and recover behavior equivalent to the feature being absent.

A resting rigid-contact solver island is frozen together: forward dynamics, the constraint solve, and position integration are skipped until the island is disturbed. The implementation is built on DART's existing solver-island partition (`ConstrainedGroup`) with no architectural change.

## Default-on safety

- **Rigid contact islands only**: sleeping is limited to rigid contact islands. Joint limits, motors, explicit dynamic joint constraints, soft contacts, and other non-contact constraints stay awake so existing force/torque query semantics are preserved.
- **Contact correction gate**: contact islands only become sleep-eligible once penetration correction has essentially converged, avoiding premature freezing during residual contact correction.
- **Wake behavior**: contact with a moving body, applied forces, and commands wake a resting island; gravity alone does not.
- **Low active-scene overhead**: unconstrained active scenes skip disturbance/rest detection because they cannot sleep.
- **Opt-out API**: `World::setDeactivationOptions()` remains the C++ opt-out path and is now exposed through dartpy as well.

## Evidence

Local verification on this branch:

- `cmake --build build/default/cpp/Release --target test_IslandDeactivation test_JointForceTorque test_Issue1184 dartpy --parallel 5`
- `ctest --test-dir build/default/cpp/Release -R 'test_(IslandDeactivation|JointForceTorque|Issue1184)$' --output-on-failure -j 1`
- `cmake --build build/default/cpp/Release --target py_test_joint_force_torque --parallel 5` (`53 passed`)
- `cmake --build build/default/cpp/Release --target check-format --parallel 5`

Performance smoke check (`DISABLED_Benchmark*`, Release build): resting grids still speed up by roughly **1.8x-2.0x** after settling, and the active no-sleep benchmark is within measurement noise on rerun.

## CI follow-up

Also fixes the linked Windows CI configure failure: the release branch still used `windows-latest` with hardcoded `Visual Studio 17 2022`, while the runner resolved to the VS 2026 image and CMake could not find VS 2022. Windows CI now pins the VS 2026 runner and selects it with the standard `CMAKE_GENERATOR=Visual Studio 18 2026` override; the win-64 Pixi tasks no longer hard-code a Visual Studio generator, so DART 6 continues to work with VS2022 installs while allowing newer Visual Studio generators.

## Demo

`examples/sleeping` visualizes independent solver islands, dims sleeping islands, and provides controls to shoot/drop bodies to wake islands or toggle the feature off.

## Scope

Targets the **DART 6.19 minor** (base `release-6.19`). No DART 7 dual-PR; DART 7 will implement deactivation natively in its experimental compute-graph engine.


jslee02:
@codex review

PR #3034, opened 2026-06-15
ci(freebsd): provision Eigen 5.0 in the VM

jslee02:
## Problem

DART requires **Eigen >= 5.0** (`cmake/dart_find_eigen3.cmake` is a hard `FATAL_ERROR` with no FetchContent fallback), but FreeBSD ports ships Eigen 3.4. The FreeBSD CI job fails at configure time.

(The companion CMake >= 4.2.3 provisioning already landed on `main`; this PR adds the Eigen half on top of it.)

## Fix

Add `ensure_eigen()` to the VM bootstrap (`scripts/freebsd.py`), called after `pkg install`:

- Reads the installed Eigen world version from `Macros.h`; **no-op** when ports already ships >= 5.
- Otherwise fetches the release tarball and installs Eigen's headers + CMake config (header-only, so no compile).
- Version overridable via `FREEBSD_VM_EIGEN_VERSION`.

Mirrors the existing `ensure_cmake()` pattern. Self-heals once FreeBSD ports catches up to Eigen 5.

### DIFF
--- a/scripts/freebsd.py
+++ b/scripts/freebsd.py
@@ -23,6 +23,14 @@ DEFAULT_TEST_REGEX = ""
 DEFAULT_CTEST_TIMEOUT = 1200
 DEFAULT_CTEST_STOP_ON_FAILURE = True
 DEFAULT_PORTS_PATCH_DIR = "docker/freebsd/ports-patches"
+# DART's top-level cmake_minimum_required (CMakeLists.txt). FreeBSD ports still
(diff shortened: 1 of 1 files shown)


jslee02:
Superseded by #3032, which consolidates all origin/main CI fixes (CodeQL disk, Coverage OOM, GUI Smoke Clang, and this FreeBSD Eigen 5.0 provisioning) into a single PR to minimize CI and review time. The `freebsd.py` change here is included verbatim in #3032.

PR #3036, opened 2026-06-15
ci(gui-smoke): fix Clang std::expected build (libstdc++ concepts-macro gap)

jslee02:
## Problem

`DART GUI Smoke (Clang)` has been chronically red — clang 18 on `ubuntu-latest` fails to compile `std::expected` (`dart/common/result.hpp`), while the GCC variant passes.

## Root cause (confirmed by an in-CI diagnostic)

```
__cplusplus      = 202302L      # C++23 on
__cpp_concepts   = 201907L      # clang 18 reports the OLD value
__cpp_lib_expected = (undefined)
```

clang 18 uses **libstdc++ 14** (which ships `<expected>`), but libstdc++ gates `std::expected` behind `__cpp_concepts >= 202002L`. Clang 18 conservatively still reports `201907L`, so the guard fails and `std::expected` is hidden. g++ reports `202002L`, which is why the GCC variant builds.

## Fix

For the Clang matrix entry only, advertise the final C++20 concepts value:

```
-D__cpp_concepts=202002L -Wno-builtin-macro-redefined
```

(`-Wno-builtin-macro-redefined` is required because the build uses `-Werror`.) This keeps the build on **libstdc++** — ABI-matching the libstdc++ conda dependencies (sdformat/urdfdom/…) — rather than switching to `-stdlib=libc++`, which would clash at link time. GCC carries an empty flag and is unchanged.

The earlier diagnostic step that established the root cause has been removed; the green Clang job is the proof.

### DIFF
--- a/.github/workflows/ci_ubuntu.yml
+++ b/.github/workflows/ci_ubuntu.yml
@@ -555,6 +555,33 @@ jobs:
       - name: Configure environment for compiler cache
         uses: ./.github/actions/configure-compiler-cache
 
+      # Diagnostic only: the Clang variant fails to find std::expected. Print
+      # how the compiler resolves the C++ standard library inside the pixi env
+      # so the real fix can be grounded in evidence rather than guesswork.
+      # Read-only and continue-on-error so it never affects the job result.
+      - name: Diagnose Clang stdlib resolution
+        if: matrix.compiler.name == 'Clang'
+        continue-on-error: true
+        env:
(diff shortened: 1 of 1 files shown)


jslee02:
Superseded by #3032, which consolidates all origin/main CI fixes into a single PR. The Clang `std::expected` fix (`-D__cpp_concepts=202002L`) validated here — GUI Smoke (Clang) passed — is included verbatim in #3032.

PR #3096, opened 2026-06-20
fix(simulation): eliminate per-step heap allocations in boxed-LCP fallback contact assembly

jslee02:
## Problem
`World.BakedBoxedLcpFallbackContactStepsDoNotMallocOnHeap` (ctest `test_world_raw_malloc`) regressed: the boxed-LCP fallback contact path did **20 real heap allocations per step** (expected 0), violating the no-malloc guarantee hardened under WP-122/PLAN-091. The other no-malloc paths (RigidIpc, default, variational, velocity-actuated) were already clean.

## Root cause (located via `malloc` backtraces, not assumption)
All 20 allocations came from one site: **`assembleRigidBodyContactProblemInto`** (`dart/simulation/compute/rigid_body_constraint.cpp`), on the real BoxedLcp hot path (`UnifiedConstraintStage::execute → assembleProblemIntoScratch`). It unconditionally `resize(0,0)`'d the **persistent, already-primed** `scratch.rigidProblem` Eigen members at the top of the function, then rebuilt them with `Eigen::MatrixXd::Zero(size, size)` — freeing and reallocating the delassus/rhs/lo/hi/findex buffers every step, even though `UnifiedConstraintStage::prepare()` had primed them at `enterSimulationMode()` precisely for reuse. Larger "mixed fallback" systems → larger reallocations → the 20 mallocs.

## Fix (behavior-preserving — memory sourcing only)
- Drop the unconditional top-of-function `resize(0,…)` that freed the primed buffers.
- The no-system early-return resizes the members to 0 there (preserves the documented "no system" post-condition).
- The populate path uses in-place `resize(size, size)` + `setZero()`, so Eigen reuses the existing allocation when the contact shape is unchanged. The Delassus block is still fully overwritten and the Jacobian still zeroed before its `+=` accumulation — **identical numerics**.

## Before / after
- Before: 20 allocations → **FAIL**.
- After: **0 allocations** → `test_world_raw_malloc` passes.

## Validation
- Full Release `-L simulation` suite: **79/79 pass, 0 failed** (incl. golden/parity/regression: `test_world_default_step_golden`, `test_world_contact_parity`, `test_world_dart7_regression`) — no numerical/behavioral regression.
- Compiles clean in Debug and Release.

_Single-file change (`rigid_body_constraint.cpp`, +20/−8). Found while verifying the dartpy split (#3094); it was the lone pre-existing failure in `test-all`._

### DIFF
--- a/dart/simulation/compute/rigid_body_constraint.cpp
+++ b/dart/simulation/compute/rigid_body_constraint.cpp
@@ -320,33 +320,30 @@ void assembleRigidBodyContactProblemInto(
   }
 
   const std::size_t n = problem.constraints.size();
-  if (!options.populateSystem && !options.populateJacobian) {
-    // No dense system requested: release any stale buffers so the members keep
-    // their documented "no system" post-condition. resize(0, ...) is a no-op
-    // when already empty, so the common populateSystem=false callers stay
-    // allocation-free across repeated same-shape steps.
-    problem.delassus.resize(0, 0);
-    problem.rhs.resize(0);
-    problem.lo.resize(0);
-    problem.hi.resize(0);
-    problem.findex.resize(0);
-    problem.jacobian.resize(0, 0);
-    return;
-  }
-
   const auto size
       = static_cast<Eigen::Index>(n * RigidBodyContactProblem::kRowsPerContact);
+
+  // Each optional output is handled independently: cleared when its flag is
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

jslee02:
@codex review

PR #3098, opened 2026-06-20
Add strict docs/ai frontmatter type-legend guardrail

jslee02:
## Summary

- Adds one additive strict guardrail to the `docs/ai` frontmatter pilot: the
  `check-docs-policy` gate now fails when a `docs/ai/*.md` file declares a
  frontmatter `type` outside the closed legend of the eight identity types the
  real docs already use. A typo or drifted identity header now fails fast
  instead of passing silently. No existing advisory is promoted to blocking.

## Motivation / Problem

- PLAN-121's "Now" work is to extend the pilot-scoped guardrails with low-risk
  strict checks, then watch for one clean advisory cycle before deciding whether
  any advisory should become blocking. The `docs/ai` frontmatter pilot already
  enforced `type` presence and `owner` resolution, but `type` accepted any
  value, so an out-of-legend identity string would slip through unnoticed.
- The advisory→blocking flip is **deferred**: `check-docs-policy` currently emits
  17 north-star freshness advisories (cited evidence paths changed
  2026-06-17..06-19 against `evidence-last-verified=2026-06-16`) while still
  exiting 0, so the one-clean-cycle precondition is demonstrably unmet today.
  This PR therefore makes the bounded additive change only and records the
  deferral rationale in the owner doc and dashboard.

## Changes / Key Changes

- `scripts/check_docs_policy.py`: add a closed `DOCS_AI_TYPE_LEGEND` constant
  (`ai-principles`, `ai-entrypoint`, `ai-component-policy`,
  `ai-verification-policy`, `ai-workflow-map`, `ai-session-policy`,
  `ai-operating-model`, `ai-north-star`) and a strict check in
  `check_ai_doc_frontmatter` that fails when `type` is set but not in the
  legend, mirroring the closed-value pattern used for the papers catalog.
- `tests/test_check_docs_policy.py`: migrate the existing fixtures off the
  non-legend `type: ai-policy` placeholder to a legend value, and add one
  net-new meta-test (`test_docs_ai_frontmatter_rejects_out_of_legend_type`)
  covering both rejection of an out-of-legend value and a clean roster.
- `docs/plans/121-ai-docs-knowledge-graph.md`: record the slice as WP-121.8 and
  note why the advisory→blocking flip stays deferred.
- `docs/plans/dashboard.md`: update the PLAN-121 Next step.

## Before / After

| Dimension          | Before                                              | After                                                              |
| ------------------ | --------------------------------------------------- | ------------------------------------------------------------------ |
[...]

PR #3108, opened 2026-06-20
Add DART_USE_SYSTEM_FMT with a FetchContent fmt fallback for the Alt Linux repro

jslee02:
## Summary

- The `Alt Linux repro (Docker)` CI job builds against the distro's system
  packages, but Alt Linux Sisyphus currently ships a rolling `libfmt-devel`
  whose packaged CMake config references files it did not install, so
  `find_package(fmt)` aborts configuration ("the installation package was
  faulty ... `fmt-targets.cmake` ... but not all the files it references").
  This adds a `DART_USE_SYSTEM_FMT` CMake option (default **ON**, so nothing
  changes for any existing build) that, when set **OFF**, builds fmt from
  source via FetchContent. The Alt Linux repro now uses it instead of the
  broken system package.

## Motivation / Problem

- The breakage is upstream and environmental, not in DART — `find_package(fmt)`
  fails before any DART code is reached, on a non-required canary job that has
  otherwise been green on `main`. A previous targeted repair (`scripts/altlinux.py`
  re-created the missing `libfmt-c.a` archive) no longer matches the package's
  current broken reference, so chasing each new missing file is brittle.
- Building fmt from source for the repro is robust against the distro package
  churn and makes the repro use the **same fmt version DART builds against
  everywhere else** (pixi pins `fmt >=12.1.0,<13`), rather than the distro's
  arbitrary rolling version.

## Changes / Key Changes

- `CMakeLists.txt`: add `dart_option(DART_USE_SYSTEM_FMT ... ON ...)`, mirroring
  the existing `DART_USE_SYSTEM_ODE`/`DART_USE_SYSTEM_BULLET` polarity (default
  ON; set OFF only when the system package is unavailable or broken).
- `cmake/dart_find_fmt.cmake`: when `DART_USE_SYSTEM_FMT` is OFF, fetch fmt
  `12.1.0` via FetchContent (mirroring `dart_find_spdlog.cmake` /
  `dart_find_entt.cmake`). `FMT_INSTALL` stays ON so the fetched fmt targets
  belong to an export set — DART links `fmt::fmt-header-only` PUBLIC and
  validates `install(EXPORT)` at generate time, which fails otherwise.
- `scripts/altlinux.py`: drop the broken `libfmt-devel` package, pass
  `-DDART_USE_SYSTEM_FMT=OFF`, and remove the now-obsolete fmt-archive repair
  hack.
- `CHANGELOG.md`: note the new option under Build, Packaging, and Developer
  Tooling.

## Before / After

| Dimension                | Before                                                        | After                                                                 |
| ------------------------ | ------------------------------------------------------------- | --------------------------------------------------------------------- |
| Alt Linux repro config   | `find_package(fmt)` aborts on the broken rolling `libfmt-devel` | Builds fmt 12.1.0 from source; configure + generate succeed           |
| Existing builds          | `find_package(fmt)` (system/conda/vcpkg/brew)                 | Unchanged — `DART_USE_SYSTEM_FMT` defaults ON                          |
| fmt version in the repro | distro's arbitrary rolling fmt                                | 12.1.0 — same as DART's canonical build (pixi `>=12.1.0,<13`)          |

## Testing

Reproduced in the actual `altlinux/base:sisyphus` Docker container
(`pixi run altlinux-test`):

- **Before** this change: `find_package(fmt)` → `Configuring incomplete, errors
  occurred!` (broken `libfmt-devel` export).
- **After**: `DART_USE_SYSTEM_FMT : OFF`, fmt 12.1.0 fetched via FetchContent
  (`-- {fmt} version: 12.1.0`, `Looking for libfmt - version 12.1.0 found`),
  `-- Configuring done`, `-- Generating done`, no export-set error. The broken
  `/usr/lib64/cmake/fmt` is no longer installed (only the harmless `libfmt12`
  runtime is pulled transitively). Full `cmake --build --target tests && ctest`
  in the same container builds and links DART against the fetched fmt 12.1.0 and
  passes end-to-end: **100% tests passed, 0 failed out of 189**.
- `pixi run lint` passes.
- The `Alt Linux repro (Docker)` CI job on this PR exercises the real
  environment end-to-end (non-required, so it does not gate merge).

## Breaking Changes

- [x] None — `DART_USE_SYSTEM_FMT` defaults ON, preserving current behavior.

- Follow-up to the non-required Alt Linux canary failure first observed on
  #3098's CI. None / no backport (DART 7 `main` only).


jslee02:
@codex review

PR #3111, opened 2026-06-20
Constrain unsafe parallel island solves

jslee02:
## Summary

- Keep opt-in parallel constraint-island solving on the serial path when island independence is uncertain.
- Preserve the current DART 6.20 public API and the #3086 `setNumSimulationThreads` path while adding safety gates for custom/manual/shared-state cases.

## Motivation / Problem

- PR #3071 is related, but it predates the current #3086 thread-pool API and now conflicts heavily with `release-6.20`.
- The useful part to carry forward is the correctness constraint: independent group solves should only run in parallel when the solver and constraints are known to be independent.
- This PR is a fidelity/safety follow-up, not an RTF speedup claim. It may intentionally reduce opt-in parallelism for unsafe configurations.

## Changes / Key Changes

- Require exact built-in Dantzig or non-randomized built-in PGS boxed LCP solvers before parallel group solving is allowed.
- Force serial group solving when manual constraints or custom `ContactConstraint` subclasses are present.
- Force serial group solving when groups share non-reactive bodies or skeletons through contact, soft contact, dynamic joint, joint, mimic motor, or Coulomb friction constraints.
- Add integration coverage for safe independent groups, randomized PGS, derived/custom LCP solvers, manual constraints, custom contact constraints, and shared non-reactive dependencies.
- Record the high-level stack decision in `docs/dev_tasks/issue_3056_dart6_performance/README.md`.

## Before / After

| Dimension | Before | After |
| --- | --- | --- |
| Public API | `setNumSimulationThreads` opt-in path from #3086 | Unchanged |
| Parallel eligibility | Could run parallel group solves for some custom/manual/shared-state cases | Only runs for exact built-in solver combinations and groups without known shared non-reactive dependencies |
| Performance claim | None for this safety follow-up | None; correctness gate before further parallel speed work |

## Testing

- `pixi run lint`
- `pixi run build`
- `pixi run test` (100/100 tests passed on current `origin/release-6.20` after merging #3105)
- `cmake --build build/default/cpp/Release --target test_ConstraintSolver test_World --parallel 5`
- `ctest --test-dir build/default/cpp/Release -R '^(test_ConstraintSolver|test_World)$' --output-on-failure -j 1`
- `git diff --check origin/release-6.20...HEAD`

## Breaking Changes

- [x] None

## Related Issues / PRs (backports)

- Related to #3056, #3071, and #3086.
- This targets `release-6.20`; a `main` follow-up should be considered if the same unsafe parallel eligibility exists there.

---

[...]

PR #3089, opened 2026-06-19
Add contact benchmark example

jslee02:
## Summary

- Add `contact_benchmark`, a baseline-compatible performance and GUI verifier for the DART 6 primitive-heavy contact scene.

## Motivation / Problem

- Part of #3056. The optimization stack needs a reusable PR0 benchmark before changing collision, contact, or sleeping behavior.
- This PR intentionally avoids later optimization APIs so it measures the current `release-6.20` baseline with the existing DART 6 physics pipeline.

## Changes / Key Changes

- Add a generated 3-lane primitive scene with spheres, boxes, and cylinders dropped from configurable height onto a finite ground box.
- Add CLI controls for collision detector, sleep thresholds, generated object count, GUI target RTF, reset/drop state, final-state scene dump, and profiling.
- Add final-state digest/hash/sums so benchmark loops cannot be optimized away.
- Add OSG GUI HUD for RTF, simulation time, contacts, candidate/resting counts, and color legend: awake red, candidate yellow, resting blue, static gray.
- Make the example usable with no arguments: it opens a paused default generated scene, with `--gui-start` available for immediate playback.

## Baseline Performance

Measured on PR head `85284a30273` with `build/default/cpp/Release` on a 13th Gen Intel(R) Core(TM) i7-13800H host. These are PR0 baseline numbers for later optimization PRs; this PR is not claiming a speedup.

Headline workload is the original #3056 report: `gz-sim` [`3k_shapes.sdf`](https://github.com/gazebosim/gz-sim/blob/main/examples/worlds/3k_shapes.sdf), 10,000 steps, 0.001 s timestep, 3,004 skeletons / 18,024 DOFs. The SDF emits the same missing-`<inertial>` warning as the issue report.

Command base:

```bash
curl -L https://raw.githubusercontent.com/gazebosim/gz-sim/main/examples/worlds/3k_shapes.sdf -o /tmp/3k_shapes.sdf
./build/default/cpp/Release/bin/contact_benchmark /tmp/3k_shapes.sdf 10000 --checkpoint 1000
```

| Original #3056 SDF case | RTF | Wall time | Avg step | Contacts | Resting | Frame check | Final state |
| --- | ---: | ---: | ---: | ---: | ---: | --- | --- |
| Default DART 6 path, deactivation enabled | `0.020227` | `494.389 s` | `49.439 ms` | `1000` cap hit | `1 / 3003` | `10000 / 10000` | finite, `0xc88d715069dc93ed` |
[...]

PR #3086, opened 2026-06-19
Enable resting-world deactivation by default

jslee02:
## Summary

- Enable automatic resting-world deactivation by default for DART 6.20 so users get the large resting-scene speedup by upgrading.
- Keep the legacy always-active path available with `DeactivationOptions::mEnabled = false` for downstream integrations that need it.

## Motivation / Problem

- Part of #3056, now based directly on `release-6.20` after #3085 and #3088 merged.
- The target primitive-heavy scene spends most time rechecking and resolving independent objects after they have genuinely settled.
- The useful default performance win is to skip collision, constraint, dynamics, and integration work only after every mobile skeleton in the relevant island is resting and wake invalidation remains correct.
- A direct single-body LCP shortcut was measured but is disabled here because it changed the explicit `--disable-deactivation` baseline. This PR keeps default speedup tied to resting-world deactivation, not to a solver shortcut without fidelity proof.

## Changes / Key Changes

- Make `DeactivationOptions::mEnabled` default to `true` and keep explicit opt-out support.
- Tighten sleep eligibility contact-depth gating to `1e-5` so default-on sleeping does not freeze visibly under-settled contacts.
- Add default-on fidelity regressions comparing the sleeping path against the always-active path and covering sleep transition pose/velocity, unconverged contacts, and multi-body settling.
- Preserve the final sleep-transition pose while still refreshing solved contact-force caches, and zero body velocity caches when entering rest.
- Clear stale sleep candidacy when an active contact group cannot physically rest.
- Allow independent quiet contact islands to sleep while unrelated bodies remain awake or moving; same-island atomicity remains enforced by solver group candidacy.
- Key the all-resting snapshot on the solver collision-filter pointer and built-in BodyNode-pair blacklist revision so solver-filter edits wake sleepers while visual-only updates remain cached no-ops.
- Track solver collision-detector identity, collision-group identity, and collision-group content revisions in resting snapshots so detector swaps and direct group membership edits cannot reuse a stale no-contact fast path.
- Track collision-option scalar settings in resting snapshots so contact-generation option edits wake sleepers instead of reusing stale fast paths.
- Track between-step resting-world contact state so support removal, collidability/filter edits, and other contact-relevant edits wake resting bodies even before the first no-contact fast-path snapshot is built.
- Track per-world skeleton structure and kinematic revisions before the first no-contact fast-path snapshot exists, so support pose edits made immediately after sleep cannot preserve stale resting state.
[...]

PR #3085, opened 2026-06-19
Add primitive plane and contact-cap collision support

jslee02:
## Summary

- Add primitive plane support and per-pair contact-cap plumbing across DART 6 collision backends.
- Keep the per-pair contact cap explicit, not default-on: it improves the original #3056 SDF workload RTF, but changes contact/resting behavior and final state.
- Fix FCL shared-object refresh bookkeeping and keep `CollisionResult` body/frame caches independent of collision-object lifetime.

## Motivation / Problem

Part of #3056 and now based directly on `release-6.20` after #3089 landed. PR0 added the reproducible `contact_benchmark` workload and final-state checks. This PR adds collision-layer building blocks needed for performance investigation while preserving DART 6/gz-physics compatibility by default.

The key correctness constraint is that speedups must not silently change contact generation, final state, or sleep/wake behavior. The per-pair cap is therefore available for explicit measurement and targeted follow-up work, but it is not defaulted on in this PR.

## Changes / Key Changes

- Add `CollisionOption::maxNumContactsPerPair` and backend support for DART, Bullet, FCL, and ODE collision detectors.
- Add primitive plane handling for FCL instead of treating infinite planes as finite mesh-like fallback geometry.
- Add SDF parser options/tests for plane-shape loading.
- Extend `contact_benchmark` with `--max-contacts-per-pair` so PR1 comparisons can be run against the PR0 workload.
- Preserve FCL backend object identity across shape refresh and remove objects based on each group's local registration state.
- Restore eager `CollisionResult` body/frame cache population so retained results do not dereference collision objects after producer refresh/destruction.
- Add regression coverage for FCL halfspace distance, shared-object refresh, and collision-result cache lifetime.

## Before / After

Measured on benchmark-relevant code head `24dff43d9f7` with `build/default/cpp/Release` on a 13th Gen Intel(R) Core(TM) i7-13800H host. Current PR head `aa36ebb4932` only adds follow-up planning notes and does not change benchmark behavior. The workload is the original #3056 report: `gz-sim` `3k_shapes.sdf`, 10,000 steps, 0.001 s timestep, 3,004 skeletons / 18,024 DOFs.

Command base:

```bash
curl -L https://raw.githubusercontent.com/gazebosim/gz-sim/main/examples/worlds/3k_shapes.sdf -o /tmp/3k_shapes.sdf
./build/default/cpp/Release/bin/contact_benchmark /tmp/3k_shapes.sdf 10000 --checkpoint 1000
```

[...]

PR #3112, opened 2026-06-21
Accelerate settled imported worlds

jslee02:
## Summary

- Accelerate the exact #3056 large imported-world case by letting zero-velocity mobile bodies that begin in shallow support contacts reach the all-resting fast path promptly.
- Keep the DART 6 / gz-physics API surface compatible; the benchmark uses Bullet only as the collision detector while DART remains the dynamics, constraint, deactivation, and LCP pipeline.
- Add focused correctness coverage and downstream gz-sim validation for the optimized path.

## Motivation / Problem

- Issue #3056 reports that the DART 6 path is far below realtime on the gz-sim 3k primitive-shape scene where the objects are already settled on the ground.
- Earlier PRs added the baseline benchmark and explored contact-count caps, but contact reduction changed final state on the original workload and cannot be defaulted as a fidelity-preserving speedup.
- This PR targets the higher-leverage case from the profiler: imported worlds that start at rest in shallow support contact should not keep paying full collision and constraint cost for thousands of no-op steps.

## Changes / Key Changes

- Add initial dwell credit for large imported worlds whose mobile skeletons start at zero velocity in shallow inactive support contacts; deep initial penetration, unsupported mobile contacts, and wall contacts are excluded, and the optimized path still requires a solved impulse before sleeping.
- Tighten and speed up the all-resting readiness check by reusing unchanged snapshot guards and scanning only generalized velocities when the cached all-resting state is still valid.
- Allow deactivation-active independent constraint groups to solve in parallel only under guarded conditions: built-in deterministic boxed LCP solvers, no manual constraints, no custom contact constraints, and no unsafe shared non-reactive dependencies.
- Track shared fixed supports before exempting pure single-reactive contact groups so mixed groups sharing the same fixed dependency stay on the serial solve path.
- Move parallel solve scratch storage onto the solver object instead of static thread-local vectors.
- Extend `contact_benchmark` with `--world-threads` and `--sdf-plane-shapes`, and pass the requested collision detector into SDF parsing so the exact issue scene can use primitive planes with Bullet collision detection.
- Fix `scripts/run_gz_sim_task.sh` so the local gz-sim validation gate finds source-built gz-physics engine plugins installed under `lib64`.

## Before / After

Measured on the original #3056 `3k_shapes.sdf` workload: 3003 mobile sphere/box/cylinder bodies plus a static ground plane, 3000 benchmark steps, 0.001 s timestep. The current optimized command was run through the project example task:

```bash
pixi run ex contact_benchmark -- /tmp/3k_shapes.sdf --steps 3000 --warmup 0 --checkpoint 0 --quiet --collision bullet --sdf-plane-shapes --world-threads 16 --max-contacts 12000
```

| Case | RTF | Wall time | Contacts / pairs | Resting state | Final state |
| --- | ---: | ---: | ---: | ---: | --- |
| PR0 baseline (#3089), default DART 6 path on the original report workload | `0.020227` | `494.389 s` for 10000 steps | global `1000` cap hit | `1 / 3003` resting | finite, `0xc88d715069dc93ed` |
| Same exact-scene command with this PR but `--disable-deactivation` | `0.0559-0.0605` | about `49.6-53.6 s` for 3000 steps | `7005 / 3003` final | deactivation off | finite, `0x8aa3ab9f1fe495c6` |
| This PR, default deactivation enabled | `1.51609` | `1.97877 s` for 3000 steps | `0 / 0` final after sleep | `3003 / 3003` resting | finite, `0x2375f1927218cd43` |

Final-state comparison against the always-active path on the same scene showed max position delta about `3.4e-6 m` and mean position delta about `1.8e-6 m`; this is the key fidelity check behind enabling the optimization by default.

The current profiler shape is three full collision/constraint solves followed by 2997 all-resting cached steps. The remaining hotspot is the all-resting readiness scan, not full LCP solving.

## Testing

- `pixi run lint`
- `pixi run cmake --build build/default/cpp/Release --target test_IslandDeactivation test_ConstraintSolver --parallel $(python3 scripts/parallel_jobs.py)`
- `pixi run ./build/default/cpp/Release/tests/integration/test_IslandDeactivation`
  - Passed: 52 / 52.
- `pixi run ./build/default/cpp/Release/tests/integration/test_ConstraintSolver`
  - Passed: 19 / 19.
- `pixi run config-coverage`
- `pixi run cmake --build build/default/cpp/Debug --target test_IslandDeactivation --parallel $(python3 scripts/parallel_jobs.py)`
- `pixi run ./build/default/cpp/Debug/tests/integration/test_IslandDeactivation '--gtest_filter=IslandDeactivation.WakeOnCollisionOptionChange*'`
  - Passed: 2 / 2; verifies the coverage/assertion failure path from hosted CI no longer aborts.
- Exact #3056 benchmark command listed above
  - RTF `1.51609`; frame/time advanced `3000 / 3000`; final state finite; final hash `0x2375f1927218cd43`.
- `pixi run --dry-run ex contact_benchmark -- /tmp/3k_shapes.sdf --steps 3000 --warmup 0 --checkpoint 0 --quiet --collision bullet --sdf-plane-shapes --world-threads 16 --max-contacts 12000`
  - Confirmed the project task builds the `contact_benchmark` example target and passes the SDF/options through to the executable.
- `pixi run -e gazebo test-gz-sim`
  - Passed `INTEGRATION_entity_system` against the source-built DART plugin.

## Breaking Changes

- [x] None

- Part of #3056.
- Builds on the merged #3089 baseline benchmark and the follow-up #3085, #3086, and #3111 performance/safety work.
- Targets `release-6.20`; a main/DART 7 comparison can be handled separately because DART 7 has a different collision/backend architecture.

---

#### Checklist

- [x] Milestone set: DART 6.20.0
- [x] N/A - no new methods or classes to document
- [x] N/A - no dartpy API changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,13 @@
     groups that share non-reactive bodies or skeletons across islands:
     [#3111](https://github.com/dartsim/dart/pull/3111)
 
+  * Accelerate large imported worlds that begin with zero-velocity bodies on
+    shallow support contacts by allowing the initial contact pass to consume the
+    configured quiet dwell before the normal final solved impulse freezes the
+    island, improving the exact 3003-body issue scene while preserving
+    micrometer-scale agreement with the always-active path:
+    [#3056](https://github.com/dartsim/dart/issues/3056)
+
 ### [DART 6.19.2 (2026-06-19)](https://github.com/dartsim/dart/milestone/100?closed=1)
 
 DART 6.19.2 is a patch release on the DART 6 LTS line. It keeps automatic
--- a/dart/constraint/ConstraintSolver.cpp
+++ b/dart/constraint/ConstraintSolver.cpp
@@ -1456,151 +1456,160 @@ void ConstraintSolver::solveConstrainedGroups()
 {
   DART_PROFILE_SCOPED;
 
-  if (!mDeactivationActive) {
-    auto solveGroupAt = [&](std::size_t i) {
-      solveConstrainedGroup(mConstrainedGroups[i]);
-    };
+  auto solveGroupAt = [&](std::size_t i) {
+    solveConstrainedGroup(mConstrainedGroups[i]);
+  };
 
-    auto solveGroupsSerial = [&]() {
-      for (std::size_t i = 0; i < mConstrainedGroups.size(); ++i)
-        solveGroupAt(i);
-    };
+  auto solveGroupsSerial = [&](const auto& solve) {
+    for (std::size_t i = 0; i < mConstrainedGroups.size(); ++i)
+      solve(i);
+  };
 
-    auto hasCustomContactConstraint = [&]() {
-      for (const auto& group : mConstrainedGroups) {
-        for (std::size_t i = 0; i < group.getNumConstraints(); ++i) {
-          const auto* constraint = group.getConstraint(i).get();
-          const auto* contact
-              = dynamic_cast<const ContactConstraint*>(constraint);
-          if (contact != nullptr
-              && !isExactDynamicType<ContactConstraint>(contact))
+  auto hasCustomContactConstraint = [&]() {
+    for (const auto& group : mConstrainedGroups) {
+      for (std::size_t i = 0; i < group.getNumConstraints(); ++i) {
+        const auto* constraint = group.getConstraint(i).get();
+        const auto* contact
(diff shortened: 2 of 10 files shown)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3189, opened 2026-06-25
docs: capture DART 6.20 backport lessons (retrospect)

jslee02:
Retrospect on the completed DART 6.20 backport effort: promote its durable lessons to their owner docs on `main` (routing per `docs/ai/components.md`; `/dart-retrospect` workflow). Onboarding/AI docs kept lean — each lesson extends an existing section, no new files, no ephemeral PR/commit details.

- **`docs/ai/verification.md`** — run the **full** `pixi run check-lint` aggregate (CI runs it whole in `ci_lint.yml`) for lint-tooling/formatting/many-file changes, not just the sub-check you edited.
- **`docs/onboarding/ci-cd.md`** (platform gotchas the default Linux/gcc build misses) — clang `-Werror` on macOS-arm64/FreeBSD flags `-Wdeprecated-declarations` (wrap deliberate deprecated-API use in `DART_SUPPRESS_DEPRECATED_*`) and `-Wpotentially-evaluated-expression` (raw-pointer `typeid`); MSVC non-zero-init heap → `std::bad_alloc` on Windows only; FreeBSD `docker/freebsd/ports-patches/` apply silently-skip when a patched `CMakeLists.txt` is reformatted.
- **`docs/ai/orchestration.md`** — sibling-lane file-collision avoidance + temporary required-gate carve-outs with a removal condition (aligned with the `remove_by`/`tracking` convention in `api-boundaries.md`).

Verified the AI-doc gates: `check-lint-spell`, `check-lint-md`, `check-docs-policy`, `check-ai-commands` all pass.

@codex review

PR #3195, opened 2026-06-26
tests: raise Coverage (Debug) timeout for world allocation shards (fix flaky main CI)

jslee02:
Fixes the flaky `Coverage (Debug)` failure on `main`.

**Symptom:** `Coverage (Debug)` failed on `main` (currently red on `dcb5c688260`/#3189) — CTest exit 8, `test_world_raw_malloc` **Timeout at 3600.13s**.

**Root cause (not a code regression):** `test_world_raw_malloc` is one of five sharded world allocation gates that already get a coverage-specific `TIMEOUT 3600` (the gcov-instrumented build is much slower). It ran ~0.13s over the limit — i.e. it sits right at the boundary and load-flakes. The same `Coverage (Debug)` job also failed on the preceding commit **#3127** (an `actions/checkout` version bump) and passed on the two before that; neither that bump nor the docs-only #3189 can affect test runtime. So this is a flaky coverage timeout, surfaced by runner load, not anything those commits introduced.

**Fix:** raise the coverage-only timeout for the world allocation shards from 3600s → **5400s** (50% headroom). Coverage-only; no effect on normal CI timing or runtime behavior.

Verified: `check-lint-cmake` (gersemi) and codespell clean. The PR's own `Coverage (Debug)` run exercises the new headroom.

@codex review

### DIFF
--- a/tests/unit/simulation/CMakeLists.txt
+++ b/tests/unit/simulation/CMakeLists.txt
@@ -339,7 +339,10 @@ if(DART_SIMULATION_BUILD_TESTS)
       test_world_rest
     )
       if(TEST ${_dart_world_shard})
-        set_tests_properties(${_dart_world_shard} PROPERTIES TIMEOUT 3600)
+        # The gcov-instrumented build runs these allocation-gate corpora close to
+        # the 3600s ctest limit; test_world_raw_malloc has timed out at ~3600.1s on
+        # loaded Coverage (Debug) runners. Give the shards extra coverage headroom.
+        set_tests_properties(${_dart_world_shard} PROPERTIES TIMEOUT 5400)
       endif()
     endforeach()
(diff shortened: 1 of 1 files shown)


jslee02:
Closing in favor of #3196, which fixes this at the root instead of raising the timeout.

The timeout bump here was a band-aid. The real cause is that `BakedBasicDeformableRowsDoNotMallocOnHeap` bundled ~13 deformable scenes into one atomic gtest test, one of which (a 17×17 matrix-free self-contact production grid) runs ~3429s under the Debug+gcov raw-malloc interposer and alone consumed the shard budget. #3196 splits the monolith into focused gates and right-sizes that one scene to 8×8 under `DART_CODECOV` (the property is grid-size independent; full 17×17 stays in normal CI), taking the shard from ~3600s to 358s locally.

PR #3196, opened 2026-06-26
tests: right-size the deformable raw-malloc coverage gate (fix flaky Coverage (Debug) timeout)

jslee02:
Fixes the flaky `Coverage (Debug)` failure on `main` (`test_world_raw_malloc` **Timeout at 3600s**) at its root.

## Root cause

`test_world_raw_malloc` runs every `World.*DoNotMallocOnHeap` gate. After excluding the rigid-IPC gates that already `GTEST_SKIP` under coverage, one test dominated: `BakedBasicDeformableRowsDoNotMallocOnHeap` was a single monolithic gate bundling **12 deformable sub-scenes**, one of which — a **17×17 matrix-free self-contact "production" grid** — is by far the most expensive scene in the shard under the Debug + gcov build, so its four post-bake steps dominated the runtime. A gtest `TEST` is atomic to ctest, so neither a bigger timeout nor ctest-filter splitting could fix it.

## Fix (keeps full-size gcov coverage)

- **Split** the monolith into focused one-scene gates, matching the one-scene-per-gate convention used by every other deformable allocation gate.
- Add an `int postBakeSteps = 4` parameter to the raw-malloc gate helpers and run the matrix-free production-grid gate for a **single post-bake step under `DART_CODECOV`** (the full four-step assertion stays in Release/Debug CI). The grid **keeps its full 17×17 size**, so the Coverage job still exercises the matrix-free self-contact solver at production scale — only the redundant extra steps are dropped.

> Updated per Codex review: the first revision shrank the grid to 8×8 under coverage, but Codex correctly noted the full 17×17 scene would then run **nowhere** under the Coverage job (the sibling no-growth/global-heap gates `GTEST_SKIP`, and the `…IsActive` correctness test compiles out, under `DART_CODECOV`). Reducing the step count instead keeps the full production grid and its gcov line coverage; only this gate runs it under coverage.

## Verification (local, real `-O0` + gcov build)

| | before | after |
|---|---|---|
| `World.*DoNotMallocOnHeap` shard | ~570s | **274s** |
| 17×17 production-grid gate | ~416s (4 steps) | **104s** (1 step, full size) |
| result | timeout on CI | **17 passed, 5 coverage-skipped, 0 failed** |

Release suite for these gates: **22 passed** at the full 17×17 size. `pixi run check-lint` passes. (Local desktop is several× faster than the CI runner for this gcov workload, which is why the same shard hit the 3600s limit on CI; the PR’s own `Coverage (Debug)` job confirms the absolute.) New gates match `World.*DoNotMallocOnHeap`, so they flow into the existing shard automatically — no CMakeLists change.

@codex review

### DIFF
--- a/tests/unit/simulation/world/test_world.cpp
+++ b/tests/unit/simulation/world/test_world.cpp
@@ -1114,7 +1114,8 @@ template <typename ConfigureScene>
 HeapAllocationSnapshot countRawHeapAllocationsDuringFirstPostBakeSteps(
     std::string_view scene,
     ConfigureScene&& configureScene,
-    bool requireInitialContact = false)
+    bool requireInitialContact = false,
+    int postBakeSteps = 4)
 {
   namespace sx = dart::simulation;
 
@@ -1128,7 +1129,7 @@ HeapAllocationSnapshot countRawHeapAllocationsDuringFirstPostBakeSteps(
   world.enterSimulationMode();
 
   ScopedRawHeapAllocationCounter rawCounter;
-  for (int i = 0; i < 4; ++i) {
+  for (int i = 0; i < postBakeSteps; ++i) {
     world.step();
   }
   rawCounter.stop();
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

Issue #2668, opened 2026-05-19
Bug Report: Null Pointer Dereference in ImGui::GetCurrentWindow via ColorPicker3 Due to Missing UI Context Validation in Test/GUI Boundary


ShilngYu:
> For questions, use [Discussions](https://github.com/dartsim/dart/discussions) instead.
> For platform-specific build/test failures, use the [Build/Test Failure template](https://github.com/dartsim/dart/issues/new?template=build-test-failure.yml).

**Before submitting:**

- [x] Checked [documentation](https://dart.readthedocs.io/) and existing issues

**Environment:**

- DART version:  6.16.4 and later
- OS: 22.04 LTS
- Installation: Source build
- Compiler (if source): Clang 17.0.0

> **Note:** Bug fixes are only accepted for `main` and `release-6.16`. If using 6.15 or older, please upgrade to 6.16.x first and verify the issue still exists.

**Expected vs Current Behavior:**
**Expected :**
When the visualization framework or automated GUI test suites (such as ColorPickerOptionsPopupTest) interact with the embedded third-party UI widgets (dart/external/imgui), the DART-GUI interaction boundary should perform an implicit state validation. If no active ImGuiContext or window stack exists, the rendering call should be safely bypassed or standard logs should emit an early warning, preventing unisolated mutation inputs or headless fuzzers from abruptly terminating the test execution.
**Current Behavior:**
The testing harness rawly invokes the immediate-mode UI component ImGui::ColorPicker3 outside of any active window lifecycle. Inside imgui_widgets.cpp, this component directly queries the active drawing canvas via ImGui::GetCurrentWindow(). Since ImGui::Begin() was never invoked to initialize the internal window stack, GetCurrentWindow() returns a nullptr. Lacking any safety validation at this integration layer, the routine proceeds to read from this zero-page memory address, triggering an unrecoverable segmentation fault (AddressSanitizer:DEADLYSIGNAL / SIGSEGV) and forcing an ungraceful crash (Exit Code 1).

**Steps to Reproduce:**
1.Build DART from source with AddressSanitizer enabled (-fsanitize=address).

2.Run the isolated fuzzing harness or execution target that invokes the standalone ColorPickerOptionsPopup component in a headless or bare environment.

3.Observe the immediate DEADLYSIGNAL crash intercepted by ASan during the GetCurrentWindow() lookup.

**Code to Reproduce:**

```cpp
#include <dart/dart.hpp>
#include <imgui//imgui.h>
[...]

PR #3198, opened 2026-06-26
Relocate bundled ImGui to dart/gui/imgui and bump to v1.92.8

jslee02:
## Summary

After the `convhull_3d`/`ikfast`/`lodepng`/`odelcpsolver` removals, `dart/external`
held only the bundled ImGui build driver. This relocates it and refreshes the
bundled ImGui:

- **Relocate** the `external-imgui` driver + its DART compatibility patch from
  `dart/external/imgui` to `dart/gui/imgui` (configured from `dart/gui` before the
  `osg` subdirectory), and **delete the now-empty `dart/external` source directory**.
- **Bump** the bundled FetchContent ImGui `v1.91.9 → v1.92.8` and **rebase** the DART
  compatibility patch onto it.
- Source builds keep `DART_USE_SYSTEM_IMGUI` defaulting **OFF** (self-contained
  FetchContent — building from source needs no system ImGui, which is not packaged on
  every platform); Pixi and the conda-forge recipe opt into the system package via
  `-DDART_USE_SYSTEM_IMGUI=ON`.

## Why the patch can't just be dropped

All four DART fixes are still **absent upstream** in v1.92.8, so the patch is rebased,
not removed:

| Fix | Where |
| --- | --- |
| issue #2516 — null/empty font input guards (×3) | `imgui_draw.cpp` |
| issue #2668 — `ColorPicker4` missing-context guard | `imgui_widgets.cpp` |
| hidden-visibility `ImFormatString`/`V` | `imgui_internal.h` |
| OpenGL2 client-attrib save/restore | `backends/imgui_impl_opengl2.cpp` |

## Compatibility

The public surface is unchanged: the `external-imgui` component, the
`dart-external-imgui` target, and the installed `include/dart/external/imgui`
compatibility include path all stay. All DART code routes through the single
`dart/gui/osg/IncludeImGui.hpp` shim, so **no consuming source needed editing**.

(The relocated `CMakeLists.txt` also leaves the `lint_cmake.py` `external` skip list, so
it is now gersemi-formatted — hence the larger reformat diff on that file.)

## Verification (Linux)

- **FetchContent path** (`-DDART_USE_SYSTEM_IMGUI=OFF`): configure fetches v1.92.8 and
  applies the rebased patch cleanly; `dart-external-imgui` and `dart-gui-osg` build.
- **System path** (`-DDART_USE_SYSTEM_IMGUI=ON`): configure + `dart-external-imgui` build.
- **Regression tests** `test_Issue2516`, `test_Issue2668`, `test_Issue2671`: **3/3 pass**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,16 @@
     an explicit system ImGui opt-in path:
     [#3081](https://github.com/dartsim/dart/pull/3081)
 
+  * Relocate the bundled `external-imgui` build from `dart/external/imgui` to
+    `dart/gui/imgui` and remove the now-empty `dart/external` source directory,
+    update the bundled FetchContent ImGui to `v1.92.8`, and rebase the DART
+    compatibility patch onto it. Source builds default to the FetchContent copy
+    (`DART_USE_SYSTEM_IMGUI=OFF`) so building from source needs no system ImGui
+    package; Pixi and the conda-forge recipe opt into the system ImGui. The
+    `external-imgui` component and the installed `include/dart/external/imgui`
+    compatibility include path are unchanged:
+    [#3198](https://github.com/dartsim/dart/pull/3198)
+
   * Update the math user-defined literal declarations to the C++23 spelling
     accepted by newer AppleClang warning-as-error builds.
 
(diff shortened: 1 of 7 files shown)


jslee02:
@codex review

jslee02:
@codex review

PR #3200, opened 2026-06-26
Stop fmt 12.2 deprecation warning leaking through DART logging headers

jslee02:
## Problem

When DART is built against **fmt >= 12.2.0**, downstream consumers get a
`-Wdeprecated-declarations` warning that originates inside DART's public,
header-only logging templates:

```
spdlog/logger.h:80: warning: 'operator const fmt::basic_string_view<char>&' is deprecated
  ... in instantiation of dart::common::warn<...> requested from GenericJoint.hpp
```

Reported downstream as gazebosim/gz-physics#1018.

### Root cause

`fmt::format_string -> string_view` conversion that fmt now marks deprecated
("use `format_string::get()` instead"). Because the logging functions are
header-only and instantiated by downstream code, the warning surfaces in
every consumer compiled against fmt >= 12.2.0 — not in DART's own CI, which
pins `fmt < 12`.

The defect ultimately lives in spdlog (it still uses the deprecated conversion
on its fmt backend, while already using `.get()` on its std::format backend),
but DART is the right layer to stop leaking it to consumers, and can do so
without waiting on an spdlog release.

## Fix

Stop handing spdlog a format string. `detail::log()` now formats the message
with fmt up front and passes spdlog a ready-made `std::string` via its
non-format `log(level, msg)` overload, so the deprecated conversion is never
instantiated — a structural fix rather than a `-Wdeprecated-declarations`
suppression (which would be include-order fragile and compiler-dependent).

While here:

- **Gate on `should_log()` first**, exactly as spdlog's own
  `SPDLOG_LOGGER_CALL` macros do, so a disabled level builds no prefix, makes
  no allocation and does no formatting. This is also cheaper than the previous
  path, which always built the format string before spdlog's internal level
  check. Trade-off (matching those macros): messages below the active level are
  not fed to spdlog's backtrace ring buffer.
- **Guard the runtime `format()`** so a malformed format string logs a fallback
  instead of throwing out of a logging call (the format is not a compile-time
  constant, so this can only be caught at runtime).
- Remove the now-unused `DART_SPDLOG_RUNTIME` macro.

Public API and all six logging signatures are unchanged.

## Verification

Built a faithful reproduction against **fmt 12.2.0 + spdlog 1.15.3** through
DART's real `<dart/common/logging.hpp>`:

- `-Wdeprecated-declarations`: **2 → 0** on g++ 15.2 and clang 22.1.
- Compiles clean; no regression against fmt < 12 (conda spdlog); non-spdlog
  config (`DART_HAVE_spdlog=0`) clean under `-Wall -Wextra`.
- Runtime check confirms correct formatting, level gating (disabled levels emit
  nothing), the no-throw fallback on a bad format, and that braces in formatted
  data are not re-parsed (i.e. the plain-message overload is selected).

CI is the place to confirm the MSVC and macOS/AppleClang lanes, which the local
environment could not exercise.

## Follow-up

A backport to `release-6.20` will be a separate PR — that branch predates the
logging refactor, so the change there is structurally different.

### DIFF
--- a/dart/common/detail/logging-impl.hpp
+++ b/dart/common/detail/logging-impl.hpp
@@ -43,13 +43,11 @@
   #include <concepts>
   #include <type_traits>
   #include <utility>
-  // Check if spdlog is using fmt or std::format backend
-  #if defined(SPDLOG_USE_STD_FORMAT)
-    // spdlog uses std::format - no need for runtime wrapper
-    #define DART_SPDLOG_RUNTIME(str) str
-  #else
-    // spdlog uses fmt - need its compatibility macro for non-compile-time
-    // strings across supported fmt versions.
+  // These fmt helpers back normalize() below and are needed for the fmt
+  // backend, which is the configuration DART builds against. spdlog's
+  // std::format backend is not currently supported here (normalize() is
+  // referenced unconditionally at the call site below).
+  #if !defined(SPDLOG_USE_STD_FORMAT)
     #include <fmt/ostream.h>
     #include <spdlog/fmt/fmt.h>
     #include <spdlog/fmt/ostr.h>
@@ -109,7 +107,6 @@ auto normalize(T&& arg)
 
 } // namespace detail_log_arg
 
-    #define DART_SPDLOG_RUNTIME(str) SPDLOG_FMT_RUNTIME(str)
   #endif
 #else
   #include <iostream>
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3201, opened 2026-06-26
Stop fmt 12.2 deprecation warning leaking through DART logging headers (6.19.3)

jslee02:
Backport of #3200 to `release-6.19`, targeted at the **DART 6.19.3** patch
release.

## Problem

Building DART against **fmt >= 12.2.0** emits a `-Wdeprecated-declarations`
warning that originates inside DART's public, header-only logging templates and
therefore leaks into every downstream consumer (reported as
gazebosim/gz-physics#1018):

```
spdlog/logger.h: warning: 'operator const fmt::basic_string_view<char>&' is deprecated
  ... in instantiation of dart::common::warn<...>
```

`fmt::format_string -> string_view` conversion that fmt now marks deprecated.

## Fix

Route all six logging functions through a new `detail::logToSpdlog()` helper
that formats the message with fmt up front and hands spdlog a ready-made
`std::string` via its non-format `log(level, msg)` overload, so the deprecated
conversion is never instantiated. The helper also:

- gates on `should_log()` first, so a disabled level does no formatting or
  allocation (trade-off, matching spdlog's own `SPDLOG_LOGGER_CALL` macros:
  below-level messages are not added to spdlog's backtrace ring);
- guards the runtime `format()` so a malformed format string logs a fallback
  instead of throwing out of a logging call.

The now-unused `DART_SPDLOG_RUNTIME` macro is removed. Public API and all six
logging signatures are unchanged.

> Note: on `main` the logging code was refactored into a single `detail::log()`
> sink, so #3200 carries the same fix in a slightly different shape; this branch
> matches the 6.19 layout (six call sites + a shared helper).

## Verification

Reproduced against **fmt 12.2.0 + spdlog 1.15.3** through DART's real
`<dart/common/Logging.hpp>`:

- `-Wdeprecated-declarations`: **2 → 0** on g++ and clang.
- Compiles clean; no regression against fmt < 12; non-spdlog config clean under
  `-Wall -Wextra`.
- Runtime check confirms correct formatting, level gating, the no-throw fallback
  on a bad format, and that braces in formatted data are not re-parsed.

CI is the place to confirm the MSVC and macOS/AppleClang lanes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,17 @@
     or corrupting the LCP solve with NaN/Inf:
     [gazebosim/gz-physics#1010](https://github.com/gazebosim/gz-physics/issues/1010)
 
+* Build
+
+  * Stop a deprecation warning from leaking out of DART's header-only logging
+    templates when building against fmt 12.2.0 or newer. spdlog's variadic
+    `log()` routes the format string through a `fmt::format_string` ->
+    `string_view` conversion deprecated in fmt 12.2.0, so the warning surfaced
+    in every downstream that instantiates DART's logging. DART now pre-formats
+    messages and hands spdlog a ready-made string, sidestepping the deprecated
+    conversion:
+    [gazebosim/gz-physics#1018](https://github.com/gazebosim/gz-physics/issues/1018)
+
 ### [DART 6.19.2 (2026-06-19)](https://github.com/dartsim/dart/milestone/100?closed=1)
 
 DART 6.19.2 is a patch release on the DART 6 LTS line. It keeps automatic
--- a/dart/common/detail/Logging-impl.hpp
+++ b/dart/common/detail/Logging-impl.hpp
@@ -36,14 +36,60 @@
 #include <dart/common/Logging.hpp>
 
 #if DART_HAVE_spdlog
+  #include <fmt/format.h>
   #include <spdlog/spdlog.h>
-  #define DART_SPDLOG_RUNTIME(str) SPDLOG_FMT_RUNTIME(str)
+
+  #include <exception>
 #else
   #include <iostream>
   #include <string>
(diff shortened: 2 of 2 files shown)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
Addressed the fmt<8 finding. The logging helper now wraps the format string with spdlog's `SPDLOG_FMT_RUNTIME(...)` (= `fmt::runtime` on fmt ≥ 8, raw string on fmt 6/7) instead of an unconditional `fmt::runtime`, restoring the fmt version range the previous `SPDLOG_FMT_RUNTIME`-based path supported.

Verified the fmt<8 fallback (`fmt::format(raw_string, args)`) compiles and runs on fmt 7.1.3, and the fmt 12.2.0 path still emits **0** deprecation warnings. The same fix is on the 6.20 backport (#3202).

(For the record: spdlog ≥ 1.9.2 — DART's minimum — references `fmt::basic_runtime`/`fmt::format_string` unguarded in its own headers, so the spdlog + fmt<8 combination isn't independently buildable today; the guard is restored regardless since it's zero-cost and matches prior behavior.)

jslee02:
@codex review

jslee02:
Follow-up: replaced the `SPDLOG_FMT_RUNTIME` wrapper from my previous comment with `fmt::vformat(format_str, fmt::make_format_args(args...))`. `vformat` takes a runtime format string on every supported fmt version and is independent of spdlog's backend, which also fixes the `SPDLOG_USE_STD_FORMAT` + fmt ≥ 8 case Codex raised on #3202. Verified on fmt 7.1.3 and fmt 12.2.0 (0 warnings).

jslee02:
@codex review

PR #3202, opened 2026-06-26
Stop fmt 12.2 deprecation warning leaking through DART logging headers (6.20)

jslee02:
Backport of #3200 to `release-6.20` (DART 6.20.x).

## Problem

Building DART against **fmt >= 12.2.0** emits a `-Wdeprecated-declarations`
warning that originates inside DART's public, header-only logging templates and
therefore leaks into every downstream consumer (reported as
gazebosim/gz-physics#1018):

```
spdlog/logger.h: warning: 'operator const fmt::basic_string_view<char>&' is deprecated
  ... in instantiation of dart::common::warn<...>
```

`fmt::format_string -> string_view` conversion that fmt now marks deprecated.

## Fix

Route all six logging functions through a new `detail::logToSpdlog()` helper
that formats the message with fmt up front and hands spdlog a ready-made
`std::string` via its non-format `log(level, msg)` overload, so the deprecated
conversion is never instantiated. The helper also:

- gates on `should_log()` first, so a disabled level does no formatting or
  allocation (trade-off, matching spdlog's own `SPDLOG_LOGGER_CALL` macros:
  below-level messages are not added to spdlog's backtrace ring);
- guards the runtime `format()` so a malformed format string logs a fallback
  instead of throwing out of a logging call.

The now-unused `DART_SPDLOG_RUNTIME` macro is removed. Public API and all six
logging signatures are unchanged.

> Note: on `main` the logging code was refactored into a single `detail::log()`
> sink, so #3200 carries the same fix in a slightly different shape; this branch
> matches the 6.19 layout (six call sites + a shared helper).

## Verification

Reproduced against **fmt 12.2.0 + spdlog 1.15.3** through DART's real
`<dart/common/Logging.hpp>`:

- `-Wdeprecated-declarations`: **2 → 0** on g++ and clang.
- Compiles clean; no regression against fmt < 12; non-spdlog config clean under
  `-Wall -Wextra`.
- Runtime check confirms correct formatting, level gating, the no-throw fallback
  on a bad format, and that braces in formatted data are not re-parsed.

CI is the place to confirm the MSVC and macOS/AppleClang lanes.


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
Addressed. Switched the helper to `fmt::vformat(format_str, fmt::make_format_args(args...))`, which accepts a runtime format string directly — so it drops the `fmt::runtime`/`SPDLOG_FMT_RUNTIME` wrapper entirely. That makes it (a) compile on fmt 6/7 and fmt ≥ 8 alike, and (b) independent of spdlog's formatting backend (no `SPDLOG_USE_STD_FORMAT` mismatch with the `fmt::format` call). Verified the `std::string` runtime-format path (the `LoggingTest.RuntimeFormatString` case), brace-safety, no-throw fallback, and level gating, with **0** deprecation warnings on fmt 12.2.0 and a compile+run check on fmt 7.1.3. Same change applied to #3201.

jslee02:
@codex review

PR #3207, opened 2026-06-27
ci: persist ccache across runs on self-hosted runners (fix recurring CUDA Build timeout)

jslee02:
Stops the `CUDA Build` job from being **cancelled on nearly every PR run** (mine and unrelated PRs alike) while `push`-to-main passed.

## Root cause (in the shared cache action, not the timeout)

The shared `configure-compiler-cache` action points `CCACHE_DIR` at `RUNNER_TEMP`, which the Actions runner wipes each job. The dartsim self-hosted runners are **long-lived** (`restart: always`, jobs run sequentially on the same instance), so a cache under `RUNNER_TEMP` is **cold every single job** — and a full from-scratch build of the CUDA library + kernels + tests runs ~60 min, landing right on `timeout-minutes: 60`. (Cancelled run build log: objects compile at ~10–20s each = cache misses, killed at `[348/350]` linking, ~99% done.)

## Fix

- **`configure-compiler-cache`**: on self-hosted runners, keep ccache on the persistent per-runner tool-cache dir (`RUNNER_TOOL_CACHE`, confirmed persistent across jobs on these runners) instead of the per-job `RUNNER_TEMP`. Builds become incremental (~15 min warm vs ~60 cold), bounded by the existing `CCACHE_MAXSIZE=5G`. GitHub-hosted runners stay on `RUNNER_TEMP`. **`ci_cuda` is the only self-hosted consumer of this action today** (`ci_freebsd` does not use it), so the blast radius is contained.
- **`ci_cuda.yml`**: `timeout-minutes` 60 → 90 as a safety net for genuinely-cold runs (first build after the runner image is recreated).

Validated: both YAMLs parse, `bash -n` clean on the action script, `pixi run check-lint` clean. The runner-side behaviour is validated by this PR's own CUDA Build run (first run warms the persistent cache; subsequent runs should be fast).

@codex review

### DIFF
--- a/.github/actions/configure-compiler-cache/action.yml
+++ b/.github/actions/configure-compiler-cache/action.yml
@@ -89,8 +89,19 @@ runs:
         echo "CMAKE_C_COMPILER_LAUNCHER=${CMAKE_COMPILER_LAUNCHER_VALUE}" >> "${GITHUB_ENV}"
         echo "CMAKE_CXX_COMPILER_LAUNCHER=${CMAKE_COMPILER_LAUNCHER_VALUE}" >> "${GITHUB_ENV}"
 
+        # Self-hosted runners are long-lived but the Actions runner wipes
+        # RUNNER_TEMP every job, so a ccache there is always cold and each build
+        # compiles from scratch. Keep the cache on the persistent per-runner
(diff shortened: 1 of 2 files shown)


jslee02:
@codex review

PR #3199, opened 2026-06-26
Complete issue 3056 performance task

jslee02:
## Summary

- Records the final issue #3056 resting-scene performance outcome in `CHANGELOG.md`.
- Removes the completed `docs/dev_tasks/issue_3056_dart6_performance/` task folder after promoting the durable result.
- Retargets the remaining native-collision scoping reference that used to point at the deleted task folder.
- Keeps the active/no-deactivation workloads as benchmark guardrails rather than treating them as the settled-scene completion claim.

## Motivation / Problem

The issue #3056 performance stack has landed through #3194 on `release-6.20`, and the original default-sleeping 3003-body target is now well above real time. The transient dev-task README still described #3194 as open and held the completion evidence, so the repository needed a final owner-doc entry plus task-folder cleanup before the work could be called complete.

## Changes / Key Changes

- Adds a DART 6.20 changelog bullet summarizing the completed 3003-body resting-scene target and the final-state guardrails.
- Deletes `docs/dev_tasks/issue_3056_dart6_performance/README.md`; the task folder is now empty and removed from the tracked tree.
- Updates the native-collision port scoping doc to reference `CHANGELOG.md` and the rerunnable `examples/contact_benchmark` harness instead of the deleted task folder.
- Leaves executable code unchanged from the merged #3194 `release-6.20` head.

## Performance Comparison

RTF is higher-is-better. Baseline `382254394d7` is the #3172 baseline used by the recent performance PRs. Parent `165d072b576` is the merged #3194 release head. This PR head is `3bc23f45170`; it changes only changelog/dev-task documentation, so executable code is identical to the parent. The rows below were rerun sequentially after rebuilding `contact_benchmark` on code-equivalent PR head `1f15052a9c9`; the later `3bc23f45170` commit only retargets a Markdown reference. Final hashes, contacts, and pair counts matched the prior guardrails.

Primary active issue-scene command:

```bash
build/default/cpp/Release/bin/contact_benchmark .deps/gz-sim/examples/worlds/3k_shapes.sdf --steps 300 --warmup 0 --checkpoint 0 --quiet --collision dart --sdf-plane-shapes --world-threads 16 --max-contacts 12000 --max-contacts-per-pair 4 --disable-deactivation
```

| Scenario | Command difference from primary command | #3172 baseline `382254394d7` | #3194 parent `165d072b576` | This PR `3bc23f45170` | Final state on this PR |
| --- | --- | ---: | ---: | ---: | --- |
[...]

Issue #3056, opened 2026-06-17
Poor headless simulation performance (low RTF) with large number of objects (3000+ shapes)


azeey:
**Before submitting:**

- [x] Checked [documentation](https://dart.readthedocs.io/) and existing issues

**Environment:**

- DART version: v6.19.0
- OS: macOS
- Installation: Source build (via pixi)
- Compiler (if source): AppleClang 17.0.0.17000013

**Expected vs Current Behavior:**

- **Expected Behavior:** Headless simulation of a scene with a large number of shapes (e.g., 3,000 shapes) sitting on a ground plane should run close to or faster than real-time (RTF > 1.0). For comparison, other physics engines supported by Gazebo (such as Bullet and MuJoCo) are able to achieve an RTF > 2.0 under similar conditions once autosleep kicks in.
- **Current Behavior:** DART experiences extremely low simulation rates, achieving a Real-Time Factor (RTF) of only ~0.05.

**Steps to Reproduce:**

1. Clone DART and checkout the profiling tool branch:
   ```bash
   git clone https://github.com/azeey/dart.git
   cd dart
   git checkout perf/sdf-perf-test
   ```
2. Build DART from source:
   ```bash
   pixi run config
   cmake --build build/default/cpp/Release --target sdf_perf_test
   ```
3. Download the `3k_shapes.sdf` world file from the `gz-sim` repository ([3k_shapes.sdf](https://github.com/gazebosim/gz-sim/blob/main/examples/worlds/3k_shapes.sdf)).
4. Run the compiled executable on the SDF file:
   ```bash
   ./build/default/cpp/Release/bin/sdf_perf_test /path/to/3k_shapes.sdf 10000
   ```

**Code to Reproduce:**

The code for the custom headless benchmarking tool is available on the `perf/sdf-perf-test` branch of the `azeey/dart` fork:
- [SDF Performance Testing Tool Code](https://github.com/azeey/dart/tree/perf/sdf-perf-test/examples/sdf_perf_test)

**Additional Info:**

Here is the sample output of running the reproduction tool on macOS with `3k_shapes.sdf` (3,004 objects, 18,024 DOFs):

```text
Loading SDF world file: /Users/addisuzt/ws/jetty/src/gz-sim/examples/worlds/3k_shapes.sdf
[2026-06-17 11:49:19.691] [warning] [SdfParser] Link [link] is missing <inertial>; using default mass/inertia (1 kg, unit inertia). Specify <inertial> to avoid unstable simulation.
[...]

PR #3209, opened 2026-06-27
Add contact-rich container benchmark

jslee02:
## Summary

Adds a deterministic, contact-rich "container" workload (mixed spheres, boxes, and cylinders dropped into an open box) shared by the `contact_benchmark` example and a new `BM_INTEGRATION_contact_container` Google Benchmark target, so #3056 optimization work can be measured on an active scene instead of only the settled `3k_shapes.sdf` case. The GUI was consolidated into a single auto-sized ImGui panel, two GUI building blocks moved into `dart-gui-osg` for reuse, and the scene was hardened based on root-cause investigation of sleep and tunneling behavior (findings below).

## Key Changes

**Contact container harness**
- `ContactContainerScene.hpp`: deterministic scene builder (seeded jitter, layers, container size/walls, drop height, initial speeds, joint damping) reused by the example and the benchmark.
- `contact_benchmark --generate-container N` plus `--container-*` knobs; existing `--generate-objects` / `--generate-capsules` paths unchanged.
- `BM_INTEGRATION_contact_container`: DART-native and ODE detectors × single/multi-threaded rows, 200 steps per iteration with setup/teardown excluded from timing; reports final contacts/resting/mobile counters so workload composition is visible.

**GUI: one ImGui panel, no text overlay**
- The osgText HUD is gone; time/frame, smoothed RTF with lo/hi, smoothed contact counts, sleep-state summary, pipeline line, and key/color legends all live in the panel's Status section.
- The panel auto-sizes to its content (no clipped labels, no scrolling) and scales with `--gui-scale N` / `DART_GUI_SCALE`.
- Live scene-rebuild controls with optional auto-apply; `--gui-capture-exercise-widget` drives the rebuild paths headlessly at capture time.

**Reusable `dart-gui-osg` components** (API-additive)
- `RealTimeWorldNode::getSmoothedRealTimeFactor()`: ~1 s exponential moving average of the per-refresh RTF, for readable HUD-style displays.
- `dart::gui::osg::applyDefaultCameraPose(viewer, world)`: +Z-up, ~30°-down framing of the world's mobile bodies that always stays above the ground plane; used for the initial view, scene rebuilds, and the Recenter button.

**Scene hardening (each traces to a root cause below)**
- Container bodies get joint-space damping (defaults: linear 0.02, angular 0.1; `--container-linear-damping` / `--container-angular-damping`) so rolling primitives can settle — DART's contact model has no rolling resistance.
- An outer ground apron catches bodies ejected over the walls so they settle instead of free-falling forever.
- Container scenes default to 20000 total / 4 per-pair contact caps (same as the benchmark) unless overridden.

## Root-Cause Findings

Investigated with 20 s headless runs (`--generate-container 71 --checkpoint ...`); numbers in the validation section.

1. **Dense piles never sleep.** Sleep candidacy requires smoothed speeds below a hardcoded final-quiet gate (`1e-3` m/s / `1e-2` rad/s in `World::updateRestStates`), independent of the configured thresholds — raising thresholds 5×/3× accumulates dwell (13.1 s vs 7.7 s) but leaves resting counts identical. Fixed in #3226 (the gate now scales with the configured thresholds). The "candidate for a frame, then awake again" flicker visible with sleep-state colors is the island-rest veto clearing candidacy each solve (finding 2).
2. **Penetration creep under load.** Contact error correction is capped at `DART_MAX_ERV = 1e-3` m/s; a 4-body tower recovers penetration at exactly ~1 mm/s, while the 71-body pile compacts faster than the cap corrects, so max penetration grows ~5–10 mm/s without bound (e.g. 0.15 m → 0.35 m over 20 s). The signature reproduces with DART-native, FCL, and ODE detectors, with both Dantzig and PGS, with 4- and 8-contact per-pair caps (the DART detector emits ≤ 4 per pair regardless), and with or without spawn overlap (`--generated-spacing 1.4`) — pointing at large-island LCP convergence / ERV budget rather than any detector. Since the island-rest veto requires every contact to reach ≤ 1e-5 m penetration, creeping piles can never freeze; this harness is the deterministic reproducer for that follow-up under #3056.
3. **ODE cylinder fallback floods the contact cap.** When the runtime probe rejects native ODE cylinder normals, the trimesh fallback emits 100+ contacts per pair, overflowing the previous 1000-contact default; silently dropped contacts let bodies tunnel out of the container (observed as a permanent free-faller reaching 160 m/s). Mitigated here by the caps + apron; the wrapper-level manifold reduction is follow-up.

## Validation

- `pixi run lint` clean; `contact_benchmark`, `BM_INTEGRATION_contact_container`, and `dart-gui-osg` build clean.
- Benchmark row `BM_ContactContainerActive/60/0/1`: 200-step window ends with `contacts=96` for 60 objects, confirming the timed region is contact-rich (machine-relative timings only; CPU scaling enabled locally).
- GUI captures (initial, settled 3 s, post-widget-exercise rebuild, and default 3-lane scene): +Z-up ~30° framing from above in every state, no text overlay, apron visible behind the translucent walls.
- `--gui-capture-exercise-widget` drives the container/live-rebuild widget paths headlessly without errors. The ImGui panel itself does not render into offscreen captures, so panel layout was verified by code path plus a live-GUI look is recommended.
- ODE cap fix A/B (71 objects, 20 s): before — `max_pair_contacts` up to 248, total contacts pinned at the 1000 default cap, one body tunneled out and reached 161 m/s in permanent free fall; after — 4 per pair, ~290 total, all 71 bodies islanded, `max_lin ≤ 0.42` m/s.
- Damping A/B (71 objects, DART detector, 20 s): the pile reaches the quiet band roughly twice as fast; resting stays 0 due to findings 1–2, as documented.

## Breaking Changes

- [x] None (library changes are API-additive; example defaults changed only for the new `--generate-container` path)

## Related Issues / PRs

- Contributes to #3056
- Follows #3199
- Follow-up: #3226 (deactivation final-quiet gate scaling)

---

#### Checklist

- [x] Milestone set (DART 6.20.0 for `release-6.20`)
- [x] CHANGELOG.md updated
- [x] Add unit tests for new functionality (benchmark/example harness; covered by build, capture, and headless runs — library sleep-gate change is tested in the follow-up PR)
- [x] Document new methods and classes (CLI help, Doxygen comments on new `dart-gui-osg` APIs, changelog)
- [x] Add Python bindings (dartpy) if applicable (not applicable)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3215, opened 2026-06-27
Scope publish_dartpy concurrency by event name (DART 6 LTS)

jslee02:
The `publish_dartpy` concurrency group was `${{ github.workflow }}-${{ github.ref }}`, so a manual `workflow_dispatch` publish and the push-triggered build of the **same** `release-*` ref shared one group with `cancel-in-progress: true` and cancelled each other. During the v6.19.3 release this killed the dispatch publish mid-run (before `upload_pypi`), and left `cancelled` (red ✗) checks on the branch tip.

Add `-${{ github.event_name }}` to the group so `push` and `workflow_dispatch` on a ref no longer collide; same-event runs on a ref still supersede each other (`cancel-in-progress` unchanged).

### DIFF
--- a/.github/workflows/publish_dartpy.yml
+++ b/.github/workflows/publish_dartpy.yml
@@ -36,7 +36,10 @@ permissions:
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

PR #3266, opened 2026-07-04
Add rigid IPC two-wall tunnel fixture-row coverage

jslee02:
## Summary

- Add DART-owned runtime coverage for the audited upstream two-wall rigid IPC tunnel unit-test fixture and mark that PLAN-082 manifest row implemented. The three-wall, four-wall, and 8K tunnel rows are left planned with a recorded reason (see below).

## Motivation / Problem

- PLAN-082 (rigid implicit-barrier contact) is tracked against an upstream fixture manifest, and the tunnel unit-test fixtures exercise a core guarantee: a fast body confined in a tight fixed passage must stay intersection-free (anti-tunneling) while advancing along the tunnel axis. The two-wall row had no DART-owned coverage.

## Changes / Key Changes

- `tests/unit/simulation/contact/test_rigid_ipc_paper_experiments.cpp`: add `TwoWallTunnelCubeStaysBetweenWallsFixtureRow`, reconstructing `fixtures/3D/unit-tests/tunnel/2-walls.json`. A unit cube flies down a tight (1 mm) fixed floor/ceiling corridor via the opt-in `RigidIpcContactStage`; it asserts finite state, forward progress (`startX + 0.1` margin), active rigid IPC contact, and cube-corner clearance against the fixed inner wall faces (avoiding static wall-vs-wall `collide()` contamination).
- `scripts/generate_rigid_ipc_fixture_manifest.py` + regenerated `docs/plans/082-rigid-implicit-barrier-contact/rigid_ipc_fixture_manifest.json`: mark that row implemented (planned 339 → 338, implemented 459 → 460).
- `docs/dev_tasks/rigid_ipc_solver/README.md`, `RESUME.md`: progress-log updates, including the deferral rationale below.

## Why only the two-wall row

A pre-PR review flagged that a bare `> startX` progress check could pass a frozen cube. Adding a `startX + 0.1` margin then exposed that at the *faithful* 0.1 mm clearance of `3-walls.json` / `4-walls.json`, the conservative curved-CCD line search limits each step to about the separation distance, so a fast cube is arrested (intersection-free but only ~3 mm / ~56 mm of travel) rather than traversing. Loosening the gap would be an overclaim, since the tight tolerance is those fixtures' defining feature, so they stay planned as a conservative-CCD / performance gap. The two-wall fixture's 1 mm clearance is above that crawl threshold, so its cube genuinely traverses. The 8K variant additionally holds active contact against two 8192-triangle walls every step (~minutes/step), too slow for a unit test.

## Testing

Verified on the branch after bringing in the latest `main` (the intervening `main` commits are docs/plan-only and leave the simulation sources unchanged):

- `test_rigid_ipc_paper_experiments` full suite — 46/46 pass (including the new two-wall tunnel test)
- `pixi run python scripts/generate_rigid_ipc_fixture_manifest.py --upstream-dir <rigid-ipc@23b6ba6>` — regeneration is idempotent against the committed manifest
- `pixi run python scripts/check_rigid_ipc_fixture_manifest.py` (offline and `--upstream-dir`) — OK, 798 entries
- `pixi run pytest tests/test_rigid_ipc_fixture_manifest_tools.py` — 50/50
- `pixi run lint` — pass

Not run for this change: full `pixi run test-all` / `-e cuda test-all` — the change is isolated to one test binary plus the manifest data file, fully covered by the gates above.

## Breaking Changes

- [x] None

### DIFF
--- a/docs/dev_tasks/rigid_ipc_solver/README.md
+++ b/docs/dev_tasks/rigid_ipc_solver/README.md
@@ -210,6 +210,18 @@
         through the opt-in runtime stage. A rotated cube with a large time-step
         velocity toward a fixed wall remains intersection-free, and the stage
         reports a conservative CCD line-search hit.
+  - [x] Add audited two-wall tunnel corridor fixture coverage through the
+        opt-in runtime stage. A unit cube flies down a tight (1 mm) fixed
+        floor/ceiling corridor at high speed, activates rigid IPC contact, stays
+        finite, keeps advancing along the tunnel axis, and stays within the
+        fixed wall clearances. The three-wall, four-wall, and 8K tunnel rows
+        remain planned: at their tighter 0.1 mm clearance the conservative
+        curved-CCD line search limits each step to about the separation
+        distance, so a fast cube is arrested (intersection-free but ~3 mm of
+        travel) rather than traversing the channel, and the dense 8K walls are
+        too slow for a unit test pending dense-contact performance work. These
+        are next gaps for the conservative-CCD/performance climb, not
+        test-authoring gaps.
   - [x] Add the first audited tessellated-plane fixture coverage through the
         opt-in runtime stage. A cube falls onto a fixed two-triangle mesh plane,
         activates contact, stays finite, and preserves nonnegative clearance.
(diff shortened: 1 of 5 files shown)


jslee02:
@codex review

PR #3274, opened 2026-07-04
Add a matrix-free-vs-sparse-CG contact-heavy crossover benchmark

jslee02:
## Summary

PLAN-081 M7 measurement slice. Produces the evidence a future **automatic
large-mesh matrix-free selection** policy needs, without changing any default
path (both the matrix-free and iterative solvers stay opt-in).

**Benchmark** `BM_DeformableContactCube{Cg,MatrixFreeCg}Step`: an unpinned FEM
cube settles on a static ground barrier, so the projected-Newton solve is
dominated by the stiff clamped-log barrier Hessian — the ill-conditioned regime
where matrix-free block-Jacobi CG and sparse Jacobi-preconditioned CG diverge. The
same scene runs under both solvers at increasing resolutions (27/64/125/216
nodes), emitting per-step **CG iterations**, **matrix-free solves**,
**steepest-descent fallbacks** (the key convergence-regression signal), and
**assembled-Hessian footprint**. Measured sample: sparse-CG footprint grows
2k→12k nnz while matrix-free stays 0.

**Regression** `MatrixFreeLinearSolverMatchesSparseCgOnScaledGroundContact`: a
mid-size 64-node cube (192 DoF, above the 128-DoF dense-direct cap, so the direct
path is never used) settling on a ground barrier reaches the same equilibrium
under sparse CG and matrix-free CG, with matrix-free keeping zero assembled-Hessian
footprint — proving parity holds as the contact mesh scales past where a dense
direct solve is viable.

## Scope

Test/benchmark only — no library or default-solver-path change.

## Verification

- `pixi run test-all` — all 6 categories PASSED (the new regression runs in the
  Simulation Tests suite).
- Benchmark smoke confirms both paths run and emit the crossover counters.
- The regression's position-parity tolerance is `1e-5` — the two preconditioners
  agree to ~2e-10 over the run (both solve the same system to the same CG
  tolerance), so `1e-5` is non-flaky yet tight enough to catch a genuine
  (mm-scale) divergence. It also asserts the cube actually descends into the
  barrier band (a real contact-heavy solve, not free-fall).

## Changelog

**No entry.** Internal benchmark/test evidence tooling; no user-facing change.

### DIFF
--- a/tests/benchmark/simulation/bm_deformable_body.cpp
+++ b/tests/benchmark/simulation/bm_deformable_body.cpp
@@ -327,6 +327,76 @@ sx::DeformableBodyOptions makeFemCubeOptions(
 }
 
 //==============================================================================
+// A solid cellsPerSide^3 FEM cube, unpinned and started just above a static
+// ground barrier (top face at z == 0). Under gravity it drops and settles on
+// the barrier, so the projected-Newton solve is dominated by the near-singular
+// clamped-log barrier Hessian on the contacting bottom nodes -- the stiff,
+// ill-conditioned regime where the matrix-free block-Jacobi preconditioner and
+// the sparse incomplete-Cholesky preconditioner diverge in CG iterations and
+// per-step cost. This is the contact-heavy fixture behind the matrix-free vs
+// sparse-CG crossover benchmarks; the crossover it measures (per-step time, CG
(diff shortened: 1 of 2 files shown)


jslee02:
@codex review

jslee02:
@codex review

PR #3275, opened 2026-07-04
Isolate Pixi binary paths in CI

jslee02:
## Summary
- configure setup-pixi to install Pixi under the per-job runner temp directory
- apply the same pixi-bin-path across CI workflows so stale ~/.pixi/bin/pixi files cannot break setup

## Root cause
The release-6.20 push run failed in DART 6 Performance Dashboard and CI gz-physics before project code ran. Both jobs failed inside prefix-dev/setup-pixi because /home/runner/.pixi/bin/pixi already existed on the runner. Using runner.temp gives each job an isolated Pixi binary path while preserving setup-pixi caching for the project environment.

## Validation
- pixi run check-lint
- git diff --check

### DIFF
--- a/.github/workflows/api_doc.yml
+++ b/.github/workflows/api_doc.yml
@@ -71,6 +71,7 @@ jobs:
         uses: prefix-dev/setup-pixi@v0.9.6
         with:
           cache: true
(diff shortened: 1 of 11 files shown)


jslee02:
@codex review

jslee02:
@codex review

PR #3281, opened 2026-07-05
Port native collision math core (phase 1, internal-only)

jslee02:
## Summary

- Execute **phase 1** of the native-collision-port plan
  (`docs/dev_tasks/dart6_dependency_minimization/03-native-collision-port-scoping.md`,
  merged in #3234): port the DART 7 native collision **algorithm core** to
  `release-6.20` as C++17, with no EnTT and no new dependency,
  **internal-only** — no installed headers, no public component, no
  behavior change. FCL remains the default detector.

## Motivation / Problem

The native collision port is the only remaining lever that can make
FCL/Bullet/ODE optional on DART 6.20 (the actual dependency win, phase 7).
Phase 0's baseline packet (#3271) froze the incumbent evidence envelope;
this PR lands the first non-default code slice per the phased plan: the
pure-math core with deterministic unit coverage.

## Changes / Key Changes

- **`dart/collision/native/` (new, 21 library files):** AABB,
  contact point, contact manifold, shared types, shape interface + SDF
  interface header, GJK (+inline support), MPR, box-box SAT/face-clip/
  contact-reduction + dispatcher-level box-box, sphere-sphere — ported
  **verbatim** from `origin/main` @ `dbe6fcccb` (DART 7 PLAN-035 lineage:
  #2652/#2688/#2700), namespace `dart::collision::native`, snake_case
  files (same precedent as the `dart/simd` backport #3229).
- **Review follow-up:** `CollisionResult` now has custom move construction and
  move assignment that invalidate cached manifold/contact views and leave the
  moved-from result empty, preventing stale raw contact pointers after moves.
- **Review follow-up:** box-box face clipping now preserves each clipped
  point's penetration depth from the reference-face signed distance, with
  regression coverage for tilted face patches and updated contact-depth
  invariants.
- **`dart/collision/native/detail/span.hpp` (new):** minimal C++17
  `span` replacing `std::span` — the **only** C++20 feature in the
  entire ported set (verified by exhaustive grep of the DART 7 tree).
- **`tests/unit/collision/native/` (new, 7 suites):**
  `UNIT_collision_native_{aabb,types,shapes,box_box,gjk,gjk_degenerate,sphere_sphere}`
  with the DART 7 port coverage plus DART 6 supplementary regression coverage,
  including binary sphere checks, moved `CollisionResult` cache invalidation,
  and per-point box contact depths.
- **Internal-only CMake:** explicit source lists via
  `dart_add_core_headers/sources`; **no `install()` rule, no generated
[...]

PR #3283, opened 2026-07-05
Honor enableContact in native sphere-sphere binary checks

jslee02:
## Summary

- Fix `dart/collision/native/narrow_phase/sphere_sphere.cpp` ignoring
  `CollisionOption::enableContact`: binary checks
  (`CollisionOption::binaryCheck()`) previously still called
  `result.addContact(...)` on every hit path, mutating the caller's
  `CollisionResult`, unlike `box_box.cpp` which gates contact creation
  before adding contacts.

## Motivation / Problem

Found while porting the native collision core to `release-6.20`
(dartsim/dart#3281): review flagged that sphere-sphere is the only
narrow-phase pair in the ported set that doesn't honor the binary-check
contract. Verified here: `box_box.cpp` checks `option.enableContact` at
four sites; `sphere_sphere.cpp` had zero such checks.

## Changes / Key Changes

- Both `collideSpheres` entry points (centers overload via
  `collideSphereCenters`, and the `SphereShape`/transform overload) now
  return the boolean overlap result without touching the
  `CollisionResult` when `enableContact` is false. Binary checks also no
  longer consume the `maxNumContacts` budget, matching box-box's
  behavior (`collided = true; if (!option.enableContact) return true;`
  precedes the cap check there).
- New shared overlap predicate keeps the boolean semantics identical to
  the contact paths (inclusive at exact touching distance).
- Regression test `SphereSphere.BinaryCheckReportsHitWithoutContacts`
  covering hit + miss on both overloads.
- Regression test coverage for `maxNumContacts == 0` with
  `enableContact == false` and the compound narrow-phase binary-check
  expectation.

## Testing

- `cmake --build build/default/cpp/Release --target test_narrow_phase test_sphere_sphere -j 24`
- `ctest --test-dir build/default/cpp/Release -R '^(test_narrow_phase|test_sphere_sphere)$' --output-on-failure`
- `pixi run lint`
- Commit hook: `pixi run check-lint-quick`

## Breaking Changes

- [x] None — binary checks previously returned the same booleans; they
  just no longer emit contacts (the documented intent of
  `enableContact`).

## Related Issues / PRs

- Dual-PR counterpart of the same fix in the DART 6.20 native-collision
  port: dartsim/dart#3281.


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3065, opened 2026-06-18
Convert live SKEL fixtures to SDF

jslee02:
## Summary

- Converts the SKEL fixtures that are still used by examples, tests, benchmarks, and Python tests into SDF fixtures so the DART 7 SKEL removal can proceed without losing coverage.
- Keeps runtime SKEL support intact in this PR; the follow-up removal slice owns deleting the parser, dispatch, bindings, stubs, parser tests, and data/skel.

## Motivation / Problem

- DART 7 should remove legacy SKEL support before release rather than redesigning SKEL as YAML. The first safe step is to move live coverage and sample-resource references away from data/skel while preserving the scenarios those tests exercise.

## Changes / Key Changes

- Added SDF replacements for the referenced single_pendulum, cube, shapes, and test_shapes SKEL fixtures under data/sdf/test/.
- Updated C++ IO, simulation skeleton-loader, resource-retriever, and URDF negative-path tests to use the converted SDF or URDF assets.
- Updated Python simulation and URDF tests to avoid SKEL sample paths; remaining Python SKEL mentions are parser/stub-specific follow-up work.
- Guarded SDF-backed URI-loader tests so no-sdformat builds skip that coverage instead of failing when SDF parsing is intentionally disabled.
- Updated onboarding, readthedocs, module guidance, and the skel_format_evolution dev task to describe SKEL as a DART 6 compatibility format and track DART 7 removal.

## Before / After

| Area | Before | After |
| --- | --- | --- |
| Live test/resource fixtures | Several tests loaded dart://sample/skel or data/skel assets. | Live C++/Python tests, examples, tutorials, and benchmarks no longer reference dart://sample/skel or data/skel. |
| SKEL removal plan | Docs still described deprecated/legacy SKEL support as retained. | Docs and dev-task notes point to removal before DART 7, with YAML left as a separate optional format decision. |
| Runtime support | SKEL parser and dispatch still exist. | Still exists; parser/API removal is intentionally left for the next slice. |

## Testing

- `pixi run cmake --build build/default/cpp/Release --target test_io_readNone INTEGRATION_io_Read test_skeleton_loader INTEGRATION_utils_LocalResourceRetriever INTEGRATION_utils_DartResourceRetriever INTEGRATION_io_UrdfParser --parallel "$JOBS"`
[...]

PR #3288, opened 2026-07-05
Remove legacy SKEL format from DART 7

### LINKED ISSUE #496: Proposal: Use YAML for .skel files
This would be a huge change, so I don't necessarily expect it to gain traction, but I think it would make everyone's lives better to use YAML instead of XML for `.skel` files. There are a few reasons I believe this:

(1) We use an awful lot of machinery just to parse XML files. In addition to using TinyXML2, we also have a ton of functions for interpreting the content of each element. But YAML has a C++ library called `yaml-cpp` which has a very nice API and which would not require any extra machinery on our end to read in everything we need. This library is cross-platform. I've already been using YAML and this library to construct and save things like Eigen data structures, so I'm confident that it would serve our needs.

(2) It is common for users to manually write, read, and manipulate robot model files. XML is a terrible and ugly format which only barely passes as human-readable. YAML on the other hand is extremely human readable. I think this would be especially advantageous since our skel format is a custom format of our own creation, so there does not exist **any** support right now for automatically generating `.skel` files, meaning everyone currently needs to generate the files by hand.

Another -- perhaps more reasonable -- option would be to deprecate the `.skel` format and come up with a new name for files that describe DART Worlds and Skeletons. This would allow us to maintain backwards compatibility.

jslee02:
## Summary

- Remove the legacy SKEL model format from DART 7 and route users to URDF,
  SDF, MJCF, or a `release-6.*` branch for old SKEL assets.

## Motivation / Problem

- SKEL is a DART-only legacy XML format. DART 7 should not carry it forward or
  redesign it as YAML while the maintained interchange paths are URDF, SDF,
  MJCF, and future project/scene serialization.

## Changes / Key Changes

- Removed `SkelParser`, `ModelFormat::Skel`, `.skel` / `<skel>` format
  inference, and direct SKEL dispatch from `dart::io`.
- Removed `.skel` sample assets, parser-specific tests, dartpy bindings/stubs,
  and user-facing SKEL docs.
- Moved the non-SKEL Kima Collada meshes from `data/skel/kima/` to
  `data/mesh/kima/`.
- Added coverage that `.skel` files are not supported through
  `dart::io::readSkeleton()`.
- Updated DART 7 docs and changelog guidance for migrating SKEL assets to URDF,
  SDF, or MJCF.

## Testing

- `pixi run python -m pytest python/tests/unit/test_check_dartpy_import_layout.py python/tests/unit/utils/test_utils_stub_import.py python/tests/unit/test_check_dart7_world_promotion_blockers.py`
- `pixi run check-ai-commands`
- `pixi run check-dartpy-import-layout`
- `pixi run run-cpp-target INTEGRATION_io_Read`
- `pixi run run-cpp-target UNIT_dynamics_MeshShape`
- `pixi run run-cpp-target test_mesh_loaderNone`
- `pixi run lint`
- `pixi run build`
- `pixi run test-all`
- `pixi run -e cuda test-all`
- After merging current `origin/main` into the PR branch: `pixi run lint`

## Breaking Changes

- [ ] None
- Removes DART 7 SKEL loading and the `SkelParser` API. Migrate assets to URDF,
  SDF, or MJCF, or use the active `release-6.*` branch for legacy SKEL
  compatibility.

- Closes https://github.com/dartsim/dart/issues/496
- Phase 1 fixture conversion: https://github.com/dartsim/dart/pull/3065

---

#### Checklist

- [x] Milestone set (DART 7.0 for `main`, branch-matching DART 6.x release
      milestone for the active DART 6 LTS branch)


jslee02:
@codex review

PR #3298, opened 2026-07-06
Fix native sphere-sphere zero-contact-limit short-circuit

jslee02:
## Summary

- Fix a contract regression in the native sphere-sphere pair introduced by
  #3281: `CollisionOption::maxNumContacts == 0` must short-circuit
  detection and return false, but the `enableContact` binary-check early
  return ran before the cap check, so a `{enableContact=false,
  maxNumContacts=0}` caller reported an overlap instead of `false`.

## Motivation / Problem

The documented `CollisionOption` contract
(`dart/collision/collision_option.hpp`): *"A value of 0 short-circuits
collision detection and returns false immediately."* DART 7's original
`sphere_sphere.cpp` honored this via its leading
`numContacts() >= maxNumContacts` check (`0 >= 0` -> false). The
`enableContact` fix ported in #3281 added an earlier return that bypassed
that check for the zero-limit-with-contacts-disabled combination. Flagged
by review on the main-branch dual PR (#3283).

## Changes / Key Changes

- Both `collideSpheres` entry points now short-circuit `maxNumContacts == 0`
  to `false` **before** the `enableContact` binary-check path.
- Regression test `SphereSphere.ZeroContactLimitShortCircuitsEvenWithoutContact`
  covering both overloads with an overlapping pair.
- Scope is limited to sphere-sphere (the pair changed by #3281). The
  documented-contract inconsistency in other native pairs
  (`box_box`/`convex_convex`/`mesh_mesh` return `true` on
  `maxNumContacts == 0`) is pre-existing upstream DART 7 behavior and is
  intentionally not widened here.

## Testing

- `pixi run lint`: clean.
- `ctest -R UNIT_collision_native_sphere_sphere`: 1/1 pass (includes the
  new regression test).

## Breaking Changes

- [x] None — restores the documented contract; `binaryCheck()`
  (`maxNumContacts = 1`) and contact-generating paths are unchanged.

## Related Issues / PRs

- Dual-PR counterpart on `main`: #3283.
- Follows up the phase-1 native core port #3281 (internal-only).

### DIFF
--- a/dart/collision/native/narrow_phase/sphere_sphere.cpp
+++ b/dart/collision/native/narrow_phase/sphere_sphere.cpp
@@ -119,6 +119,13 @@ bool collideSphereCenters(
   const double dy = center2.y() - center1.y();
   const double dz = center2.z() - center1.z();
 
+  // A zero contact limit short-circuits detection and returns false (see the
+  // CollisionOption::maxNumContacts contract); this must precede the
+  // enableContact binary-check path.
+  if (option.maxNumContacts == 0) {
+    return false;
+  }
+
   if (!option.enableContact) {
     return spheresOverlap(dx, dy, dz, radius1, radius2);
   }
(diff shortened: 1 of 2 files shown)


jslee02:
@codex review

PR #3300, opened 2026-07-06
Fix dart-gui-osg mouse-handler unregistration and InteractiveFrameDnD tool ownership

jslee02:
Two `dart::gui::osg` interaction-teardown bugs, surfaced while building the consolidated demos app (they are latent in any app that registers/destroys mouse handlers or interactive-frame gizmos at runtime).

## Bugs

1. **`DefaultEventHandler::addMouseEventHandler` never wires the unregister path.** The header documents that destroying a registered `MouseEventHandler` auto-unregisters it, but `addMouseEventHandler` only calls `handler->addSubject(this)` — the handler observes the event handler, not the reverse — so `DefaultEventHandler::handleDestructionNotification` never fires and the destroyed handler stays in `mMouseEventHandlers` as a dangling pointer. The next mouse event calls `triggerMouseEventHandlers()` → `update()` on freed memory. Fix: also `addSubject(_handler)` so the documented contract actually runs.

2. **`InteractiveFrameDnD` leaks its nine `InteractiveToolDnD`.** The constructor `new`s nine sub-DnDs into `mDnDs` but the destructor was `= default`, leaking them on every teardown. Fix: out-of-line destructor that deletes them (ABI-safe — no member/layout change).

## Scope / policy

- Release-6.20 only: `dart/gui/osg` does not exist on `main` (DART 7 rewrote the GUI), so no dual-PR applies.
- No public header or package-component change; not Gazebo-facing.

## Verification

- `pixi run build` + `pixi run lint` green.
- `drag_and_drop`-style interactive smoke and the consolidated `dart-demos --cycle-scenes` (which exercises gizmo attach/detach and mouse-handler teardown across 31 scene switches) run clean.

Companion PR: the consolidated demos work (`feature/dart6-consolidated-demos`) builds on these fixes for its runtime gizmo dragging.

### DIFF
--- a/.github/workflows/ci_simd.yml
+++ b/.github/workflows/ci_simd.yml
@@ -14,7 +14,9 @@ on:
       - "release-*"
     paths:
       - "dart/simd/**"
+      - "dart/collision/dart/**"
       - "tests/unit/simd/**"
+      - "tests/unit/collision/test_DARTCollisionDetector.cpp"
       - "tests/benchmark/simd/**"
       - ".github/workflows/ci_simd.yml"
   pull_request:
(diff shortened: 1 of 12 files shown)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3305, opened 2026-07-06
Use overflow-safe hypot in native sphere binary overlap

jslee02:
## Summary

- Make the native sphere-sphere binary-check overlap predicate overflow-safe.
  `spheresOverlap` used a squared-distance comparison
  (`dx²+dy²+dz² <= (r1+r2)²`) that overflows to `inf` for large-but-finite
  inputs, so a clearly-separated pair at radii ~1e154 spuriously reports a hit
  the (linear, overflow-safe) axis-aligned contact path rejects.

## Motivation / Problem

Follow-up to #3298 (which landed `spheresOverlap` for the `maxNumContacts==0`
short-circuit). Flagged by review on the main-branch dual PR (#3283): the
binary `enableContact=false` path now runs `spheresOverlap` before the linear
axis checks, so its overflow diverges from the contact path for finite inputs.

## Changes / Key Changes

- Replace the squared comparison with `std::hypot(dx, dy, dz) <= r1 + r2`
  (C++17 3-arg `std::hypot`, overflow-safe; `<cmath>` already included).
- Regression test `SphereSphere.BinaryCheckDoesNotOverflowOnLargeSeparation`
  (separated miss + genuine overlap at ~1e154 magnitude).

## Testing

- `pixi run lint`: clean.
- `ctest -R UNIT_collision_native_sphere_sphere`: 1/1 pass (incl. new test).

## Breaking Changes

- [x] None — identical results for all physical inputs; only removes a
  spurious hit at non-physical (~1e154) magnitudes.

## Related Issues / PRs

- Dual-PR counterpart on `main`: #3283 (same commit).
- Follows #3298 / the phase-1 native core #3281.

### DIFF
--- a/dart/collision/native/narrow_phase/sphere_sphere.cpp
+++ b/dart/collision/native/narrow_phase/sphere_sphere.cpp
@@ -103,8 +103,9 @@ bool collideSpheresOnAxis(
 bool spheresOverlap(
     double dx, double dy, double dz, double radius1, double radius2)
 {
-  const double sumRadii = radius1 + radius2;
-  return dx * dx + dy * dy + dz * dz <= sumRadii * sumRadii;
+  // Use std::hypot (overflow-safe) rather than a squared comparison so
+  // large-but-finite inputs agree with the linear axis-aligned contact path.
+  return std::hypot(dx, dy, dz) <= radius1 + radius2;
 }
 
 bool collideSphereCenters(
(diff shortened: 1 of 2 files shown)


jslee02:
@codex review

jslee02:
Withdrawing: the `std::hypot` approach here rounds one ULP differently than the squared-distance predicate the contact path uses, so it can make a binary (`enableContact=false`) check disagree with full contact detection for ordinary near-tangent inputs (raised on #3283). The correct fix is to keep the squared predicate, which release-6.20 already has via #3298, so no change is needed here. The main-branch counterpart #3283 has been reverted to the squared form for the same reason.

PR #3309, opened 2026-07-06
Fix scalar macOS broadphase build (DART 6 LTS)

jslee02:
## Summary

- Fix the release-6.20 macOS scalar build by making the DART-native broadphase SIMD overlap helper instantiate only when the SIMD path is compiled.

## Motivation / Problem

- AppleClang warning-as-error builds on release-6.20 fail on arm64 scalar configurations because `computeFiniteOverlapMask4()` is an internal function whose only call sites are discarded by `if constexpr (kUseNativeBroadphaseSimd)`.
- The root issue is helper instantiation in scalar builds, not the warning itself, so the fix keeps the helper unavailable unless the SIMD path actually uses it.

## Changes / Key Changes

- Convert the fixed-width internal overlap helper into a width-parameterized function template.
- Instantiate it only from the existing native-SIMD broadphase branches.
- Preserve the scalar broadphase path and SIMD behavior.

## Testing

- `pixi run clang-format -i dart/collision/dart/DARTCollisionDetector.cpp`
- `pixi run config`
- `pixi run cmake --build build/default/cpp/Release --target dart -j 12`
- `pixi run lint`

## Breaking Changes

- [x] None

- Addresses release-6.20 `CI macOS` run 28766439435 job 85291604587, plus superseded jobs 85289560052 and 85289560128 from run 28765744101 with the same AppleClang error.
- Main has a separate writer-test CI fix in the companion PR.

---

#### Checklist

- [x] Milestone set (DART 7.0 for `main`, branch-matching DART 6.x release milestone for the active DART 6 LTS branch)
      - No new entry required: this is a narrow internal build/CI hardening fix for the existing DART-native broadphase implementation already represented in the release collision changelog section.
      - N/A: no behavior change; this fixes scalar-build compilation of existing code.
      - N/A: no public API.
      - N/A: no binding changes.

### DIFF
--- a/dart/collision/dart/DARTCollisionDetector.cpp
+++ b/dart/collision/dart/DARTCollisionDetector.cpp
@@ -1962,26 +1962,27 @@ bool overlaps(const BroadphaseEntry& entry1, const BroadphaseEntry& entry2)
 }
 
 //==============================================================================
-std::uint32_t computeFiniteOverlapMask4(
+template <std::size_t Width>
+std::uint32_t computeFiniteOverlapMask(
     const BroadphaseEntry& entry,
     const SortedBroadphaseBounds& bounds,
     std::size_t begin)
 {
-  using Vec4d = dart::simd::Vec<double, kBroadphaseSimdWidth>;
+  using Vecd = dart::simd::Vec<double, Width>;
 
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

PR #3297, opened 2026-07-05
Add DART 6 simulation allocation hardening

jslee02:
## Summary

- Add DART 6 World-owned simulation memory management and optional `World::enterSimulationMode()` preparation.
- Route DART-owned profiler, solver, Dantzig, and `World::step` scratch away from steady-state heap allocation on native same-shape simulation loops.
- Add allocation-counting regression gates for native DART collision and downstream-compatible base-allocator gates for Bullet/ODE paths.

## Motivation / Problem

DART 6 release and wheel builds keep the built-in profiler enabled, and contact-heavy simulation loops were still doing thousands of DART-owned heap allocations per step. That makes step latency less predictable and hides allocator regressions in code paths used by downstream Gazebo/gz-physics users.

This backports the relevant DART 7 memory-management pattern to the DART 6 LTS branch without requiring existing users or gz-physics to call a new API. Explicit preparation is available for C++ callers that want it; ordinary `World::step()` performs the same preparation lazily on the first step.

## Changes / Key Changes

- Add `FrameAllocator` and extend `MemoryManager` with frame scratch capacity and free-list growth options.
- Add World-owned `MemoryManager` configuration, `World::enterSimulationMode()`, `World::isInSimulationMode()`, and `World::getMemoryManager()`.
- Reuse `World::step` scratch vectors and route Dantzig / boxed-LCP solver scratch through retained buffers.
- Make the built-in text profiler avoid per-scope steady-state heap allocation.
- Add an allocation-counting integration harness covering `operator new`, raw malloc-family calls where available, and DART `MemoryAllocator` traffic.
- Add maintainer testing guidance for simulation allocation gates in `docs/onboarding/testing.md`.
- Retire the temporary allocation-hardening dev-task folder after promoting durable facts to changelog, API comments, and onboarding docs.

## Before / After

| Area | Before | After |
| --- | --- | --- |
| Native DART same-shape step allocations | Contact-heavy native-collision runs reported 5835.020 `operator new` calls and 5862.020 raw malloc-family calls per measured step. | Strict native gates report zero `operator new`, zero raw malloc-family, and zero World base-allocator growth after explicit preparation or the implicit first step. |
| Public preparation API | No public way to prepare World-owned simulation scratch ahead of a hot loop. | C++ callers may call `World::enterSimulationMode()`; existing callers can keep using `World::step()` unchanged. |
[...]

PR #3315, opened 2026-07-06
Fix FrameAllocator out-of-bounds fast path for tiny arena capacities

jslee02:
## Summary

- Fix an AddressSanitizer-confirmed heap-buffer-overflow in `dart::common::FrameAllocator` when the arena is constructed with a capacity smaller than the 64-byte arena alignment. Affects any caller that constructs a `FrameAllocator` with a tiny capacity; the default `World`/`MemoryManager` sizes are unaffected in practice, but the class is public.

## Motivation / Problem

`FrameAllocator`'s constructor rounds `mBegin` up to a 64-byte boundary but sets `mEnd = mBuffer + mCapacity`. When `capacity < 64`, the rounded `mBegin` can land past `mEnd`, leaving `mCur > mEnd`. The fast path then computes `remaining = mEnd - mCur` as an underflowed `size_t` (a huge value), skips the overflow fallback, and returns storage past the end of the buffer.

Reproduced with ASan: `FrameAllocator(GetDefault(), 16)` then `allocate(8)` and a write yields `heap-buffer-overflow ... located 32 bytes after 16-byte region`. `resetSlow()` had the same latent issue on the failed-grow path.

This is the DART 7 forward-port of the same fix already shipped on `release-6.20` in #3297.

## Changes / Key Changes

- Constructor: when the buffer is too small to hold a 64-byte-aligned arena, keep the fast-path window empty (`mCur == mEnd == mBuffer`) so every request is served from the overflow path instead of underflowing `remaining`.
- Constructor and `resetSlow()`: compare integer addresses before forming the aligned `char*`, so tiny-buffer detection does not rely on unspecified out-of-array pointer comparison.
- Regression test `FrameAllocatorTest.TinyCapacityStaysInBounds` (plus a reset-and-reuse case) with two independent signals:
  - writing across every returned block trips ASan on an out-of-bounds pointer;
  - the invariant `used() <= usableCapacity()` catches the underflow in **any** build configuration (the pre-fix fast path left `usableCapacity() == 0` while advancing `used()`), so the ASan-only lane is no longer the sole guard.

## Testing

- Reproduced the pre-fix overflow and confirmed the post-fix clean write under AddressSanitizer.
- Verified the new invariant fails against the pre-fix constructor and passes against the fix.
- Current head `fa95cb092d59`:
  - `pixi run lint`
  - `cmake --build build/default/cpp/Release --target UNIT_common_frame_allocator`
  - `ctest --test-dir build/default/cpp/Release --output-on-failure -R '^UNIT_common_frame_allocator$'`

## Breaking Changes

- [x] None

- release-6.20 already carries the fix via #3297. A companion test-only PR strengthens the same regression test on `release-6.20`.


jslee02:
@codex review

PR #3335, opened 2026-07-07
Fix MSVC composite data template ownership

jslee02:
## Summary
- explicitly instantiate the zero-aspect composite state/properties template bases in the core DLL on Windows
- declare matching extern templates so component DLLs such as dart-io do not emit duplicate MSVC symbols

## Testing
- pixi run check-lint-cpp
- pixi run cmake --build build/default/cpp/Release --target dart-io --parallel


jslee02:
@codex review

jslee02:
@codex review

PR #3342, opened 2026-07-08
Fix release Windows zero-duration profile summary

jslee02:
## Summary
- Treat recorded profiler scopes as present even when their measured inclusive time is zero
- Keep zero-duration scope rows visible in text summaries instead of reporting that no scoped regions were recorded
- Add a deterministic regression for zero-duration profile scopes

## CI failure addressed
- release-6.20 Windows Release job: https://github.com/dartsim/dart/actions/runs/28909805330/job/85764648992

## Validation
- pixi run config
- cmake --build build/default/cpp/Release --target UNIT_common_Profile --parallel 4
- ctest --test-dir build/default/cpp/Release -R UNIT_common_Profile --output-on-failure
- pixi run check-lint-cpp
- git diff --check

### DIFF
--- a/dart/common/detail/profiler.cpp
+++ b/dart/common/detail/profiler.cpp
@@ -343,6 +343,22 @@ void Profiler::recordCounter(
   counter->last = value;
 }
 
+void Profiler::recordScopeForTesting(
+    std::string_view label,
+    std::string_view file,
+    int line,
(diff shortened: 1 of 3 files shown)


jslee02:
@codex review

jslee02:
@codex review

PR #3366, opened 2026-07-10
Fix dartpy DegreeOfFreedom ownership in getDofs bindings

jslee02:
## Summary

- `Skeleton.getDofs`, `ReferentialSkeleton.getDofs`, `BodyNode.getChainDofs`, and `JacobianNode.getChainDofs` were bound without a return-value policy. pybind11's default policy takes ownership of returned raw pointers, so the Python wrappers deleted the joint-owned `DegreeOfFreedom` objects when collected, corrupting the heap. The typical symptom is a SIGSEGV at interpreter exit (or any later teardown) after contact-rich simulations that touched `getDofs()`.
- Fix: `::py::return_value_policy::reference_internal` on all four bindings, matching the existing (correct) `MetaSkeleton.getDofs` binding in the same tree.
- Adds regression coverage: an in-process test that collects the `getDofs()` wrappers and asserts the skeleton's DOFs survive, plus a subprocess reproducer (2000-step torque-driven `mjcf/openai/reacher.xml` run) that must exit cleanly. The subprocess test reproduced the crash deterministically (`rc=139`) before the fix and passes after.

## Reproduction

```python
import dartpy as dart
world = dart.utils.MjcfParser.readWorld('dart://sample/mjcf/openai/reacher.xml')
dofs = [d for i in range(world.getNumSkeletons()) for d in world.getSkeleton(i).getDofs()]
for _ in range(2000):
    world.step()
# segfaults during teardown before this fix
```

## Dual-PR note

`main` is not affected: its nanobind `meta_skeleton.cpp` binding already uses `nb::rv_policy::reference_internal` (python/dartpy/dynamics/meta_skeleton.cpp:98).

## Validation

- `pixi run build-py-dev` and `python -m pytest python/tests/unit/dynamics/test_degree_of_freedom_ownership.py` → 2 passed (fails with SIGSEGV subprocess exit before the fix).
- `pixi run lint` + `pixi run check-lint` clean.

### DIFF
--- a/python/dartpy/dynamics/BodyNode.cpp
+++ b/python/dartpy/dynamics/BodyNode.cpp
@@ -971,7 +971,8 @@ void BodyNode(py::module& m)
           +[](const dart::dynamics::BodyNode* self)
               -> const std::vector<const dart::dynamics::DegreeOfFreedom*> {
             return self->getChainDofs();
-          })
+          },
+          ::py::return_value_policy::reference_internal)
       .def(
           "addExtForce",
           +[](dart::dynamics::BodyNode* self, const Eigen::Vector3d& _force) {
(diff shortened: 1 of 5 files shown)


jslee02:
@codex review

PR #3371, opened 2026-07-11
Add active camera control, debug layers, and claim-tied PR evidence for agent visual verification

jslee02:
## Summary

- Agents (and humans) capturing headless evidence can now control the camera actively, measure whether a view is adequate before shipping it, render world-derived debug layers offscreen, select a minimal claim-tied artifact set, and generate a PR "Visual verification" section with GitHub-hosted media — closing the gaps a review of the merged visual-verification lineage (#2690, #2984, #2992, #3304, #3313, #3314, #3320) showed every prior PR shared: none embedded its visual evidence, cameras were fixed presets, debug overlays never reached the headless path, and artifacts were not tied to claims.
- **Core-first per maintainer directive**: every rendering/GUI capability lives in `dart::gui` C++ and is exposed through dartpy; the Python layer only orchestrates. Agent needs drove real engine capabilities: world-aware debug extraction (wiring flags that were dead since #2984), a shared projection primitive, headless debug-label rendering, view-adequacy assessment, and public rigid-body enumeration on `World`.

## Motivation / Problem

- The agent-verification stack (#3313/#3314/#3320) renders and judges images but has no notion of view adequacy: a capture can be cropped, too far, occluded, or ambiguous and still pass "non-blank".
- Debug producers from #2690/#2984 (contacts, joint axes, velocities) were viewer-only; `OffscreenRenderer` had no debug channel, nothing aggregated layers from a DART 7 `World` (`drawBodyFrames`/`drawCentersOfMass` were dead flags, labels ImGui-only, no trajectory layer), so headless captures could not show internal state.
- Evidence never reached reviewers: every lineage PR cites local `/tmp/*.png` paths ("attaching transient images requires the web PR editor" — #2984), and no tooling ranked artifacts, recorded why each was chosen, or generated the PR verification section.

## Changes / Key Changes

**Core (`dart::gui`, `dart::simulation`) — the capabilities**

[...]

PR #3378, opened 2026-07-11
Add interactive memory and layout diagnostics to consolidated demos

jslee02:
## Summary

- Replaces the earlier grouped logical-capacity grids with a default-open **2D host virtual-memory address map** for DART 7 Worlds. Each row is one real contiguous allocator backing allocation, bytes stay in relative-address order, and separate allocations remain visibly separate.
- Combines exact free-list/frame-allocator state with exact typed EnTT payload overlays where DART can prove a pointer and byte extent. Semantic colors distinguish model, geometry, state, control, solver/contact, cache, scratch, entity/index, metadata, and unknown live allocations; patterns independently distinguish allocated, free, reserved, padding, metadata, live, tombstone, and spare state.
- Keeps `World` as a thin public facade. The allocator traversal and ECS-aware collection remain in internal detail components beside opaque World storage; raw addresses, allocator headers, EnTT types, and live pointers do not cross the provider boundary.
- Establishes an accepted cross-branch design contract and evidence ladder for DART 7 and DART 6. Address geometry is presented as a locality diagnostic, not as measured cache behavior.

## Motivation / Problem

The first version of this PR answered “how much capacity is active, holed, or reserved?” It did **not** answer the requested question: “where are the bytes within the host allocations, and which simulation data types are adjacent?” Grouping all live slots first and all spare slots second discarded address order and could not reveal layout or locality opportunities.

The corrected view starts from what the current architecture can prove:

- DART 7 has multiple real allocations, not one fictional global slab.
- A free-list allocator can own multiple chunks; the frame allocator has a primary arena plus independent overflow allocations.
- EnTT component payloads are paged, and packed entity arrays are separate allocations.
- Virtual-address adjacency can motivate a cache-locality hypothesis, but it cannot establish access order, working set, cache misses, TLB behavior, physical adjacency, or speedup.

### Provider and ownership

```mermaid
flowchart LR
  MM[World MemoryManager] --> FL[Free-list backing chunks]
  MM --> FA[Frame arena and overflow allocations]
  FL --> P[Exact allocator-state partitions]
  FA --> P
  ECS[Opaque World ECS registry] --> T[Known component pages and packed entity buffers]
  T --> O[Exact typed payload overlays]
  P --> MAP[2D region-relative address map]
  O --> MAP
  RSS[Whole-process RSS] --> TABLE[Separate process metrics]
```

- Adds a friend-based, non-mutating internal inspector for free-list and frame allocators without changing allocator object layouts.
- Validates region bounds, pointer arithmetic, linked-list continuity, cycles, exact partition coverage, and overflow before publishing a region. Invalid metadata fails the affected capture closed.
- Reports:
  - every free-list backing chunk with allocator headers and allocated/free payload extents;
  - the frame arena with alignment padding, cursor-consumed backing, and reserved capacity;
  - every live frame-overflow backing allocation as a separate region;
  - exact known EnTT component-page runs and initialized packed entity-index runs when their extents are directly observable.
- Leaves uncorrelated allocated bytes explicitly `Unclassified`. It does not guess ownership from size or allocation order.
- Uses region-relative offsets only. Raw bases and inter-region address gaps are removed before public DTOs, dartpy, UI state, logs, or screenshots.

### Orthogonal map semantics

| Axis | Answers | Examples |
| --- | --- | --- |
| Allocator state | What the allocator proves about the bytes | metadata, allocated, free, reserved, padding |
| Logical use | How a typed allocated range is occupied | live, tombstone, spare, not applicable |
| Data category | What the payload means to simulation | model, geometry, state, control, contact/solver, cache, scratch, entity/index, metadata, unknown |
| Evidence | Why the range is drawn | allocator bookkeeping or exact typed payload overlay |

An ECS tombstone is therefore still allocated component-page memory; it is not rendered as an allocator-free hole. Unused frame-arena capacity is reserved scratch backing, not free-list space.

### Interactive diagnostics

- Makes **Address map (actual regions)** the first, default-open map section and moves the old grouped view under collapsed **Capacity composition**.
- Provides 32–4096 display cells per region with address order preserved at every zoom.
- Represents multiple exact ranges within one display cell as ordered proportional segments instead of a blended “mixed” color.
- Adds persistent text legends plus non-color patterns/borders for grayscale and color-vision-deficiency use.
- Adds a bounded exact relative-range table carrying label, offset, size, state, logical use, category, evidence, source, and limitation.
- Keeps map payload current-sample-only; it is not copied into the 180-sample history or baseline. The current slice therefore uses an explicitly snapshot-local address-order ordinal, not a persistent region ID.

### API and dartpy

- Extends `WorldMemoryDiagnosticsOptions` with opt-in exact memory-layout collection.
- Adds sanitized region/span DTOs and branch-neutral state/category/evidence/logical-use enums.
- Preserves the no-argument summary path and existing storage-layout option.
- Extends `PanelBuilder::blockGrid` with optional per-block patterns and ordered weighted segments after the existing arguments.
- Exposes equivalent dartpy enums/DTOs, `PanelBlockPattern`, owning `PanelBlockSegment`, stubs, validation, and headless rendering coverage.
- Normalizes segment weights by their largest finite positive value so even multiple `DBL_MAX` weights cannot produce `inf/inf` draw coordinates.

### Scrolling renderer hardening

The corruption reported while scrolling was in the Filament ImGui adapter, not in the memory-map model:

- the adapter flattened raw 16-bit `ImDrawIdx` values while ignoring each command's `VtxOffset`, so a dense scrolled panel wrapped indices after 65,535 vertices;
- it also discarded `ImDrawCmd::ClipRect`, allowing scrolled-off geometry to escape the panel viewport;
- capacity-driven buffer replacement could race in-flight uploads.

Every GUI build now advertises `ImGuiBackendFlags_RendererHasVtxOffset`, flatten commands into 32-bit indices as `vertexBase + VtxOffset + rawIndex`, preserve command index ranges, and submit one ordered Filament primitive/material instance per visible command with its own scissor. Vertex/index buffers grow monotonically, fence only before replacement, and keep asynchronous staging bytes alive through upload completion. DART's ImGui coordinates are already framebuffer-pixel coordinates, so scissors are not multiplied by `FramebufferScale` a second time.

Both defects are correctness bugs for any dense Dear ImGui panel, not only the memory one, so the fix is unconditional and the legacy 16-bit path is deleted rather than gated on a diagnostics build option. This deliberately retires the earlier claim that a default build executes the original overlay instruction stream: the default build now carries the fix, and a default-configuration headless capture was inspected to confirm ordinary panels, clipping, and glyphs still render correctly.

### Cost and disabled behavior

Two independent build options, each default **OFF**, with a one-way implication:
`DART_BUILD_DEMOS_MEMORY_DIAGNOSTICS` force-enables
`DART_BUILD_MEMORY_DIAGNOSTICS`, because the panel cannot report anything
without the library instrumentation. The reverse is deliberately allowed, so a
library user can instrument their own application without building the example
panel.

- **`DART_BUILD_MEMORY_DIAGNOSTICS=OFF` (default) is literally zero on the step
  path.** Following the `DART_BUILD_PROFILE` pattern, the option is exposed as
  `#cmakedefine01` in `dart/config.hpp` and guards both the `WorldStorage`
  member and every step-path call site. `World::resetFrameScratchForStep()`
  compiles down to the functional frame-allocator reset alone -- the reset tally
  and the `m_storage` load are gone, and it now tail-calls `resetSlow()`.
  `stepPipelineOnce()` contains no diagnostics reference at all, and `World`
  stores no diagnostics state.
- `World::getMemoryDiagnostics()` stays available in that configuration and
  still reports every value it can read live from the allocator and the
  registry; only `frameScratchPeakUsedBytes` and `frameScratchResetCount`, which
  require a per-step sample, report zero. This mirrors how
  `isStepProfilingEnabled()` behaves under `DART_BUILD_PROFILE=OFF`.
- With the option **ON**, the step path costs three cross-library allocator
  calls, down from five, after removing cached fields that `collect()`
  overwrote with live reads before any caller could observe them. The
  always-resident tracker is 16 B per World, versus 168 B for
  `WorldMemoryDiagnostics` on `origin/main`; the 208 B rich DTO and all
  ECS/region vectors are built only by an explicit query.
- The allocator debug snapshot is no longer refreshed on simulation step/reset
  paths. The only production `getDebugDiagnostics()` call under `dart/` is
  inside explicit `collect()`.
- `DART_BUILD_DEMOS_MEMORY_DIAGNOSTICS=OFF` keeps the panel, session/model
  sources, scene hooks, environment probe, and diagnostic strings out of
  `dart-demos` and its link graph. A post-link artifact test rejects any OFF
  executable containing diagnostics sources, symbols, or schema/UI markers.
- The enhanced ImGui adapter is no longer gated. It is a general renderer
  correctness fix and is present in every GUI build; see the note above.
- If the panel is compiled in, the runtime checkbox still returns before OS
  probes, allocator walks, ECS scans, formatting, history mutation, or map
  construction. The checkbox controls collection; the compile-time options
  control whether any of it exists.

## Before / After

| Surface | Before | After |
| --- | --- | --- |
| `World::getMemoryDiagnostics()` | No-argument summary only | No-argument call unchanged; new overload takes `WorldMemoryDiagnosticsOptions` for opt-in storage-layout and exact memory-layout detail |
| `WorldMemoryDiagnostics` | Allocator counters plus aggregate ECS counters | Adds sanitized `memoryRegions` (region-relative spans carrying state, category, evidence, and logical use) and per-storage layout detail; both built only on explicit request |
| Always-resident `World` state | `WorldMemoryDiagnostics` member, 168 B per World, unconditional | Nothing when `DART_BUILD_MEMORY_DIAGNOSTICS=OFF`; 16 B opaque `WorldStorage::MemoryDiagnosticsTracker` when ON. The 208 B rich DTO is constructed per query |
| Step and reset path | `getDebugDiagnostics()` refreshed on every step and every frame-scratch reset -- 53 instructions and 6 cross-library calls, twice per step | Nothing at all when the option is OFF (default); one 22-instruction, 3-call peak sample when ON. No allocator debug query on either path |
| `PanelBuilder::blockGrid` | `label`, `colors`, `tooltips`, `preferredColumns` | Same leading arguments; optional per-block `patterns` and ordered weighted `segments` appended |
| dartpy | `get_memory_diagnostics()` summary | Same, plus options, branch-neutral enums, sanitized DTOs, `PanelBlockPattern`, owning `PanelBlockSegment`, and matching stubs |
| Demos memory panel | Grouped logical-capacity grids only | Address-ordered region map is first and default-open; the grouped view moves under collapsed **Capacity composition** |
| Library build surface | n/a (instrumentation was unconditional) | New default-OFF `DART_BUILD_MEMORY_DIAGNOSTICS` gates all World instrumentation, in the style of `DART_BUILD_PROFILE` |
| ImGui overlay | 16-bit indices, `VtxOffset` ignored, `ClipRect` discarded | 32-bit indices with per-command `VtxOffset` and scissor, in **every** GUI build |
| `dart-demos` build surface | n/a (new) | `DART_BUILD_DEMOS_MEMORY_DIAGNOSTICS` defaults **OFF**; the panel, session, model, probes, and enhanced renderer path are absent from the OFF executable and its link graph |

## Design and research basis

The accepted contract is documented in `docs/design/memory_layout_diagnostics.md`. The design separates five non-interchangeable evidence levels:

1. actual contiguous virtual-memory region;
2. actual allocation/state range;
3. layout-derived locality;
4. measured access locality;
5. measured cache/TLB/latency behavior.

Primary sources that directly shaped the implementation:

- [Unity Memory Profiler memory map](https://docs.unity.cn/Packages/com.unity.memoryprofiler%400.2/manual/memory-map): preserve virtual-address order, expose adjustable row scale, filters, and range details.
- [AMD Radeon Memory Visualizer](https://gpuopen.com/manuals/rmv_manual/): keep heap, allocation, and typed-resource evidence distinct rather than collapsing everything into utilization.
- [EnTT ECS storage documentation](https://github.com/skypjack/entt/wiki/Entity-Component-System): component storage is paged and deletion policies can leave tombstones, so logical holes and allocator-free bytes are different facts.
- [MuJoCo simulation framework](https://mujoco.readthedocs.io/en/3.3.1/programming/simulation.html): explicit typed arrays, alignment, padding, and arenas are useful only when the owning runtime exposes their real ranges.
- [LLVM `BumpPtrAllocatorImpl`](https://llvm.org/docs/doxygen/classllvm_1_1BumpPtrAllocatorImpl.html): an arena can span slabs and separate large allocations; “arena” does not imply one contiguous address range.
- [Intel VTune Memory Access analysis](https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2024-0/memory-access-analysis.html) and [Linux `perf mem`](https://man7.org/linux/man-pages/man1/perf-mem.1.html): cache, TLB, latency, and data-source claims require separate sampled/hardware evidence.
- [WCAG 2.2 use of color](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color): state is also encoded with text, pattern, border, or shape rather than hue alone.

## Testing

Validated at head `35c2de12d3f`, which merges the EnTT 4.0 upgrade (#3414). That upgrade matters here because the collector reads EnTT storage internals directly -- `component_traits::page_size`, `storage->raw()` payload pages, the packed entity array, and `extent()`. The exact-partition, page-slot-order, and typed-page-range tests all pass against 4.0, so the address map is not silently mis-reporting under the new storage implementation. Both build configurations are exercised, because the
zero-overhead claim and the feature itself live in different ones.

An automated review round on `f67113b971c` found that the uninstrumented build
substituted the current frame-scratch usage for `frameScratchPeakUsedBytes`,
so a field documented as a maximum since construction could decrease after the
next reset. That is fixed: the uninstrumented path now reports the documented
zero, and the World test helper asserts that contract instead of the value the
code happened to produce.

**Default configuration (`DART_BUILD_MEMORY_DIAGNOSTICS=OFF`)**

- `pixi run test-all` — **passed**, 6/6 phases.
- `pixi run -e cuda test-all` — **passed**, 7/7 phases with the NVIDIA runtime
  detected, including 8/8 CUDA device tests.
- Step path is empty by inspection of the generated code, not by assertion.
  `World::resetFrameScratchForStep()` disassembles to the frame-allocator reset
  alone and tail-calls `resetSlow()`; the reset-tally increment and the
  `m_storage` load are absent. `stepPipelineOnce()` contains no diagnostics
  reference. For comparison, `origin/main`'s `refreshMemoryDiagnostics()` is 53
  instructions with 6 cross-library calls plus a stack canary, and ran **twice**
  per step.
- `World` memory tests pass against the inert contract (peak reports live usage,
  reset tally reports zero).
- Default-configuration headless capture inspected: ordinary demo panels,
  clipping, and text render correctly under the now-unconditional overlay.

**Diagnostics configuration (`DART_BUILD_DEMOS_MEMORY_DIAGNOSTICS=ON`)**

- The demos option force-enables `DART_BUILD_MEMORY_DIAGNOSTICS`; verified from
  a clean configure that sets only the demos override.
- Focused gate — **5/5 passed**, including the positive
  `EXAMPLE_dart_demos_memory_diagnostics_included` artifact test and
  `UNIT_dartsim_ui_MemoryActions`.
- The same 9 `World` memory/frame-scratch tests pass against the **instrumented**
  contract, so both contracts are covered rather than either being dropped.
- With the option on, the per-step sample is 22 instructions and 3 calls, down
  from 31 instructions and 5 calls, after removing cached fields that
  `collect()` overwrote with live reads.

**Both configurations**

- Full `test_world` binary — **432/432 passed**. CTest registers this suite with
  `DISABLED TRUE` (pre-existing on `origin/main` too), so it does not run in
  `test-all` or CI; it was run directly because it owns the frame-scratch peak
  and reset assertions.
- Focused AddressSanitizer run over the allocator-walking code —
  `UNIT_common_frame_allocator`, `UNIT_common_free_list_allocator`, and
  `test_memory_diagnostics` — **3/3 passed, zero findings**, with
  `detect_leaks`, `detect_stack_use_after_return`, `strict_string_checks`, and
  `check_initialization_order` enabled. ASAN moved off the PR CI tier in July
  2026, so nothing in the required gate otherwise exercises this pointer
  arithmetic under a sanitizer.
- Resident sizes measured with a compiled `sizeof` probe: tracker 16 B, rich DTO
  208 B, versus 168 B for `WorldMemoryDiagnostics` on `origin/main`.
- `git grep getDebugDiagnostics -- dart/` finds exactly one caller, inside
  `collect()`.
- Large-draw regression — 20,000 filled rectangles (>80,000 vertices), nonzero
  command-local `VtxOffset`, flattened maximum index above the 16-bit
  `ImDrawIdx` range while below `TotalVtxCount`.
- Clip/scissor regression — distinct command clip rectangles preserved, a fully
  offscreen command omitted, and `DisplayFramebufferScale=2` covered without
  double-scaling DART's pixel-coordinate scissors.
- `git diff --check` clean; `pixi run check-lint` clean with AI adapters in sync.

## Visual verification

**Environment**: Linux x86_64, Filament/OpenGL through llvmpipe, headless capture at 2000x1600. The two hosted images below were captured at Release head `b35f6d9e865`. The address map was re-captured at the current head `f67113b971c` and is functionally identical (same regions, legend, and populated grid; only the frame counter and process RSS differ run to run), so the hosted image remains representative. The post-scroll image is not re-runnable per head because its wheel injection is a manual step; its automated oracle is re-run instead.

### Current 2D address map

Current DART 7 address-ordered memory map

The free-list backing allocation is rendered as a true 2D byte-ordered grid with independent color/category and pattern/state encoding. It is populated in this capture; it is not the earlier empty-memory image.

Independently re-captured at head `f67113b971c` with the documented command below: 2000x1600, PPM SHA-256 `659bd1d8999be3e84656dc4978b548329a6bfd07942c2ebfb325f26712451b99`, automated verdict **PASS** for non-blank and contrast. Direct inspection confirms address-ordered cells, per-cell borders, the text legend, the pattern legend (`solid=allocated/live`, `cross=tombstone or metadata`, `dots=spare or reserved`, `/=free`, `\=padding`), ordered proportional sub-cell segments at range boundaries, and no raw address anywhere in the panel. The panel source confirms the section defaults the capture shows: address map `true` (open), capacity composition `false` (collapsed).

### Deeply scrolled renderer regression

Memory diagnostics after 30 wheel ticks and 4000 rendered frames

This second image is the critical regression evidence: it was captured after 30 discrete vertical wheel ticks inside the memory panel at head `b35f6d9e865`. Text, tables, section bars, and clipping remain structurally correct with no giant triangles or smeared index-wrap geometry. The wheel injection is a manual step, so this capture is not re-run per head; its automated oracle `UNIT_gui_ImGuiOverlayIndices` re-passed at `f67113b971c` in both the default-OFF and diagnostics-ON configurations.

### Claims and evidence

- **C1** — Each map row is one real contiguous World-owned backing allocation in relative address order. ✔
- **C2** — Allocator state and known simulation-data categories remain independently visible within the same cells. ✔
- **C3** — Multiple regions remain separate; no synthetic global slab or raw process address is exposed. ✔
- **C4** — Dense scrolling remains correct beyond the 16-bit draw-index range and respects command-local clipping. ✔
- **C5** — The visualization is layout evidence only; no cache-miss or performance-improvement claim is made. ✔

C1, C2, C3, and C5 were re-confirmed by direct inspection of the `f67113b971c` re-capture. C4 keeps the `b35f6d9e865` screenshot; its automated large-draw and clip/scissor regressions re-passed at `f67113b971c`.

| Artifact | Resolution | SHA-256 | Automated verdict |
| --- | ---: | --- | --- |
| Current address map | 2000x1600 | `14fab15d190a12e8a79063579a48025d919da4676ca062e8f3e0f8d026a2bffc` | [PASS, non-blank + contrast](https://github.com/dartsim/dart/releases/download/verification-media-memory-diagnostics/memory-map-main-b35f6d9e865-verdict.json) |
| Post-scroll readback | 2000x1600 | `9a3e6492c5a777a10c9809b09e29cd0d7f06e7b0402c60a002a6f47e360c884c` | [PASS, non-blank](https://github.com/dartsim/dart/releases/download/verification-media-memory-diagnostics/memory-scroll-main-b35f6d9e865-verdict.json) |
| Address map re-capture at `f67113b971c` | 2000x1600 | `659bd1d8999be3e84656dc4978b548329a6bfd07942c2ebfb325f26712451b99` (PPM) | PASS, non-blank + contrast (local `scripts/image_verdict.py`) |

**What the images do not prove**: allocator arithmetic, disabled-path overhead, cross-platform behavior, physical-page placement, access frequency, cache misses, TLB events, or speedup. Automated collector/model tests prove the arithmetic; compile-graph, binary, object-text, and hot-path audits prove the disabled behavior.

<details><summary>Reproduce</summary>

```bash
DART_BUILD_DEMOS_MEMORY_DIAGNOSTICS_OVERRIDE=ON pixi run config
cmake --build build/default/cpp/Release --target dart-demos

DISPLAY=:0 \
LIBGL_ALWAYS_SOFTWARE=1 \
DART_DEMOS_MEMORY_DIAGNOSTICS=1 \
build/default/cpp/Release/bin/dart-demos \
  --scene rigid_body \
  --frames 300 \
  --width 2000 \
  --height 1600 \
  --show-ui \
  --render-backend opengl \
  --screenshot /tmp/memory-map.ppm
```

For the scroll regression, use the same command with `--frames 4000`, keep the pointer over the memory panel, inject 30 downward wheel ticks, and capture the final frame to `/tmp/memory-scroll.ppm`.

</details>

## Breaking Changes

- [x] None.
- Existing summary APIs remain; the exact layout query and panel rendering options are additive and opt-in.

- DART 6 LTS counterpart: https://github.com/dartsim/dart/pull/3379
- Visual verification base merged before publication: https://github.com/dartsim/dart/pull/3371
- Related issue: none.

---

#### Checklist

- [x] Milestone set (DART 7.0)
- [x] CHANGELOG.md updated per `docs/onboarding/changelog.md`


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3374, opened 2026-07-11
Add active camera control, view quality, and debug-overlay evidence workflow for DART 6

jslee02:
## Summary

- Ports the DART 7 agent visual-evidence workflow (main PR #3371) to the DART 6 LTS branch — core-first per maintainer directive: debug overlays render through the engine's own OSG pipeline via a new `dart::gui::osg::DebugOverlay` viewer attachment (unlit, always-on-top lines + osgText labels), view adequacy uses core raycasts and the newly bound `math::BoundingBox`, and active camera control drives the WP-ASV `captureOffscreen` path. Scripts only orchestrate core APIs.

## Motivation / Problem

- #3314 gave DART 6 headless capture and image verdicts, but the camera is the fixed `defaultAgentCamera` pose: a capture can be cropped, too far, occluded, or ambiguous and still pass "non-blank". Nothing exposes internal state (contacts, frames, velocities) in captures, and nothing ties artifacts to claims or generates the PR verification section.
- Maintainer directive: agent visual capabilities must live in the core rendering/GUI codebase so agent needs drive engine capability and quality — not in parallel Python that papers over engine gaps. Two such gaps this PR fixes in core: `math::BoundingBox` was unbound in dartpy (Shape.getBoundingBox() returned an unusable opaque object), and DART 6 had no way to render always-on-top debug text/lines.

## Changes / Key Changes

- **New core GUI capability: `dart::gui::osg::DebugOverlay`** (`dart/gui/osg/DebugOverlay.{hpp,cpp}` + dartpy binding) — a ViewerAttachment (GridVisual-pattern peer) rendering debug line geometry (GL_LINES, per-vertex colors) and world-anchored osgText labels unlit, depth-test-off, in a late render bin — the DART 6 mirror of DART 7's core debug-overlay treatment, usable by any OSG consumer. `osgText` added to the OSG CMake components.
- **Core bindings**: `math::BoundingBox` registered in dartpy (getMin/getMax/setters/center/extents) — `Shape.getBoundingBox()` is now usable from Python.
[...]

PR #3379, opened 2026-07-11
Add opt-in memory diagnostics to dart-demos for DART 6

jslee02:
## Summary

- Adds the same corrected 2D memory-layout concept to DART 6 without pretending that the classic object graph lives in one arena.
- Shows every exact `World::MemoryManager` free-list/frame backing allocation as an address-ordered allocator map, and separately shows a typed sparse virtual-address atlas for traversed Skeleton, BodyNode, Joint, Shape, PointMass, degree-of-freedom, frame, and constraint objects.
- Preserves DART 6 ABI: allocator classes gain friend access only; no data member, virtual function, public World responsibility, or class layout changes.
- Keeps process RSS, allocator ownership, shallow object extents, layout-derived locality, and unavailable allocation/cache instrumentation visibly separate.

## Motivation / Problem

The original DART 6 panel showed useful process and graph counters, but the requested diagnostic is spatial: which real host-memory chunks exist, where allocator state changes within them, and whether related simulation data appears clustered or scattered.

DART 6 requires an honest two-view answer:

- its World `MemoryManager` owns real free-list and frame backing allocations, but the classic Skeleton/BodyNode/Joint/Shape graph is generally **not** allocated from them;
- the classic graph can therefore be shown only as observed address points or explicitly shallow `sizeof` lower bounds over sparse virtual-address page runs;
- gaps in that atlas are unobserved process address space, not free memory or fragmentation;
- virtual-address geometry is not cache-miss evidence.

### Two truthful providers

```mermaid
flowchart LR
  MM[World MemoryManager] --> FL[Free-list backing chunks]
  MM --> FR[Frame arena and overflow allocations]
  FL --> EXACT[Exact allocator region maps]
  FR --> EXACT
  WG[Classic World graph traversal] --> OBS[Typed points and proven shallow bounds]
  OBS --> ATLAS[Sparse host-page address atlas]
  EXACT --> UI[Memory tab]
  ATLAS --> UI
  RSS[Whole-process RSS] --> TABLE[Separate process metrics]
```

#### Exact allocator maps

- Adds detail-level, non-mutating snapshots of:
  - every free-list backing chunk, block header, allocated payload extent, and available payload extent;
  - frame backing alignment/tail padding, cursor-consumed backing, reserved-available backing, and live overflow allocations.
- Validates list cycles, global and cross-region `mPrev` invariants for both split and unsplit growth, pointer arithmetic, ordering, overlap, and full partition coverage.
- Distinguishes cursor-consumed `Allocated/Scratch` from unused `Reserved/None`; cursor-consumed bytes explicitly include alignment/cache-color/internal padding and are not presented as requested payload.
- Labels these as World reservation arenas. They do not claim ownership of the classic graph or current legacy solver temporaries.

#### Typed classic-object address atlas

- Captures exact object addresses at a safe traversal point and groups them into contiguous host virtual-page runs using the runtime page-size query.
- Emits:
  - point observations when the allocation extent is unknown;
  - shallow `sizeof` lower bounds only when runtime type exactly matches a known concrete type.
- Preserves overlapping/same-address observations independently. One mark can no longer clip or overwrite another.
- Renders page-run background as dotted **unobserved** space, never allocator-free memory.
- Keeps address count and adjacent virtual-address gap proxies available even if host page size is unavailable; only page grouping becomes unavailable.

### Renderer and interaction

- Address contiguity is explicit in both providers. Every row-group header carries a `contiguous backing bytes` (allocator maps) or `contiguous host-page run` (typed atlas) tag, and a full-width zigzag break separates consecutive row groups -- labeled `address discontinuity` between allocator regions (a separate backing allocation) and `unobserved address gap` between atlas page runs, because those are different facts. The zigzag shape is distinct from the in-cell hatch and dot patterns, which encode storage state within one contiguous group, and the legends define the encoding.
- Adds adjustable 4–48 detail rows per region while preserving address order.
- Uses hue for semantic category and independent patterns/borders/opacity for storage/evidence state.
- Uses ordered per-cell evidence lanes for every intersecting partition and typed observation.
- Builds cells in `O(cells + intersections)` using address-ordered source emission and a linear merge; a 512-overlap fixture covers dense composition.
- Uses responsive one-column legends in the real 360 px DART 6 scene panel, wrapped scale/evidence guidance, lighter patterns, and capped point markers.
- Adds a bounded textual relative-range table with horizontal scrolling, practical fixed widths, a frozen region column, and explicit source/limitation text.

### Safety, cost, and compatibility

- `DART_BUILD_DEMOS_MEMORY_DIAGNOSTICS` is default **OFF**. In that configuration the diagnostics sources, `DemoHost` state/calls/tab, environment probe, diagnostic markers, and Windows `Psapi` dependency are absent from the application compile/link graph.
- The OFF post-link checker rejects diagnostic symbols, strings, source edges, frame-path calls, or a fourth inspector tab. Independently compiled current and `origin/release-6.20` OFF `DemoHost` `.text` sections are byte-identical (SHA-256 `082cb7cdfbc27bfbea1a38fbe2485c972183d907882b6834856fbf9219469d62`). This is the strict zero-runtime-overhead mode.
- When the feature is compiled in, the runtime checkbox returns before OS probes, World traversal, allocator inspection, sorting, formatting, or history mutation. That checkbox disables collection; the default compile-time option is what removes every frame-path instruction and link dependency.
- Raw process addresses are used only transiently to sort/correlate a capture; published models and UI contain region/page-run-relative offsets only.
- No allocation hook or global `new`/`malloc` interposition is installed; active allocation count and DART-owned heap bytes remain explicitly unavailable.
- Renderer work is bounded for the legacy OpenGL2/16-bit ImGui backend: per-region eight-row child viewports cull invisible cells, the address map stops below a 56,000-vertex ceiling, and hovered details stop after 32 entries or 40,000 tooltip vertices.
- Public DART 6 class layouts and vtables are unchanged. The implementation is example/detail-local and header-only at the allocator inspection boundary.

## Design and research basis

This branch implements the DART 6 half of the cross-branch contract documented by the DART 7 counterpart in `docs/design/memory_layout_diagnostics.md`.

Evidence remains layered:

1. actual contiguous virtual-memory region;
2. actual allocator state or known pointer-plus-extent range;
3. layout-derived virtual-address locality;
4. measured access locality;
5. measured cache/TLB/latency behavior.

Primary-source influences:

- [Unity Memory Profiler memory map](https://docs.unity.cn/Packages/com.unity.memoryprofiler%400.2/manual/memory-map): retain virtual-address order, adjustable scale, discontinuities, and range details.
- [AMD Radeon Memory Visualizer](https://gpuopen.com/manuals/rmv_manual/): distinguish heaps/allocations/resources and preserve aliases/overlaps instead of reducing them to one utilization number.
- [MuJoCo simulation framework](https://mujoco.readthedocs.io/en/3.3.1/programming/simulation.html): typed contiguous layouts are meaningful when the owning runtime exposes their real arrays, alignment, and padding.
- [LLVM `BumpPtrAllocatorImpl`](https://llvm.org/docs/doxygen/classllvm_1_1BumpPtrAllocatorImpl.html): arenas can own multiple slabs and independent large allocations.
- [Intel VTune Memory Access analysis](https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2024-0/memory-access-analysis.html) and [Linux `perf mem`](https://man7.org/linux/man-pages/man1/perf-mem.1.html): cache, TLB, latency, NUMA, and access-hotness claims require separate sampling/hardware evidence.
- [WCAG 2.2 use of color](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color): patterns, borders, text, and point/range geometry accompany color.

## Testing

Final regression head: `d7a439e6211`. It makes address contiguity explicit in both map providers (see Changes) and adds a hidden `--debug-scroll-memory-panel <ticks>` verification hook (following the existing undocumented debug flags): `runHeadlessShot()` hovers the scene panel and injects one wheel-down tick per frame through the viewer event queue -- the same `osgGA` -> `ImGuiHandler` path as interactive input -- and captures one frame per tick beside the `--shot` image, whose final state is therefore a deep-scrolled readback. The hook compiles out with the diagnostics option: the default-OFF binary rejects the flag as an unknown option, and the OFF `DemoHost.cpp` `.text` section remains byte-identical to `origin/release-6.20` **with the harness in the source** (SHA-256 `082cb7cdfbc27bfbea1a38fbe2485c972183d907882b6834856fbf9219469d62`, 70,005 bytes).

- Headless scroll regression at `d7a439e6211` -- **passed**: 120/120 tick frames captured at 1600x1000; the motion proof found **zero** identical consecutive frames (sampled per-pair diffs min/median/max 98/185/331), so every tick visibly scrolled the panel. The first, middle, and last tick frames and the final post-scroll shot all pass the automated verdicts -- including the contrast gate on the mid-scroll frame, which the zigzag separators now supply with bright structure.
- `pixi run lint` and the full `pixi run test-all` re-passed at `d7a439e6211`: 155/155 C++ tests and 271 Python tests.
- The previous head `0e7d2cc15e0` merged current `release-6.20` (11 commits, including the volumetric FEM removal, the collision-detection
consolidation, and two new deformable demo scenes).

- Restored the unconditional `<cstring>` include in `DemoHost.cpp`. `runHeadlessShot()` calls `std::strerror(errno)` on a path that is not guarded by `DART_BUILD_DEMOS_MEMORY_DIAGNOSTICS`, but this PR had moved the include inside that guard; `release-6.20` has it unconditional. A default OFF build therefore compiled only where `<cstring>` arrived transitively, which conforming C++17 does not guarantee.
- `pixi run test-all` at the merged head — **passed**: 155/155 C++ tests and 271 Python tests.
- Diagnostics-ON focused gate — **2/2 passed** (`UNIT_common_AllocatorMemoryLayout`, `UNIT_examples_demos_MemoryDiagnostics`).
- Default-OFF build re-verified, including the post-link compile-out check.

- `pixi run lint` — **passed** after the final source changes.
- Default-OFF build and artifact audit — **passed**: the app has no diagnostics source/library edge, symbol, schema/UI/environment marker, diagnostics-only dependency, frame-path call, or extra Memory tab.
- OFF `DemoHost` instruction proof — current and `origin/release-6.20` `.text` are byte-identical with SHA-256 `082cb7cdfbc27bfbea1a38fbe2485c972183d907882b6834856fbf9219469d62`.
- Focused C++ — **2/2 passed**: `UNIT_common_AllocatorMemoryLayout` and `UNIT_examples_demos_MemoryDiagnostics`.
- Dense renderer regression — 512 overlapping observations exercise more than 32,000 vertices while remaining below both the enforced 60,000-test ceiling and the 16-bit `ImDrawIdx` maximum; the hovered tooltip path is included.
- Standalone package compatibility — **passed** against an installed DART `v6.19.3` SDK (capability disabled) and an installed feature-capable SDK (capability enabled), using package-matched ImGui headers.
- Windows CMake signature probe — **passed**: diagnostics-enabled targets retain private `Psapi`, while OFF executables retain their original link signature.
- Broad validation before the final compile-out hardening passed all 153 C++ tests, the complete Python suite, and `pixi run -e gazebo test-gz` (199 gz-physics tests plus gz-sim `INTEGRATION_entity_system`). Hosted checks for the exact final head are now running.
- C++/CMake formatting, codespell, `git diff --check`, and two independent final reviews passed; ABI review confirmed friend declarations only.

## Visual verification

**Environment**: Linux x86_64, OpenSceneGraph/ImGui with software rendering, exact 1600x2400 headless capture. The map image below was captured at the final head `d7a439e6211` (SHA-256 `b3dfc9deeaf06831304701916b629aa0e2627e7d8b45c268058839d72623fdb2`, automated verdict **PASS** for non-blank and contrast). Earlier captures at `f6cb11e0d64` and `0e7d2cc15e0` remain on the release for history.

Current DART 6 exact allocator maps and typed address atlas

The right-side Memory tab shows three separate exact World allocator regions followed by the independently labeled typed classic-object address atlas.

Read the two providers differently. The allocator regions are mostly free in this scene, and that is the truthful result rather than an empty map: DART 6's classic dynamics graph does not allocate from the World `MemoryManager`, so a preallocated 4.00 MiB free-list chunk legitimately reports almost entirely free bytes. The populated evidence for classic objects is the typed atlas below it. That split is the whole reason this PR ships two providers instead of one. Contiguity is now explicit at a glance: headers assert the contiguous extent of each row group, and labeled zigzag breaks mark every address discontinuity between allocator regions and every unobserved gap between atlas page runs, so a contiguous layout and a fragmented one cannot be confused.

### Scrolling video (renderer integrity under scroll)

The panel-cropped animation plays inline; every frame is one injected wheel
tick. The zigzag discontinuity breaks and `contiguous backing bytes` headers
stay correct throughout the traversal.

Scrolling the Memory tab: 120 wheel ticks across all allocator regions into the typed atlas

Full-resolution video:
[**memory-scroll-release620-d7a439e6211.mp4**](https://github.com/dartsim/dart/releases/download/verification-media-memory-diagnostics/memory-scroll-release620-d7a439e6211.mp4)
-- 1600x1000, 30 fps, exactly 120 frames (SHA-256
`5a3d58d24e11ecb3459a95caddcae349b84500bcff204760e26b2cd939e2759b`), assembled
losslessly from the per-tick headless captures with ffmpeg.

The traversal starts at the Memory tab header, crosses all three exact
allocator regions (free-list backing regions 1 and 2, then the frame arena)
with a labeled address-discontinuity break at each boundary, and reaches the
typed classic-object atlas. Because the injected pointer stays over the panel,
the deep-scroll frames also exercise the live hover path: a cell tooltip
renders correctly at maximum scroll depth. Text, headers, grids, and clipping
stay structurally intact in every frame; there are no index-wrap triangles,
smeared geometry, or dropped glyphs. First/last tick frames and the
post-scroll readback are hosted beside the video:
[first](https://github.com/dartsim/dart/releases/download/verification-media-memory-diagnostics/memory-scroll-first-release620-d7a439e6211.png),
[last](https://github.com/dartsim/dart/releases/download/verification-media-memory-diagnostics/memory-scroll-last-release620-d7a439e6211.png),
[post-scroll shot](https://github.com/dartsim/dart/releases/download/verification-media-memory-diagnostics/memory-postscroll-release620-d7a439e6211.png)
([verdict](https://github.com/dartsim/dart/releases/download/verification-media-memory-diagnostics/memory-postscroll-release620-d7a439e6211-verdict.json)).

### Claims and evidence

- **C1** — Every allocator row is one real contiguous `World::MemoryManager` backing allocation with exact relative-byte partitioning. ✔
- **C2** — The typed classic-object atlas is visually and textually separate from allocator ownership. ✔
- **C3** — Multiple same-cell observations remain visible through ordered evidence lanes; categories and point/range evidence are distinguishable. ✔
- **C4** — Gaps are unobserved virtual-address space, not allocator-free memory. ✔
- **C5** — Map and tooltip tessellation stay below the legacy 16-bit renderer budget. ✔
- **C6** — 120 real wheel events scrolled through every region and the atlas without corrupting the renderer, and the hover tooltip stays correct at deep scroll. ✔ (video above)
- **C7** — Contiguity is explicit: row-group headers assert the contiguous extent, and labeled zigzag breaks mark address discontinuities between allocator regions and unobserved gaps between atlas page runs; the video shows the breaks during traversal. ✔

| Artifact | Resolution | SHA-256 | Automated verdict |
| --- | ---: | --- | --- |
| Current full GUI/map | 1600x2400 | `072702787847cc48e86b3be387624325ccb60963c8811fc7f9dceafe7ba1e04e` | [PASS, non-blank + contrast](https://github.com/dartsim/dart/releases/download/verification-media-memory-diagnostics/memory-map-release620-f6cb11e0d64-verdict.json) |

**What the image does not prove**: physical-page adjacency, complete classic-object allocation extents, heap ownership of atlas observations, access frequency, cache misses, speedup, cross-platform accounting, or disabled-path overhead. Automated range/model fixtures prove the map invariants; the compile-graph, binary, and `.text` comparisons prove the default-OFF behavior.

<details><summary>Reproduce</summary>

```bash
pixi run cmake -S . -B build/default/cpp/Release \
  -DDART_BUILD_DEMOS_MEMORY_DIAGNOSTICS=ON
cmake --build build/default/cpp/Release --target dart-demos

DISPLAY=:0 \
LIBGL_ALWAYS_SOFTWARE=1 \
DART_DEMOS_MEMORY_DIAGNOSTICS=1 \
build/default/cpp/Release/bin/dart-demos \
  --scene soft_bodies \
  --headless \
  --shot /tmp/memory-map-release620.png \
  --steps 5 \
  --width 1600 \
  --height 2400 \
  --gui-scale 1.0

# Scroll regression + video (diagnostics builds only; the flag does not
# exist in default-OFF builds):
DISPLAY=:0 LIBGL_ALWAYS_SOFTWARE=1 DART_DEMOS_MEMORY_DIAGNOSTICS=1 \
build/default/cpp/Release/bin/dart-demos \
  --scene soft_bodies --headless --shot /tmp/mem.png \
  --steps 5 --width 1600 --height 1000 --gui-scale 1.0 \
  --debug-scroll-memory-panel 120
pixi exec -s ffmpeg ffmpeg -framerate 30 -i /tmp/mem.png.scroll%04d.png \
  -pix_fmt yuv420p -crf 20 /tmp/memory-scroll.mp4
```

</details>

## Breaking Changes

- [x] None.
- No DART 6 public layout, vtable, or World API change.

- DART 7 counterpart: https://github.com/dartsim/dart/pull/3378
- Visual verification base merged before publication: https://github.com/dartsim/dart/pull/3374
- Related issue: none.

---

#### Checklist

- [x] CHANGELOG.md updated per `docs/onboarding/changelog.md`
- [x] Add Python bindings (N/A — example/detail-only DART 6 implementation with no new public API)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3369, opened 2026-07-10
MJCF parser fidelity: stacked joint compositions, contype/conaffinity filtering, and geom friction

jslee02:
## Summary

One cohesive MJCF parser-fidelity PR (per the fewer-PRs direction), in three commits plus follow-ups:

### 1. Stacked hinge/slide joint compositions
MJCF bodies listing multiple `<joint>` elements beyond the existing special cases failed with *unsupported joint composition*, rejecting `humanoid.xml`, `humanoidstandup.xml`, `walker2d.xml`, `hopper.xml`, `half_cheetah.xml`, `swimmer.xml`. Remaining hinge/slide stacks are now translated into chains of single-DOF joints through massless intermediate BodyNodes; per the MuJoCo XML reference each DART joint factor `T_CBJ · Q(q) · T_CBJ⁻¹` is exactly MuJoCo's local-frame screw about the undisplaced anchor, so the chained product reproduces MuJoCo's composition for arbitrary q. Existing compositions match first and stay byte-identical.

### 2. Parser hardening (review follow-ups + a fixture-exposed crash)
- Unnamed joints on root bodies no longer dereference a null parent BodyNode.
- Unnamed root bodies keep inheriting a generated BodyNode name (skipping synthetic `__mjcf_dof` links).
- **Release builds segfaulted on any MJCF file without a `<default>` element** (`Defaults::getRootDefault()` returned null; the `DART_ASSERT` compiles out). It now falls back to MuJoCo's built-in attribute defaults.

### 3. contype/conaffinity collision filtering + geom friction
`contype`/`conaffinity` and `friction` were parsed (with correct defaults and `<default>` inheritance) but never consumed:
- Geoms with `contype==0 && conaffinity==0` get visual-only ShapeNodes (no collision aspect) — e.g. striker's goal-root marker and its density-1e6 coaster cylinder, whose spurious table contact previously produced a ~1e6 mass-ratio LCP that made the Dantzig primary solver fail (and fall back to PGS) **every step**.
- The full MuJoCo pair rule `(ctA & caB) || (ctB & caA)` is enforced by a new `detail::GeomCollisionFilter` that subclasses `BodyNodeCollisionFilter` — deliberately, because `ConstraintSolver::updateConstraints` `dynamic_cast`s the installed filter to `BodyNodeCollisionFilter` for the resting-contact optimization; a generic composite would silently defeat it. Base filtering (self-collision/adjacency/resting pairs) is applied first; unregistered ShapeFrames are unaffected.
[...]

PR #3384, opened 2026-07-12
Guard non-finite LCP solutions and format agent scripts (release-6.20 CI)

jslee02:
Fixes the three red `release-6.20` CI jobs on the current HEAD (#3374):
**coverage**, **Asserts enabled (no -DNDEBUG)**, and **Check format**.

## 1. Non-finite LCP solutions (coverage + Asserts enabled)

Both asserts-enabled builds aborted in `test_MjcfParser` (the `Striker`
subtest, whose stepping loop was added in #3369):

```
test_MjcfParser: dart/dynamics/BodyNode.cpp:2283:
  BodyNode::addConstraintImpulse(...): Assertion `!math::isNan(_constImp)' failed.
```

**Root cause (verified with a local diagnostic):** a near-singular contact —
striker.xml's thin, hard, frictionless 1 mm goal walls, effective mass ~0 along
the normal — makes the primary Dantzig solver report success while emitting an
*infinite* normal impulse (confirmed: every step logged `success=1 anyInf=1`
for a single-row `n=1` LCP). `BoxedLcpConstraintSolver` only rejected NaN
entries, so the infinite value propagated into the spatial contact Jacobian,
where `inf * 0` in the moment-arm rows produced NaN body impulses. Release
builds were unaffected only because `NDEBUG` compiled the assertion out while
the NaN silently corrupted the free ball's state.

**Fix:** treat both NaN and infinite solutions as non-finite failures at each
solve site (velocity and position phases). The primary result is rejected so
the secondary (PGS) fallback / fail-safe zeroing runs, and the scrub loop now
zeroes infinite as well as NaN entries — guaranteeing finite constraint
impulses.

## 2. black formatting (Check format)

`scripts/agent_capture.py`, `scripts/agent_view_quality.py`, and
`scripts/evidence_select.py` (added by #3374) were unformatted and failed the
`black --check` gate in the Publish dartpy "Check format" job. Reformatted with
the pinned `black`.

## Verification (local, asserts-enabled `BUILD_TYPE=None`)

- `test_MjcfParser`: **22/22 passed** (was aborting on `Striker`).
- `test_ConstraintSolver`, `test_ContactConstraint`: **passed** (no regression
  from the solver change).
- `black --check` clean on the three scripts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -297,6 +297,15 @@
 
 * Dynamics
 
+  * Fix `BoxedLcpConstraintSolver` propagating non-finite constraint impulses:
+    a near-singular contact (e.g. a thin, hard, frictionless geometry) could
+    make the primary solver report success while emitting an *infinite* impulse
+    that slipped past the NaN-only sanity check and produced NaN body impulses
+    (`inf * 0` in the spatial Jacobian moment-arm rows). Both NaN and infinite
+    LCP solutions are now rejected so the secondary/fail-safe zeroing path runs,
+    guaranteeing finite impulses:
+    [#3384](https://github.com/dartsim/dart/pull/3384)
+
   * Add a default-off `BoxedLcpConstraintSolver` matrix-free contact solver
     option, plus dartpy bindings and `contact_benchmark` flags, for explicitly
     benchmarking large supported contact-only islands without changing the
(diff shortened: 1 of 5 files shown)


jslee02:
@codex review

jslee02:
@codex review

PR #3399, opened 2026-07-23
fix: load SDF `<soft_shape>` links as SoftBodyNode

jslee02:
## Summary

`dart::utils::SdfParser` never created `SoftBodyNode`s: its internal
`createJointAndNodePair<NodeType>` template called
`skeleton->createJointAndBodyNodePair<JointType>(...)` with only the *joint*
template argument, so the body-node type defaulted to rigid
`dynamics::BodyNode` and the `SoftBodyNode::Properties` argument was sliced to
`BodyNode::Properties`. **Every SDF `<soft_shape>` link has therefore loaded
as a rigid body.**

## Fix

Forward `NodeType` as the second template argument at all seven joint sites
(`createJointAndBodyNodePair<JointType, NodeType>`). A `<soft_shape>` link
(NodeType == `SoftBodyNode`) now creates a real `SoftBodyNode`; rigid links are
unchanged because `NodeType == BodyNode` reproduces the prior default. This is an
internal `.cpp` template change only — no public header, signature, or ABI
change.

## Test

Adds `SdfParser.SoftShapeLinkLoadsAsSoftBodyNode`: the soft-feet Atlas asset now
reports two `SoftBodyNode`s (`l_foot`/`r_foot`), while its rigid variant reports
zero. The regression fails before the fix (0 soft bodies) and passes after.

## Impact

The only in-tree SDF asset with `<soft_shape>` is
`atlas_v3_no_head_soft_feet.sdf`; `.skel` soft bodies use `SkelParser` and are
unaffected. Local `test_SdfParser` passes 7/7 and `test_ConstraintSolver`
passes 55/55 with the fix.

## Related

The same defect exists on `main`; a matching dual-PR follows per the dual-PR
policy.

## Breaking Changes

- [x] None (behavior correction for a previously-broken path; no API/ABI change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -311,6 +311,13 @@
 
 * Dynamics
 
+  * Fix `dart::utils::SdfParser` loading every SDF `<soft_shape>` link as a
+    rigid `BodyNode`: the parser forwarded only the joint type to
+    `createJointAndBodyNodePair`, so the body-node type defaulted to `BodyNode`
+    and `SoftBodyNode::Properties` was sliced away. Soft SDF links now load as
+    `SoftBodyNode`s; rigid links are unchanged and there is no API/ABI change:
+    [#3399](https://github.com/dartsim/dart/pull/3399)
+
   * Fix `BoxedLcpConstraintSolver` propagating non-finite constraint impulses:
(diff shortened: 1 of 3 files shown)


jslee02:
@codex review

PR #3382, opened 2026-07-12
Improve DART 6 deformable-body performance and soft contact

jslee02:
## Summary

- Add opt-in adaptive soft-contact activation and opt-in soft face-interior
  contact coverage for DART 6.
- Extend deformable-body correctness, stability, determinism, allocation,
  benchmark, and demo coverage.
- Retain measured performance improvements while preserving existing public
  layouts and default behavior.

This is a representative DART 6 release slice, not a claim of complete paper
parity, successful multicore scaling, or a native-default deformable backend.

## Motivation / Problem

DART 6 `SoftBodyNode` had a useful point-mass soft-body model, but lacked
several release-quality pieces:

- contact-local adaptive activation modeled after Jain and Liu;
- coverage for primitives intersecting a soft-face interior without enclosing
  a vertex;
- representative equation, energy, force/CoP, determinism, and allocation
  gates;
- reproducible headless performance tooling and flagship demonstrations; and
- durable evidence separating completed work from paper-scale, scaling, and
  native-default follow-ups.

The implementation must preserve DART 6 API/ABI compatibility. New behavior
therefore remains opt-in and implementation state stays out of public class
layouts.

## Before / After

| Area | Before | After |
| --- | --- | --- |
| Soft-contact activation | Every soft point mass remained active. | Users can opt into contact-local activation with ring, linger, and tolerance controls; default-off behavior retains the all-active path. |
| Soft face-interior contact | A primitive could intersect a soft face without enclosing a vertex and miss the contact. | Users can opt into DART-owned, BVH-pruned sphere/box face-interior coverage. |
| Representative demos | There were no adaptive-contact or soft-worm flagship scenes. | `adaptive_soft_contact` and `soft_worm` are integrated into `dart-demos`; no new standalone GUI executables remain. |
| Release evidence | Performance, correctness, and deferred parity claims were not collected in one release packet. | #3382 carries a tested representative slice while keeping failed winner, scaling, native-default, and paper-parity gates explicit. |

## Changes / Key Changes

- **Adaptive soft-contact activation (WP-DB.05, opt-in)**
  - Adds non-virtual `SoftBodyNode` controls for contact seeding, ring
    expansion, linger, and rest tolerances.
  - Treats inactive points as rigidly lumped mass and inertia on their parent.
  - Preserves the previous all-active path when disabled.

- **Soft face-interior contacts (WP-DB.08, opt-in)**
  - Adds BVH-pruned sphere/box feature tests for contacts occurring between
    soft vertices.
[...]

PR #3404, opened 2026-07-30
Add elastic element forces to the volumetric FEM body

jslee02:
## Summary

The volumetric FEM body added earlier carried a tetrahedral mesh, lumped masses,
and an integration pipeline, but no constitutive model — so it fell as a rigid
cloud of points. This adds the element forces that make it deform.

## Changes

- **`ElasticModel`**: small-strain `Linear` and co-rotated `Corotational`,
  selectable per body, defaulting to co-rotated.
- Per-element deformation gradient `F = Ds · Dm⁻¹`, using the inverse rest shape
  matrix the mesh already cached for this purpose.
- First Piola-Kirchhoff stress for both models and nodal force assembly
  (`H = -volume · P · Dm⁻ᵀ`), with node zero's force following from the element
  exerting no net force on itself.
- Polar decomposition via 3x3 SVD, including the reflection fix that maps an
  inverted element to the nearest true rotation.
- Elastic potential energy, node pinning (`setNodeFixed`) so a body can be
  anchored without a coupling constraint, and validation of Young's modulus and
  Poisson's ratio now that both are consumed.

Integration keeps the mass-proportional implicit damping and adds the elastic
acceleration, so all fourteen earlier gates hold **unchanged**: a freely falling
or uniformly translating body keeps `F == I` and feels no elastic force.

## Tests — 21 gates, 7 new

The gate that justifies the co-rotated model puts both side by side under a pure
rigid rotation: co-rotated peak nodal force and stored energy are both below
`1e-8`, while small-strain elasticity manufactures both above `1.0` from a
motion that did not deform the body at all.

Also: rest has no force or energy; rigid translation produces none; internal
forces sum to below `1e-9` of their peak; elastic energy is **exactly** quadratic
in strain and **exactly** linear in stiffness (ratios 4.0 and 2.0 to `1e-9`); a
stretched body relaxes without gaining energy; and a pinned body hangs, sagging
further when softer.

Full suite: **155/155**, no regressions.

## Known limitation

Elastic forces are integrated explicitly, matching the reference papers'
integration choice, which carries a CFL-style bound of roughly
`dt < h·sqrt(rho/E)`. The defaults are comfortably stable on the test meshes,
but a stiffer material or finer mesh needs a smaller step and **nothing warns
when the bound is exceeded**. Documented rather than guarded; damping remains
unconditionally stable. The per-element SVD is the dominant elastic cost and is
what selective diagonalization targets next.

## Compatibility

- No pre-existing runtime source file is touched, and no pre-existing object
  references `dart::dynamics::fem::` — rigid-body simulation runs identical
  instructions, with **zero** added per-step cost.
- The subsystem stays internal-only: no installed header, not in the generated
  `dynamics.hpp` aggregate, excluded from Doxygen. Public API and ABI unchanged.
- gz-physics and gz-sim gates pass on this base (199/199 functional, 4/4
  performance, 1/1 gz-sim integration).

Evidence: `docs/dev_tasks/dart6_deformable_body_performance/14-fem-elastic-forces.md`.

## Breaking Changes

- [x] None

## Visual verification

The change is visually verifiable, so here it is in motion. Same scene, same
camera, same seed; only simulated time advances.

The new `fem_cantilever` demo scene puts two identical tetrahedral boxes side by
side, both pinned along their left face and left to hang. The **near (orange)**
body carries a stiffness so low that elastic forces are negligible, which is how
the body behaved before this PR — nothing couples the nodes. The **far (blue)**
body uses co-rotated elasticity. Pinned nodes are drawn dark.

Animation: both boxes start identical; the uncoupled body disintegrates and falls away from its anchors while the co-rotated body holds together and settles into a sagging cantilever

*18 frames, t = 0 to 0.48 s. Animated PNG — GitHub plays it inline.*

Without a constitutive model the unpinned nodes are uncoupled: they separate
from the pinned face and fall away, leaving the dark anchors behind holding
nothing. With co-rotated elasticity the body stays a body — it holds its shape,
transfers load to the anchors, and settles.

| At t = 0.22 s | max displacement | stored elastic energy |
| --- | ---: | ---: |
| Negligible stiffness (near, orange) | **0.165 m** drop, detached | 1.7e-03 J |
| Co-rotated elasticity (far, blue) | **0.095 m** sag, intact | 2.81 J |

Side-by-side still ([full size](https://github.com/dartsim/dart/releases/download/verification-media-dart6-agent-evidence/fem_elastic_before_after.png)), which also
shows the on-screen readouts:

Before and after: identical start at t=0, then the uncoupled body falls away while the co-rotated body sags and holds

Individual frames:
[t = 0.00 s](https://github.com/dartsim/dart/releases/download/verification-media-dart6-agent-evidence/fem_cantilever_t000.png) ·
[t = 0.22 s](https://github.com/dartsim/dart/releases/download/verification-media-dart6-agent-evidence/fem_cantilever_t220.png)

Reproduce the frames:

```bash
pixi run demos --scene fem_cantilever            # interactive
for i in $(seq 0 17); do                         # frame sequence
  ./build/default/cpp/Release/bin/dart-demos --scene fem_cantilever \
    --headless --steps $((i*28)) --shot f$(printf %02d $i).png
done
python3 scripts/image_compose.py side-by-side \
  --labels "t=0.00s identical start" "t=0.22s after elastic forces" \
  --out fem_elastic_before_after.png f00.png f08.png
```

The stills pass `scripts/image_verdict.py` (the composite with no failing
checks). Note a tooling gap: `--headless` writes one PNG per invocation and the
repo has no animation assembler, so the APNG above was built from those frames
with a small stdlib script. A reusable equivalent under `scripts/` would be
worth adding separately. The imagery corroborates the numeric gates; it is not
the correctness oracle, which remains the 21 tests.

## Review follow-up

**Non-finite material parameters.** Validation accepted a non-finite Young's
modulus, because `!(value > 0.0)` is false for positive infinity. Both Lame
parameters then became infinite and the rest pose multiplied them by a zero
strain, so the first step filled every node with NaN. Reproduced before fixing —
the body reported NaN forces and NaN elastic energy at rest. Every material
scalar is now required to be finite as well as in range, not just the one
flagged, and the **derived** Lame parameters are validated too, which catches
individually admissible inputs whose combination still overflows (for example
`E = 1e308` with `nu = 0.4999999999999999`). The Lame formula moved into a
shared helper so validation checks exactly the values the body then uses.

**Sag measured from the wrong reference.** The demo compared every node against
the body origin, but the box mesh is centered on that origin, so free nodes
begin half a box thickness below it and the readout showed 0.06 m of sag before
anything had moved. Each node's starting height is now recorded and sag measures
displacement from it. This corrected the published figures at 0.22 s from
0.220 m / 0.153 m down to **0.165 m / 0.095 m**; the numbers and imagery above
are the corrected ones.


jslee02:
@codex review

Addressed the P2 finding on non-finite Young's modulus in `8d593d06ae1`. Reproduced it first (infinite modulus was accepted and produced NaN forces and NaN elastic energy at rest), then required every material scalar to be finite as well as in range, and additionally validated the derived Lame parameters so that individually admissible inputs whose combination overflows are rejected too. The Lame formula moved into a shared helper so validation checks exactly the values the body uses. Regression covers infinity and NaN across all four scalars plus the overflow case.

Also added the `fem_cantilever` demo scene and before/after visual evidence to the PR description.

jslee02:
@codex review

Addressed the P2 sag-reference finding in `65197020a42`. You were right that the box mesh is centered on its origin, so comparing every node against `visual.origin[2]` reported half the box thickness as sag before anything moved. Each node's starting height is now recorded when its marker is created and sag measures displacement from that. The published figures at 0.22 s corrected from 0.220 m / 0.153 m to 0.165 m / 0.095 m, and the PR description and imagery were regenerated with the corrected values.

The earlier non-finite-modulus finding remains fixed in `8d593d06ae1`.

The visual evidence is now an animation (18 frames, t = 0 to 0.48 s) rather than stills.

jslee02:
Closing this rather than merging it, following a scope decision.

The two reference papers need different discretizations. Jain & Liu 2011 is surface point masses attached to articulated rigid bodies — that **is** `SoftBodyNode`, which DART 6 already has and which #3382 improved. Kim & Pollard 2011 is a reduced **volumetric** FEM over tetrahedra with an embedded surface, which cannot be expressed in the point-mass model, so it required a second deformable architecture beside `SoftBodyNode`.

Carrying two parallel deformable subsystems on a compatibility release branch is the wrong shape, and it explains why this work needed such an awkward ABI posture: internal-only staging, uninstalled headers, and a Doxygen exclusion purely to avoid freezing an unfinished API into 6.20.

The volumetric FEM lane therefore leaves DART 6 and is retargeted to DART 7, where it can be designed without the compatibility contract. Whether a reduced FEM in this style is still the right DART 7 target is deliberately left open — newer solvers such as AVBD may be a better choice, and that should be decided on DART 7's own terms rather than inherited from a 2011 method.

Nothing is lost. The work stays in branch history: the foundation in `wp-db-fem-foundation` and this PR's elastic element forces in `wp-db-fem-elastic`. The reviews here found two real defects that were fixed and are worth carrying forward if the work is revived — a non-finite Young's modulus producing NaN forces at rest, and a sag metric measured from the wrong reference.

Findings that are facts about DART 6 rather than about FEM are retained in `docs/dev_tasks/dart6_deformable_body_performance/11-fem-integration-seam.md`: notably that `World::step()` returns early without calling `ConstraintSolver::solve()` once automatic deactivation puts the rigid scene to rest, and that external forces applied mid-solve are cleared before the next step. Both constrain any future per-step extension of DART 6.

The removal from `release-6.20` is in its own PR. The Jain & Liu lane continues as DART 6 work.

PR #3385, opened 2026-07-12
Upgrade DART 6.20 AI infrastructure and visual verification

jslee02:
## Summary

- Modernizes the DART 6.20 repository-local AI contributor infrastructure while preserving Claude Code, Codex, and OpenCode capability parity.
- Makes text-first plus claim-tied visual verification the enforced path whenever simulation, scene, physics, rendering, or GUI behavior can be inspected visually.
- Adds blocking OpenSceneGraph end-to-end checks that prove contacts, collision bounds, and labels each change rendered pixels independently.

## Motivation / Problem

- The visual tooling introduced and strengthened by #3371 and #3374 made reproducible offscreen evidence available, but contributor workflows and infrastructure checks did not consistently require or exercise it.
- Codex discovery and runtime support had drifted from the current `.agents/skills`, project-agent, hook, and scenario conventions.

## Changes / Key Changes

- Migrates generated Codex skill adapters from the legacy `.codex/skills` path to `.agents/skills`, with branch-profile-aware generation and drift checks.
- Adds project agents, Codex configuration/hooks, setup and doctor tooling, deterministic agent scenarios, staged-path enforcement, and Windows hook smoke coverage.
- Routes applicable build, test, I/O, Python, CI, PR/review, downstream, backport, and autonomous workflows through `dart-verify-sim`, with an explicit documented exception contract.
- Strengthens OSG capture verification with combined and per-layer A/B checks that hold the scene and camera fixed while independently checking contacts, collision bounds, and labels.
- Makes every repository-text read in the infrastructure checker explicitly UTF-8 so the Windows locale cannot decode tracked UTF-8 content as CP1252.
- Accepts both `tool_input.command` and `tool_input.cmd` through the native Windows bridge and shared shell guard; the canonical PowerShell wrapper now forwards pipeline JSON as BOM-less UTF-8, and Windows CI exercises both the exact Codex `-Command` path and raw-stdin `-File` fallback.
- Lets validated non-commit Windows payloads bypass Git Bash discovery while commit-like payloads remain fail-closed; the shared classifier recognizes Windows `git` and `git.exe` executable casing consistently.
- Captures the native bridge exit code through PowerShell 5.1's global automatic variable, avoiding a script-scope shadow while retaining null/nonzero fail-closed behavior.
- Explicitly enumerates the Windows PowerShell input stream so pipeline JSON always reaches the native bridge.
- Forces repository text synchronization through explicit UTF-8 reads and ASCII-safe status output, including a non-UTF-8-locale regression.
[...]

PR #3405, opened 2026-07-30
Re-arm the black formatting gate in pixi lint tasks (release-6.20)

jslee02:
## Summary

- Re-arms the black formatting gate in the `pixi` lint tasks, which has been excluding every Python file in the repository, and applies the one formatting fix it reports.

## Motivation / Problem

`Publish dartpy` → `Check format` is failing on `6c88ac1d774` because
`scripts/compare_soft_body_performance.py` is not black-formatted.

The more important problem is why nothing caught it earlier. `lint-py` and
`check-lint-py` passed `--exclude '\..*'` to black. Black matches `--exclude`
with `re.search` against the candidate path, so `\..*` matched the literal dot
in every `.py` suffix and excluded the whole repository:

```console
$ black --check . --exclude '\..*'      # what check-lint-py ran
No Python files are present to be formatted. Nothing to do 😴

$ black --check .                       # what publish_dartpy runs
1 file would be reformatted, 43 files would be left unchanged.
```

So the black half of the required CI Linux **"Check Lint"** context — and of the
`pixi run lint` that `AGENTS.md` asks every contributor to run before
committing — has been a silent no-op. That is why "Check Lint" passed while
`Publish dartpy` failed on the same commit. The flag was also redundant:
`--exclude` *overrode* the `[tool.black]` excludes already declared in
`pyproject.toml`.

This is not a one-off. The last four `Publish dartpy` failures on this branch
are all the identical `Check format` step — `6c88ac1d774`, #3388, #3385, and
#3374 — and #3390 was a whole PR spent cleaning up drift the dead gate let
through.

## Changes / Key Changes

- `pixi.toml`: `--exclude '\..*'` → `--extend-exclude '/\.[^/]+'` in `lint-py`
  and `check-lint-py`, in both `[tasks]` and `[target.win-64.tasks]`. This is
  exactly what `main` already carries: additive, so the `pyproject.toml`
  excludes stay in effect, and anchored on `/` so it only matches dot-entries at
  a path boundary. The original "exclude hidden folders" intent from
  `d8396ee0d64` is preserved — the gate now scans 44 files instead of 0.
- `scripts/compare_soft_body_performance.py`: the single formatting fix the
  re-armed gate reports (one implicit-concat f-string pair joined).

## Testing

- `pixi run check-lint` — full aggregate, exit 0, with `check-lint-py` now
  reporting `44 files would be left unchanged` instead of `No Python files are
  present to be formatted`.
- `pixi run black --check --diff .` — the exact failing step, now passes.
- `pixi run isort --check --diff .` — the `Check import` step that was skipped
  behind the format failure and therefore never validated on this commit; also
  passes, so it will not be the next failure.

No C++ or Python library behavior is touched, so no build/test gates are
implicated. `scripts/compare_soft_body_performance.py` is a developer benchmark
script and the change is formatting-only.

## Breaking Changes

- [x] None

The re-armed gate does make the required CI Linux "Check Lint" context genuinely
stricter than it has been. The full aggregate passes on this branch, so the tree
is already clean under it.

- No companion `main` PR needed: `main` already carries
  `--extend-exclude '/\.[^/]+'` (4 occurrences, 0 broken). This is a
  release-branch-only catch-up.
- Prior failures with this root cause: #3388, #3385, #3374; cleanup PR #3390.

---

#### Checklist

- [x] CHANGELOG.md updated if required — not required; internal lint tooling with
      no user-visible change, matching the #3390 precedent and
      `docs/onboarding/changelog.md`
- [ ] Add unit tests for new functionality — n/a, no new functionality
- [ ] Document new methods and classes — n/a
- [ ] Add Python bindings (dartpy) if applicable — n/a

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -78,7 +78,7 @@ lint-cpp = { cmd = """
 """, depends-on = ["config"], env = { BUILD_TYPE = "Release" } }
 
 lint-py = { cmd = """
-    black . --exclude '\\..*' \
+    black . --extend-exclude '/\\.[^/]+' \
     && isort . --skip-glob '.*'
 """, depends-on = ["config"] }
 
@@ -138,7 +138,7 @@ check-lint-cpp = { cmd = """
 """, depends-on = ["config"], env = { BUILD_TYPE = "Release" } }
 
 check-lint-py = { cmd = """
-    black . --check --exclude '\\..*' \
+    black . --check --extend-exclude '/\\.[^/]+' \
     && isort . --check --skip-glob '.*'
 """, depends-on = ["config"] }
 
@@ -801,7 +801,7 @@ config = { cmd = """
 """, env = { DART_VERBOSE = "OFF" } }
 
 lint-py = { cmd = """
-    black . --exclude '\\..*' && isort . --skip-glob '.*'
+    black . --extend-exclude '/\\.[^/]+' && isort . --skip-glob '.*'
 """, depends-on = ["config"] }
 
 lint-spell = { cmd = "codespell --write-changes --config .codespellrc" }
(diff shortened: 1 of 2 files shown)


jslee02:
@codex review

PR #3407, opened 2026-07-30
Remove the volumetric FEM subsystem from DART 6

jslee02:
## Summary

Removes `dart/dynamics/fem/` from DART 6 and retargets the Kim & Pollard lane to
DART 7. DART 6 keeps a single deformable model.

## Why

The two deformable reference papers need **different discretizations**:

| Paper | Model | Fits DART 6? |
| --- | --- | --- |
| Jain & Liu 2011 | Surface point masses on articulated rigid bodies, vertex/edge springs, adaptive active vertices | **Yes** — this *is* `SoftBodyNode`, improved by #3382 |
| Kim & Pollard 2011 | Reduced **volumetric** FEM over tetrahedra with an embedded surface | **No** — needs a second, parallel deformable architecture |

Supporting Kim & Pollard meant standing a whole second deformable subsystem
beside `SoftBodyNode`. Carrying two parallel deformable architectures on a
compatibility release branch is the wrong shape.

It also explains the posture that subsystem needed just to exist here: sources
compiled into `libdart` while every header stayed **uninstalled**, absent from
the generated `dynamics.hpp` aggregate, and **excluded from Doxygen** — all
purely to avoid freezing an unfinished API into 6.20. Those were symptoms of the
code being in the wrong place, not problems to engineer around.

## What is removed

- `dart/dynamics/fem/{TetMesh,DeformableBody}.{hpp,cpp}` and its `CMakeLists.txt`
- `tests/integration/test_FemDeformableBody.cpp` and its test registration
- `add_subdirectory(fem)` in `dart/dynamics/CMakeLists.txt`
- the `DOXYGEN_EXCLUDE` entry that existed only for this subsystem
- `13-fem-foundation.md`, the specification for the removed code

Unaffected: `SoftBodyNode`, `PointMass`, `SoftContactConstraint`, adaptive
contact activation, soft face-interior contacts — the entire Jain & Liu lane.

## Verification

- **No public API, ABI, or default-behavior change.** The subsystem never had an
  installed surface, so nothing downstream could depend on it.
- **Nothing referenced it**: zero matches for the FEM types across `dart/`,
  `tests/`, `examples/`, and `python/` after removal.
- **Full suite: 154/154 passing**, zero build errors.

## Where the work went

Preserved in branch history rather than discarded: `wp-db-fem-foundation` (the
foundation) and `wp-db-fem-elastic` / #3404 (elastic element forces, closed
rather than merged). Two real defects that review found there — a non-finite
Young's modulus producing NaN forces at rest, and a sag metric measured from the
wrong reference — are fixed in that history and worth carrying forward if the
work is revived.

Whether a reduced FEM in the Kim & Pollard style is still the right **DART 7**
target is deliberately left open: newer solvers such as AVBD may suit DART 7
better than a method from 2011, and that should be decided on DART 7's own terms.

## Knowledge retained

`11-fem-integration-seam.md` is kept and reframed, because its findings are facts
about **DART 6**, not about FEM:

- `World::step()` returns early **without calling `ConstraintSolver::solve()`**
  once automatic deactivation puts the rigid scene to rest — so anything hung off
  the solver silently stops running in an ordinary settled scene.
- External forces applied mid-solve are cleared before the next step.

Both constrain any future per-step extension of DART 6, deformable or otherwise.
The decision and its reasoning are recorded in `decisions.md`.

## Breaking Changes

- [x] None


jslee02:
@codex review

Addressed the P2 in `078ff593c7d`. You were right that the scope note alone was not enough: the instructions that actually drive the next session still pointed at the removed subsystem, so anyone resuming would have rebuilt it.

- `docs/plans/dashboard.md` — PLAN-622's scope now states DART 6 carries one deformable model (Jain/Liu), and the next step is landing this removal and continuing with PR-3a soft-foot SIMBICON, with an explicit "do not restart the volumetric FEM subsystem on release-6.20".
- `RESUME.md` — the M2.0/M2.1 next-build items are replaced by a statement that the tet/FEM lane is not DART 6 work, with pointers to the branches that preserve it; the next build step is now PR-3a.
- `11-fem-integration-seam.md` — the closing "proceed with the additive FEM backend on release-6.20" recommendation is marked withdrawn, keeping only the DART 6 mechanism facts that outlive the FEM code.

Verified no remaining instruction anywhere under `docs/` tells a reader to build the tet/FEM subsystem.

jslee02:
@codex review

Addressed the P2 in the latest commit. You were right that the disclaimer alone left the plan self-contradictory while the dashboard still links it as the execution plan.

The plan is now split rather than annotated:

- **Part A** is the active DART 6 plan, covering the Jain/Liu lane only. Its sequencing section now runs land-the-removal, then PR-3a soft-foot SIMBICON, then the remaining Jain/Liu rows.
- **Part B** holds the Kim/Pollard material, banner-marked as retired and kept as DART 7 reference rather than instruction. The section that still read "M2.0 must prototype and pick this seam before anything else" is rewritten in past tense as a record of what was found.
- The status line no longer names M2.0 as the first build step; PR-2 in `## 6. PR structure` is marked retired; the two branch-strategy/seam risks in the risk list are marked settled; sections were renumbered so Part A reads continuously; and the stale `§7/§11` cross-references in `11-fem-integration-seam.md` and `12-pr3a-soft-foot-simbicon.md` now point at Part B and Part A.

Verified no active instruction to build the tet/FEM subsystem remains anywhere under `docs/`.

jslee02:
@codex review

Both addressed in the latest commit.

**P1 — durable owner.** You were right that this was the important one: the decision lived only in the temporary dev-task folder while `docs/design/dart6_deformable_body.md` still published the 2026-07-11 deferral list, so the obsolete policy would have been the surviving one once that folder is retired. The design owner now records both halves — Kim/Pollard out of DART 6 scope and retargeted to DART 7 with the AVBD question left open, and the Jain/Liu controller and hand rows as active DART 6 work rather than deferrals. The dashboard now links that owner rather than only the working record.

**P2 — completion rule.** Also correct: Part A was Jain/Liu-only, but its north star defined done as every row of `02-paper-parity-matrix.md` with no deferrals, and that matrix contains the Kim/Pollard rows, so the retired lane was still transitively required. Completion is now defined over the Jain/Liu rows, and the matrix's Kim/Pollard section is banner-marked as retargeted to DART 7 rather than as DART 6 acceptance targets.

Swept for the contradiction class you have been finding: no doc under `docs/` now calls the Jain/Liu SIMBICON or hand rows deferred, and none requires Kim/Pollard rows for DART 6 completion.

jslee02:
@codex review

Addressed in the latest commit, and thank you for staying on this thread — my previous sweep only covered `docs/design` and `docs/plans`, which is exactly why these survived.

Retracted in the three remaining places:

- `02-paper-parity-matrix.md` — the approved-deferral mapping is replaced by an explicit retraction: the 2026-07-11 list is withdrawn, Kim/Pollard rows are out of DART 6 scope, and every Jain/Liu row is active DART 6 work. The Jain/Liu acceptance gates no longer read "Add or defer", since deferring is no longer available.
- `README.md` closeout bullet — the paper-scale deferrals are recorded as retracted rather than maintainer-approved.
- `README.md` work-packet rows WP-DB.03 and WP-DB.09 — the four-link flexible-foot row is stated as active DART 6 work rather than "neither implemented nor deferred".

All four now point at `docs/design/dart6_deformable_body.md` as the durable owner rather than at the temporary task folder.

The sweep now covers all of `docs/` and returns nothing: no surviving statement defers any Jain/Liu row.

jslee02:
@codex review

Both addressed.

**Matrix acceptance rule** — it applied to "each row" and offered a maintainer-approved deferral as a passing outcome, so it could still be aimed at the retired Kim/Pollard rows or satisfied by deferring. It now applies to the Jain/Liu rows only, says explicitly that the Kim/Pollard rows are not acceptance targets here, and the deferral route is removed.

**RESUME.md** — correct that the stale instructions came hundreds of lines before the do-not-restart note, so a resuming session hit them first. The header now reads `Updated: 2026-07-29` with the current scope plus a read-this-first banner (one deformable model, do not restart Kim/Pollard, next step PR-3a, durable owner named, pre-2026-07-29 sections are historical). The 2026-07-23 section is marked SUPERSEDED, keeping only what still holds from it: the deferral-list retraction and the zero-rigid-overhead requirement.

I also swept proactively rather than reactively this time, across all of `docs/`, for three things: any doc describing DART 6 scope as both papers, any instruction to build the FEM/PR-2 lane, and any place offering deferral as an outcome. The two that turned up were fixed in this commit — the same annotation on the 2026-07-23 entry in `decisions.md`, and a stale "deferred to the first real PR-2 milestone" reference in the retained seam document. Everything else returned is a banner or a heading whose superseded note is on the adjacent line.

jslee02:
@codex review

Both addressed in `edfff4b3e0d`, and the pattern you keep surfacing is the real finding: I have been fixing the *later* sections of each document while its opening still stated the old scope, so a top-down reader met the stale version first.

**Execution plan, Part A** — the retired lane's actions are gone from the active plan. Multi-core scaling (WP-DB.07) now points at the contact-heavy Jain/Liu scenes, noisy-floor locomotion and the hand rows, rather than larger FEM workloads. The SIMD item targets the point-mass and soft-contact kernels rather than FEM element/modal kernels. Model authoring covers the hand, arm, and four-link foot assets and explicitly notes no tet meshes are needed. The Kim/Pollard gap ledger was **moved out of Part A into Part B** as B.2, so §5 in the active plan is Jain/Liu-only. Part A's remaining mentions of FEM are all statements about the removal itself.

**Task README** — the scope now sits at the objective rather than only in the later table and closeout. The objective states the Jain/Liu-only scope, names `docs/design/dart6_deformable_body.md` as the durable owner, and says not to restart the lane. The reference-scope section splits the two papers into "Active for DART 6" and "Not DART 6 work — retargeted to DART 7" instead of listing both unqualified, and the paragraph that followed now refers to the remaining Jain/Liu rows.

PR #3408, opened 2026-07-30
Add soft-foot SIMBICON locomotion scene and push-recovery gate

jslee02:
## Summary

Reproduces the Jain & Liu 2011 soft-contact claim that a biped on **deformable
feet maintains more ground contact** than the same biped on rigid feet. This is
the first paper row of the DART 6 deformable lane, which is now scoped to
Jain & Liu only (#3407).

## Why this is small

Almost everything needed already exists. The `atlas_simbicon` controller in the
demos tree is **GUI-free**, finds the feet by name, and its terminal condition
already reads `BodyNode::isColliding()`, which works for a `SoftBodyNode` foot.
The soft-feet asset `atlas_v3_no_head_soft_feet.sdf` also already exists — and
now actually loads as soft bodies, thanks to #3399. So this increment is
**assembly and measurement**, not new dynamics.

## Changes

- `SoftFootSimbiconModel` — GUI-free model driving the rigid or soft-feet Atlas
  with the existing SIMBICON controller; exposes upright state, foot-contact
  count, finiteness, a position checksum, and a push-threshold sweep.
- `soft_foot_simbicon` `dart-demos` scene with feet toggle, push keys, and a
  live panel.
- `test_SoftFootSimbiconModel` — three gates, wired into `test_ConstraintSolver`
  behind a `dart-utils`/`dart-utils-urdf` target guard, since the model loads
  URDF and SDF and those components are optional.

## Visual verification

Same scene, same camera, same scripted push at step 650; only the foot geometry
differs. The soft-foot biped through the push:

Soft-foot Atlas biped absorbing a lateral push and recovering under SIMBICON control

*12 frames, t = 0.60 to 1.10 s. Animated PNG — GitHub plays it inline.*

At the identical frame (870, t = 0.87 s), with the on-screen readouts legible so
the numbers are verifiable in the image rather than only in this description:

Rigid rest-mesh control versus soft feet at the same instant: 24 versus 47 foot contacts

| At frame 870 | Feet | Foot contacts | Upright |
| --- | --- | ---: | --- |
| Rigid | `Rigid (rest-mesh control)` | **24** | yes |
| Soft | `Soft (SoftBodyNode)` | **47** | yes |

Both stay upright, so the difference is in **contact count**, not a staged
collapse — which is precisely the mechanism the paper describes.

### Making the two models comparable

Three things had to be corrected before the comparison meant anything, all on
the loaded instance (before it joins the world) rather than in the shared
`dart://sample` asset:

- **Mass.** A `<soft_shape>`'s `<total_mass>` is added *on top of* its link mass,
  so the soft feet were a kilogram heavier, and extra ground load raises contact
  counts by itself. The link mass and inertia are scaled down so link plus point
  masses equals the rigid foot; both bipeds weigh 146.554 kg.
- **Collision surfaces.** Parsing `<soft_shape>` adds a collidable
  `SoftMeshShape` while the link's original rigid STL collision mesh is still
  there, so a soft foot presented two overlapping surfaces against the rigid
  foot's one. Only the soft surface stays collidable.
- **Collision tessellation.** The rigid control collided with `l_foot.stl` /
  `r_foot.stl` while the soft feet collide with the surface generated from
  `<soft_shape>`, and different representations produce inherently different
  contact manifolds — an analytic box caps at a few corner contacts while a
  mesh emits per-vertex contacts. The control's feet now collide as a **rigid
  triangle mesh with the soft feet's exact rest tessellation**, generated by
  the same helper, friction carried over, STL kept for visuals. A gate checks
  the control's mesh vertex-for-vertex against the soft feet's actual
  point-mass rest positions.

  The shared narrow phase is *measured*, not argued: the comparison is pinned
  to FCL (where both foot types are BVH meshes against the ground primitive;
  the native `dart` detector dispatches the two shape types differently and is
  asserted away), and an equivalence gate collides both models undeformed at
  an identical penetrating pose, requiring the same foot-ground manifold — it
  reports **48 = 48 contacts with a worst pairwise distance of 9 nanometers**
  (the residual is the soft mesh's single-precision vertex storage). Whatever
  the walking gates then measure between the two models is the soft contact
  pipeline at work, not collision bookkeeping.

A gate holds all of these properties. Representation alone moves the rigid contact
count by an order of magnitude (STL mesh 44.4, analytic box 3.6, matched mesh
**18.45**), which is why the earlier 52-vs-44 and 52-vs-3.6 figures both said
more about collision geometry than about compliance. Against the matched
control, soft's unchanged 52.0 average is a **2.8x spread attributable to
deformability alone**, and the gate now enforces the advertised margin at
soft ≥ 1.5x rigid rather than a bare ordering.

### Push recovery does not reproduce

With the models matched, soft recovers **4000 N** against the rest-mesh
control's **12000 N** (the sweep was widened to 24000 N after the control
saturated the old 12000 N ceiling, so this is a measured threshold, not a
clamp). The gate measures and prints both thresholds instead of asserting the
paper's comparison.

**Two maintainer decisions block the claim.** Full options and trade-offs in
`docs/dev_tasks/dart6_deformable_body_performance/12-pr3a-soft-foot-simbicon.md`
§ Open maintainer decisions.

- **Soft-foot stiffness** — the shipped feet are too compliant for a 147 kg
  Atlas. Recoverable push by `kv`/`damp`: 5e4/1e3 (shipped) 4000 N · 1e6/1e3
  6000 N · 1e6/5e3 **12000 N**, which matches the mesh control's measured
  threshold but no longer exceeds it. Retuning the shared
  `atlas_v3_no_head_soft_feet.sdf` changes behavior for `test_SdfParser` and
  downstream callers; overriding scene-locally is contained but makes the demo
  differ from the asset it loads.
- **Foot inertia and COM** — equal mass is not equal dynamics. The point masses
  add inertia at their rest positions, leaving the soft foot at ~1.8-2.0x the
  rigid foot's principal moments with a shifted COM. Matching the combined
  inertia may be unrealizable without redesigning the foot geometry, since the
  point masses alone already exceed the rigid moments.

I have not picked the values that would make the assertion pass.

## Gates — 3/3, full suite 155/155

1. **Stands and is deterministic** — soft biped finite and upright for 1000
   steps, minimum pelvis-above-feet gap 0.800 m, exactly equal full state
   vectors across two runs (378 components), and Reset is bit-exact against a
   fresh model over 300 lockstep steps.
2. **Soft maintains at least rigid ground contact** (the paper's claim) — over
   window [500, 1000]: **soft avg 52.0 (min 31) vs rigid avg 18.45 (min 5)**,
   gate-enforced at soft ≥ 1.5x rigid.
3. **Recoverable push, measured** — rigid **12000 N**, soft **4000 N**. This
   is a negative result and the gate does not assert the comparison; it asserts
   only that both thresholds are real measurements inside the sweep, neither
   zero nor saturated. See "Push recovery does not reproduce" below.

## Compatibility

Purely additive: no existing runtime source is modified, no public API, ABI, or
default behavior changes. The controller sources are reused as-is;
`AtlasSimbiconScene.cpp` is deliberately excluded from the test target as the
only file there pulling in OSG and ImGui.

## Breaking Changes

- [x] None


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3413, opened 2026-07-30
Scale demo contact-force arrows to the scene

jslee02:
Contact-force arrows in the demos were converted from newtons to meters with a
fixed `0.1 m/N` constant carried over from `RigidCubesScene`, where it had been
tuned against 1 kg blocks. On anything heavier every arrow ran past the 5 m
safety cap and came out at exactly 5 m, so they became identical hairline
streaks whose heads sat meters from the bodies they belonged to, with the
magnitude encoding gone.

On the Atlas SIMBICON demo the contacts carry roughly 1800 N, which is 176 m of
arrow before clamping. What that looks like in practice is a bundle of lines
pointing off toward the horizon that never appears to follow the robot — the
symptom that prompted this fix.

## Visual difference

Same scene, same steps (frames 900–1340 of `atlas_simbicon`), same camera.

| Before | After |
| --- | --- |
| before | after |

Close-up on the feet:

| Before | After |
| --- | --- |
| before closeup | after closeup |

Reproduce:

```bash
for s in 900 940 980 1020 1060 1100 1140 1180 1220 1260 1300 1340; do
  ./build/default/cpp/Release/bin/dart-demos --scene atlas_simbicon \
    --headless --steps $s --shot frame_$s.png
done
```

## What changed

`ContactArrowLayout` derives both of its references from the scene:

- **Length** — a quarter of the mobile skeletons' diagonal, so arrows stay
  proportional to whatever is being simulated. Static geometry is excluded, or a
  20 m ground box would size every arrow to the floor. Degrees of freedom rather
  than `isMobile()` decide that, because a welded ground still reports mobile.
- **Force** — the peak contact force actually observed, tracked instantly and
  released over half a second, so the longest arrow on screen is the largest
  force on screen, and resting contacts become readable again shortly after an
  impact rather than staying crushed by a spike seconds in the past. A floor at
  5% of the supported weight stops a contact-free scene from normalizing
  numerical noise up to full-length arrows.

Length and color now come from the same normalized value, so the two agree. The
arrow shaft is sized from the same reference. A scene-relative length bounds the
arrow inherently, which is what the old 5 m cap existed to do.

The layout is free of OSG and ImGui so it can be gated headlessly; only the
toggle stays in `ContactVisualizer`.

## Gates

`tests/integration/test_ContactArrowLayout.cpp`, five cases covering the three
properties the saturation violated plus two robustness ones:

| Case | Asserts |
| --- | --- |
| `ArrowsStayAnchoredToTheirContactPoints` | over 400 steps of a sliding 180 kg box, every arrow tail is a contact point of that step and every arrow is within one reference length |
| `ReferenceLengthTracksTheSceneNotTheForce` | a 3 m box gets >5x the arrow of a 0.3 m box of equal mass; the small scene stays under 1 m; and the old constant would still saturate here (176 m > the 5 m cap) |
| `ArrowLengthOrdersByForceMagnitude` | halving the force halves the arrow; the peak draws at exactly the reference length; color matches |
| `DropsNonFiniteAndNegligibleContacts` | NaN/Inf points and forces and zero-force contacts are dropped, not drawn |
| `ReferenceForceRecoversAfterASpike` | after a 5000 N spike a 25 N resting contact is briefly a sliver, then readable again once the reference decays |

Measured: 0.3 m box → 0.075 m arrows, 3 m box → 0.75 m arrows, versus 176 m
(clamped to 5 m) under the old constant.

Also verified: `test_ConstraintSolver` + the soft demo-model suites, 63 tests
passing, and `dart-demos --cycle-scenes` over all 36 demos.


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3423, opened 2026-08-01
Resolve the soft-foot comparability decisions: matched rest inertia and properly damped feet

jslee02:
Resolves the two decisions PR #3408 left open, both by measurement. The
Jain/Liu push-recovery claim now **reproduces and is gate-asserted**:
recoverable push **18000 N soft vs 8000 N rigid**, alongside contacts
**51.2 vs 15.64** (3.3x, gate-enforced at 1.5x). A Codex round-2 finding
(the reused controller's COM sensor was blind to point masses) is fixed at
the root and A/B'd below — conclusions unchanged.

## Visual difference

The same 6000 N scripted push (step 650, 100 steps), soft feet, corrected
COM sensor (see the round-2 section), identical scene — the only change is
the asset's `<damp>`:

| damp 1000 (shipped) — topples | damp 4000 (this PR) — recovers |
| --- | --- |
| before | after |

End state at t = 2.25 s, panels legible (`Upright: NO (fallen)` vs
`Upright: yes`):

end state

Reproduce:

```bash
DART_DEMO_SOFT_FOOT_FEET=soft DART_DEMO_SOFT_FOOT_PUSH_STEP=650 \
DART_DEMO_SOFT_FOOT_PUSH_N=6000 \
  ./build/default/cpp/Release/bin/dart-demos --scene soft_foot_simbicon \
  --headless --steps 2250 --shot end.png
```

## Decision 1 — the control carries the soft rest-pose inertia

Measured mismatch before the fix: soft combined moments **1.98 / 1.79 /
1.84x** the control's, COM shifted **5.9 mm**, at equal mass. Matching by
thinning the soft link is unrealizable (negative inertia), but the control is
defined as *the soft foot with deformation frozen*, so the realizable match is
the other direction: `normalizeRigidFoot()` sets the control to the soft
link (down-scaled exactly as `normalizeSoftFoot()` does) plus every point
mass at rest via the parallel-axis theorem. The comparability gate recomputes
the soft side from the live body and requires equality to 1e-9 — during
development it caught a real bug (a placeholder total mass in the mesh
generator call), which is what a gate computed through an independent path is
for.

**A/B:** the matched tensor alone moved the control's threshold
**12000 → 8000 N** and its settled contacts 18.45 → 15.64. The compact SDF
tensor had been making the control artificially stable.

## Decision 2 — the feet were under-damped, not under-stiff

4x2 screening grid on the soft arm (contacts / recoverable push):

| kv \\ damp | 1e3 (shipped) | 5e3 |
| --- | --- | --- |
| 5e4 (shipped) | 51.98 / 4000 N | 48.4 / **18000 N** |
| 2e5 | 37.8 / 4000 | 53.5 / 4000 |
| 5e5 | 28.9 / 10000 | 50.4 / 4000 |
| 1e6 | 24.3 / 6000 | 37.7 / 12000 |

Raising stiffness erodes the contact-spreading claim; raising damping at the
shipped stiffness resolves recovery — an under-damped surface returns the
push energy, a damped one absorbs it, which is the paper's own mechanism.

3x5 basin map around the candidate (recoverable push):

| kv \\ damp | 3e3 | 4e3 | 5e3 | 6e3 | 7e3 |
| --- | --- | --- | --- | --- | --- |
| 4e4 | 18000 | 18000 | 18000 | 6000 | 6000 |
| 5e4 | 18000 | **18000** | 18000 | 6000 | 6000 |
| 6e4 | 4000 | 18000 | 18000 | 18000 | 18000 |

Chosen: **kv unchanged, `<damp>` 1000 → 4000** — the interior of the
plateau (all four adjacent cells also 18000 N; contacts 51.2 with the
corrected sensor), not a knife-edge. `test_SdfParser` asserts structure only, so the shared-asset
change is behavioral for this demo alone.

## Why the shared asset, not a scene-local override

Changing `dart://sample/sdf/atlas/atlas_v3_no_head_soft_feet.sdf` alters
default behavior for every consumer, and AGENTS.md protects that **unless a
maintainer approves otherwise** — which is the case here: the maintainer
delegated these decisions with an explicit direction toward clean, long-term,
root-cause fixes over local workarounds.

The root cause is the asset itself: its shipped `<damp>` makes the very biped
it was authored for fall at a 4000 N push in its own bundled demo. A
scene-local `setDampingCoefficient()` override was considered and rejected —
it would keep the invalid value for every other consumer of the asset and make
the demo silently diverge from the file it claims to load. The only in-repo
behavioral consumers are this demo and `test_SdfParser` (structure-only
assertions), and the retuned value sits at the interior of a measured
stability basin rather than a hand-picked point.

## Round 2 — the reused controller sensed a COM without the point masses

Codex flagged that the reused SIMBICON `State` fed its balance feedback from
`Skeleton::getCOM()`/`getCOMLinearVelocity()`, which iterate body nodes only —
point masses are invisible to them. On the soft arm that sensor missed the
0.5 kg per foot this PR moves into point masses, while on the rigid arm it
was exact: a per-arm sensor bias, i.e. a third confound.

Root-cause fix in `State`: `getCOM()`/`getCOMVelocity()` now accumulate body
nodes *and* point masses (mass-weighted world positions/velocities) — a no-op
for rigid skeletons. The comparability gate computes the observed COM through
an independent path and asserts (a) the body-only sensor really did diverge
on the soft arm (> 1 mm, non-vacuity) and (b) both arms' observed rest COMs
agree to 1e-9.

**A/B (sensor fix alone):** thresholds unchanged (18000 / 8000 N), soft
contacts 51.4 → 51.2 — the confound was real but every conclusion survives
it. The evidence strips above were re-captured under the corrected sensor.

## Gates

All six pass; `MeasuresRecoverablePushForBothFeet` now asserts the paper's
comparison (soft ≥ rigid) on same-run measurements, on top of its
non-vacuous/non-saturated checks. The comparability gate adds the per-foot
inertial equality (mass, COM, full moment tensor to 1e-9). Task spec's
*Open maintainer decisions* section is now *Resolved decisions* with this
evidence; the parity matrix row is updated.

Also verified: `dart-demos --cycle-scenes` over all 37 demos, full
`check-lint`, clang syntax check on the changed demo sources.


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3451, opened 2026-08-26
Validate default Python lock policy

jslee02:
## Summary

- Make scheduled Pixi updates fail before opening a PR when the default environment loses its explicit, cross-platform Python minor-version contract.

## Motivation / Problem

- PR #3449 repaired a generated lockfile where the default environment was not explicitly pinned and one platform silently resolved an older Python line. The maintenance workflow should detect that class of drift at generation time instead of relying on review to notice it.

## Changes / Key Changes

- After `pixi update`, enumerate every declared workspace platform.
- Require exactly one explicit default-environment Python package on each platform.
- Require a uniform `major.minor.*` requested spec and verify each resolved version matches it.
- Stop before `create-pull-request` when any platform violates the policy.

## Testing

- `pixi run lint`
- Ran the guard against `main`: Python 3.14.7 with `3.14.*` on linux-64, linux-aarch64, osx-64, osx-arm64, and win-64.
- Ran the guard against the landed `release-6.20` fix: Python 3.14.7 with `3.14.*` on linux-64, osx-64, osx-arm64, and win-64.
- Confirmed the guard rejects the pre-fix #3449 lockfile because its Python package has no explicit requested spec.
- Product build and unit tests are not applicable to this workflow-only validation change.

## Breaking Changes

- [x] None

- Follow-up to #3449

---

#### Checklist

- [x] CHANGELOG.md not required for workflow-only maintenance
- [x] Workflow behavior covered by positive and negative local checks
- [x] No public methods or classes to document
- [x] No dartpy binding changes

### DIFF
--- a/.github/workflows/update_lockfiles.yml
+++ b/.github/workflows/update_lockfiles.yml
@@ -69,6 +69,62 @@ jobs:
         run: |
           set -o pipefail
           pixi update --json --no-install | pixi exec pixi-diff-to-markdown >> diff.md
+      - name: Validate default Python lock policy
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          expected_spec=""
+          mapfile -t platforms < <(
+            pixi workspace platform list --json |
+              jq -r '.platforms[] | select(.name != "current") | .subdir'
+          )
+
+          if (( ${#platforms[@]} == 0 )); then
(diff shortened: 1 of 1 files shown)


jslee02:
@codex review

PR #3474, opened 2026-09-02
Auto-detect the GPU runner CUDA architecture instead of pinning it

jslee02:
## Summary

- Replace the hard-coded CUDA architecture in `ci_cuda.yml` with auto-detection on the GPU-runner lane, so the kernels always carry native SASS for whatever GPU the self-hosted runner has. #3475 already unblocked CI by re-pinning `75` -> `89`; this removes the class of failure rather than its current instance.

## Motivation / Problem

- CI CUDA failed all 41 GPU runtime tests with "the provided PTX was compiled with an unsupported toolchain" (https://github.com/dartsim/dart/actions/runs/33572219705). Root cause: the workflow pinned sm_75 while the runner is an RTX 5000 Ada (sm_89), so there was no native SASS for the device and the driver had to JIT the embedded compute_75 PTX — which its CUDA 13.2 driver rejects because pixi ships nvcc 13.3.
- The pin was stale because the runner GPU changed under it. A static value re-stales the same way on the next runner swap, and the failure is invisible until a run actually reaches the GPU (superseded-cancelled runs hid it for days here).
- The repo already owns the durable mechanism: with `DART_CUDA_ARCHITECTURES` empty, the cuda config auto-detects compute capability via `nvidia-smi` and emits real SASS — documented in testing.md and implemented in `scripts/cmake_config.py::resolve_cuda`. Only the workflow env pin was overriding it.

## Changes / Key Changes

- `.github/workflows/ci_cuda.yml`: `DART_CUDA_ARCHITECTURES` is empty on the GPU-runner lane (auto-detect) and an explicit `89` on the GPU-less fork-PR fallback, which only compiles and has nothing to detect. The comment keeps #3475's root-cause explanation and records why the GPU lane is intentionally unpinned.
- Detection cannot silently fall back on that lane: the job's "Show GPU (nvidia-smi)" step runs before the build, so a missing `nvidia-smi` fails loudly first.

## Testing

- `pixi run lint`; YAML parse plus an assertion that the resolved env carries the expected conditional expression.
- `scripts/cmake_config.py::resolve_cuda` contract probe: empty env -> auto-detect, non-empty env -> pinned.
- The auto-detect -> sm_89 SASS path is validated end to end on identical hardware (local RTX 5000 Ada: `pixi run -e cuda test-all` fully green, plus 155/155 CUDA demo scenes).
- This same-repo PR runs CI CUDA on the GPU runner, which is the end-to-end check.
- CHANGELOG: no entry — CI workflow configuration.

## Breaking Changes

- [x] None

- Follows #3475, which fixed the immediate breakage by re-pinning the arch; `main` is merged in and its explanatory comment is preserved. Close this PR instead if you prefer to keep an explicit pin and accept re-pinning on each runner GPU change.


jslee02:
@codex review

PR #3478, opened 2026-09-05
Improve task retrospectives and harness validation

jslee02:
## Summary

Make `dart-retro` use the original request and session outcomes to choose supported improvements to future task execution, including unsuccessful tasks.

## Motivation / Problem

A merged PR can complete an authorized slice while leaving the original goal unmet. Retrospectives need to distinguish harness causes, existing coverage, and external constraints, and avoid accumulating instructions after every task.

## Changes / Key Changes

- Rewrite the retrospective method around observed decisions, expected future behavior, and evidence that could disprove a proposed benefit; retain `Skip If` and make tool-specific reading conditional.
- Clarify semantic ownership checks during documentation consolidation and verification of formatted startup scripts on supported runtimes in existing owner docs.
- Regenerate Codex/OpenCode adapters and update the workflow catalog and changelog. No new tracked files or always-loaded rules.

## Before / After

- Before: the workflow described capturing durable lessons from completed work.
- After: it explicitly compares original intent with actual outcomes, tests whether a harness change would help similar tasks, and skips unsupported or already-covered edits.

## Testing

- Passed `pixi run test-all`: builds, 229 core and 81 simulation CTest cases, 2,015 Python tests passed (19 expected configuration/dependency skips), and docs build.
- Passed `pixi run -e cuda test-all`, including CUDA runtime tests and benchmark smoke.
- Passed `pixi run check-ai-commands`, `pixi run check-ai-infra`, `pixi run test-ai-infra` (550 tests), `pixi run exercise-agent-scenarios`, `pixi run check-docs-policy`, `pixi run check-lint-md`, and `pixi run check-lint-spell`. Full lint ran in both complete checks. The 11 existing report-only north-star freshness advisories remain.
- Three fresh Astra Max sessions compared the original workflow, the initial rewrite, and the shorter final version on identical archived inputs: the completed harness upgrade, an unsatisfactory CPU/GPU task with an authorized smaller closeout, and a routine docs fix. All retained outcome/scope and skip decisions. Revised versions made falsification and contrasting-task checks explicit; these bounded qualitative results do not establish general quality or speed gains.
- Ownership/discovery and outcome/scope review passes plus the principle audit found no remaining issue. No simulation/model/GUI behavior changes, so visual verification is N/A.

## Breaking Changes

- [x] None

## Related Issues / PRs (backports)

- Follow-up to #3477.
- Backports: none; this is a DART 7 harness workflow enhancement.

---

#### Checklist

- [x] Milestone set: DART 7.0
[...]

PR #2328, opened 2025-12-17
dartpy: unify vector conversions and release GIL

jslee02:
- Use eigen_utils helpers for vector-like Python inputs across bindings
- Accept (1, N) NumPy arrays in toVector
- Release the GIL for stepping/collision-heavy calls (convert args before release)
- Retire completed docs/dev_tasks/py_bind and capture conventions in onboarding

---

#### Before creating a pull request

- [x] Run `pixi run test-all` to lint, build, and test your changes
- [x] Add unit tests for new functionality (N/A: no new public functionality added)
- [x] Document new methods and classes (N/A: no new APIs)
- [x] Add Python bindings (dartpy) if applicable

PR #3485, opened 2026-09-06
Re-enable the CI compiler cache and fix the doc-only CI skip

jslee02:
## Summary

- Re-enable the compiler cache on every GitHub-hosted CI job and make the doc-only CI skip actually work. Both had silently regressed: since GitHub's cache-service v2 migration every Linux/macOS/Windows build ran cold, and since the shared code filter was centralized (#2655) every documentation- and AI-harness-only PR still ran the full platform and wheel matrix. No job is added to or removed from the PR tier; required check names are unchanged.

## Motivation / Problem

- `configure-compiler-cache` gated sccache on `ACTIONS_CACHE_URL`, which the runner and `mozilla-actions/sccache-action` stopped exporting with cache service v2 (only `ACTIONS_RESULTS_URL` exists now). Every hosted job logged `GitHub Actions cache URL is unavailable; disabling sccache launcher.` and built from scratch. Builds are >95% of job time: Linux Release Tests spent 53 min compiling for 21 s of ctest, Windows ~100 min for 32 s.
- `.github/filters/ci-code.yml` used `**` plus `!` exclusions, but `dorny/paths-filter` ORs patterns under its default `predicate-quantifier: some`, so a negated pattern matches everything that is *not* excluded and `code` was always `true`. #3478 (nine markdown files) ran 27 jobs; the run log shows every file, including `docs/ai/README.md`, matched `code`.
- The 10 GB Actions cache was full (10.28 GB): two CodeQL C++ TRAP entries of ~3 GB and ~3.3 GB of pixi environment caches. Even with the launcher fixed, sccache would have been evicted immediately.
- `SCCACHE_NO_DAEMON=1` (introduced without rationale in #2328) makes a compile whose output is piped, i.e. every ninja job, block until the sccache server's idle timeout closes the inherited pipe. Reproduced locally with sccache 0.17.0: 20 s stall with `SCCACHE_IDLE_TIMEOUT=20`, 0 s without the variable.

## Changes / Key Changes

- `.github/actions/configure-compiler-cache`: detect the cache service through `ACTIONS_RESULTS_URL` or the legacy variable; export the backend variables in the current shell *before* the sanity compile (that compile spawns the daemon, which fixes its storage at spawn) and assert via `sccache --show-stats` that the running server uses the GitHub Actions backend; drop `SCCACHE_NO_DAEMON`, disable the idle timeout so the post-step stats are cumulative; write a "Compiler cache" block (launcher, backend, storage, runner) to the step summary; **fail GitHub-hosted jobs that would build without a launcher or with a local-disk server** (`require-cache`, default true; macOS keeps its documented best-effort policy with `require-cache: "false"`). The Windows `Tests (Release)` job drops its PowerShell copy and does not run the action: the pixi Windows tasks use the Visual Studio generator and MSBuild ignores `CMAKE_<LANG>_COMPILER_LAUNCHER`, so sccache never ran there and cannot with that generator; the follow-up (Ninja under a VS developer environment, or sccache's MSBuild `CMAKE_VS_GLOBALS` integration) is recorded in the dev task.
- `.github/actions/setup-pixi-ci`: pixi environment cache off by default (`pixi install --locked` is well under a minute; the entries consumed a third of the cache budget); `sccache-action` v0.0.11 with the post-step hit/miss stats enabled so every job records evidence. `codeql.yml`: `trap-caching: false`, and `DART_DISABLE_COMPILER_CACHE=ON` for the traced C++ build (the extractor cannot see compilations that a detached sccache server performs).
- `.github/filters/ci-code.yml` + every `dorny/paths-filter` step: `predicate-quantifier: some-with-excludes`; exclusions widened to `.agents/`, `.claude/`, `.codex/`, `.opencode/`, `docs/`, prose formats, scheduled-only and maintenance workflows, `docker/`, `.devcontainer/`, repository metadata, and the AI-infra Python files only `pixi run test-ai-infra` exercises. `tests/test_ci_code_filter.py` (in `test-ai-infra`) fails if a workflow reads the filter without the quantifier or a literal exclusion goes stale. The Windows hook smoke moved to its own ~2 min job gated by a `hooks` filter so `.claude/hooks/**` changes still get their Windows check.
- `permissions: contents: read` on every workflow (`pull-requests: read` on the `changes` jobs for the PR file API) and `timeout-minutes` on every job; `changes` jobs no longer clone full history (PR events use the API, push events fetch the `before` commit on demand).
- `publish_dartpy.yml`: the `wheel_matrix` job and `wheel_workflow` filter emitted the same Py3.14 matrix on both branches; replaced by one static matrix.
- `.github/dependabot.yml`: `directories: ["/", "/.github/actions/*"]` so composite-action pins are bumped too (Dependabot ignores them under `directory: "/"`).
- `docs/onboarding/ci-cd.md`: cache section rewritten around the composite actions with the regression history and a cache-budget rule, wheel and filter sections updated, stale task narrative and contradictory Debug-tier statements removed, measured baseline recorded. `docs/dev_tasks/ci-optimization/` holds the initiative state; PR-2 (job consolidation and YAML de-duplication) deletes it.

## Testing

- `pixi run lint-yaml`, `lint-md`, `lint-py`, `lint-toml`; `pixi run check-lint-yaml`, `check-lint-md`, `check-lint-py` clean. `pixi run lint` itself stops at the C++ formatter's CMake configure in this checkout (Filament not installed in the local default env), unrelated to this change.
- `pixi run test-ai-infra` (555 passed, including the new filter guard), `pixi run check-ai-infra`, `pixi run check-docs-policy`, `pixi run check-diff-workflow`, `pixi run check-phase5-cuda-workflow`; every workflow, action, and filter file parses as YAML.
- Local reproduction of the `SCCACHE_NO_DAEMON` pipe stall with the pixi sccache binary (see Motivation).
- The compiler-cache action script dry-run locally under a simulated runner environment for five cases: hosted runner with a rejected GitHub backend (fails with the `::error::` guard), hosted runner without a cache-service URL with and without `require-cache` (fails / continues with `launcher: none`), self-hosted Linux (ccache), and hosted runner with a local-disk sccache (fails). The ccache fallback is limited to self-hosted runners because on an ephemeral hosted runner it is always cold and would hide a broken sccache setup behind a green job.
- Upstream sources verified for the exact contracts relied on: `dorny/paths-filter@v4` (`some-with-excludes`, PR events via API with `pull-requests: read`, push events fetch `before` on demand), `mozilla-actions/sccache-action` exports (`ACTIONS_RESULTS_URL`, `ACTIONS_RUNTIME_TOKEN`, `ACTIONS_CACHE_SERVICE_V2`), `codeql-action/init` `trap-caching`, `setup-pixi` `cache` semantics, Dependabot `directories` globbing, and GitHub's rule that a job skipped by `if:` satisfies a required check.
- Independent multi-lens review of the diff (GitHub Actions semantics, cache runtime, path filter, guard scripts/docs, security/ops) with three-way adversarial verification per finding; the ten confirmed findings are fixed here (daemon spawned before the backend variables were exported; MSBuild ignoring launchers on Windows; `package.xml`, `codecov.yml`, `.gitattributes`, `.gitignore` wrongly excluded from the code filter; idle-timeout stats; four doc contradictions; dev-task pointers).
- Live evidence on this PR: (1) first push builds cold and populates the cache; (2) a docs-only push must report `Filter code = false` and skip every platform and wheel job with required checks green; (3) a second code push must show non-zero sccache hits in the post-step stats on Linux/macOS jobs and the macOS/Windows wheels, with materially shorter build steps. Results recorded in the "Expected CI Times" table of `ci-cd.md` before merge.
- CHANGELOG: entry added under Build, Packaging, and Developer Tooling.

## Breaking Changes

- [x] None

- Follows the July 2026 tiering work (#3222). PR-2 will consolidate the redundant Linux Release builds and de-duplicate the workflow YAML on top of the warm-cache numbers this PR produces.

---

#### Checklist

- [x] Milestone set (DART 7.0 for `main`, branch-matching DART 6.x release
      milestone for the active DART 6 LTS branch)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3486, opened 2026-09-06
Probe: documentation-only change against the CI cache/filter branch

jslee02:
Evidence probe for #3485: a documentation-only change whose pull-request run must report `Filter code = false` in every `changes` job and skip every platform and wheel job while the required checks stay green. Not for merging; closed once the run is recorded.

Base: `ci/fix-compiler-cache-and-path-filter`.


jslee02:
Probe complete: every workflow's `changes` job on this documentation-only PR logged `Filter code = false` with `Matching files: none`, and all platform and wheel jobs were skipped while Lint/Documentation ran (evidence for #3485).

PR #3490, opened 2026-09-06
Add the living DART 7 architecture map with archify views and liveness gates

jslee02:
## Summary

- **Problem.** The published DART 7 architecture map was a hand-drawn ASCII diagram, changed 10 times since June while `dart/simulation` changed 92 times, and only its availability table was machine-checked. The assessment's standing audit rule named no diagram to update, so the picture of the framework could drift silently behind the code. The handbook's architecture entry was still the 1085-line DART 6 analysis, loaded as always-on context by four module `AGENTS.md` files.
- **Fix.** Four typed archify views under `docs/assets/architecture/` (simulation framework, `World::step()` data flow, compute graph, library context) are now the source of truth. `scripts/render_architecture_map.py` renders them into the architecture page at docs-build time from a pinned archify checkout, stamping the built commit so every `SRC` marker opens the right revision. `pixi run check-architecture-map` (in `check-lint`) fails when a cited path, line, or symbol disappears or when a `dart/simulation` module, step-stage slot, solver family, `WorldStepStage` subclass, or `dart/<module>` is missing from its view. A recording-executor probe test and `pixi run check-architecture-map-runtime` report schedule drift (advisory, `--strict` available). The classic DART 6 page shrinks to a pointer page, the `dart-architecture` skill and owner docs carry the update procedure, and the decisions live in `docs/design/architecture_map.md` (PLAN-130 archived).
- **Keep in mind.** Docs builds now expect Node.js 18+ (pixi already locked it through prettier; `.readthedocs.yml` declares `build.tools.nodejs`); without it the page degrades to text fallbacks with a warning. The views are English like code identifiers (archify's viewer supports `en`/`zh-CN`); the page prose around each embed stays translatable, so the Korean build keeps working.

## Visuals

<details>
<summary>Before: the ASCII pipeline on <code>main</code></summary>

```text
WORLD FACADE — construct · control · step · inspect · checkpoint
  Current: DART-owned method options and resolved configuration
  Planned: versioned selection policy over tested compatible methods
                         │
PHYSICAL SCENE — domains + dimensions + representations + interactions
  Current: rigid/free and articulated paths; bounded deformable paths
  Planned: qualified bidirectional coupling; thin structures and fluid models
                         │
SOLVER PORTFOLIO — each method owns a declared supported envelope
  Current: SI / boxed-LCP contact; rigid IPC; VBD / AVBD;
           articulated semi-implicit / variational; deformable methods
  Planned: versioned variants and validated shared/partitioned combinations
                         │
DATA AND COLLISION — model · state · controls · contacts · continuation
  Current: split components, dense rigid/multibody model, native collision,
           metrics, replay and bounded binary snapshots
  Partial: independent shared-model states and complete portable checkpoints
                         │
COMPUTE — semantic dependency graph → executable plan → runtime adapter
  Current: ordered World stages, explicit graphs within selected stages,
           Taskflow CPU execution, selected resident CUDA kernels
  Planned: qualified ownership/completion and full CUDA rigid/contact step pipeline
```

</details>

After: the published page embeds the rendered views (headless Chrome, 1440px wide, local `pixi run docs-build` of this branch).

Architecture page with the embedded simulation framework view

The four views as their standalone interactive pages (`?theme=light`, 1440×900 desktop viewport, archify showcase mode with zero diagnostics):

Simulation framework view

World::step() data flow view with three guided variants

Compute graph view

Library context view

## Testing

- `pixi run lint` and `pixi run check-lint` (full gate, including `check-architecture-map`, `check-ai-commands`, `check-ai-infra`) on the branch merged with `main`.
- `pixi run docs-build`: the conf.py hook renders the four views and the page embeds them; the two Sphinx warnings are the pre-existing autodoc "cannot import dartpy" baseline.
- `pixi run test-ai-infra`: **658 passed** (render driver, checker, runtime checker, and page-lint tests; the page-lint test was previously unwired and stale at eight rows, the page has nine).
- Built and ran `test_architecture_probe` (Release); `pixi run check-architecture-map-runtime --strict` agrees with the views and a fresh probe dump. On first use the advisory check caught that the fused guided view omitted `deformable_dynamics`; fixed.
- Codex round 1 (commit `4bc9c8ab28`): fixed qualified-symbol resolution (a `dart::a::b::Name` must be declared under `dart/a/b/`), strict mode now fails when the probe cannot run, guided views are checked for stage order and the deformable stage moved into the advance column, the page's source-evidence promise is narrowed to the three architecture views, and the implicit string concatenation is gone. The P1 about `except OSError, subprocess.CalledProcessError:` is PEP 758 syntax that black 26 emits for the pinned Python 3.14: verified with `ast.parse` and a module import on Python 3.14.7, and `test-ai-infra` imports the module on every run.
- Codex round 2 (commit `11394f2d2c`): the split and IPC guided views now name the standalone multibody and deformable stages that run alongside the rigid path, probe discovery accepts the Windows `.exe`, and every view's JSON shape is validated before rendering or text fallbacks so a malformed view fails the docs build in every environment. The repeated P1 syntax claim and the repeated qualified-symbol and strict-probe findings were already resolved in round 1 (see above).
- Codex round 3 (commit `e0b02d55a2`): qualified symbols now resolve against the namespace that encloses the declaration (`dart::simulation::WorldStepStage` is rejected, `dart::simulation::compute::WorldStepStage` accepted, nested and compact namespace forms both handled); the driver runs the structural checker before rendering or fallbacks; the probe derives its stage list from `makeBuiltInWorldStepSchedule()` so it passes with `DART_BUILD_PROFILE=OFF` and cross-checks the profile only when compiled in, and it records only the profiled step (fixture regenerated: one kinematics graph); Markdown summaries of each view are generated and included through `only` blocks so PDF and EPUB builds keep the map; the split guided view names the optional domain stages; the classic page no longer claims the DART 6 `World` exists on `main`. The 15 `myst.xref_missing` warnings visible in a full docs build come from `papers.md` links to design and plan docs on `main` and predate this PR.
- Codex round 4 (commit `b56bd146f3`): the probe dump now states whether its stage list is an execution trace (profiling compiled in, the pixi default) or only the built-in schedule, and the runtime checker refuses to treat a schedule-only dump as fresh evidence or to regenerate the fixture from one without `--allow-schedule-only`; the checker and renderer reject duplicate logical view names across suffixes; every node must carry a status tag. Nine re-listed findings were already resolved in rounds 1 to 3.
- Codex round 5 (commit `055c7b35b0`): the probe lookup now covers the multi-config layout that Visual Studio builds use (`build/<env>/cpp/bin/<Config>/`) and the active pixi environment, and the runtime checker rejects a fixture or fresh dump with an empty stage list or no executed compute graph, in the checks and in `--regenerate`, so the evidence can no longer be erased silently.
- Codex round 6 (commit `5ff50850a5`): the split guide drops `multibody_forward_dynamics`, which `makeBuiltInWorldStepSchedule()` only selects under the combined IPC branch; the Skeleton loading bridge is drawn solid because `dart/simulation/io/skeleton_loader.cpp` implements it; the engine region is labelled as public only through the installed headers (with a matching page sentence) instead of "never public"; the blocking gate now sweeps every public selector `enum class` in `world_options.hpp` and `multibody/multibody_options.hpp` and a framework card lists every enumerator; the `win-64` `lint` and `check-lint` aggregates run the map gate, guarded by a test over every pixi target. The view screenshots above were recaptured from this commit.
- Codex round 7 (commit `d28090ff99`): a line-cited source must be labelled with the symbol declared at the cited lines and the gate fails when those lines no longer contain it (so an insertion that shifts a declaration breaks the build instead of silently retargeting the `SRC` link); the World-to-sensor relationship is drawn as planned because nothing under `dart/simulation` uses `SensorManager` and `Sensor` still binds classic frames; fcl, bullet, and ode are described as compatibility facades on the classic page and in the library-context view, matching `dart/collision/AGENTS.md`; the design doc's deferred list moved to a parked `PLAN-131` dashboard entry per `docs/design/AGENTS.md`. The library-context screenshot above was recaptured from this commit.
- Codex round 8 (commit `f6531ea8a4`): the combined IPC stage moved into the second column of the step-flow grid so the IPC guide renders `RigidIpcContact` before both standalone multibody stages (the rigid and joint position nodes moved up to keep the column geometry, and the fan-out flows from the sync node leave from its bottom edge); the gate validates symbols named in cards, boundaries, stage labels, and guide notes; qualified symbols resolve only through return-type-led function declarations, so `dart::simulation::move` no longer resolves through `std::move(...)` calls; the stage scan walks the whole compute tree (headers and sources); the runtime check matches normalized compute-node stems against whole vocabulary tokens, so `kinematic_level_0_chunk_0` is a finding. The step-flow screenshot above was recaptured from this commit.
- Codex round 9 (commit `304ded114b`): the render driver writes a marker file into the archify checkout it clones and replaces only a directory carrying that marker (or an empty one), refusing any other `--deps-dir` with a message instead of deleting it; a qualified line-cited label such as `World::step` now requires the named owner to be the class, enum, or namespace enclosing the cited lines (`Wrong::step` fails, template headers open no scope); the step-flow view's two non-slot nodes are listed as `STEP_VIEW_BOOKENDS` with reasons, every other node must be a snake-cased slot, and the page, skill, and design doc state that distinction.
- Codex round 10 (commit `b890b81a93`): a pinned archify checkout is reused only when `git diff --quiet HEAD` is clean, otherwise a marked cache is re-cloned and any other directory is refused; the stage scan matches `class` or `struct` under any access specifier and any qualification of `WorldStepStage` (comments stripped, names without a `Stage` suffix included) and still finds the same 14 stage classes; recorded compute-graph edges must be `[from, to]` pairs of recorded nodes before a dump is accepted or regenerated; every architecture component must cite at least one source; the framework view now draws compute graphs from the rigid solver families (the rigid integration stage builds one) and labels the multibody edge as `parallelFor`, with the card text saying which stages emit graphs. The framework screenshot above was recaptured from this commit.
- Codex round 11 (commit `a12228236f`): the runtime fixture must name a guided view (a missing or null `guided_view` is a finding, so a reversed stage order can no longer pass `--strict`); every top-level view array (components, nodes, connections, flows, cards, boundaries, stages) must hold objects only, checked by the gate and by the renderer's shape check so the text fallback cannot crash on a scalar; the collision relationship is redrawn as the scheduled stages querying native collision (`makeBuiltInWorldStepSchedule()` never sees contacts), with the page saying so. The framework screenshot above was recaptured from this commit.
- Codex round 12 (commit `a2c8058cad`): `execution_trace` must be a JSON boolean in fixtures and probe dumps (a string such as `"false"` is a finding and blocks `--regenerate`), and only `true` counts as observed execution; `--regenerate` validates the guided view it records against the step-flow view's `meta.views`, refusing a stale id and accepting `--guided-view <id>` (a null or missing id records the validated default); the Markdown summaries used by PDF and EPUB builds now append `(planned)` to dashed relationships. The `x86_64 avx2` CI failure on the previous head was a transient conda download error in environment setup; the re-run passed.
- Codex round 13 (commit `50f02ec66a`): the HTML fallback marks dashed relationships as planned like the Markdown summaries; the gate validates every guided-view `focus` entry against the view's node ids; the classic page orders the layers Foundation, Dynamics, Collision, Constraint because collision consumes dynamics types, and names `dart/math/lcp` (`DantzigSolver`, `PgsSolver`) as the LCP owner with `dart/lcpsolver` as the deprecated compatibility surface; `dart/common/AGENTS.md` points its first See Also entry at the living architecture page instead of duplicating the allocator design link.
- Codex round 14 (commit `46650687eb`): the stage scan accepts both `virtual public` and `public virtual` base specifiers; a line-cited span must contain a declaration of its labelled symbol (type, function or member led by its type, or enumerator) rather than a mention in a comment or call; the default render path fails when one of the four required views is missing instead of rendering the rest; card titles and items are included in the Markdown summaries and the HTML fallback, so the public-selector list survives PDF and EPUB builds.
- Codex round 15 (commit `1024dd0ad4`): each public selector's enumerators are validated inside its own `Enum: A, B` card item (a token surviving in another card no longer covers it, and a missing item is an error); every edge needs a unique id; the page check accepts only an `<iframe src=...>` as the embed of a view, not a link; recorded compute graphs are rejected on self-edges and cycles; `compare_dumps` reports a drifted reference scene; the probe lookup honors `BUILD_TYPE`/`CMAKE_BUILD_TYPE` and otherwise picks the newest binary; the step-flow card now states the built-in schedule selects at most seven slots (the longest fused configuration plus kinematics).
- Codex round 16 (commit `b49985d507`): with `BUILD_TYPE`/`CMAKE_BUILD_TYPE` set, a missing probe of that configuration yields no probe instead of another build's binary; a selector card item may list only enumerators the enum declares (a stale `Legacy` value is an error); every runtime dump must carry a nonempty `scene`; the fixture's stage list must equal its guided view's scheduled entries (focus minus the two bookends) in order, so a deleted stage is a finding; Markdown summaries and the HTML fallback cite each node's sources as `path:line-end`.
- No browser-based archify `visual-check` (declared non-goal); the screenshots above are plain headless-Chrome captures of the delivered HTML.

---

<details>
<summary>Checklist</summary>

- [x] Milestone: DART 7.0.
- [x] Changelog updated.
- [x] Unit tests: driver, checker, runtime checker, and the C++ probe test.
- [x] Docs: design doc, skill, handbook pointers; API docs and Python bindings N/A (docs tooling and a test only).
- [x] Architecture map views updated (this PR introduces them).

</details>


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3508, opened 2026-09-15
Fix World::getIndex stale DOF boundaries after skeleton removal or topology changes

### LINKED ISSUE #3497: World::getIndex returns stale DOF boundaries after removal or topology changes
**Before submitting:**

**Environment:**

- OS: Ubuntu 24.04.5 LTS

`World::getIndex(i)` documents cumulative DOF boundaries for `i` in `[0, getNumSkeletons()]`, including the total DOF count at the terminal index. I expect those values to reflect the world's current skeletons and their current widths. After removing a middle skeleton, the terminal boundary remains stale. Changing the topology of a skeleton already in the world also leaves cached boundaries stale.

For initial widths `[2, 3, 4]`, removing the middle skeleton should produce `[0, 2, 6]`; the current branch reports `[0, 2, 5]`. Adding one DOF to the first remaining skeleton should then move the following boundaries, but they remain unchanged.

**Steps to Reproduce:**

1. Check out the commit above and build DART in Debug mode.
2. Compile `poc/repro.cc` using the command in `poc/README.md`.
3. Run the executable and observe `BUG_REPRODUCED=true` and `TOPOLOGY_BUG_REPRODUCED=true`.

**Code to Reproduce:**

The complete minimal program is attached as `poc/repro.cc`. Its essential sequence is to add three differently sized skeletons, remove the middle one, inspect every legal `getIndex` boundary, then add a joint to a remaining skeleton and inspect again.

**Additional Info:**

`removeSkeleton` stops its index-repair loop one entry early before popping the old terminal boundary. The separate topology case exists because `mIndices` is updated on world membership changes but not when a contained skeleton's DOF count changes.

The proposed patch repairs the complete removal range and computes the public boundary from current skeleton widths, with a regression for both sequences.

[world-get-index-stale-after-topology-change-poc.zip](https://github.com/user-attachments/files/32205534/world-get-index-stale-after-topology-change-poc.zip)

jslee02:
## Summary

- `World::getIndex()` now returns correct cumulative DOF boundaries after `removeSkeleton()` and after a skeleton already in the world gains or loses joints. Previously the terminal total kept the removed skeleton's width, and topology changes stayed invisible until the next `step()`.
- No API or header change; the fix lives in `World.cpp` plus one regression test.

## Motivation / Problem

- Fixes #3497. With skeleton widths `[2, 3, 4]`, removing the middle one reported `[0, 2, 5]` instead of `[0, 2, 6]`, and adding a joint to a remaining skeleton left every later boundary unchanged.
- Root cause: the repair loop in `removeSkeleton()` stopped one entry early, and the cached prefix sums were only rebuilt on membership changes or when entering simulation mode.

## Changes / Key Changes

- `getIndex()` sums the skeletons' current DOF counts instead of reading the cache, so it tracks topology changes. It is O(n) in skeleton count; a comment names the cache key to use if that ever matters.
- `addSkeleton()` and `removeSkeleton()` rebuild the cache through the existing `refreshSkeletonDofIndices()` helper. That retires the delta-subtraction repair, which was also wrong whenever the removed skeleton's width had changed since it was added.
- New regression test `World.GetIndexTracksRemovalAndTopologyChanges`.

## Before / After

| Scenario | Before | After |
| --- | --- | --- |
| `getIndex(2)` after removing the middle of `[2, 3, 4]` | 5 | 6 |
| Boundaries after adding a joint to a contained skeleton | stale until next `step()` | correct immediately |
| Cost of `getIndex(i)` | O(1) cached read | O(i) sum |

## Testing

- Negative check: the new test fails on the parent commit with exactly the four expected mismatches (5 vs 6, 9 vs 10, 5 vs 6, 5 vs 7) and passes with the fix.
- The reporter's `repro.cc` from the issue attachment, compiled against the fixed Release build, prints `BUG_REPRODUCED=false` and `TOPOLOGY_BUG_REPRODUCED=false` and exits 0.

## Breaking Changes

- [x] None

- Fixes #3497. Thanks to @peachtree0222 for the analysis and reproduction.
- DART 6 only: `main` has no `World::getIndex` or `mIndices`, so no companion PR.
- Left unchanged: `Recording` keeps its own per-skeleton widths (refreshed only on add/remove), so `bake()` recordings still assume a fixed layout across frames.

---

#### Checklist

- [x] Document new methods and classes (N/A, none added)
- [x] Add Python bindings (dartpy) if applicable (N/A, existing binding unchanged)


jslee02:
@codex review

jslee02:
@codex review

PR #3509, opened 2026-09-15
Refresh both collision groups in two-group collide and distance queries

### LINKED ISSUE #3499: Binary collision queries automatically update only the receiver group
**Before submitting:**

**Environment:**

- OS: Ubuntu 24.04.5 LTS
- Compiler: GCC 11.5.0, Debug builds

With automatic update enabled on two subscribed `CollisionGroup` objects, I expect a two-group collision or distance query to refresh both operands before querying the detector. Currently only the receiver calls `update()`.

After a collision shape is added to the second group's subscribed body, `first.collide(&second)` returns false and leaves the second group at zero shapes. Reversing the same symmetric query to `second.collide(&first)` updates the second group and returns true. The result therefore depends on operand order.

**Steps to Reproduce:**

1. Build either reviewed branch.
2. Compile the branch-specific source (`poc/repro-dart6.cc` or `poc/repro-main.cpp`) using `poc/README.md`.
3. Run it and compare the `forward` and `reverse` lines.

**Code to Reproduce:**

The attached program creates two subscribed groups, performs a baseline query, adds a collidable shape only to the second source, and queries in both orders.

[binary-collision-query-skips-other-group-update-poc.zip](https://github.com/user-attachments/files/32208987/binary-collision-query-skips-other-group-update-poc.zip)

jslee02:
## Summary

- `CollisionGroup::collide(otherGroup, ...)` and
  `CollisionGroup::distance(otherGroup, ...)` now refresh the other operand as
  well as the receiver, each according to its own automatic-update setting.
- Adds a regression test in `test_CollisionGroups` covering both operand
  orders, the two-group `distance` overload, and a group that opted out of
  automatic updates.
- Reported by @peachtree0222 with a minimal reproduction and proposed patch
  (#3499); the fix here follows the same idea.

## Motivation / Problem

With automatic updates enabled on two subscribed groups, a two-group query only
called `update()` on the receiver. The other operand went to the detector with
stale subscriptions, so a collision shape added to a body it subscribed to was
invisible: `first.collide(&second)` returned `false` and left `second` at zero
shape frames, while `second.collide(&first)` returned `true`. The result
depended on operand order. Every engine's two-group entry point only refreshes
poses of already-known objects (`updateEngineData()`), so the subscription
refresh has to happen in the `CollisionGroup` layer.

## Changes / Key Changes

- `dart/collision/CollisionGroup.cpp`: after the receiver's own update, call
  `otherGroup->update()` when `otherGroup` is non-null, not `this`, and has
  automatic updates enabled. Two-line change in each of the two overloads; no
  header or ABI change.
- `tests/integration/test_CollisionGroups.cpp`: new
  `PairwiseQueryUpdatesOtherGroup` parameterized test.

## Before / After

| Query after adding a shape to `second`'s body | Before | After |
| --- | --- | --- |
| `first.collide(&second)` | `false`, `second` has 0 frames | `true`, `second` has 1 frame |
| `second.collide(&first)` | `true`, `second` has 1 frame | `true`, `second` has 1 frame |
| `first.distance(&second)` | `second` stays stale | `second` refreshed first |
| `second.setAutomaticUpdate(false)` then `first.collide(&second)` | `second` untouched | `second` untouched (unchanged) |

## Testing

- Negative check: the new test fails on the unfixed tree at the forward,
  reverse, and distance assertions on both `dart` and `fcl`, and passes after
  the fix.
- Reporter's `repro-dart6.cc` compiled against this build: `forward=true
  second_count=1`, `reverse=true second_count=1`, `BUG_REPRODUCED=false`.
- The same sequence run directly on the `dart`, `fcl`, `bullet`, and `ode`
  detectors agrees on every engine (the `test_CollisionGroups` binary skips
  bullet/ode at runtime because the linker drops those unreferenced engine
  libraries; pre-existing, not changed here).

## Breaking Changes

- [x] None
- Two-group queries now also refresh the other operand when it has automatic
  updates enabled (the default): a version check per subscribed source when
  nothing changed, and the collision-object work the caller expected otherwise.

- Fixes #3499
- DART 7 `main` has the same defect in `dart/collision/collision_group.cpp`;
  a matching `main` PR follows this one (dual-PR policy).

---

#### Checklist

- [x] Milestone set (DART 6.20.0 for `release-6.20`)
- [x] Document new methods and classes (no new API)
- [x] Add Python bindings (dartpy) if applicable (existing binding inherits the fix)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3512, opened 2026-09-15
Honor DART_SKIP_spdlog when spdlog is installed

### LINKED ISSUE #3503: DART_SKIP_spdlog=ON still discovers and enables spdlog
**Before submitting:**

- [x] Checked the build documentation and existing issues

**Environment:**

- OS: Ubuntu 24.04.5 LTS
- Compiler/build tools: GCC 11.5.0 and CMake 4.2.3

The option description says `DART_SKIP_spdlog=ON` should avoid using spdlog even when it is installed, and the DART 7 build-system guide lists the same skip option. I expect core to configure without `DART_HAVE_spdlog` in that case.

Both maintained branches declare the option but call `find_package(spdlog ...)` unconditionally afterward. On a host with spdlog installed, an explicit skip configuration still generates `DART_HAVE_spdlog=1`. This can also force an installed spdlog built against an incompatible external fmt into a build that explicitly requested isolation from it.

**Steps to Reproduce:**

1. On a host where CMake can discover spdlog, configure DART with `-DDART_SKIP_spdlog=ON`.
2. Inspect the core target's generated compile definitions or build verbosely.
3. Observe `DART_HAVE_spdlog=1` and spdlog/fmt definitions despite the skip request.

**Code to Reproduce:**

The exact configure and inspection commands are in `poc/README.md`, with baseline and fixed compile-definition excerpts.

**Additional Info:**

Guarding package discovery with `if(NOT DART_SKIP_spdlog)` and explicitly setting `spdlog_FOUND` false in the skip branch produces `DART_HAVE_spdlog=0`. A clean DART 7 core build then completes without spdlog. The same correction applies to the active DART 6 LTS dependency file.

[skip-spdlog-option-ignored-poc.zip](https://github.com/user-attachments/files/32209649/skip-spdlog-option-ignored-poc.zip)

jslee02:
## Summary

- Make `-DDART_SKIP_spdlog=ON` actually skip spdlog: the option was declared but `find_package(spdlog ...)` ran unconditionally afterwards, so on a host with spdlog installed a skip build still got `DART_HAVE_spdlog=1`. Reported by @peachtree0222 in #3503, including the fix shape used here.

## Motivation / Problem

- `cmake/DARTFindDependencies.cmake` declared `DART_SKIP_spdlog` ("do not use spdlog even if it is found") and then called `dart_find_package(spdlog)` on the next line; nothing ever read the option. `dart/CMakeLists.txt` keys the `DART_HAVE_spdlog` definition, the spdlog link, and the exported `dart` component dependency off `spdlog_FOUND`, so the skip was a no-op wherever CMake could find spdlog. It also pulled an installed spdlog (and its external fmt linkage) into builds that explicitly asked to be isolated from it.

## Changes / Key Changes

- Probe for spdlog only when `DART_SKIP_spdlog` is OFF; when ON, set `spdlog_FOUND` to FALSE so a value inherited from a parent project cannot re-enable it. One file, no change to the default configuration.

## Before / After (`-DDART_SKIP_spdlog=ON`, spdlog installed)

| | Before | After |
| --- | --- | --- |
| Core compile definitions | `-DDART_HAVE_spdlog=1`, `-DSPDLOG_FMT_EXTERNAL`, `-DSPDLOG_FWRITE_UNLOCKED` | `-DDART_HAVE_spdlog=0`, no `SPDLOG_*` |
| Exported `dart` component deps | `assimp;Eigen3;fcl;fmt;octomap;spdlog` | `assimp;Eigen3;fcl;fmt;octomap` |
| `spdlog_DIR` in a fresh cache | set | absent (never probed) |

## Testing

- Reproduced on a host with spdlog 1.16.0 / fmt 11.2.0 (CMake 4.3, GCC 15.2); CI never configures with the skip option.
- Skip-ON: built the `dart` target with the branch-default warnings-as-errors (0 warnings) and ran `UNIT_common_Logging`, which passes through the iostream fallback path.
- Skip-OFF: reconfigured the default tree; the core compile line and exported component list are identical to before, and `dart` builds.

## Breaking Changes

- [x] None (only builds that explicitly set `DART_SKIP_spdlog=ON` change, and they now do what the option always promised).

- Fixes #3503
- DART 7 `main` has the same bug in `cmake/dart_find_dependencies.cmake`; a matching `main` PR follows this one per the dual-PR policy.

---

#### Checklist

- [x] CHANGELOG.md updated if required (follow-up commit once the PR number exists)
- [ ] Add unit tests for new functionality (n/a: build-system fix, covered by the skip-ON build and the existing logging test)
- [ ] Document new methods and classes (n/a)
- [ ] Add Python bindings (dartpy) if applicable (n/a)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3513, opened 2026-09-15
Fix Recording reinterpreting old frames after the skeleton DOF layout changes

### LINKED ISSUE #3498: Recording reinterprets historical frames after the skeleton schema changes
**Before submitting:**

**Environment:**

- OS: Ubuntu 24.04.5 LTS

I expect a baked frame to retain an unambiguous relationship to the skeleton layout under which it was packed, or to be invalidated when that layout changes. Instead, `Recording` retains old packed vectors while replacing its single per-skeleton DOF schema.

If two one-DOF skeletons are baked at positions `11` and `22`, removing the first skeleton makes the recording report one skeleton, but historical row zero still returns `11`. The only live skeleton is the former row one at `22`, so the old value is silently attributed to the wrong object. Growing an existing skeleton's topology before the next bake creates the same schema mismatch.

**Steps to Reproduce:**

1. Check out and build the commit above.
2. Compile `poc/repro.cc` as documented in `poc/README.md`.
3. Run it and observe one retained historical frame, one reported skeleton, `historical_row_zero=11`, `surviving_position=22`, and `BUG_REPRODUCED=true`.

**Code to Reproduce:**

The complete program is attached as `poc/repro.cc`.

**Additional Info:**

The proposed fix treats a changed vector of per-skeleton DOF counts as a recording epoch boundary, clears incompatible frames, and refreshes the schema at the start of `World::bake`. The regression covers removal and in-place topology growth.

DART 7's new replay design documents explicit rejection of incompatible static layouts and does not use this implementation. I would therefore request a DART 6 LTS-only scope exception and maintainer feedback on clear-on-change versus an explicit error for the legacy API.

[recording-schema-reinterprets-history-poc.zip](https://github.com/user-attachments/files/32205595/recording-schema-reinterprets-history-poc.zip)

jslee02:
## Summary

- Make `Recording` treat a changed per-skeleton DOF layout as a new recording
  epoch and refresh that layout at the start of `World::bake()`, so baked
  frames are never re-sliced under a layout they were not packed with.
  Fixes #3498. Thanks to @peachtree0222 for the public-API repro and the
  proposed patch.

## Motivation / Problem

- `Recording` keeps one per-skeleton DOF width vector for every baked frame.
  `World::addSkeleton`/`removeSkeleton` overwrote it while keeping the old
  frames, so after removing the first of two 1-DOF skeletons baked at 11 and
  22 the recording reported one skeleton whose history read 11, the removed
  skeleton's value. Growing a contained skeleton before the next bake produced
  the same mismatch, and `bake()` never refreshed the layout at all.

## Changes / Key Changes

- `Recording::updateNumGenCoords` compares the new widths with the stored
  ones; when they differ and frames exist it drops the frames and warns with
  the dropped count. Unchanged layouts are a no-op, so setup-time
  `addSkeleton` calls and repeated bakes stay silent.
- `World::bake()` refreshes the layout before `addState`, so in-place joint
  additions or removals are detected.
- Regression test `World.RecordingDropsFramesWhenSkeletonDofLayoutChanges`
  covers removal, in-place growth, unchanged-layout accumulation, and adding
  a skeleton while frames exist.
- No public header changes; `Recording.hpp` and the `updateNumGenCoords`
  signature are untouched. No dartpy change (`Recording` is not bound).

| Scenario | Before | After |
| --- | --- | --- |
| Remove the first of two baked 1-DOF skeletons | frame kept, row 0 reads 11 for the survivor | frame dropped with a warning; next bake reads 22 |
| Grow a contained skeleton, then bake | recording reports 1 DOF for a 2-DOF frame | layout refreshed to 2 DOFs, stale frame dropped |
| Bake repeatedly with the same layout | frames accumulate | unchanged, no warning |

## Testing

- Negative check: the new test fails on the unfixed code at five
  expectations (frame retained after removal, row 0 reads 11, stale width 1
  after growth).
- The reporter's `repro.cc` compiled against this branch prints
  `BUG_REPRODUCED=false` and exits 0.
- Verified the growth case needs #3508: with the pre-#3508 `getIndex`,
  `bake()` sizes the frame from the stale cache and the new DOF reads back 0.

## Breaking Changes

- [x] None
- Behavior change for a previously undefined case: changing the skeleton DOF
  layout while frames exist now discards those frames with a warning instead
  of keeping frames sliced with the wrong layout.

- Fixes #3498
- Builds on #3508 (#3497, merged), which sizes frames correctly after topology changes
- DART 6 only: DART 7's replay design already rejects incompatible layouts,
  so there is no matching `main` PR.

---

#### Checklist

- [x] Document new methods and classes (none added)
- [x] Add Python bindings (dartpy) if applicable (not applicable)


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3432, opened 2026-08-02
Add VBD and AVBD paper-parity foundation

jslee02:
## Summary

Foundation PR for the PLAN-104 VBD/AVBD paper-parity contract. This head
closes the zero-trust audit findings and reseals the Figure 13 evidence on
the final bytes. No paper-parity row is claimed complete: the 88 VBD and 88
AVBD contract rows stay fail-closed.

Highlights since the last review head:

- Rigid AVBD: persistent manifold row scratch for every caller (the
  six-argument builder overload that cold-started continuation is gone);
  all ten public AVBD joint tests assert the paper's Eq. 18 schedule
  (5 % of a pre-existing violation corrected per step under alpha = 0.95);
  warm-started friction duals are projected onto their Coulomb cone;
  multi-point manifolds keep continuation while material points move by less
  than a tenth of the manifold's point separation and cold-start (counted in
  the manifold scratch) on any non-finite point; compatibility distance
  springs ramp and warm-start on one paper-profile schedule under every
  rigid-body family, so the continuation preserved across AVBD <-> SI
  crossings is schedule-consistent.
- Bounded, fail-closed contact storage for deformable and rigid pipelines
  (exact valid-pair bound up to 65,536 candidates, budgeted reserve above;
  automatic rigid shape-pair envelopes with 65,536-pair / 16,384-contact
  reserve budgets).
- Evidence pipeline: capture provenance records working-tree cleanliness and
  ignored paths (v2); encoder identity binds ffmpeg/libx264 versions; the
  CMake build-configuration record (v2) adds launcher/linker/cache/path
  variables and drops the retired DART_FETCH_FILAMENT option; schema-6 packets
  seal their sources through one shared helper; the capture harness no longer
  discards a run because its own process lazily mapped a trusted Python
  extension module.
- Legacy packets: every allowlisted historical packet carries its claim
  boundary inside the packet (three were stamped in this range with
  `scripts/bound_legacy_avbd_packet.py`); the two breakable scale packets are
  regenerated at schema 6 and, like the three Figure 13 packets, have left
  the legacy schema pins so a downgrade of current evidence is rejected.

### Figure 13 finding (maintainer decision recorded)

Under the immutable paper profile the public AVBD wall no longer fractures
the way the private contact configuration did: alpha = 0.95 now applies to
contact rows, so a 24 m/s ball that starts its first contact step deeply
penetrating keeps 95 % of that penetration per step and lodges in the wall;
the anchored wall transmits the impulse to the ground, 36 joints break
[...]

PR #3516, opened 2026-09-16
Fix Python 3 except syntax in the build-time provenance script

### LINKED ISSUE #3502: Source build fails on invalid Python 3 exception syntax in provenance script
**Before submitting:**

- [x] Checked the documentation, existing issues, and open PRs

**Environment:**

- DART version: `main` at `deb98693d479f12d6f1ae4b257479584448584b9`
- OS: Ubuntu 24.04.5 LTS
- Compiler/build tools: GCC 11.5.0, Python 3.12, CMake 4.2.3

I expect a normal source build to generate the source-provenance header and proceed to C++ compilation. The build instead stops while parsing `scripts/capture_source_provenance.py` because lines 677 and 822 use Python 2-style multi-exception syntax: `except OSError, RuntimeError, ValueError:`.

Python 3 requires those exception types to be parenthesized. The first malformed clause masks the second until it is corrected.

**Steps to Reproduce:**

1. Configure the reviewed `main` commit using the documented source-build flow.
2. Build any DART target that depends on `dart_capture_source_provenance_header`.
3. Observe `SyntaxError: multiple exception types must be parenthesized` at line 677.

Alternatively, run `python3 -m py_compile scripts/capture_source_provenance.py` to reproduce the parser failure directly.

**Code to Reproduce:**

```text
python3 -m py_compile scripts/capture_source_provenance.py
```

**Additional Info:**

Changing both clauses to `except (OSError, RuntimeError, ValueError):` makes Python compilation pass and lets the DART core target build successfully. PR #3432 introduced the file. The script is absent from the maintained DART 6 line, so this requires a `main`-only scope exception.

[provenance-script-python-syntax-error-poc.zip](https://github.com/user-attachments/files/32209464/provenance-script-python-syntax-error-poc.zip)

jslee02:
## Summary

- A C++-only source build of `main` fails while generating
  `capture_source_provenance.hpp` whenever CMake finds a Python older than
  3.14: `scripts/capture_source_provenance.py` used PEP 758
  `except A, B:` syntax in two clauses, a `SyntaxError` on Ubuntu 24.04's
  Python 3.12. Both clauses are now parenthesized.
- Parenthesizing alone is undone by `pixi run lint`: black inferred a py314
  target from `requires-python` and rewrote the clauses back. `[tool.black]`
  now pins `target-version` to py310 through py314 (Ubuntu 22.04, the oldest
  platform in the support table, ships 3.10), so the formatter cannot emit
  3.14-only syntax into build-time scripts; every other file is left
  unchanged.
- `docs/onboarding/building.md` now states the build-time floor under
  Prerequisites: the Python 3 that CMake discovers must be 3.10 or newer
  because it generates the source-provenance header.
- A unit test parses every script that runs outside the Pixi environment
  with the grammar of its floor (`ast.parse(feature_version=...)`): the
  provenance script, its `avbd_packet_schema` import, and the headless-smoke
  analyzer at 3.10 (CMake/ctest), the pre-commit agent hook and its Windows
  PreToolUse bridge at 3.11, and the
  community-signals dashboard generator at 3.10 (workflow runner Python). The
  regression therefore fails `pixi run test-py` even though every Pixi
  environment pins 3.14.

Reported with a PoC by @peachtree0222. Main-only: the script came in with
#3432 and does not exist on `release-6.20`. The changelog entry is in this PR
with the issue link; the PR link is added in a follow-up commit once the PR
number exists, per `docs/onboarding/changelog.md`.

Fixes #3502

## Testing

- `python3.12 -m py_compile scripts/capture_source_provenance.py` on CPython
  3.12.13 and 3.11.15: `SyntaxError: multiple exception types must be
  parenthesized` (line 677) before, clean after. CI has no pre-3.14
  interpreter, so this cannot be shown there; the 3.10 floor is checked with
  `ast.parse(feature_version=(3, 10))`.
- Ran the script with CMake's exact command line (`--repo-root`,
  `--benchmark-source`, `--write-cpp-header`) under CPython 3.12 and 3.14:
  exit 0 and byte-identical headers.
- `pixi run lint` twice on the branch: idempotent, black leaves all 235 files
  unchanged; without the pin (`--target-version py314`) black would revert
  the fix.
- The new test fails on the pre-fix script (`except expressions without
  parentheses are only supported in Python 3.14 and greater`) and passes
  after.

---

<details>
<summary>Checklist</summary>

- [x] Milestone set (DART 7.0)
- [x] CHANGELOG.md updated per `docs/onboarding/changelog.md` (PR link
      follows in a separate commit once the PR number exists)
- [ ] Document new methods and classes: N/A, no API change
- [ ] Add Python bindings (dartpy) if applicable: N/A
- [ ] Architecture map views updated: N/A, no `dart/simulation/**` change

</details>


jslee02:
@codex review

jslee02:
@codex review

PR #3518, opened 2026-09-16
Refresh both collision groups in two-group collide and distance queries

### LINKED ISSUE #3499: Binary collision queries automatically update only the receiver group

jslee02:
Title: Refresh both collision groups in two-group collide and distance queries (DART 7)
Base: main | Milestone: DART 7.0

## Summary

- `CollisionGroup::collide(otherGroup, ...)` and `distance(otherGroup, ...)`
  only called `update()` on the receiver, so a collision shape added to a body
  the *other* group subscribed to stayed invisible and `first.collide(&second)`
  disagreed with `second.collide(&first)`. Both overloads now refresh the other
  operand too, according to its own automatic-update setting.
- `main` port of the release-6.20 PR #3509. Reported by @peachtree0222
  with a reproduction and a proposed patch in #3499.

**Behavior**: the native detector's two-group entry points only validate the
groups and refresh engine data, never subscriptions, so the refresh belongs in
the `CollisionGroup` layer. When nothing changed, the extra cost is one version
check per subscribed source on the other group. A group queried against itself
is not updated twice, and a group with `setAutomaticUpdate(false)` is left
alone in either position. No header, ABI, or dartpy change.

## Testing

- New `CollisionGroupTests.PairwiseQueryUpdatesOtherGroup` (native detector)
  covers the forward order, the reverse order, the two-group `distance`
  overload, and the opt-out case. On the unfixed tree it fails the forward,
  reverse, and distance assertions; with the fix the whole
  `UNIT_collision_CollisionGroup` binary passes (26 tests).
- Reporter's `repro-main.cpp` compiled against this build prints
  `forward=true second_count=1`, `reverse=true second_count=1`,
  `BUG_REPRODUCED=false` and exits 0.
- `pixi run lint` clean; `pixi run build` / `test-unit` left to the maintainer
  gate (only the affected test target was built here).
- Not run: the reference-engine (FCL/Bullet) configuration of this test file.
  The fix is engine-agnostic and the release-6.20 PR covers all four engines.
- Pre-existing, unrelated: the dartpy `collide`/`raycast` bindings call the
  detector directly and skip automatic updates entirely.

Fixes #3499

---

<details>
<summary>Checklist</summary>

- [ ] Milestone set (DART 7.0)
- [ ] CHANGELOG.md updated per `docs/onboarding/changelog.md` if required
      (entry drafted; added once the PR number exists)
- [x] Document new methods and classes (N/A: no new API)
- [x] Add Python bindings (dartpy) if applicable (N/A: no new API)
- [x] Architecture map views updated (N/A: no `dart/simulation/**` change)

</details>


jslee02:
@codex review

jslee02:
@codex review

jslee02:
@codex review

PR #3511, opened 2026-09-15
Fix constraint parameter setters storing invalid values after clamping warnings

jslee02:
## Summary

- Make the class-wide parameter setters of seven DART 6 constraint types store the bound they warn about instead of the invalid argument. Reported by @peachtree0222 in #3501.

## Motivation / Problem

- `setErrorAllowance`, `setErrorReductionParameter`, `setMaxErrorReductionVelocity`, and `setConstraintForceMixing` on `DynamicJointConstraint`, `JointConstraint`, `JointLimitConstraint`, `SoftContactConstraint`, `ContactConstraint`, `JointCoulombFrictionConstraint`, and `ServoMotorConstraint` warn that an invalid value "is set to" the bound, assign the bound in the invalid branch, and then overwrite the member with the original argument. The getter returns the invalid value, and since the members are static, one bad call affects every constraint of that type for the rest of the process.

## Changes / Key Changes

- Clamp the local argument in each of the 26 invalid branches so the existing final assignment stores the bound. No warning text, header, signature, or default changes.
- Add `ConstraintSolver.ParameterSettersClampInvalidValues` to `tests/integration/test_ConstraintSolver.cpp`: it drives every affected setter past each bound, expects the bound from the getter, and restores the previous class-wide value.
- Every other bounded `set*(double)` in `dart/constraint` (`CouplerConstraint`, `MimicMotorConstraint`, `ContactConstraint::setMaxErrorReductionVelocity`, `ConstraintSolver::setAutomaticSleepingContactPenetrationTolerance`) already keeps the bound and is untouched.

## Before / After

| Call, then getter | Before | After |
| --- | --- | --- |
| `JointConstraint::setErrorReductionParameter(1.25)` | `1.25` | `1.0` |
| `ContactConstraint::setConstraintForceMixing(0.0)` | `0.0` | `1e-9` |

## Testing

- Negative check: the new test fails on the unfixed branch with 26 `EXPECT_DOUBLE_EQ` failures, one per invalid branch, and passes after the fix.
- The reporter's 26-case probe (`repro.cc` from the issue) built against this branch prints `failed_checks=0` and `BUG_REPRODUCED=false`.

## Breaking Changes

- [x] None
- Values that were previously stored out of range are now clamped, which is what the warnings always claimed. Valid arguments behave exactly as before.

## Related Issues / PRs (backports)

- Fixes #3501
- Related clamp cleanup: #2641
- Correction (added after merge): the claim that DART 7 `main` was unaffected was wrong. A later audit
[...]

PR #3524, opened 2026-09-19
Fix constraint parameter setters storing invalid values after clamping warnings

### LINKED ISSUE #3501: Constraint parameter setters overwrite clamped values with invalid input
**Before submitting:**

**Environment:**

- OS: Ubuntu 24.04.5 LTS

When a setter warns that an invalid parameter “is set to” its lower or upper bound, I expect the getter to return that bound. In seven DART 6 constraint implementations, the invalid branch assigns the bound to the member and then the method unconditionally overwrites the member with the original invalid argument.

The attached probe checks error allowance, error-reduction parameter, maximum error-reduction velocity, and constraint-force-mixing limits across `DynamicJointConstraint`, `JointConstraint`, `JointLimitConstraint`, `SoftContactConstraint`, `ContactConstraint`, `JointCoulombFrictionConstraint`, and `ServoMotorConstraint`. All 26 low/high checks fail on the reviewed branch even though the warnings claim clamping occurred.

**Steps to Reproduce:**

1. Build the reviewed DART 6 LTS commit.
2. Compile `poc/repro.cc` using the supplied README.
3. Run it and observe `failed_checks=26` and `BUG_REPRODUCED=true`.

**Code to Reproduce:**

The complete 26-case public-setter probe is attached as `poc/repro.cc`.

**Additional Info:**

The proposed implementation normalizes the local parameter in each invalid branch and retains the existing final assignment. The regression and complete constraint-solver test executable pass with it.

PR #2641 contains related clamp-helper cleanup, so I am linking it rather than claiming a wholly new clamp topic. The equivalent DART 7 setters already preserve bounded values.

[constraint-clamp-setters-overwrite-bounds-poc.zip](https://github.com/user-attachments/files/32209286/constraint-clamp-setters-overwrite-bounds-poc.zip)

jslee02:
## Summary

- Constraint parameter setters warned that an out-of-range argument "is set to" a
  bound, assigned that bound to the static member inside the invalid branch, then
  unconditionally overwrote the member with the original argument. The branch
  assignment was dead, so the getter returned the value the warning had just
  rejected. Each invalid branch now clamps the local parameter, so the existing
  final assignment stores the bound.

This is the DART 7 counterpart of release-6.20 PR #3511. That PR and the issue
thread both stated `main` already preserved the bound. **That assessment was
incorrect.** Auditing every bounded setter in `dart/constraint` found the identical
defect in **18 setters across 6 classes** (`DynamicJointConstraint`,
`JointConstraint`, `JointLimitConstraint`, `SoftContactConstraint`,
`JointCoulombFrictionConstraint`, `ServoMotorConstraint`), covering 22 invalid
branches. `ContactConstraint` is the one class the earlier reading got right: it
already clamps via `std::max`/`std::clamp` and is untouched, as are the DART 7-only
`MimicMotorConstraint` and `CouplerConstraint`. Warning text, signatures, headers
and defaults are unchanged; the diff is 22 lines of `mMember = <bound>;` becoming
`<param> = <bound>;`.

Behavior change: an error reduction parameter above `1.0` is now genuinely clamped,
so callers tuned against the retained value will see different solver behavior —
though they were already warned on every call. CFM above `1.0` remains allowed.

The regression test replaces `DynamicJointConstraint.CurrentCompatibilityForHighErrorReductionParameter`,
which asserted an ERP of `2.0` was stored verbatim and so pinned the defect.

Reported by @peachtree0222, whose reproducer ported to `main` unchanged apart from
header names.

Fixes #3501

## Testing

CI covers the build and the full suites, so only what CI does not run is listed.

- **Red/green**: the test was written first and run against unfixed `main`, giving
  exactly 22 failing expectations, one per invalid branch, partitioned by
  `SCOPED_TRACE` as 5/5/5/5/1/1 with **zero** traced to `ContactConstraint`. After
  the fix `UNIT_constraint_JointConstraints` passes 9/9, warnings still printed.
- **Reporter's PoC**, ported to DART 7 header names: `failed_checks=22` before,
  `0` after — differing from DART 6's 26 by exactly the four `contact.*` checks,
  which pass unfixed. Independent confirmation of the corrected scope.
- **Collateral**: `ctest -R UNIT_constraint` passes 8/8. No other caller in
  `tests/`, `python/` or `examples/` depends on the old behavior.
- `pixi run lint` passes and reformats nothing.
- Not run here: `pixi run test-all` and the CUDA gate.

---

<details>
<summary>Checklist</summary>

- [ ] CHANGELOG.md updated per `docs/onboarding/changelog.md` if required (entry follows in a separate commit once this PR has a number, per the changelog guide)
- [ ] Document new methods and classes — N/A, no new API; the documented clamping
      contract is now actually honored
- [ ] Add Python bindings (dartpy) if applicable — N/A, existing bindings are
      pass-throughs and inherit the fix
- [ ] Architecture map views updated — N/A, no `dart/simulation/**` structure,
      step-stage slot, or solver-family change

</details>


jslee02:
@codex review

jslee02:
@codex review

Issue #3526, opened 2026-09-19
Lift the filament build 1 pin once the conda-forge build 2 packages stop deadlocking the headless demo viewer


jslee02:
## Problem

Since the lockfile refresh in #3523 moved filament from the conda-forge build 1 packages (`1.76.0 h803e5e9_1` on linux-64) to build 2 (`1.76.1 h3f833be_2`; `1.76.0 h3f833be_2` behaves the same), the demo viewer deadlocks as soon as it cycles scenes headless. `Release Tests` (step "Run Release Python tests", inside `integration/test_demos_cycle.py::test_runner_cycle_returns_zero`) and `Visual Verification (ImGui docking)` hang until their timeouts on `main` (run 35456937122) and on every pull request that includes that commit (#3525). A single-scene headless run still exits normally; `python -m examples.demos --cycle-scenes --headless --frames 1` hangs at the third scene.

Native backtrace (gdb) of the hang:

- main thread: `dart::gui::detail::renderApplicationFrame` → `filament::FRenderer::beginFrame` → `FrameSkipper::shouldRenderFrame` → `OpenGLDriver::getFenceStatus` → `OpenGLDriver::fenceWait` → `pthread_mutex_lock`
- second thread: `OpenGLDriver::fenceWait` from the `FrameInfoManager::endFrame` lambda, in `pthread_cond_timedwait`
- filament driver thread idle in `CommandBufferQueue::waitForCommands`

Swapping only the filament library under identical DART binaries reproduces it: build 1 finishes the three-scene cycle in about a second, both build 2 packages hang.

## Suspected trigger

Upstream filament 1.76.1 only changes an iOS backend test and Vulkan `readPixels`. The build 2 packages follow the feedstock's 2026-09-12/13 recipe changes (native X11 rendering/swapchains on Linux) and add `libxcb`/`xorg-libx11` dependencies that build 1 lacks (the build 1 recipe built Linux with `EGL_ON_LINUX=ON`; the current recipe turns it off, so headless runs get an X11-backed OpenGL context instead of EGL); 1.77.0 carries the same host requirements, so it is expected to behave like build 2 until the cause is fixed there or in `dart::gui`'s headless swap chain setup.

## Temporary pin

`pixi.toml` pins `filament = { version = "==1.76.0", build = "*_1" }` (#3527) and `docs/onboarding/build-system.md` documents it.

## Lift condition

[...]

PR #3527, opened 2026-09-19
Pin filament to the conda-forge build 1 packages to fix the headless demo viewer deadlock

jslee02:
## Summary

- Since the lockfile refresh in #3523, the `Release Tests` and `Visual Verification (ImGui docking)` jobs hang until their timeouts on `main` (run 35456937122) and on every PR that includes it (#3525). The demo viewer deadlocks inside filament when it cycles scenes headless: `Renderer::beginFrame` → `FrameSkipper::shouldRenderFrame` → `OpenGLDriver::getFenceStatus` → `fenceWait` blocks on a driver mutex held by the frame-info thread, whose own `fenceWait` never completes because the driver thread stays idle.
- The refresh moved filament from the conda-forge build 1 packages to build 2 (`1.76.1 h3f833be_2`; `1.76.0 h3f833be_2` deadlocks the same way), so this pins `filament = { version = "==1.76.0", build = "*_1" }` and re-solves only that package (build 1 on all five platforms). Lifting the pin is tracked in #3526.

## Testing

- `python -m examples.demos --cycle-scenes --headless --frames 1` against the two build-2 packages hangs at the third scene (exit 124 under `timeout`); the three-scene cycle used by `test_demos_cycle.py::test_runner_cycle_returns_zero` completes with rc 0 against the pinned build 1, with everything else identical (same DART binaries, library swapped through `LD_LIBRARY_PATH`). On that host build 1 renders through filament's Vulkan fallback, so its OpenGL path is covered by main's earlier green `Release Tests` runs on this exact package set (for example 90c79d9f16e) and by this PR's CI.
- Upstream filament 1.76.1 notes only an iOS backend-test change and a Vulkan `readPixels` fix; the behavior change tracks the feedstock's build-2 recipe, which is what the pin excludes.

- `docs/onboarding/build-system.md` describes the pin and the lift condition. No CHANGELOG entry: the pin restores the package set of the existing filament packaging entry and has no user-facing effect.

Related: #3526 (lift condition), #3523 (lockfile refresh), #3525 (blocked PR).

---

<details>
<summary>Checklist</summary>

- [ ] CHANGELOG.md updated — N/A, build-environment pin without user-facing change
- [ ] Add unit tests for new functionality — N/A
- [ ] Document new methods and classes — N/A
- [ ] Add Python bindings (dartpy) if applicable — N/A
- [ ] Architecture map views updated — N/A

</details>


jslee02:
@codex review

